PR #6, opened 2025-09-01
Open door events

alexmillane:
## Summary

Adds a reset event to the open door tasks which resets the door to the closed position on reset

## Details
- The event calls the `Openable` interface `close`
- I needed to expand these interfaces to take `env_ids` because events are triggered per `env_id`
- Expanded our tests to cover the multi-env case.

### DIFF
--- a/isaac_arena/tests/utils/subprocess.py
+++ b/isaac_arena/tests/utils/subprocess.py
@@ -46,12 +46,7 @@ def runner(q: multiprocessing.Queue, function: Callable[[SimulationAppContext, A
     # The runner runs a function in a way that a result is returned to the main process, before
     # simulation_app.close() can ruin everything.
     # Simulation app args. For now, we just make these default + headless.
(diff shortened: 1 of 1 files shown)


isaac_arena/tests/test_open_door.py
```diff
@@ -37,7 +35,7 @@ def _test_open_door_microwave(simulation_app) -> bool:
     from isaac_arena.tasks.open_door_task import OpenDoorTask
 
     args_parser = get_isaac_arena_cli_parser()
-    args_cli = args_parser.parse_args([])
+    args_cli = args_parser.parse_args(["--num_envs", str(num_envs)])
```
viiik-inside, line 38: should we just add it in the get_isaac_arena_cli_parser?

alexmillane: The argument is in there. This syntax is just setting the argument. It looks weird because the test doesn't take arguments from the command line.

PR #11, opened 2025-09-02
Alex/feature/generalize scene

alexmillane:
## Summary
Generalizes our scene class so it can be used across multiple tasks

## Detailed description
- Previously, we had a scene type per task type, for example `OpenDoorScene` for the `OpenDoorTask`
- This doesn't work when we want multiple tasks to be able to be carried out in a single scene.
- Here we introduce a more general `Scene` object which should support many tasks
- The design is that the `Scene` is just an arbitrary composition of assets.

### DIFF
--- a/isaac_arena/environments/compile_env.py
+++ b/isaac_arena/environments/compile_env.py
@@ -49,7 +49,6 @@ class ArenaEnvBuilder:
     @classmethod
     def from_args(cls, args: argparse.Namespace) -> ArenaEnvBuilder:
         cfgs = get_environment_configuration_from_asset_registry(args.background, args.object, args.embodiment)
-        scene = Scene(assets=[cfgs["background"], cfgs["object"]])
         if args.teleop_device is not None:
             cfgs.update(get_environment_configuration_from_device_registry(args.teleop_device))
         else:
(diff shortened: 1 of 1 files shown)


isaac_arena/tests/test_asset_registry.py
```diff
@@ -52,87 +51,69 @@ def test_default_assets_registered():
...
-    env_cfg = parse_env_cfg(
-        isaac_arena_environment.name,
-        device=args_cli.device,
-        num_envs=args_cli.num_envs,
-        use_fabric=not args_cli.disable_fabric,
+        objects_in_registry.append(asset)
+        objects_in_registry_names.append(asset.name)
+    assert len(objects_in_registry) > 0
+
+    scene = Scene(assets=[background, *objects_in_registry])
```
alexmillane, line 86: Constructing a scene with a variable number of assets is now a single line rather than the monstrosity it used to be.

isaac_arena/environments/compile_env.py
```diff
@@ -44,18 +44,21 @@ def __init__(self, arena_env: IsaacArenaEnvironment, args: argparse.Namespace):
...
         if args.teleop_device is not None:
             cfgs.update(get_environment_configuration_from_device_registry(args.teleop_device))
         else:
             cfgs["device"] = None
         arena_env = IsaacArenaEnvironment(
             name=f"pick_and_place_{args.embodiment}_{args.background}_{args.object}",
             embodiment=cfgs["embodiment"],
-            scene=PickAndPlaceScene(cfgs["background"], cfgs["object"]),
-            task=PickAndPlaceTask(),
+            scene=scene,
```
viiik-inside: Nit: Maybe its nicer to keep the full definition here - `scene = Scene(assets=[cfgs["background"], cfgs["object"]])`

alexmillane: done

PR #29, opened 2025-09-10
Automatic Object Type Detection

alexmillane:
## Summary
Adds functionality to automatically detect the type of an `Object`.

## Detailed description
- This means that we can generate `Objects` with a usd path alone without knowing anything more about the asset a-priori.
- **Limitations:** This method only works if there is a single rigid-body or articulation at the shallowest depth on the prim tree, in which one of these occurs.

### DIFF
--- a/isaac_arena/assets/object.py
+++ b/isaac_arena/assets/object.py
@@ -17,8 +17,6 @@ from isaaclab.assets import ArticulationCfg, AssetBaseCfg, RigidObjectCfg
 from isaaclab.sim.spawners.from_files.from_files_cfg import UsdFileCfg
 
 from isaac_arena.assets.object_base import ObjectBase, ObjectType
-
-# from isaac_arena.assets.object_base import ObjectType
 from isaac_arena.assets.object_utils import detect_object_type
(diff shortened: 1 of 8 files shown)


isaac_arena/assets/object.py
```diff
@@ -25,22 +26,34 @@ class Object(ObjectBase):
...
+        scale: tuple[float, float, float] = (1.0, 1.0, 1.0),
         initial_pose: Pose | None = None,
         **kwargs,
     ):
-        assert self.name is not None and self.usd_path is not None
-        super().__init__(name=self.name, object_type=self.object_type, **kwargs)
+        assert name is not None and usd_path is not None
+        # Detect object type if not provided
+        if object_type is None:
+            object_type = detect_object_type(usd_path=usd_path)
```
viiik-inside, line 42: should we just do it always?

alexmillane: I was wondering that. I'm not sure basically. I think that there may be assets which have articulations and rigid objects where that might fail. Basically I'm going to start using it everywhere. We may eventually retire the explicit specification.

PR #34, opened 2025-09-15
Add example environment with the lightwheel kitchen.

alexmillane:
## Summary
Adds an example environment with the lightwheel kitchen where the task is to deliver the pot to the stove.

## Detailed description
- I put one of the lightwheel kitchen USDs on the nucleus and added this example such that we can work in the kitchen, without `lwlab` such that we're not blocked by outages in the lightwheel API.
- _NOTE: We should now ship this example. We can't ship lightwheels kitchen_
- Runnable with: `python policy_runner.py --policy_type zero_action lightwheel_kitchen_pot_pick_and_place`

### DIFF
--- a/isaac_arena/assets/background_library.py
+++ b/isaac_arena/assets/background_library.py
@@ -83,7 +83,7 @@ class LightwheelKitchenBackground(Background):
     tags = ["background"]
     default_robot_initial_pose = Pose.identity()
     usd_path = "omniverse://isaac-dev.ov.nvidia.com/Projects/isaac_arena/lightwheel_assets_for_deletion/robocasakitchen-4-2/scene.usd"
-    initial_pose = Pose(position_xyz=(-1.20, 1.70, -0.87), rotation_wxyz=(1.0, 0, 0, 0))
+    initial_pose = Pose(position_xyz=(-1.20, 1.70, -0.92), rotation_wxyz=(1.0, 0, 0, 0))
     object_min_z = -0.2
 
     def __init__(self, robot_initial_pose: Pose = default_robot_initial_pose):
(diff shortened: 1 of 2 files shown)


isaac_arena/assets/background_library.py
```diff
@@ -66,3 +66,26 @@ class GalileoBackground(Background):
...
+@register_asset
+class LightwheelKitchenBackground(Background):
+    """
+    Encapsulates the background scene for the lightwheel kitchen.
+    """
+
+    name = "lightwheel_kitchen"
+    tags = ["background"]
+    default_robot_initial_pose = Pose.identity()
+    usd_path = "omniverse://isaac-dev.ov.nvidia.com/Projects/isaac_arena/lightwheel_assets_for_deletion/robocasakitchen-4-2/scene.usd"
```
peterd-NV, line 85: The nucleus drive is private to NV internal but I think at some point prior to releases (e.g. Isaac Lab 2.3) the contents on the drive are automatically synced to a staging S3 bucket which is publicly accessible. Might be good to check with @sheikh-nv just to make sure the contents in the `isaac_arena` are excluded from this.  Just wanted to note that in case you weren't aware.

alexmillane: Hmm, I was certainly not aware of this.

In the past the contents of our `nvblox` folder `omniverse://isaac-dev.ov.nvidia.com/Projects/nvblox/` haven't been shipped (at least not that I'm aware of). We had to move them to some staging ground before they were ship. But I'll be honest, I'm not super confident.

Perhaps @viiik-inside knows better.

For now I'll merge, and we can investigate this in parallel.

PR #35, opened 2025-09-16
Few improvements to docker setup

myurasov-nv:
## Summary
Changes to `run_docker.sh` to improve user experience running docker setup.

## Changes
* -r flag for forcing the image rebuilding (otherwise existing one is launched, which is fine for testing)
* -v for verbose mode (useful for seeing the resulting docker run command)
* Add some color and [Isaac Arena] to bash prompt inside container so it's obvious we're in
* Map Omniverse config dir so we have connection to the dev Nucleus via token obtained with SSO login (OMNI_PASS is considered deprecated afaik)
* Minor refactoring: move files that are copied to docker rather than ran directly to separate folder; extract entry point code to entrypoint.sh.

## Impact
Should be backwards-compatible, no breaking changes

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,2 @@
+    [mypy]
+    mypy_path=submodules/IsaacLab/source/isaaclab


docker/Dockerfile.isaac_arena
```diff
@@ -18,7 +18,7 @@ RUN if [ -e "/usr/share/vulkan" ] && [ -e "/etc/vulkan" ]; then \
 # Optional: Install CUDA 12.8
 ################################
 # Copy CUDA 12.8 installation script
-COPY docker/install_cuda.sh /tmp/install_cuda.sh
+COPY docker/res/install_cuda.sh /tmp/install_cuda.sh
```
alexmillane: Is `res` an abbreviation for something? Consider renaming `res` to the full word :).

docker/run_docker.sh
```diff
@@ -176,9 +145,20 @@ else
...
 
+    # map omniverse auth or config so we have connection to the dev nucleus
+    if [ -n "$OMNI_PASS" ]; then
+        DOCKER_RUN_ARGS+=("--env" "OMNI_USER=\$omni-api-token")
+        DOCKER_RUN_ARGS+=("--env" "OMNI_PASS=$OMNI_PASS")
+    else
+        if [ -d "$HOME/.nvidia-omniverse" ]; then
+            DOCKER_RUN_ARGS+=("-v" "$HOME/.nvidia-omniverse:/home/$(id -un)/.nvidia-omniverse")
+        fi
+    fi
```
alexmillane, line 158: Amazing! Thanks!

docker/run_docker.sh
```diff
@@ -69,68 +82,24 @@ done
...
+    echo "Docker image $DOCKER_IMAGE_NAME already exists. Not rebuilding."
+    echo "Use -r option to force the rebuild."
+else
+    docker build --pull \
+        --build-arg INSTALL_GROOT=$INSTALL_GROOT \
+        --build-arg GROOT_DEPS_GROUP=$GROOT_DEPS_GROUP \
+        -t ${DOCKER_IMAGE_NAME} \
+        --file $SCRIPT_DIR/Dockerfile.isaac_arena \
+        $SCRIPT_DIR/..
+fi
```
alexmillane, line 102: Thanks!

docker/Dockerfile.isaac_arena
myurasov-nv: I meant "resources", which is a bit generic I guess. How about "setup" so it's a subtle hint that it is needed during setup? :)

alexmillane: great!

Merge at will!

alexmillane: Ah I see, workflows don't run on your MRs. I just triggered one. I guess potentially you also don't have rights to merge?

Let's talk offline, we can add you as a developer if your forsee making more contributions.

PR #37, opened 2025-09-16
add onnxruntime as dockerfile deps

xyao-nv:
onnxruntime is needed for running WBC policy, type_extension is needed for run torch. Two deps are small, so installation in base dockerfile.

### DIFF
--- a/docker/Dockerfile.isaac_arena
+++ b/docker/Dockerfile.isaac_arena
@@ -63,9 +63,9 @@ RUN /isaac-sim/python.sh -m pip install --upgrade pip && \
   /isaac-sim/python.sh -m pip install \
   pytest \
   jupyter \
-  # NOTE(xinjieyao, 2025-09-15): Needed for torch
+  # NOTE(xinjieyao, 2025-09-15): Needed for inheritance of torch nn.module, saving/loading model
   typing_extensions \
-  # NOTE(xinjieyao, 2025-09-15): Needed for onnx runtime policy
+  # NOTE(xinjieyao, 2025-09-15): Needed for WBC running onnx compiled policy
   onnxruntime
 
 # Install isaac-arena


docker/Dockerfile.isaac_arena
```diff
@@ -62,7 +62,11 @@ RUN ${ISAACLAB_PATH}/isaaclab.sh -i
 RUN /isaac-sim/python.sh -m pip install --upgrade pip && \
   /isaac-sim/python.sh -m pip install \
   pytest \
-  jupyter
+  jupyter \
+  # NOTE(xinjieyao, 2025-09-15): Needed for torch
+  typing_extensions \
```
alexmillane: Suggestion to alter the comment slightly. The comment at the moment indicates that if you remove this, torch will break, which isn't the case (?). Suggestion to modify to say

"Needed for type hinting in pytorch"

docker/Dockerfile.isaac_arena
```diff
@@ -62,7 +62,11 @@ RUN ${ISAACLAB_PATH}/isaaclab.sh -i
...
+  # NOTE(xinjieyao, 2025-09-15): Needed for onnx runtime policy
+  onnxruntime
```
alexmillane, line 69: Should this be in the optional gr00t dependencies? Or do we forsee needing anyway?

alexmillane: Ok I think I see now. This is actually needed for the WBC, which we probably want in the base deps so we can do loco-manipulation teleop. Is that right?

Then it's all good! Bring it in! Maybe modify the comment just add that we're bringing it in for the Whole Body Controller (WBC)

xyao-nv: See updated comments.

docker/Dockerfile.isaac_arena
xyao-nv: One example could be, it's for any class inheriting torch.nn.module and running saving/loading ckpts. Added comments.

PR #38, opened 2025-09-16
Add optional camera registration

viiik-inside:
## Summary
Add the option of registering cameras and the corresponding observation to the environment.

### DIFF
--- a/isaac_arena/tests/test_camera_observation.py
+++ b/isaac_arena/tests/test_camera_observation.py
@@ -72,6 +72,8 @@ def _test_camera_observation(simulation_app) -> bool:
             # Make sure the camera observation contains values other than 0
             assert camera_observation.any() != 0, "Camera observation contains only 0s"
 
+    # Close the environment.
+    env.close()
     return True
 
 


isaac_arena/cli/isaac_arena_cli.py
```diff
@@ -43,3 +43,4 @@ def add_isaac_lab_cli_args(parser: argparse.ArgumentParser) -> None:
         help="Disable Pinocchio.",
     )
     isaac_lab_group.add_argument("--mimic", action="store_true", default=False, help="Enable mimic environment.")
+    isaac_lab_group.add_argument("--camera_tag", type=str, default=None, help="Name of the camera tag.")
```
alexmillane: Consider making this `help` string more helpful. It's not clear to me what a camera tag is.

isaac_arena/utils/cameras.py
```diff
@@ -0,0 +1,115 @@
...
+from copy import deepcopy
+from typing import Any
+
+from isaaclab.envs import mdp
+from isaaclab.managers import ObservationGroupCfg as ObsGroup
+from isaaclab.managers import ObservationTermCfg as ObsTerm
+from isaaclab.managers import SceneEntityCfg
+
+# if you already have these utilities in your repo, reuse them
+from isaac_arena.utils.configclass import make_configclass
```
alexmillane: I'm not sure what this means? I would suggest removing this comment? Or maybe I'm missing something.

isaac_arena/embodiments/gr1t2/gr1t2.py
```diff
@@ -114,20 +115,29 @@ class GR1T2SceneCfg:
...
-    #             rot=(0.62, 0.32, -0.32, -0.63),
-    #             convention="opengl",
-    #         ),
-    #     )
+
+@configclass
+class GR1T2CameraCfg:
+    """Configuration for cameras."""
+
+    observation_cameras: dict = {
```
alexmillane: why "observation" cameras? Are we expecting another type of cameras? Otherwise change it to `cameras`

isaac_arena/tests/test_camera_observation.py
```diff
@@ -0,0 +1,83 @@
...
+    # Compile an IsaacLab compatible arena environment configuration
+    builder = ArenaEnvBuilder(isaac_arena_environment, args_cli)
+    env = builder.make_registered()
+    env.reset()
+
+    for _ in tqdm.tqdm(range(NUM_STEPS)):
+        with torch.inference_mode():
+            actions = torch.zeros(env.action_space.shape, device=env.device)
+            obs, _, _, _, _ = env.step(actions)
+            print("obs", obs)
```
alexmillane: I presume that there are camera images inside obs? Could we perform a simple test on those images? For example that they have the correct shape and the pixels are not all black?

alexmillane: Also consider making the print statement a bit more informative. "Get an observation containing an image of size: ..."

isaac_arena/tests/utils/subprocess.py
```diff
@@ -42,17 +42,24 @@ def run_subprocess(cmd, env=None):
...
     # Simulation app args. For now, we just make these default + headless.
     # TODO(alexmillane, 2025.09.01): We're eventually going to want a way to override the args.
     parser = get_isaac_arena_cli_parser()
     simulation_app_args = parser.parse_args([])
     simulation_app_args.headless = headless
+    simulation_app_args.enable_cameras = enable_cameras
     # Launch the simulator
     with SimulationAppContext(simulation_app_args) as simulation_app:
-        # Run the function
+        # Run the functionWW
```
alexmillane: remove `WW`

isaac_arena/cli/isaac_arena_cli.py
alexmillane: Ok we're using this for teleop vs. not teleop? Do we forsee any other camera tags?

If not I would move this to a binary distinction. I'm wondering if this should be a command-line argument at all. Basically if the `IsaacArenaEnvironment.teleop_device` is not none, we're in teleop mode and environment compilation should switch to teleop mode.

I would suggest that we remove this flag entirely.

isaac_arena/embodiments/gr1t2/gr1t2.py
```diff
@@ -114,20 +115,29 @@ class GR1T2SceneCfg:
...
+                spawn=sim_utils.PinholeCameraCfg(focal_length=18.15, clipping_range=(0.01, 1.0e5)),
+                offset=CameraCfg.OffsetCfg(
+                    pos=(0.12515, 0.0, 0.06776),
+                    rot=(0.62, 0.32, -0.32, -0.63),
+                    convention="opengl",
+                ),
+            ),
+            "tags": ["teleop"],
+        },
+    }
```
alexmillane: This structure, of a dict of tagged cameras, make the environment compilation complicated later on.

I would suggest that we do this a simpler way. We make two `GR1T2CameraCfg` structs:
- `GR1T2CameraCfg`
- `GR1T2TeleopCameraCfg`

These are normal `isaaclab`  `configclass`es. `CameraCfg`s on the first level. And we just return the right one, based on a flag `teleop: bool = False` passed to `get_camera_cfg()` by the `ArenaEnvBuilder`

isaac_arena/utils/cameras.py
```diff
@@ -0,0 +1,115 @@
...
+        observation_dict[name] = meta
+        # each camera becomes a field on the Scene config
+        fields_spec.append((name, type(cam_cfg), cam_cfg))
+
+    if not fields_spec:
+        return make_configclass("EmptyCamerasSceneCfg", [])(), make_configclass("EmptyCameraObsCfg", [])()
+
+    CamerasSceneCfg = make_configclass("CamerasSceneCfg", fields_spec)
+
+    return CamerasSceneCfg(), make_camera_observations_cfg(registered_cameras=observation_dict)
```
alexmillane: I think we can get rid of this function entirely by removing the camera tags, and moving the logic into the embodiments. I think we'll end up with a system which much more closely resembles the other configclass combination we do in `compile_env`

isaac_arena/environments/compile_env.py
```diff
@@ -44,12 +45,25 @@ def compose_manager_cfg(self) -> tuple[IsaacArenaManagerBasedRLEnvCfg, str | typ
...
+        camera_scene_cfg, camera_observation_cfg = add_camera_to_environment_cfg(
+            self.arena_env.embodiment.get_camera_cfg(),
+            enable_cameras=self.args.enable_cameras,
+            tag=self.args.camera_tag,
+        )
+        observation_cfg = combine_configclass_instances(
+            "ObservationCfg",
+            self.arena_env.embodiment.get_observation_cfg(),
+            camera_observation_cfg,
+        )
```
alexmillane: See above an below. This sort of break the pattern that we have elsewhere. I think we can get around this breakage by passing a flag to `self.arena_env.embodiment.get_camera_cfg()`, namely `teleop:bool = False`

isaac_arena/embodiments/gr1t2/gr1t2.py
viiik-inside: so ideally we would not need a camera during teleop. We would need it when generating data or in RL during the training loop. We might have various different settings in the same environment, for example when using Franka, we can have two or three different settings - single cam, multi cam etc. And in this case i thought it would be nicer to use a tag instead of a conditional statement.

isaac_arena/embodiments/gr1t2/gr1t2.py
viiik-inside: fair point

isaac_arena/environments/compile_env.py
viiik-inside: I'll try to get this from embodiment.

isaac_arena/tests/test_camera_observation.py
viiik-inside: yup

isaac_arena/utils/cameras.py
viiik-inside: removed

isaac_arena/utils/cameras.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def make_camera_observation_cfg(
+    camera_cfg: Any,
+    normalize: bool = False,
+):
+    """
+    Build a configclass instance that adds one ObsTerm per selected camera.
+    The SceneEntity name equals the camera field name used in the Scene.
+    """
```
alexmillane: Expand the comment.

PR #43, opened 2025-09-18
Few improvements to docker setup and dev environment

myurasov-nv:
Creating MR from repo branch rather than fork so that workflows work.

## Changes
1. -r (rebuild) and -v (verbose) flags for run_docker.sh
1. Add some color and [Isaac Arena] to bash prompt inside container so it's obvious we're in
1. Map Omniverse config dir so we have connection to the dev Nucleus via token obtained with SSO login
1. Minor refactoring: move files that are copied to docker rather than ran directly to separate folder; extract entry point code to entrypoint.sh.
1. Add support for `debygpy` (simply set breakpoints and run your files with `debugpy` instead of `python`)
1. Misc fixes and things nice to have :)

## Impact
Should be backwards-compatible, no breaking changes

### DIFF
--- a/docker/setup/entrypoint.sh
+++ b/docker/setup/entrypoint.sh
@@ -45,8 +45,10 @@ fi
 # change prompt so it's obvious we're inside the arena container
 echo "PS1='[Isaac Arena] \[\e[0;32m\]~\u \[\e[0;34m\]\w\[\e[0m\] \$ '" >> /home/$DOCKER_RUN_USER_NAME/.bashrc
 
-# useful aliases
+# useful aliases:
+# list files as a table, with colors, show hidden, append indicators (/ for dirs, * for executables, @ for symlinks)
 echo "alias ll='ls -alF --color=auto'" >> /home/$DOCKER_RUN_USER_NAME/.bashrc
+# go one level up
 echo "alias ..='cd ..'" >> /home/$DOCKER_RUN_USER_NAME/.bashrc
 
 set +x


docker/Dockerfile.isaac_arena
```diff
@@ -90,5 +99,15 @@ RUN echo "alias python='/isaac-sim/python.sh'" >> /etc/bash.bashrc
...
 
+# Debugging with debugpy
+# Usage:
+# 1) Set your breakpoints
+# 2) Start the script you want to debug with "debugpy" instead of "python".
+#    It will pause waiting for the debugger to attach.
+# 3) Attach to the running container with VSCode using the "Attach to debugpy session"
+#    configuration from the Run and Debug panel.
+RUN pip3 install debugpy
+RUN echo "alias debugpy='python -Xfrozen_modules=off -m debugpy --listen localhost:5678 --wait-for-client'" >> /etc/bash.bashrc
```
alexmillane, line 110: Nice!

docker/run_docker.sh
```diff
@@ -40,17 +49,21 @@ while getopts ":d:m:e:hn:g:G:" OPTION; do
             echo "Helper script to build $DOCKER_IMAGE_NAME (default)"
             echo "Usage:"
             echo "$script_name -h"
+            echo "$script_name -v"
```
alexmillane: Maybe just add a string to indicate that this triggers verbose output in `run_docker.sh`

docker/setup/entrypoint.sh
```diff
@@ -0,0 +1,59 @@
...
+if [ ! -d "/workspaces/isaac_arena/submodules/IsaacLab/_isaac_sim" ]; then
+    ln -s /isaac-sim/ /workspaces/isaac_arena/submodules/IsaacLab/_isaac_sim
+fi
+
+# change prompt so it's obvious we're inside the arena container
+echo "PS1='[Isaac Arena] \[\e[0;32m\]~\u \[\e[0;34m\]\w\[\e[0m\] \$ '" >> /home/$DOCKER_RUN_USER_NAME/.bashrc
+
+# useful aliases
+echo "alias ll='ls -alF --color=auto'" >> /home/$DOCKER_RUN_USER_NAME/.bashrc
+echo "alias ..='cd ..'" >> /home/$DOCKER_RUN_USER_NAME/.bashrc
```
alexmillane: Request to add a comment to what these command do. (maybe I'm slow but It's not obvious to me).

docker/Dockerfile.isaac_arena
myurasov-nv: Added, help message is now like this:
```
$ docker/run_docker.sh -h
Helper script to build and Isaac Arena docker environment.

Usage:
run_docker.sh [options]

Options:
  -v (Verbose output)
  -d <datasets directory> (Path to datasets on the host. Default is "/home/myurasov/datasets".)
  -m <models directory> (Path to models on the host. Default is "/home/myurasov/models".)
  -e <evaluation directory> (Path to evaluation data on the host. Default is "/home/myurasov/eval".)
  -n <docker name> (Name of the docker image that will be built or used. Default is "isaac_arena".)
  -r (Force rebuilding of the docker image.)
  -g (Install GR00T with base dependencies.)
  -G <deps_group> (Install GR00T with dependency group: base, dev, orin, thor, deploy.)
```

PR #55, opened 2025-09-23
add optional root-user-action=ignore for pip command

xyao-nv:
pip install --root-user-action is available only for pip>=22.1. The pip version inside docker does not guarantee to be 22.1, pip from sim uses 21.0 from my observation.

Set it available only if pip >=22.1.


docker/setup/entrypoint.sh
myurasov-nv, line 1: Really nice! https://github.com/isaac-sim/isaac_arena/pull/43/commits/f6f17668bb87d81c9a6174389e8322334a7e041d ensures we have newest pip, but in case it’s a prebuilt image this check makes sense.

PR #57, opened 2025-09-24
Feature/mimic

viiik-inside:
## Summary
Enable mimic for articulation objects and multiple embodiments

## Detailed description
- Fix a bug in gr1 mimic configuration
- Enable mimic for articualtion objects

### DIFF
--- a/isaac_arena/embodiments/franka/franka.py
+++ b/isaac_arena/embodiments/franka/franka.py
@@ -37,6 +37,7 @@ from isaaclab_tasks.manager_based.manipulation.stack.mdp import franka_stack_eve
 from isaaclab_tasks.manager_based.manipulation.stack.mdp.observations import ee_frame_pos, ee_frame_quat, gripper_pos
 
 from isaac_arena.assets.register import register_asset
+from isaac_arena.embodiments.common.mimic_utils import get_rigid_and_articulated_object_poses
 from isaac_arena.embodiments.embodiment_base import EmbodimentBase
 from isaac_arena.geometry.pose import Pose
 
@@ -309,10 +310,10 @@ class FrankaMimicEnv(ManagerBasedRLMimicEnv):
         # last dimension is gripper action
         return {list(self.cfg.subtask_configs.keys())[0]: actions[:, -1:]}
 
-    # Have to implement this to consider articulated objects as well
+    # Implemented this to consider articulated objects as well
     def get_object_poses(self, env_ids: Sequence[int] | None = None):
         """
-        Gets the pose of each object relevant to Isaac Lab Mimic data generation in the current scene.
+        Gets the pose of each object(rigid and articulated) in the current scene.
         Args:
             env_ids: Environment indices to get the pose for. If None, all envs are considered.
         Returns:
@@ -323,15 +324,6 @@ class FrankaMimicEnv(ManagerBasedRLMimicEnv):
 
         state = self.scene.get_state(is_relative=True)
 
-        def pose_from(obj_state) -> "torch.Tensor":
-            rp = obj_state["root_pose"][env_ids]  # (..., 7): [x,y,z, qx,qy,qz,qw]
-            pos, quat = rp[..., :3], rp[..., 3:7]
-            return PoseUtils.make_pose(pos, PoseUtils.matrix_from_quat(quat))  # (..., 4, 4)
-
-        groups = ("rigid_object", "articulation")
-
-        object_pose_matrix = {
-            name: pose_from(obj_state) for group in groups for name, obj_state in state.get(group, {}).items()
-        }
+        object_pose_matrix = get_rigid_and_articulated_object_poses(state, env_ids)
 
         return object_pose_matrix
--- a/isaac_arena/embodiments/gr1t2/gr1t2.py
+++ b/isaac_arena/embodiments/gr1t2/gr1t2.py
@@ -34,6 +34,7 @@ from isaaclab_assets.robots.fourier import GR1T2_CFG
 from isaaclab_tasks.manager_based.manipulation.pick_place.pickplace_gr1t2_env_cfg import ActionsCfg as GR1T2ActionsCfg
 
 from isaac_arena.assets.register import register_asset
+from isaac_arena.embodiments.common.mimic_utils import get_rigid_and_articulated_object_poses
 from isaac_arena.embodiments.embodiment_base import EmbodimentBase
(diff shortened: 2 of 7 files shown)


isaac_arena/embodiments/franka/franka.py
```diff
@@ -308,3 +308,30 @@ def actions_to_gripper_actions(self, actions: torch.Tensor) -> dict[str, torch.T
...
+        def pose_from(obj_state) -> "torch.Tensor":
+            rp = obj_state["root_pose"][env_ids]  # (..., 7): [x,y,z, qx,qy,qz,qw]
+            pos, quat = rp[..., :3], rp[..., 3:7]
+            return PoseUtils.make_pose(pos, PoseUtils.matrix_from_quat(quat))  # (..., 4, 4)
+
+        groups = ("rigid_object", "articulation")
+
+        object_pose_matrix = {
+            name: pose_from(obj_state) for group in groups for name, obj_state in state.get(group, {}).items()
+        }
```
alexmillane: This is quite difficult to read because it's a double dictionary comprehension. Consider splitting across multiple lines for readability.

isaac_arena/embodiments/franka/franka.py
```diff
@@ -308,3 +308,30 @@ def actions_to_gripper_actions(self, actions: torch.Tensor) -> dict[str, torch.T
         """
         # last dimension is gripper action
         return {list(self.cfg.subtask_configs.keys())[0]: actions[:, -1:]}
+
+    # Have to implement this to consider articulated objects as well
```
alexmillane: Suggestion to remove this comment. From what I can tell, we've now implemented this for articulations.

isaac_arena/embodiments/franka/franka.py
```diff
@@ -308,3 +308,30 @@ def actions_to_gripper_actions(self, actions: torch.Tensor) -> dict[str, torch.T
...
+    def get_object_poses(self, env_ids: Sequence[int] | None = None):
+        """
+        Gets the pose of each object relevant to Isaac Lab Mimic data generation in the current scene.
```
alexmillane: Suggestion to update the name of this function and the corresponding comment.

`get_rigid_body_and_articulation_poses`

I think the comment is slightly wrong; it doesn't just get the mimic-relavent objects right? It gets the poses of all rigid-bodies and all articulations in the scene.

isaac_arena/embodiments/gr1t2/gr1t2.py
```diff
@@ -305,3 +305,30 @@ def actions_to_gripper_actions(self, actions: torch.Tensor) -> dict[str, torch.T
...
+            pos, quat = rp[..., :3], rp[..., 3:7]
+            return PoseUtils.make_pose(pos, PoseUtils.matrix_from_quat(quat))  # (..., 4, 4)
+
+        groups = ["rigid_object", "articulation"]
+
+        object_pose_matrix = {
+            name: pose_from(obj_state) for group in groups for name, obj_state in state.get(group, {}).items()
+        }
+
+        return object_pose_matrix
```
alexmillane: This function is an exact copy of the one above no? Suggestion to put it somewhere common. Maybe `isaac_arena/embodiments/common/mimic_utils.py`

isaac_arena/examples/example_environments/kitchen_pick_and_place_environment.py
```diff
@@ -54,6 +55,7 @@ def get_env(self, args_cli: argparse.Namespace):  # -> IsaacArenaEnvironment:
             name="destination_location",
             prim_path="{ENV_REGEX_NS}/kitchen/Cabinet_B_02",
             parent_asset=background,
+            object_type=ObjectType.RIGID,
```
alexmillane, line 59: Could you add the following comment above this line:

```
TODO(alexmillane, 2025.09.24): Add automatic object type detection of ObjectReferences.
```

isaac_arena/tasks/open_door_task.py
```diff
@@ -73,3 +77,120 @@ def __init__(self, openable_object: Openable):
...
+
+    embodiment_name: str = "franka"
+
+    openable_object_name: str = "openable_object"
+
+    def __post_init__(self):
+        # post init of parents
+        super().__post_init__()
+        # # TODO: Figure out how we can move this to the MimicEnvCfg class
+        # # The __post_init__() above only calls the init for FrankaCubeStackEnvCfg and not MimicEnvCfg
```
alexmillane: This comment is no longer true no? We only have a single parent class?

isaac_arena/tasks/pick_and_place_task.py
```diff
@@ -196,7 +196,7 @@ def __post_init__(self):
                 # TODO(alexmillane, 2025.09.02): This is currently broken. FIX.
                 # We need a way to pass in a reference to an object that exists in the
                 # scene.
-                object_ref="destination_object",
+                object_ref="destination_location",
```
alexmillane, line 196: In a later MR we should pass in the name of the destination location. Right now it's hardcoded, which will break pretty much immediately when anyone uses it on another scene. It is now possible (because I added object references) to fix this.

Maybe let's get a test going, and then fix this in a follow up MR.

isaac_arena/embodiments/franka/franka.py
viiik-inside: okay its misleading. I wanted to tell that I am overriding the base class version to consider articulations. I will rewrite the comment to Implemented this to consider blah blah

isaac_arena/embodiments/franka/franka.py
viiik-inside: but get object poses is used everywhere in Isaaclab core functions. So I would let this be the name of the function. We override it from the base class

isaac_arena/embodiments/franka/franka.py
viiik-inside: black reformats it. I'll try to avoid this

isaac_arena/embodiments/gr1t2/gr1t2.py
viiik-inside: done.

isaac_arena/tasks/open_door_task.py
viiik-inside: true

isaac_arena/tasks/pick_and_place_task.py
viiik-inside: yup

PR #65, opened 2025-09-25
bring configurations to init and cleanup

viiik-inside:
## Summary
bring all hidden configurations to init so its user accessible.

Clean up code and move around for it make sense

### DIFF
--- a/isaac_arena/affordances/openable.py
+++ b/isaac_arena/affordances/openable.py
@@ -18,7 +18,6 @@ from isaaclab.envs.manager_based_env import ManagerBasedEnv
 from isaaclab.managers import SceneEntityCfg
 
 from isaac_arena.affordances.affordance_base import AffordanceBase
-
 from isaac_arena.utils.joint_utils import get_normalized_joint_position, set_normalized_joint_position
 
 


isaac_arena/embodiments/embodiment_base.py
```diff
@@ -69,7 +69,7 @@ def get_observation_cfg(self) -> Any:
                 )
         return self.observation_config
 
-    def get_event_cfg(self) -> Any:
+    def get_events_cfg(self) -> Any:
         return self.event_config
```
alexmillane, line 77: Thanks for changing this.

Could I request another change? Could we add `@abstractmethod`s to populate the required members? I.e. something like:

```
def __init__(self):
    self.scene_config = self._make_scene_cfg()
    ...

@abstractmethod
def _make_scene_cfg(self) -> Any:
    pass
```

This will have the type system enforce what we expect: that these members exist for all embodiments.

isaac_arena/assets/asset_registry.py
```diff
@@ -13,14 +13,13 @@
...
 
-from isaac_arena.assets.asset import Asset
-from isaac_arena.teleop_devices.teleop_device_base import TeleopDeviceBase
 from isaac_arena.utils.singleton import SingletonMeta
 
-# TODO(alexmillane, 2025.09.19): Before release we should sort out this
-# circular import issue.
+if TYPE_CHECKING:
+    from isaac_arena.assets.asset import Asset
+    from isaac_arena.teleop_devices.teleop_device_base import TeleopDeviceBase
```
alexmillane, line 22: I wonder/hope that this solves all our circular import issues :sweat_smile:

isaac_arena/scene/scene.py
```diff
@@ -28,6 +28,9 @@ class Scene:
 
     def __init__(self, assets: list[Asset] | None = None):
         self.assets: dict[str, Asset] = {}
+        self.observation_cfg = None
+        self.events_cfg = None
+        self.termination_cfg = None
```
alexmillane, line 34: Is the idea here that someone may want observations, events, and terminations associated with the scene, but by default we dont have any? If so consider adding a comment to indicate this; that we are initializing these members to give users a chance to override them from the outside.

isaac_arena/scene/scene.py
```diff
@@ -51,10 +54,10 @@ def get_scene_cfg(self) -> Any:
...
-        pass
+        return self.observation_cfg
 
     def get_events_cfg(self) -> Any:
-        pass
+        return self.events_cfg
 
     def get_termination_cfg(self) -> Any:
-        pass
+        return self.termination_cfg
```
alexmillane, line 64: Should we move these functions to the base class to match what we have in the embodiment (and also to avoid repeating them in every task).

isaac_arena/tasks/open_door_task.py
```diff
@@ -35,11 +35,17 @@ def __init__(
...
+        self.termination_cfg = self.make_termination_cfg()
 
     def get_scene_cfg(self):
-        pass
+        return self.scene_config
 
     def get_termination_cfg(self):
+        return self.termination_cfg
+
+    def make_termination_cfg(self):
```
alexmillane, line 48: Similar to my comment in embodiments. We move populating of the members (like `self.events_cfg`) to the base class and then enforce initialization in the constructor by having `@abstractmethod`s which handle the construction of the sub-objects.

isaac_arena/assets/asset_registry.py
viiik-inside: most of it. I had to add another type checking and it seems to have solved all of them

isaac_arena/embodiments/embodiment_base.py
viiik-inside: yup

isaac_arena/scene/scene.py
viiik-inside: yes, will add a comment

isaac_arena/tasks/open_door_task.py
viiik-inside: ill do this in a separate PR. Its getting big

PR #67, opened 2025-09-25
install ffmpeg for ffprobe for gr00t policy docker

xyao-nv:
`ffprobe` from `ffmpeg` is needed for hdf5 to lerobot data format conversion. Install it in the optional gr00t deps install script.


docker/setup/install_gr00t_deps.sh
```diff
@@ -18,6 +18,10 @@ echo "PATH=$PATH"
 echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
 echo "TORCH_CUDA_ARCH_LIST=$TORCH_CUDA_ARCH_LIST"
 
+# Installing dpes for system-level media libraries
```
alexmillane: nit: dpes -> deps

PR #68, opened 2025-09-25
Add tool to convert hdf5 to Lerobot

xyao-nv:
Convert dataset from MimicGen HDF5, to GR00T-Lerobot compatible format.
(Migrate and refactor from IssacLabEvalTask Repo, with expansion to gr1-locomanip)

1. Add configs for g1, gr1 specifying joint orders and joint groups
2. Add data utils tools to support image/data preprocessing
3. Add workflow cfg yaml to support g1-locomanip, and gr1-tabletopmanip data conversion
4. Add video processing & waiting mechanisms to prevent corruption or reading incomplete/non-existent files.

Make sure you are inside the gr00t-deps installed docker
(`./docker/run_docker.sh -g true  -n arena_cuda_gr00t`)

`python isaac_arena/policy/data_utils/convert_hdf5_to_lerobot.py --yaml_file isaac_arena/policy/config/g1_locomanip_config.yaml`

`python isaac_arena/policy/data_utils/convert_hdf5_to_lerobot.py --yaml_file isaac_arena/policy/config/gr1_tabletopmanip_config.yaml`


isaac_arena/policy/config/dataset_config.py
```diff
@@ -0,0 +1,206 @@
...
+        default=Path("/datasets/"),
+        metadata={"description": "Root directory for all data storage."},
+    )
+    language_instruction: str = field(
+        default="", metadata={"description": "Instruction given to the policy in natural language."}
+    )
+    hdf5_name: str = field(default="", metadata={"description": "Name of the HDF5 file to use for the dataset."})
+
+    # Mimic-generated HDF5 datafield
+    # NOTE(xinjieyao, 2025-09-25): robot joint position must exist in the HDF5 file
```
cvolkcvolk, line 33: Could we add a check somewhere that throws an error warning if the joint positions are not present instead of this Note?

isaac_arena/policy/config/dataset_config.py
```diff
@@ -0,0 +1,206 @@
...
+    )
+    target_image_size: tuple[int, int, int] = field(
+        default=(480, 640, 3), metadata={"description": "Target size for images after resizing and padding."}
+    )
+
+    hdf5_file_path: Path = field(init=False)
+    lerobot_data_dir: Path = field(init=False)
+    task_index: int = field(
+        default=0, metadata={"description": "Task index for the task description in LeRobot file."}
+    )  # task index for the task description in LeRobot file
```
cvolkcvolk: Suggestion to remove the comment. Duplicates what the description already states :)

isaac_arena/policy/config/dataset_config.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def __post_init__(self):
+
+        self.hdf5_file_path = self.data_root / self.hdf5_name
+        self.lerobot_data_dir = self.data_root / self.hdf5_name.replace(".hdf5", "") / "lerobot"
+
+        # Assert all paths exist
+        assert self.hdf5_file_path.exists(), "hdf5_file_path does not exist"
```
cvolkcvolk: Suggestion to print the actual tested path. 
Eg 
`assert self.hdf5_file_path.exists(), f"{self.hdf5_file_path} does not exist."`

isaac_arena/policy/config/dataset_config.py
```diff
@@ -0,0 +1,206 @@
...
+        assert Path(self.gr00t_joints_config_path).exists(), "gr00t_joints_config_path does not exist"
```
cvolkcvolk: Same below

isaac_arena/policy/config/dataset_config.py
```diff
@@ -0,0 +1,206 @@
...
+        assert Path(self.action_joints_config_path).exists(), "action_joints_config_path does not exist"
+        assert Path(self.state_joints_config_path).exists(), "state_joints_config_path does not exist"
+        assert Path(self.info_template_path).exists(), "info_template_path does not exist"
+        assert Path(self.modality_template_path).exists(), "modality_template_path does not exist"
+        # if lerobot_data_dir not empty, throw a warning and remove
```
cvolkcvolk: You can remove those comments as most of the code is already self explanatory ;)

cvolkcvolk: As a rule of thumb, comments explain more the _why_ and less the _what_

isaac_arena/policy/config/dataset_config.py
```diff
@@ -0,0 +1,206 @@
...
+        if self.lerobot_data_dir.exists():
+            print(f"Warning: lerobot_data_dir {self.lerobot_data_dir} already exists. Removing it.")
+            # remove directory contents and the directory itself using shutil
```
cvolkcvolk:
```suggestion
```

isaac_arena/policy/config/dataset_config.py
```diff
@@ -0,0 +1,206 @@
...
+            shutil.rmtree(self.lerobot_data_dir)
```
cvolkcvolk: This might be a potential dangerous operation that is capable of nuking someones home directory.
Suggestion to ask for a user confirmation to delete it if it exists

isaac_arena/policy/config/dataset_config.py
```diff
@@ -0,0 +1,206 @@
...
+        # Prepare data keys for mimic-generated hdf5 file
+        # Minimum set of keys are state & action
+        self.hdf5_keys = {"state": self.state_name_sim, "action": self.action_name_sim}
+        # Optional keys if provided
+        if self.left_eef_pos_name_sim != "":
```
cvolkcvolk: Comparing to an empty string explicitly is not necessary as they are falsy in Python. And comparing to an empty string could cause subtle issues if the variable is _None_ for example.

You could just do
```
if self.left_eef_pos_name_sim:
    self.hdf5_keys["left_eef_pos"] = self.left_eef_pos_name_sim
```

isaac_arena/policy/config/dataset_config.py
```diff
@@ -0,0 +1,206 @@
...
+            self.hdf5_keys["left_eef_pos"] = self.left_eef_pos_name_sim
+        if self.left_eef_quat_name_sim != "":
```
cvolkcvolk: Same below

docker/setup/install_gr00t_deps.sh
```diff
@@ -18,6 +18,10 @@ echo "PATH=$PATH"
 echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
 echo "TORCH_CUDA_ARCH_LIST=$TORCH_CUDA_ARCH_LIST"
 
+# Installing dpes for system-level media libraries
```
cvolkcvolk:
```suggestion
# Installing deps for system-level media libraries
```

isaac_arena/tests/policy/test_convert_hdf5_to_lerobot.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # Clean up any existing output directory
+    if g1_ds_config.lerobot_data_dir.exists():
+
+        shutil.rmtree(g1_ds_config.lerobot_data_dir)
+
+    # Run conversion
+    convert_hdf5_to_lerobot(g1_ds_config)
+
+    # assert lerobot_data_dir is created
```
cvolkcvolk: Your assert code is expressive enough, suggestion to remove the assert commets here and below :)

isaac_arena/policy/data_utils/joints_conversion.py
```diff
@@ -0,0 +1,96 @@
...
+    for joint_name, joint_index in sim_joints_config.items():
+        match joint_name.split("_")[0]:
+            # For GR1
+            case "L":
+                joint_group = "left_hand"
+            # For GR1
+            case "R":
+                joint_group = "right_hand"
+            case "left":
+                # For G1
```
cvolkcvolk, line 72: This function looks quite specialized to work only for G1 and GR1 setups for it to live in a data_utils/ directory.

We could make this function generic in a way that it for example accepts a mapping dict. Instead of checking G1 and Gr1 specific "L", "R" "left" definitions here in the function body, we could have an expressive mapping somewhere defined for each robot. remap_policy_joints_to_sim_joints then only applies this mapping

isaac_arena/policy/data_utils/yaml_config_loader.py
```diff
@@ -0,0 +1,95 @@
...
+        return value
+
+    # Handle basic types
+    if field_type in (str, int, float, bool):
+        return field_type(value)
+
+    return value
+
+
+def load_config_from_yaml(yaml_path: str | Path) -> Gr00tDatasetConfig:
```
cvolkcvolk, line 46: Suggestion to rename.  `load_gr00t_dataset_config_from_yaml`

isaac_arena/policy/data_utils/robot_eef_pose.py
```diff
@@ -0,0 +1,64 @@
...
+import torch
+from dataclasses import dataclass
+
+
+@dataclass
+class EefPose:
+    eef_pos: torch.Tensor
+    """Eef position in meters"""
+
+    eef_quat: torch.Tensor  # wxyz
```
cvolkcvolk: Can remove comment in this line as its included in the docstrin below

isaac_arena/policy/config/g1/43dof_joint_space.yaml
```diff
@@ -0,0 +1,46 @@
...
+  "left_hand_index_1_joint": 35
+  "left_hand_middle_1_joint": 36
+  "left_hand_thumb_1_joint": 37
+  "right_hand_index_1_joint": 38
+  "right_hand_middle_1_joint": 39
+  "right_hand_thumb_1_joint": 40
+  "left_hand_thumb_2_joint": 41
+  "right_hand_thumb_2_joint": 42
+
+total_joints: 43
```
alexmillane, line 46: Are these numbers specific to the embodiment, or specific to how they're represented inside gr00t?

I notice the next file `gr00t_43dof_joint_space.yaml` seems to describe how gr00t represents the joints, which makes me thing that these numbers are related to the embodiment only? I wonder therefore if we could retrieve these numbers from the embodiment, as there seems to be similar lists of joints defined in the embodiment.

isaac_arena/policy/data_utils/io_utils.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    with open(file_path) as fp:
+        return json.load(fp, **kwargs)
+
+
+def load_robot_joints_config(yaml_path: str | Path) -> dict[str, Any]:
+    """Load robot joint configuration from YAML file"""
+    with open(yaml_path, encoding="utf-8") as f:
+        config = yaml.safe_load(f)
+    return config.get("joints", {})
```
alexmillane: This file seems to contain very general functions (great!) and then a very specific function which checks for joint on the top level and then returns the result. I would suggest moving the last function into a file which is robot/policy specific.

isaac_arena/policy/data_utils/robot_eef_pose.py
```diff
@@ -0,0 +1,64 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import torch
+from dataclasses import dataclass
+
+
+@dataclass
+class EefPose:
```
alexmillane, line 21: Suggestion that we use the `Pose` object which we already have in `isaac_arena`. From what I can tell, there's nothing about this `EefPose` which makes it specific for end effectors, so I believe we could use the existing on `Pose` object.

isaac_arena/policy/data_utils/robot_eef_pose.py
```diff
@@ -0,0 +1,64 @@
...
+    eef_pos: torch.Tensor
+    """Eef position in meters"""
+
+    eef_quat: torch.Tensor  # wxyz
+    """Eef orientation in quaternions w, x, y, z"""
+
+    eef_pose: torch.Tensor
+    """Eef pose in 6D format"""
```
alexmillane: This class contains an invariant which can be easily broken. For example the following code will generate a hard to find and confusing bug

```
pose.eef_pos = torch.tensor(0,0,0)
...
do more stuff
...
my_fuction(pose.get_eef_pose())
```

In general, invariants should be in priority order:
1) avoided
2) enforced, or
3) checked

I think here we can easily avoid it. Actually I added to our `isaac_arena` `Pose` object a function which returns what you want be doesn't introduce an invariant.

isaac_arena/policy/data_utils/robot_joints.py
```diff
@@ -0,0 +1,55 @@
...
+    @staticmethod
+    def zero(joint_order_config: dict[str, int], device: torch.device):
+        return JointsAbsPosition(
+            joints_pos=torch.zeros((len(joint_order_config)), device=device),
+            joints_order_config=joint_order_config,
+            device=device,
+        )
+
+    def to_array(self) -> torch.Tensor:
+        return self.joints_pos.cpu().numpy()
```
alexmillane, line 35: I think the return type annotation is wrong here. `-> np.ndarray`

isaac_arena/policy/data_utils/robot_joints.py
```diff
@@ -0,0 +1,55 @@
...
+@dataclass
+class JointsAbsPosition:
+    joints_pos: torch.Tensor
+    """Joint positions in radians"""
+
+    joints_order_config: dict[str, int]
+    """Joints order configuration"""
+
+    device: torch.device
+    """Device to store the tensor on"""
```
alexmillane: This also introduces a subtle invariant in the class. The invariant is `JointsAbsPosition.joint_pos.device == JointsAbsPosition.device`

It is easy for these two representations of the same piece of information (where the tensor is stored) to become inconsistent.

Is there anything to speak again removing `JointsAbsPosition.device` and relying instead on `JointsAbsPosition.joints_pos.device` to represent where that tensor is stored?

isaac_arena/tests/policy/test_convert_hdf5_to_lerobot.py
```diff
@@ -0,0 +1,90 @@
...
+    assert expected_columns.issubset(actual_columns), f"Missing columns: {expected_columns - actual_columns}"
+    # check parquet file data is the same as expected
+    assert actual_df.equals(expected_g1_parquet)
+
+    # remove lerobot_data_dir
+    shutil.rmtree(g1_ds_config.lerobot_data_dir.parent)
+
+
+if __name__ == "__main__":
+    test_g1_convert_hdf5_to_lerobot()
```
alexmillane, line 70: Thank you for adding a great test!

isaac_arena/policy/config/dataset_config.py
xyao-nv: done

isaac_arena/policy/config/dataset_config.py
xyao-nv: added user confirmation input. done

isaac_arena/policy/config/dataset_config.py
xyao-nv: I default it to empty string in the dataclass, can set to None. Then following your suggestion avoid comparing to empty string.

isaac_arena/policy/config/dataset_config.py
xyao-nv: done

isaac_arena/policy/config/dataset_config.py
xyao-nv: done

isaac_arena/policy/config/dataset_config.py
xyao-nv: added assertion in convert_hdf5_to_lerobot() func call

isaac_arena/tests/policy/test_convert_hdf5_to_lerobot.py
xyao-nv: Kept only checking files, instead of parent-dir + files.

isaac_arena/policy/config/g1/43dof_joint_space.yaml
xyao-nv: Those numbers are retrieved offline in Lab loading USD and accessing `joint_names`, and `joint_indices`. The less-wanted behavior is that Lab does not enforce this 1-1 mapping between `joint_names` and `joint_indices`. 
I was thinking of retrieving from embodiment, however, this conversion script does not bringup Lab/Arena Sim. I could add another script which retrieves the 1-1 mapping from launching Arena's embodiment and writes to this yaml. WDYT @alexmillane

isaac_arena/policy/data_utils/io_utils.py
xyao-nv: Kepts whatever before the last line as `load_dict_from_yaml()`, then define another func call `load_robot_joints_config_from_yaml()`.

isaac_arena/policy/data_utils/robot_eef_pose.py
xyao-nv: `Pose` objection uses fixed tuples: `position_xyz`: tuple[float, float, float] and rotation_wxyz: tuple[float, float, float, float]. It has **no support for arrays or batched operations**
In this use case, it needs to process a batch of `Pose`, if sticking to Pose, then it has to be a list of `Pose` object.
I would propose, how about adding a `BatchedPose `obj in geometry/pose.py? The potential use case could be, when inquiring object pose in multi-env apps, it can be done thru batched ops, like frame transformation?
@alexmillane wdyt

xyao-nv: Since my understanding of Pose obj is not directly applicable here, I will keep it as it is. Can revisit in another MR once we draw a conclusion.

isaac_arena/policy/data_utils/robot_joints.py
xyao-nv: done

isaac_arena/policy/data_utils/robot_eef_pose.py
xyao-nv: aha, I see your point! Thanks for the explanation! Will keep it in mind, remove pose as a class member and added a getter to compose the pose on the fly.

isaac_arena/policy/data_utils/robot_joints.py
xyao-nv: Totally agree with removing device from dataclass member. Thanks for the explanations!

isaac_arena/policy/data_utils/joints_conversion.py
xyao-nv: yeah I will think about it in another MR.

isaac_arena/policy/data_utils/yaml_config_loader.py
xyao-nv: changed based on Alex comments - generic yaml loading & robot specific config retrieving

isaac_arena/policy/data_utils/robot_eef_pose.py
xyao-nv: done

PR #70, opened 2025-09-26
Remove teleop devices from post init

viiik-inside:
## Summary
Remove teleop devices from post init


isaac_arena/environments/compile_env.py
```diff
@@ -64,7 +64,10 @@ def compose_manager_cfg(self) -> tuple[IsaacArenaManagerBasedRLEnvCfg, str | typ
         )
         actions_cfg = self.arena_env.embodiment.get_action_cfg()
         xr_cfg = self.arena_env.embodiment.get_xr_cfg()
-        teleop_device = self.arena_env.teleop_device
+        if self.arena_env.teleop_device is not None:
+            teleop_device = self.arena_env.teleop_device.get_teleop_device_cfg(actions_cfg=actions_cfg, xr_cfg=xr_cfg)
```
alexmillane: suggestion to name it `teleop_device_cfg` to match the `*Cfg` structures above.

isaac_arena/teleop_devices/avp_handtracking.py
```diff
@@ -25,27 +25,29 @@ class HandTrackingTeleopDevice(TeleopDeviceBase):
...
                 "avp_handtracking": OpenXRDeviceCfg(
                     retargeters=[
                         GR1T2RetargeterCfg(
-                            enable_visualization=True,
+                            enable_visualization=self.enable_visualization,
                             # number of joints in both hands
-                            num_open_xr_hand_joints=2 * self.NUM_OPENXR_HAND_JOINTS,
-                            sim_device=sim_device,
-                            hand_joint_names=actions.pink_ik_cfg.hand_joint_names,
+                            num_open_xr_hand_joints=2 * self.num_open_xr_hand_joints,
```
alexmillane: Any idea where the `2*` is coming from? It's slightly unexpected that you have two variables, both called `num_open_xr_hand_joints` and one is two times the other... You would expect them, because they have the same name, to be representing the same thing, and therefore have the same value. If they really represent the same quantity, should we change the default value from 26 to 52 and remove the `2*`?

isaac_arena/teleop_devices/avp_handtracking.py
```diff
@@ -25,27 +25,29 @@ class HandTrackingTeleopDevice(TeleopDeviceBase):
...
-        super().__init__()
+    def __init__(
+        self, sim_device: str | None = None, num_open_xr_hand_joints: int = 26, enable_visualization: bool = True
+    ):
+        super().__init__(sim_device=sim_device)
+        self.num_open_xr_hand_joints = num_open_xr_hand_joints
+        self.enable_visualization = enable_visualization
 
-    def build_cfg(self, *, sim_device: str | None = None, actions: object | None = None, xr_cfg: object | None = None):
+    def get_teleop_device_cfg(self, actions_cfg: object | None = None, xr_cfg: object | None = None):
```
alexmillane: I'm thinking that this function should just take the whole embodiment. It seems to me that in the (near) future, we'll want to switch the retargeter based on the embodiment. So we could just take the whole embodiment and extract the information we need: in this case `action_cfg` and `xr_cfg`.

This would also have the advantage that right now, we could assert that the embodiment is GR1 in this function, which, temporarily, only supports that embodiment.

isaac_arena/environments/compile_env.py
viiik-inside: yeah

isaac_arena/teleop_devices/avp_handtracking.py
viiik-inside: makes sense. Lemme do that change

PR #72, opened 2025-09-29
Add Replay Lerobot Action into policy runner

xyao-nv:
## Summary
Replay GT Lerobot dataset, by directly accessing actions as the action target in policy runner

## Detailed description
Useful to verifying HDF5->Lerobot converter, and data integrity before post-training or closedloop.
Feature including
- Add ReplayLerobotActionPolicy() to support both g1 locomanip & gr1 tabletop manip cases, where action dims are different
- Add data config to enable GR00T loading g1 locomanip data which is not supported in GR00T OSS repo
- Add test to verify policy runner using this policy with associated g1 required lerobot data

## Important
NOTE: lazy deps import is introduced on purpose, to prevent Sim app hanging and enable CI testing gr00t-related policy code in another docker. See more `in isaac_arena/examples/policy_runner_cli.py`

### DIFF
--- a/isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
+++ b/isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
@@ -100,7 +100,7 @@ class ReplayLerobotActionPolicy(PolicyBase):
         assert self.current_action_index < self.num_feedback_actions
 
         action = self.current_action_chunk[:, self.current_action_index]
-        assert action.shape == (self.num_envs, env.action_space.shape[0])
+        assert action.shape == env.action_space.shape, f"{action.shape=} != {env.action_space.shape=}"
 
         self.current_action_index += 1
         # reset to empty action chunk


isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
```diff
@@ -0,0 +1,190 @@
...
+        self.num_envs = num_envs
+        self.device = device
+        self.task_mode = TaskMode(self.policy_config.task_mode_name)
+
+        self.load_policy_joints_config()
+        self.load_sim_joints_config()
+
+    def load_policy_joints_config(self):
+        """Load the policy joint config from the data config."""
+        self.gr00t_joints_config = load_robot_joints_config_from_yaml(self.policy_config.gr00t_joints_config_path)
```
alexmillane: I'm wondering if this policy is general, or represents a replay policy for gr00t demonstration data. I would guess that given that we're loading a groot config file, this is specific to gr00t demonstration data stored in lerbot format?

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
```diff
@@ -0,0 +1,190 @@
...
+        # determine rollout how many action prediction per observation
+        self.num_feedback_actions = self.policy_config.num_feedback_actions
+        self.current_action_index = 0
+        self.current_action_chunk = None
+        self.num_envs = num_envs
+        self.device = device
+        self.task_mode = TaskMode(self.policy_config.task_mode_name)
+
+        self.load_policy_joints_config()
+        self.load_sim_joints_config()
```
alexmillane: Maybe it's a personal preference, but I would prefer to return a value from these functions, and set the member variable in the constructor. For example

```
self.gr00t_joints_config = self.load_policy_joints_config(self.policy_config.gr00t_joints_config_path)
```

This makes it obvious to the reader what changes are occurring to the class when reading the constructor.

Of course this isn't always possible, but I would have a preference for it, if it is.

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
```diff
@@ -0,0 +1,190 @@
...
+        self.gr00t_joints_config = load_robot_joints_config_from_yaml(self.policy_config.gr00t_joints_config_path)
+
+    def load_sim_joints_config(self):
+        """Load the simulation joint config from the data config."""
+
+        self.robot_action_joints_config = load_robot_joints_config_from_yaml(
+            self.policy_config.action_joints_config_path
+        )
+
+    def load_policy(self):
```
alexmillane: nit: return type annotation.

Suggestion to implement the same change as above. This function could be slightly modified to take `self.policy_config` as an argument and return a `LeRobotSingleDataset`. That way the dataflow is totally clear to the reader :)

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
```diff
@@ -0,0 +1,190 @@
...
+            video_backend_kwargs=None,
+            transforms=None,  # We'll handle transforms separately through the policy
+            embodiment_tag=self.policy_config.embodiment_tag,
+        )
+
+    def get_trajectory_length(self) -> int:
+        """Get the number of frames in one trajectory in the dataset."""
+        assert self.policy.trajectory_lengths is not None
+        assert self.trajectory_index < len(self.policy.trajectory_lengths)
+        return self.policy.trajectory_lengths[self.trajectory_index]
```
alexmillane: Suggestion to take `trajectory_index` as an argument.

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
```diff
@@ -0,0 +1,190 @@
...
+        # get new predictions and return the first action from the chunk
+        if self.current_action_chunk is None and self.current_action_index == 0:
+            self.current_action_chunk = self.get_action_chunk()
+            assert self.current_action_chunk.shape[1] >= self.num_feedback_actions
+
+        assert self.current_action_chunk is not None
+        assert self.current_action_index < self.num_feedback_actions
+
+        action = self.current_action_chunk[:, self.current_action_index]
+        assert action.ndim == 2, f"Action is {action.shape} but should be (B, action_dim)"
```
alexmillane: This check is slightly different than what the comment says. Is it possible to check what the comment says ie `assert action.shape == torch.size(C, action_dim)`

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
```diff
@@ -0,0 +1,190 @@
...
+        """Get action_horizon number of actions, as an action chunk, from the dataset"""
+        step_index = next(self.policy_iter)
+        data_point = self.policy[step_index]
+        # Support MultiEnv running
+        actions = {
+            "action.left_arm": np.tile(np.array(data_point["action.left_arm"]), (self.num_envs, 1, 1)),
+            "action.right_arm": np.tile(np.array(data_point["action.right_arm"]), (self.num_envs, 1, 1)),
+            "action.left_hand": np.tile(np.array(data_point["action.left_hand"]), (self.num_envs, 1, 1)),
+            "action.right_hand": np.tile(np.array(data_point["action.right_hand"]), (self.num_envs, 1, 1)),
+        }
```
alexmillane, line 122: Does this mean that for multiple envs, we copy the actions for each environment?

Suggestion to update the comment to indicate that the same actions are run for each environment.

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def reset(self):
+        """Resets the policy's internal state."""
+        # As GR00T is a single-shot policy, we don't need to reset its internal state
+        # Only reset the action chunking mechanism
+        self.policy_iter = self.create_trajectory_iterator(self.trajectory_index)
+        self.current_action_chunk = None
+        self.current_action_index = 0
+
+    def create_trajectory_iterator(self, trajectory_index: int = 0):
```
alexmillane: nit: return type annotation.

isaac_arena/policy/data_utils/io_utils.py
```diff
@@ -16,8 +16,12 @@
 import json
 import numpy as np
 import yaml
+from dataclasses import fields
 from pathlib import Path
-from typing import Any
+from typing import Any, TypeVar, Union
+
+# Generic type variable for configuration classes
+ConfigType = TypeVar("ConfigType")
```
alexmillane, line 24: Interesting. Elsewhere in the repo we're just type annotating `configclass`es with `Any`. I wonder if we should create a type to use across the whole repo for this.

One issue is that configclasses don't actually belong to some common base class..

isaac_arena/examples/policy_runner_cli.py
```diff
@@ -82,6 +109,20 @@ def create_policy(args: argparse.Namespace) -> tuple[PolicyBase, int]:
...
+        # NOTE(xinjie.yao, 2025-09-28): lazy import to prevent app stalling
+        # due to import GR00T py dependencies that are conflicting with omni.kit
+        # see functional import sequence here https://github.com/isaac-sim/IsaacLabEvalTasks/blob/main/scripts/evaluate_gn1.py#L38
+        from isaac_arena.policy.gr00t.replay_lerobot_action_policy import ReplayLerobotActionPolicy
+
+        policy = ReplayLerobotActionPolicy(
+            args.config_yaml_path, num_envs=args.num_envs, device=args.device, trajectory_index=args.trajectory_index
+        )
+        # Use custom max_steps if provided to optionally playing partial sequence in one trajectory
+        if hasattr(args, "max_steps") and args.max_steps is not None:
```
alexmillane: Maybe I'm missing something, but it seems to me like the argument `max_steps` should always exist right? Could we remove `hasattr(args, "max_steps")`?

isaac_arena/examples/policy_runner_cli.py
```diff
@@ -58,18 +82,21 @@ def setup_policy_argument_parser() -> argparse.ArgumentParser:
...
+        choices=["zero_action", "replay", "replay_lerobot"],
         required=True,
-        help="Type of policy to use: 'zero_action' or 'replay'",
+        help="Type of policy to use: 'zero_action' or 'replay' or 'replay_lerobot'",
     )
 
     # Add policy-specific argument groups
     add_zero_action_arguments(args_parser)
     add_replay_arguments(args_parser)
+    add_replay_lerobot_arguments(args_parser)
```
alexmillane, line 93: At some point we're going to have to think of a way where policies automatically register themselves and their arguments, such that we don't have to write a list here for all polcies. We have something like this for the example environments.

Anyway, for now, let's just roll with this.

isaac_arena/tests/utils/constants.py
```diff
@@ -30,7 +30,7 @@ def __init__(self):
 
         self.python_path = f"{self.repo_root}/submodules/IsaacLab/_isaac_sim/python.sh"
 
-        self.test_data_dir = f"{self.test_dir}/test_data/"
+        self.test_data_dir = f"{self.test_dir}test_data/"
```
alexmillane: Is this intentional? I guess that `self.test_dir` ends with a `/`? I would suggest to remove the trailing `/` in order to align with the other path roots in this file.

isaac_arena/policy/gr00t/g1_locomanip_replay_action_config.yaml
```diff
@@ -0,0 +1,12 @@
+# This file shows how to configure the ReplayLerobotActionPolicy class for G1 locomanipulation task
+dataset_path: /datasets/g1_galileo_box_npnp_mimic_generated_100_v2_action_noise_003/lerobot/
```
peterd-NV, line 3: Maybe add a note here that this is a hard coded path, the dataset name will probably change when we release.

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
xyao-nv: The ordering is specified in Lerobot thru modality.json -- so it's Lerobot general, I changed the var name as policy_joints_config to make it more generic.

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
xyao-nv: Thx for the suggestion, changed to return config.

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
xyao-nv: Makes sense. done

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
xyao-nv: done

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
xyao-nv: changed to explicitly asserting the shape

isaac_arena/policy/gr00t/replay_lerobot_action_policy.py
xyao-nv: done

isaac_arena/examples/policy_runner_cli.py
xyao-nv: yeah, right, done.

isaac_arena/policy/gr00t/g1_locomanip_replay_action_config.yaml
xyao-nv: done

isaac_arena/tests/utils/constants.py
xyao-nv: it was because of `test_dir` has trailing `/`. Removed all trailing `/`

isaac_arena/policy/data_utils/io_utils.py
xyao-nv: lemme think about it later, and see if it's necessary to do a refactor of type annotation in another MR

PR #102, opened 2025-10-07
Add locomanip Mimic data gen instruction to docs

peterd-NV:
## Summary
Add locomanip Mimic data gen instruction to docs

### DIFF
--- a/docs/pages/examples/example_locomanipulation.rst
+++ b/docs/pages/examples/example_locomanipulation.rst
@@ -25,11 +25,6 @@ from a small set of annotated demonstrations by using rigid body transformations
 object initial poses are randomized to introduce variation and boost dataset diversity. Noise is added to the robot's actions to further
 increase diversity. For more information about Isaac Lab Mimic, please visit the Isaac Lab documentation `here <https://isaac-sim.github.io/IsaacLab/main/source/overview/imitation-learning/teleop_imitation.html#teleoperation-and-imitation-learning-with-isaac-lab-mimic>`_.
 
-Before proceeding, make sure you are inside the Isaac Lab - Arena docker container by running:
-
-.. code:: bash
-
-   ./docker/run_docker.sh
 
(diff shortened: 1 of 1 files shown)


docs/pages/examples/example_locomanipulation.rst
```diff
@@ -1,41 +1,95 @@
...
 
-.. TODO::
+Before proceeding, make sure you are inside the Isaac Lab - Arena docker container by running:
 
-   (xinjie.yao, 2025-10-03): add record_demos command
+.. code:: bash
 
-Mimic Generated Data
----------------------
+   ./docker/run_docker.sh
```
xyao-nv: it uses virtual mounting of the /dataset. Maybe you can point users to download dataset there

docs/pages/examples/example_locomanipulation.rst
```diff
@@ -1,41 +1,95 @@
...
 
-Use Isaac Lab Mimic with added Mimic supports for loco-manipulation to generate demonstrations:
+Download Pre-recorded Annotated Dataset
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Download the pre-recorded annotated dataset from `TODO_INSERT_LINK <https://>`_ and place it
+under ``isaac_arena/datasets/Arena-G1-Loco-Manipulation-Task/``.
```
xyao-nv, line 32: so we can expect users will run from /datasets/...

docs/pages/examples/example_locomanipulation.rst
peterd-NV: I'll move the docker command after telling users to download dataset. This way might be more clear.

docs/pages/examples/example_locomanipulation.rst
peterd-NV: users should run commands from the top level `isaac_arena` directory

PR #119, opened 2025-10-14
Parallel env policy closedloop evaluation

xyao-nv:
## Summary
Enable policy closedloop with multi envs using TiledCamera and policy reset

## Detailed description
- Add TiledCameraCfg class and set it to GR and GR1 with joint space action term.
- Enabled GR00T observation space to take >1 env
- Fix app exiting and policy reset func
- Added test and updated doc

E.g. if policy needs 100 steps to succeed, assuming 100% SR, with num_steps = 200 
- single env: will end up running 1< num_episode <= 2,  >=1 success 
- 2 parallel envs (async): will end up running 2 < num_episodes <=4, >=2 success

### DIFF
--- a/isaac_arena/embodiments/g1/g1.py
+++ b/isaac_arena/embodiments/g1/g1.py
@@ -67,10 +67,12 @@ class G1EmbodimentBase(EmbodimentBase):
             anchor_rot=(0.70711, 0.0, 0.0, -0.70711),
         )
 
+
 # Default camera offset pose
 _DEFAULT_G1_CAMERA_OFFSET = Pose(
     position_xyz=(0.04485, 0.0, 0.35325), rotation_wxyz=(0.32651, -0.62721, 0.62721, -0.32651)
 )
+
 
 @register_asset
 class G1WBCJointEmbodiment(G1EmbodimentBase):
@@ -330,8 +332,8 @@ class G1CameraCfg:
     def __post_init__(self):
         # Get configuration from private attributes set by embodiment constructor
         # These use getattr with defaults to avoid scene parser treating them as assets
-        is_tiled_camera = getattr(self, '_is_tiled_camera', True)
-        camera_offset = getattr(self, '_camera_offset', _DEFAULT_G1_CAMERA_OFFSET)
+        is_tiled_camera = getattr(self, "_is_tiled_camera", True)
+        camera_offset = getattr(self, "_camera_offset", _DEFAULT_G1_CAMERA_OFFSET)
 
         CameraClass = TiledCameraCfg if is_tiled_camera else CameraCfg
         OffsetClass = CameraClass.OffsetCfg
@@ -356,9 +358,6 @@ class G1CameraCfg:
         )
 
         self.robot_head_cam = CameraClass(offset=offset, **common_kwargs)
-
-
-
 
 
 @configclass
--- a/isaac_arena/embodiments/gr1t2/gr1t2.py
+++ b/isaac_arena/embodiments/gr1t2/gr1t2.py
@@ -111,7 +111,6 @@ class GR1T2EmbodimentBase(EmbodimentBase):
         )
 
 
-
 @register_asset
 class GR1T2JointEmbodiment(GR1T2EmbodimentBase):
     """Embodiment for the GR1T2 robot with joint position control.
@@ -347,8 +346,8 @@ class GR1T2CameraCfg:
     def __post_init__(self):
         # Get configuration from private attributes set by embodiment constructor
         # These use getattr with defaults to avoid scene parser treating them as assets
-        is_tiled_camera = getattr(self, '_is_tiled_camera', True)
-        camera_offset = getattr(self, '_camera_offset', _DEFAULT_CAMERA_OFFSET)
+        is_tiled_camera = getattr(self, "_is_tiled_camera", True)
+        camera_offset = getattr(self, "_camera_offset", _DEFAULT_CAMERA_OFFSET)
 
         CameraClass = TiledCameraCfg if is_tiled_camera else CameraCfg
         OffsetClass = CameraClass.OffsetCfg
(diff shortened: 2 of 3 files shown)


isaac_arena/examples/policy_runner.py
```diff
@@ -45,8 +54,17 @@ def main():
             with torch.inference_mode():
                 actions = policy.get_action(env, obs)
                 obs, _, terminated, truncated, _ = env.step(actions)
-                if terminated.any() or truncated.any():
-                    obs, _ = env.reset()
+                # NOTE(xinjieyao, 2025-10-13): For policy closedloop, if any env succeeds, reset the policy; As in parallel
+                # envs, each env is running async, it may be cases where one env succeeds when others are not done.
+                # Given GR00T policy is a single-shot policy, it is safe to reset the policy. However, for other policies, we need to be careful.
+                if env.unwrapped.termination_manager.get_term("success").any():
```
peterd-NV: Should the policy also be reset if the env reaches a time out or any other termination? If a policy fails to successfully complete a task and reaches a termination or time out, the env will auto reset via the termination manager and policy should also be reset, correct?

isaac_arena/examples/policy_runner.py
```diff
@@ -45,8 +54,17 @@ def main():
...
+                    policy.reset()
+                # Only break out of the loop if all envs are done
+                # NOTE(xinjieyao, 2025-10-13): In single env, closedloop policy runs until num_steps is reached. Keep resetting if task is completed.
+                # In multi env, closedloop policy runs until all envs are done. Depending on num_steps, it could have more than 1 eposides with single env.
+                if args_cli.num_envs > 1:
+                    if terminated.all() or truncated.all():
```
peterd-NV: I don't believe this check is necessary.

If parallel envs are running, then it is very unlikely for this condition to ever be satisfied. Each env is running async so they will reset at different times. Upon a termination condition being triggered or time out, the env will auto reset and start running again. 

Even if all envs happen to terminate/timeout at the same time, they will be auto reset and continue running. In that case, I think the rollout should still continue until the number of steps has reached `num_steps`.

docs/pages/examples/create_a_new_environment.rst
```diff
@@ -74,8 +74,8 @@ The asset registry provides access to all registered assets:
 **Available Assets:**
 
 - **Backgrounds**: ``kitchen``, ``galileo_lab``
-- **Objects**: ``cracker_box``, ``mustard_bottle``, ``sugar_box``, ``tomato_soup_can``, ``microwave``, etc.
-- **Embodiments**: ``franka``, ``gr1_pink``, ``gr1_joint``
+- **Objects**: ``cracker_box``, ``mustard_bottle``, ``sugar_box``, ``tomato_soup_can``, ``microwave``, ``brown_box``, etc.
+- **Embodiments**: ``franka``, ``gr1_pink``, ``gr1_joint``, ``g1_wbc_joint``, ``g1_wbc_pink``
```
viiik-inside, line 78: nice

isaac_arena/embodiments/g1/g1.py
```diff
@@ -315,6 +318,28 @@ class G1CameraCfg:
     )
 
 
+@configclass
+class G1TiledCameraCfg:
```
viiik-inside: Lets try to do this in a differnt way. Lets take the offset as an argument to this class and keep one single class for any kind of camera variations. We should redesign this

isaac_arena/embodiments/gr1t2/gr1t2.py
```diff
@@ -120,6 +120,8 @@ def __init__(self, enable_cameras: bool = False, initial_pose: Pose | None = Non
         self.action_config = GR1T2JointPositionActionCfg()
         # Tuned arm joints pd gains, smoother motions and less oscillations
         self.scene_config = GR1T2HighPDSceneCfg()
+        # Closedloop evaluation could be done in parallel environments, tiled camera is used for efficient evaluation
+        self.camera_config = GR1T2TiledCameraCfg()
```
viiik-inside: lets try in a different way. It would be great to avoid copies

isaac_arena/examples/policy_runner.py
viiik-inside: sounds correct

isaac_arena/examples/policy_runner.py
xyao-nv: I agree with what you said, it's just I added this when seeing metrics reporting an error. But with other fix in termination, it doesn't appear anymore.

isaac_arena/embodiments/g1/g1.py
xyao-nv: See the new one. Added a factory builder method and keep these 2 classes. I don't see the value of keeping a normal camera here, as tiledCamera shall also work for usecases. Keeping both if there is anything else I miss.

isaac_arena/embodiments/gr1t2/gr1t2.py
xyao-nv: see the change. I still don't quite understand what you mean by copies here, configclass are not compiled until specified.

isaac_arena/examples/policy_runner.py
xyao-nv: yeah that makes sense. And this  env.unwrapped.termination_manager.get_term("success") is actually the last episode success -- won't change after a Success reset.

isaac_arena/embodiments/g1/g1.py
```diff
@@ -315,6 +365,82 @@ class G1CameraCfg:
...
+            vertical_aperture=0.284,  # Scaled to preserve 80 vertical FOV
+            clipping_range=(0.1, 5),
+        ),
+        offset=TiledCameraCfg.OffsetCfg(
+            pos=(0.04485, 0.0, 0.35325), rot=(0.32651, -0.62721, 0.62721, -0.32651), convention="ros"
+        ),
+    )
+
+
+def create_g1_camera_cfg(
```
viiik-inside: give me a bit of time. Ill get back for this.

isaac_arena/examples/policy_runner.py
peterd-NV: thanks, marking this policy reset as resolved

isaac_arena/examples/policy_runner.py
peterd-NV: thanks, marking "removal of break" as resolved

PR #128, opened 2025-10-16
Add termination cfg and recorder cfg to embodiment

viiik-inside:
## Summary
Add recorder and termination terms to the embodiment class


isaac_arena/environments/compile_env.py
```diff
@@ -85,6 +86,7 @@ def compose_manager_cfg(self) -> IsaacArenaManagerBasedRLEnvCfg:
             "RecorderManagerCfg",
             metrics_recorder_manager_cfg,
             self.arena_env.task.get_recorder_term_cfg(),
+            self.arena_env.embodiment.get_recorder_term_cfg(),
```
cvolkcvolk, line 89: Fantastic!

PR #143, opened 2025-10-21
Add an OSMO workflow.

alexmillane:
## Summary
Add an OSMO workflow to fine-tune gr00t GN1

## Detailed description
- This allows us to train and experiment with gr00t.
- Exposes some finetuning parameters to the command line.

## Notes
- We may not want to release this. So we may need to remove it before release, however, having it tracked in git for our internal use is useful during development.


osmo/finetune_microwave.yaml
```diff
@@ -0,0 +1,93 @@
...
+          export WANDB_ENTITY=nv-welcome
+          export WANDB_PROJECT="Isaac Lab Arena"
+          export WANDB_NAME="{{ workflow_name }}_$(date +%Y%m%d_%H%M%S)"
+
+          # Modify gr00t code to allow for naming of the run.
+          sed -i "s/^\(\s*run_name=\)None/\1\"${WANDB_NAME}\"/" scripts/gr00t_finetune.py
+
+          # Launch finetune
+          HYDRA_FULL_ERROR=1 python scripts/gr00t_finetune.py \
+            --dataset_path="{{input:0}}/lerobot_gr1/lerobot_gr1/gr1_open_microwave_50_generated/lerobot" \
```
cvolkcvolk: Shall we add this (the dataset) as an default value as well at the bottom?

alexmillane: yea that's a good idea actually.

I might actually simplify the path too :sweat_smile:

PR #146, opened 2025-10-22
Add feature for hot reloading

viiik-inside:
## Summary
Hot reload feature for development.

Draft because lightwheel sdk fails during reload intermittently.

### DIFF
--- a/isaac_arena/examples/example_env_notebook.py
+++ b/isaac_arena/examples/example_env_notebook.py
@@ -23,46 +23,9 @@ from isaaclab.app import AppLauncher
 print("Launching simulation app once in notebook")
 simulation_app = AppLauncher()
 
-# %%
-
-
-def reload_arena_modules():
-    import importlib
-    import os
-    import sys
-
-    # Clear Python's bytecode cache
-    if hasattr(importlib, "invalidate_caches"):
-        importlib.invalidate_caches()
-
-    # Get all isaac_arena modules currently loaded
-    isaac_arena_modules = [
-        (name, module) for name, module in sys.modules.items() if name.startswith("isaac_arena") and module is not None
-    ]
-
-    if len(isaac_arena_modules) == 0:
-        print("No isaac_arena modules found")
-        return
-
-    # We skip this step due to import issues in our asset registry
-    # # Reload modules from bottom to top (deepest/leaf modules first)
-    # isaac_arena_modules.sort(key=lambda x: x[0].count('.'), reverse=True)
-
-    for module_name, module in isaac_arena_modules:
-        try:
-            # Delete the .pyc cache file to force recompilation from source
-            module_cached = getattr(module, "__cached__", None)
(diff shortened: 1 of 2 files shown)


isaac_arena/examples/example_env_notebook.py
```diff
@@ -24,6 +24,47 @@
...
+        (name, module) for name, module in sys.modules.items() if name.startswith("isaac_arena") and module is not None
+    ]
+
+    if len(isaac_arena_modules) == 0:
+        print("No isaac_arena modules found")
+        return
+
+    # We skip this step due to import issues in our asset registry
+    # # Reload modules from bottom to top (deepest/leaf modules first)
+    # isaac_arena_modules.sort(key=lambda x: x[0].count('.'), reverse=True)
```
cvolkcvolk: leftover comment?

isaac_arena/examples/example_env_notebook.py
```diff
@@ -24,6 +24,47 @@
 simulation_app = AppLauncher()
 
 # %%
+
+
+def reload_arena_modules():
```
cvolkcvolk: Could we move this function out to a respective utils/ ?

isaac_arena/examples/example_env_notebook.py
```diff
@@ -62,5 +103,15 @@
         actions = torch.zeros(env.action_space.shape, device=env.unwrapped.device)
         env.step(actions)
 
-
 # %%
+
+from isaaclab.sim import SimulationContext
```
cvolkcvolk, line 71: Out of interest, why do we need these following lines now?

isaac_arena/examples/example_env_notebook.py
```diff
@@ -24,6 +24,47 @@
...
+            print(f"Reloading {module_name}")
+            importlib.reload(module)
+
+        except Exception as e:
+            print(f"[WARNING] Failed to reload {module_name}: {e}")
+
+
+# %%
+
+reload_arena_modules()
```
cvolkcvolk, line 30: Do we always need to reload our modules?

cvolkcvolk: Eg, if we should hide this behind a flag or if its needed all the time

isaac_arena/examples/example_env_notebook.py
viiik-inside: No its a valid comment. We dont know why it works and this line is kind of important. So we have to get back to it

isaac_arena/examples/example_env_notebook.py
viiik-inside: This is to clean up the simultion environment so a new one can load in the same simulation app

isaac_arena/examples/example_env_notebook.py
viiik-inside: We can. The reload will do nothing if there is nothing to reload

PR #148, opened 2025-10-23
ViewerCfg sets the initial camera position per task

cvolkcvolk:
## Summary
Spawn the initial camera position per task, focussing on the object of interest.

## Detailed description
Don't spawn to a default value far away but use the obejct of interest per task to place the initial camera view.

### DIFF
--- a/isaac_arena/environments/arena_env_builder.py
+++ b/isaac_arena/environments/arena_env_builder.py
@@ -139,6 +139,7 @@ class ArenaEnvBuilder:
                 # recorders=recorder_manager_cfg,
                 # metrics=metrics,
                 isaac_arena_env=isaac_arena_env,
+                viewer=viewer_cfg,
             )
         return env_cfg
 


isaac_arena/environments/arena_env_builder.py
```diff
@@ -116,6 +118,7 @@ def compose_manager_cfg(self) -> IsaacArenaManagerBasedRLEnvCfg:
                 recorders=recorder_manager_cfg,
                 metrics=metrics,
                 isaac_arena_env=isaac_arena_env,
+                viewer=viewer_cfg,
```
viiik-inside, line 121: I think we also want to add it to the mimic rl config right?

isaac_arena/tasks/open_door_task.py
```diff
@@ -77,6 +80,9 @@ def get_metrics(self) -> list[MetricBase]:
             ),
         ]
 
+    def get_viewer_cfg(self) -> ViewerCfg:
+        return get_viewer_cfg_look_at_object(lookat_object=self.openable_object, offset=np.array([-1.5, -1.5, 1.5]))
```
viiik-inside, line 84: nice

isaac_arena/environments/arena_env_builder.py
cvolkcvolk: Right.

PR #149, opened 2025-10-23
Use persistant simulation app for tests

viiik-inside:
## Summary
We create a persistant simulation app for tests. This gets more than 10x speedup

## Detailed description
- We dont want the simulation app to restart for every test.
- We mark tests with camera seperately as they require another simuation app.


isaac_arena/tests/test_camera_observation.py
```diff
@@ -72,13 +76,18 @@ def _test_camera_observation(simulation_app) -> bool:
...
+
     return True
 
 
+@pytest.mark.with_cameras
 def test_camera_observation():
-    result = run_simulation_app_function_in_separate_process(
+    # Safety: if someone runs this without the env var, skip with a helpful message
+    if not ENABLE_CAMERAS:
+        pytest.skip("This test requires cameras. Re-run with: WITH_CAMERAS=1 pytest -m with_cameras")
```
alexmillane, line 88: What's this environment variable `WITH_CAMERAS`?

When we run this in the CI workflow, I don't think we set this?

isaac_arena/tests/test_camera_observation.py
```diff
@@ -72,13 +76,18 @@ def _test_camera_observation(simulation_app) -> bool:
             # Make sure the camera observation contains values other than 0
             assert camera_observation.any() != 0, "Camera observation contains only 0s"
 
-    # Close the environment.
     env.close()
+
     return True
 
 
+@pytest.mark.with_cameras
```
cvolkcvolk, line 84: cool

isaac_arena/tests/test_camera_observation.py
alexmillane: Ah this is not an environment variable this is a file variable from above.

I'm not sure what the intention here. Does it makes sense to run this test without cameras?

Something is confusing here.

isaac_arena/tests/test_asset_registry.py
```diff
@@ -95,6 +96,8 @@ def _test_all_assets_in_registry(simulation_app):
 
     builder = ArenaEnvBuilder(isaac_arena_environment, args_cli)
     env = builder.make_registered()
+    # disable control on stop
+    env.unwrapped.sim._app_control_on_stop_handle = None
```
alexmillane, line 100: What does this do, and what is the effect of leaving it out? Is this to allow multiple environments to run after one another?

It's a bit brittle if we require all tests to have this, but don't enforce it in any way.

We should either:
- Have it on by default. Is there any disadvantage here?
- Hide the long variable name inside the compiler.

Something like:

```
builder.allow_env_restart = True
```

There's probably a better name, but I don't understand what this variable does.

isaac_arena/tests/test_device_registry.py
```diff
@@ -77,15 +77,22 @@ def _test_all_devices_in_registry(simulation_app):
...
-    env.close()
+        # Close the environment.
+        from isaaclab.sim import SimulationContext
+
+        simulation_context = SimulationContext.instance()
+        simulation_context._disable_app_control_on_stop_handle = True
+        simulation_context.stop()
+        simulation_context.clear_instance()
+        env.close()
+        omni.timeline.get_timeline_interface().stop()
```
alexmillane, line 88: What's going on here? Why is this required in this test but not all the others.

At a minimum, I would add a comment here. But probably better would be to encapsulate these segment of code (it's a bit hard to know _how_ to encapsulate it withouth knowing what it's doing).

isaac_arena/tests/test_camera_observation.py
viiik-inside: i removed it. It doesnt make sense

isaac_arena/tests/test_device_registry.py
viiik-inside: This is required for this test as we create new environments and run it even within this test itself. All the other tests create the environment only once. Hence its being done explicitily here. I actually am gonna switch to using the safe teardown instead. It should ideally work

PR #163, opened 2025-10-28
Docs: Simplify HuggingFace Commands

alexmillane:
## Summary
Improve the download commands.

## Detailed description
- Use the updated version of the huggingface CLI tool
- Now download to constant paths to make the instructions simpler.
- Add the huggingface CLI tool to the docker.

### DIFF
--- a/docs/pages/example_workflows/locomanipulation/index.rst
+++ b/docs/pages/example_workflows/locomanipulation/index.rst
@@ -65,11 +65,15 @@ by downloading the pre-generated output of the preceding step(s), which we provi
 Prerequisites
 ^^^^^^^^^^^^^
 
-We store data on Hugging Face, so you'll need log in to Hugging Face
-(see the `official instructions <https://huggingface.co/docs/huggingface_hub/installation>`_
-for more details).
+Start the isaaclab docker container
 
-.. code:: bash
+.. code-block:: bash
+
+    ./docker/run_docker.sh
+
+We store data on Hugging Face, so you'll need log in to Hugging Face if you haven't already.
+
+.. code-block:: bash
 
     hf auth login
 
--- a/docs/pages/example_workflows/locomanipulation/step_3_policy_training.rst
+++ b/docs/pages/example_workflows/locomanipulation/step_3_policy_training.rst
@@ -47,7 +47,7 @@ Note that this conversion step can be skipped by downloading the pre-converted L
          nvidia/Arena-G1-Loco-Manipulation-Task \
          --include lerobot/* \
          --repo-type dataset \
-         --local-dir $LEROBOT_DATASET_DIR
+         --local-dir $DATASET_DIR
 
    If you download this dataset, you can skip the conversion step below and continue to the next step.
 
--- a/docs/pages/example_workflows/static_manipulation/index.rst
+++ b/docs/pages/example_workflows/static_manipulation/index.rst
@@ -62,15 +62,20 @@ by downloading the pre-generated output of the preceding step(s), which we provi
 Prerequisites
 ^^^^^^^^^^^^^
 
-We store data on Hugging Face, so you'll need log in to Hugging Face
-(see the `official instructions <https://huggingface.co/docs/huggingface_hub/installation>`_
-for more details).
+Start the isaaclab docker container
 
-.. code:: bash
+.. code-block:: bash
+
+    ./docker/run_docker.sh
+
+We store data on Hugging Face, so you'll need log in to Hugging Face if you haven't already.
+
+.. code-block:: bash
 
     hf auth login
 
-Create the folders for the data and models:
+You'll also need to create the folders for the data and models.
+Create the folders for the data and models with:
 
 .. code:: bash
 


docs/pages/example_workflows/locomanipulation/step_2_data_generation.rst
```diff
@@ -65,23 +65,16 @@ Step 3: Validate Generated Dataset (Optional)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 To visualize the data produced, you can replay the dataset using the following command:
-(replace ``<OUTPUT_DATASET_PATH>`` with the actual path):
 
 .. code-block:: bash
 
    python isaaclab_arena/scripts/replay_demos.py \
-     --device cpu \
      --enable_cameras \
```
peterd-NV, line 73: Just a note to check that the cpu device flag get's added here after rebasing on top of latest main branch. 

It seems like this is missing the cpu flag that was added in last commit to main [here](https://github.com/isaac-sim/IsaacLab-Arena/commit/469fbcb8a69341f220a9ff7d2f2a8f64872f5f7f).

docs/pages/example_workflows/locomanipulation/index.rst
```diff
@@ -61,23 +62,35 @@ A user can follow the whole pipeline, or can start at any intermediate step
...
 
-This example workflow involves the following steps:
+Prerequisites
+^^^^^^^^^^^^^
 
-- :doc:`step_1_environment_setup`
-- :doc:`step_2_data_generation`
-- :doc:`step_3_policy_training`
-- :doc:`step_4_evaluation`
+We store data on Hugging Face, so you'll need log in to Hugging Face
```
peterd-NV: Should we add a step/command here advising users to enter the docker container?

docs/pages/example_workflows/static_manipulation/index.rst
```diff
@@ -50,34 +50,46 @@ Task Overview
...
 This tutorial covers the pipeline between creating an environment, generating training data,
 fine-tuning a policy (GR00T N1.5), and evaluating the policy in closed-loop.
 A user can follow the whole pipeline, or can start at any intermediate step
 by downloading the pre-generated output of the preceding step(s), which we provide
 (described in the relevant step below).
 
-This example workflow involves the following steps:
+Prerequisites
+^^^^^^^^^^^^^
+
```
peterd-NV, line 64: Similarly here, I think we can add a step telling users to start the container

docs/pages/example_workflows/locomanipulation/step_3_policy_training.rst
```diff
@@ -39,61 +39,51 @@ Note that this conversion step can be skipped by downloading the pre-converted L
...
    .. code-block:: bash
 
-      huggingface-cli download \
+      hf download \
          nvidia/Arena-G1-Loco-Manipulation-Task \
-         lerobot \
+         --include lerobot/* \
          --repo-type dataset \
-         --local-dir <LEROBOT_DATASET_PATH>
+         --local-dir $LEROBOT_DATASET_DIR
```
peterd-NV: I don't think `LEROBOT_DATASET_DIR` is defined in the container or any of the previous steps. 

To keep consistency with the conversion step below I think this can be changed to `$DATASET_DIR/lerobot`

docs/pages/example_workflows/locomanipulation/step_3_policy_training.rst
```diff
@@ -39,61 +39,51 @@ Note that this conversion step can be skipped by downloading the pre-converted L
...
-   To download run (replacing ``<LEROBOT_DATASET_PATH>`` with the actual path):
+   To download run:
 
    .. code-block:: bash
 
-      huggingface-cli download \
+      hf download \
          nvidia/Arena-G1-Loco-Manipulation-Task \
-         lerobot \
+         --include lerobot/* \
```
xyao-nv, line 48: Nice!

docs/pages/example_workflows/locomanipulation/step_3_policy_training.rst
```diff
@@ -152,6 +142,12 @@ We provide two post-training options:
       TBD
 
 
+.. todo::
+
+   (alexmillane, 2025-10-23): Check that the resulting model matches
+   the folder structure that we download from Hugging Face.
```
xyao-nv, line 148: No it doesn't. It saves each `save_steps` as a separate subfolder, I pushed the `max_steps` subfolder to HF. Shall we add a note telling users to use the final checkpoint subfolder?

docs/pages/example_workflows/locomanipulation/step_2_data_generation.rst
xyao-nv: Maybe we shall add a special note saying the datagen used device==cpu, thus we keep cpu flag on here for reproducibility.

alexmillane: Thanks for the comment. I hadn't seen those `--device cpu` flags.

I checked and they've been correctly merged-in in your tutorials.

However, in the static manipulation tutorials, we don't have them. I guess we're using them to make the physics maximally reproducible? Is that correct? On that basis I think I should add them to the static manipulation tutorials too.

I'll add them with this MR.

docs/pages/example_workflows/locomanipulation/step_3_policy_training.rst
alexmillane: Good idea. Do you see any easy way of unifying the locations of the fine-tuned model in the train-yourself, and downloaded model case? We could for example, change the path of the downloaded model to match where the final `save_steps` model would reside.

What do you think about that?

docs/pages/example_workflows/locomanipulation/step_3_policy_training.rst
alexmillane: thanks :)

docs/pages/example_workflows/locomanipulation/step_3_policy_training.rst
alexmillane: yea this is bug. Great catch! Thank you!

alexmillane: good catch!

docs/pages/example_workflows/static_manipulation/index.rst
alexmillane: good catch

docs/pages/example_workflows/locomanipulation/index.rst
alexmillane: done.

docs/pages/example_workflows/locomanipulation/step_2_data_generation.rst
alexmillane: Actually I'll add this in a follow up MR so we can review together exactly where we should add this :)

PR #186, opened 2025-11-03
Add RL managers

viiik-inside:
## Summary
Expose the RL managers

### DIFF
--- a/isaaclab_arena/embodiments/embodiment_base.py
+++ b/isaaclab_arena/embodiments/embodiment_base.py
@@ -80,7 +80,7 @@ class EmbodimentBase(Asset):
     def get_curriculum_cfg(self) -> Any:
         return self.curriculum_config
 
-    def get_command_cfg(self) -> Any:
+    def get_commands_cfg(self) -> Any:
         return self.command_config
 
     def get_events_cfg(self) -> Any:


isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -101,6 +106,27 @@ def compose_manager_cfg(self) -> IsaacLabArenaManagerBasedRLEnvCfg:
             bases=(RecorderManagerBaseCfg,),
         )
 
+        rewards_cfg = combine_configclass_instances(
+            "RewardsCfg",
+            self.arena_env.scene.get_rewards_cfg(),
+            self.arena_env.embodiment.get_rewards_cfg(),
+            self.arena_env.task.get_rewards_cfg(),
+        )
+
```
xyao-nv, line 115: It doesn't need to be solved by now. But I wonder what if, rewards depend on task & robot, e.g. eef approaches drawer handle < 3cm dist error, and <10 degree quat error.

viiik-inside: so ideally that will be in the task definition then. The task can take a robot as the input. But if things get too complicated we have the undesired option of using the orchestrator. We will have to think about this though

PR #199, opened 2025-11-06
Xr teleop view spawns aligned with G1

cvolkcvolk:
## Summary
Realign the cloud xr view with GR1's head and add a note that the control panel should be moved to the left before connecting for it to not get occluded. 
Adresses [5634748](https://nvbugspro.nvidia.com/bug/5634748)

### DIFF
--- a/docs/pages/example_workflows/static_manipulation/step_2_teleoperation.rst
+++ b/docs/pages/example_workflows/static_manipulation/step_2_teleoperation.rst
@@ -83,7 +83,7 @@ Follow these steps to record teleoperation demonstrations:
 .. note::
    Before proceeding and pressing the "Connect" button:
    Move the CloudXr Controls Application window to the left of you by pinching the bar at the bottom of the window.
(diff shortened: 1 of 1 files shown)


docs/pages/example_workflows/static_manipulation/step_2_teleoperation.rst
```diff
@@ -79,43 +79,52 @@ Follow these steps to record teleoperation demonstrations:
 
 1. Launch the Isaac XR Teleop app on the Apple Vision Pro
 2. Enter your workstation's IP address in the app window.
+
+.. note::
+   Before proceeding and pressing the "Connect" button:
+   Move the CloudXr Controls Application window to the left of you by pinching the bar at the bottom of the window.
+   Close objects will occlude the window otherwise making it harder to interact with the controls.
```
alexmillane: This sentence sounds a bit odd to me. Suggestion:

"Without doing this, close objects will occlude the window making it harder to interact with the controls."

PR #205, opened 2025-11-10
Alex/fix/g1 reset

alexmillane:
## Summary
Add a term to reset articulations to the G1 embodiment

## Detailed description
- Before the arms would not reset when the env reset. This was causing issues with the arms hitting things.


isaaclab_arena/embodiments/g1/g1.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, enable_cameras: bool = False, initial_pose: Pose | None = Non
         self.camera_config = G1CameraCfg()
         self.action_config = MISSING
         self.observation_config = MISSING
-        self.event_config = MISSING
+        self.event_config = G1EventCfg()
```
xyao-nv, line 59: Is there any use case of this base class directly? I expect users to use `G1WBCJoint` or `G1WBCPink` embodiment class.
 `event_config` is defined in its child classes, see `G1WBCJointEventCfg`, `G1WBCPinkEventCfg`.
 Just checking to make sure if you are aware this overwritten of base class.

alexmillane: Great catch. I was not aware of this duplicated functionaly.

For some reason I was having some issues with the arms hitting things on reset... I thought this improved matters, but I guess I must have had something else going on.

I'll close this. Thanks for catching.

PR #211, opened 2025-11-11
Task timeouts

alexmillane:
## Summary
Adds an optional specification of timeout to tasks.

## Detailed description
- We had the issue that if a policy failed at a task in a way it got stuck, it would not time-out (and therefore fail) until the default time-out, which was set at 30s. This was typically longer that than the evaluation time. In such a scenario the failure would not be registered as the failure, and not affect the success rate. A policy could get 100% success by succeeding once and then getting stuck for ~30s.
- Now each task allows optional specification of a time out.

### DIFF
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -150,7 +150,6 @@ class ArenaEnvBuilder:
                 metrics=metrics,
                 isaaclab_arena_env=isaaclab_arena_env,
                 viewer=viewer_cfg,
-                # episode_length_s=1.0,
             )
             if episode_length_s is not None:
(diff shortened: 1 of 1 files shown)


isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -148,7 +150,10 @@ def compose_manager_cfg(self) -> IsaacLabArenaManagerBasedRLEnvCfg:
                 metrics=metrics,
                 isaaclab_arena_env=isaaclab_arena_env,
                 viewer=viewer_cfg,
+                # episode_length_s=1.0,
```
cvolkcvolk: Leftover comment?

alexmillane: good catch

PR #236, opened 2025-11-26
Doc for Parallel Env Eval

xyao-nv:
Doc changes mainly to include parallel env for closedloop GR00T in 2 example workflows.

### DIFF
--- a/docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
+++ b/docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
@@ -116,7 +116,8 @@ Test the policy in 5 parallel environments with visualization via the GUI run:
      --embodiment g1_wbc_joint
 
 And during the evaluation, you should see the following output on the console at the end of the evaluation
-indicating which environments are terminated or truncated.
+indicating which environments are terminated (task-specific conditions like the brown box is placed into the blue bin),
+or truncated (if timeouts are enabled, like the maximum episode length is exceeded).
 
 .. code-block:: text
 
--- a/docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
+++ b/docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
@@ -125,7 +125,8 @@ Test the policy in 10 parallel environments with visualization via the GUI run:
      --embodiment gr1_joint
 
 And during the evaluation, you should see the following output on the console at the end of the evaluation
-indicating which environments are terminated or truncated.
+indicating which environments are terminated (task-specific conditions like the microwave door is opened),
+or truncated (if timeouts are enabled, like the maximum episode length is exceeded).
 
 .. code-block:: text
 


docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
```diff
@@ -96,6 +96,40 @@ by the quality of post-trained policy, the quality of the dataset, and number of
...
+     --policy_config_yaml_path isaaclab_arena_gr00t/g1_locomanip_gr00t_closedloop_config.yaml \
+     --num_steps 1200 \
+     --num_envs 5 \  # run the policy in 5 parallel environments
+     --enable_cameras \
+     galileo_g1_locomanip_pick_and_place \
+     --object brown_box \
+     --embodiment g1_wbc_joint
+
+And during the evaluation, you should see the following output on the console at the end of the evaluation
+indicating which environments are terminated or truncated.
```
alexmillane: What does truncated mean in this context? It might be worth adding a small explanation of that term in brackets.

docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
```diff
@@ -105,6 +105,40 @@ post-trained policy, the quality of the dataset, and number of steps in the eval
...
+     --policy_type gr00t_closedloop \
+     --policy_config_yaml_path isaaclab_arena_gr00t/gr1_manip_gr00t_closedloop_config.yaml \
+     --num_steps 2000 \
+     --num_envs 10 \  # run the policy in 10 parallel environments
+     --enable_cameras \
+     gr1_open_microwave \
+     --embodiment gr1_joint
+
+And during the evaluation, you should see the following output on the console at the end of the evaluation
+indicating which environments are terminated or truncated.
```
alexmillane: see above about the word truncated.

docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
```diff
@@ -105,6 +105,40 @@ post-trained policy, the quality of the dataset, and number of steps in the eval
...
+
+   Resetting policy for terminated env_ids: tensor([7], device='cuda:0') and truncated env_ids: tensor([], device='cuda:0', dtype=torch.int64)
+
+At the end of the evaluation, you should see the following output on the console indicating the metrics.
+You can see that the success rate is no longer 1.0 as more trials are being evaluated, and the number of episodes is more
+than the single environment evaluation because of the parallel evaluation.
+
+.. code-block:: text
+
+   Metrics: {'success_rate': 0.5833333333333334, 'door_moved_rate': 1.0, 'num_episodes': 120}
```
alexmillane, line 141: Strange that the success rate has dropped by so much. I would have thought that we'd still get 100% here. Do we think that this is a genuine measurement of the policy's performance, or represents some sort of bug in our stuff?

(This question doesn't relate to this MR, just more out of interest).

xyao-nv: Yeah that kinda surprising but somehow expected as well. I observed its failure cases are mainly around

Tabletop
1. After reset, the arm somehow closes the microwave door
2. During rollout of some envs, the arm gets stuck when door is 50%~

Locomanip
1. After reset, the arm somehow punches the box away and misses the grasp
2. During the rollout of some envs, the robot gets stuck after turning

Given how similar these 2 behaviors are, I was looking into 
1 <-> any bugs with reset behavior, it could be us or Lab or PhyX
2 <-> seems like its policy related given how random it is

I will add videos into a slack thread and keep the investigation going. WDYT?

docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
xyao-nv: done

docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
xyao-nv: done

PR #239, opened 2025-11-26
Expose Object Cfgs

viiik-inside:
Create the object cfg when an asset is called. With this one can add properties after calling the asset

## Summary
Create object configuration as soon as an asset is called.

## Detailed description
- We need to do this as users will want to edit the object configuration, earlier it was only possible after a scene is created but now it can be done as soon as the object is called

### DIFF
--- a/isaaclab_arena/tests/test_object_configuration.py
+++ b/isaaclab_arena/tests/test_object_configuration.py
@@ -3,15 +3,13 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import pytest
-
 from isaaclab_arena.tests.utils.subprocess import run_simulation_app_function
 
 HEADLESS = True
 
+
 def _test_object_initial_pose_update(simulation_app):
 
-    from isaaclab_arena.assets.object import Object
     from isaaclab_arena.assets.asset_registry import AssetRegistry
     from isaaclab_arena.utils.pose import Pose
 
@@ -20,7 +18,6 @@ def _test_object_initial_pose_update(simulation_app):
     rigid_object = asset_registry.get_asset_by_name("cracker_box")()
     # Disable debug visualization, this is True by default.
     rigid_object.object_cfg.debug_vis = False
-    initial_pose = rigid_object.get_initial_pose()
 
     # Now lets add an initial pose to the object.
     new_initial_pose = Pose(position_xyz=(5.0, 0.0, 0.0), rotation_wxyz=(1.0, 0.0, 0.0, 0.0))
@@ -28,9 +25,10 @@ def _test_object_initial_pose_update(simulation_app):
 
     # Now lets check that the initial pose has been updated and that the debug visualization is still disabled.
     assert rigid_object.get_initial_pose() == new_initial_pose
-    assert rigid_object.object_cfg.debug_vis == False
+    assert rigid_object.object_cfg.debug_vis is False
 
     return True
+
 
 def test_object_configuration():
     result = run_simulation_app_function(
@@ -38,6 +36,7 @@ def test_object_configuration():
         headless=HEADLESS,
     )
     assert result, "Test failed"
+
 
 if __name__ == "__main__":
     test_object_configuration()


isaaclab_arena/assets/object_base.py
```diff
@@ -43,7 +43,7 @@ def set_prim_path(self, prim_path: str) -> None:
     def get_prim_path(self) -> str:
         return self.prim_path
 
-    def get_cfgs(self) -> dict[str, Any]:
+    def get_cfg(self) -> dict[str, Any]:
```
xyao-nv: Return type is the config class now?

isaaclab_arena/assets/object.py
```diff
@@ -34,6 +34,7 @@ def __init__(
         self.usd_path = usd_path
         self.scale = scale
         self.initial_pose = initial_pose
+        self.object_cfg = self.get_cfg()
```
alexmillane: Suggestion to move this to `ObjectBase` such that we don't have to repeat this in `Object` and in `ObjectReference`

docs/pages/concepts/concept_scene_design.rst
```diff
@@ -37,8 +37,7 @@ It reads all the registered assets and their cfgs and combines them into a confi
          # Combine the configs into a configclass.
          fields: list[tuple[str, type, AssetCfg]] = []
          for asset in self.assets.values():
-               for asset_cfg_name, asset_cfg in asset.get_cfgs().items():
-                  fields.append((asset_cfg_name, type(asset_cfg), asset_cfg))
+               fields.append((asset.name, type(asset.object_cfg), asset.object_cfg))
```
alexmillane, line 40: Suggestion to access `object_cfg` through an accessor function (as it was), and create a new function `_update_cfg()` for updating the internal member (more on this later in the review).

docs/pages/concepts/concept_assets_design.rst
```diff
@@ -22,8 +22,9 @@ The asset system uses a hierarchical class structure:
        def __init__(self, object_type: ObjectType, prim_path: str, **kwargs):
            self.object_type = object_type
            self.prim_path = prim_path
+           self.object_cfg = self.get_cfg()
```
alexmillane, line 25: Suggestion to make this `self.object = self._init_cfg()` or something similar, such that we have one function `_init_cfg()` which is used internally to generate the cfg, and another function `get_cfg()` which is used to retrieve the cfg (in the compiler). See comments elsewhere.

isaaclab_arena/assets/object.py
```diff
@@ -75,7 +76,7 @@ def _generate_base_cfg(self) -> AssetBaseCfg:
         assert self.object_type == ObjectType.BASE
         object_cfg = AssetBaseCfg(
             prim_path="{ENV_REGEX_NS}/" + self.name,
-            spawn=UsdFileCfg(usd_path=self.usd_path),
+            spawn=UsdFileCfg(usd_path=self.usd_path, scale=self.scale),
```
alexmillane, line 80: thanks. two birds with one stone.

I would actually consider separating this into another, single line MR. So we can associate the solve of the issue with a single MR. Also we can merge this earlier as we work out what to do with the cgfs.

isaaclab_arena/assets/object_reference.py
```diff
@@ -19,9 +19,11 @@ class ObjectReference(ObjectBase):
     """An object which *refers* to an existing element in the scene"""
 
     def __init__(self, parent_asset: Asset, **kwargs):
+        # Call the parent class constructor first as we need the parent asset and initial pose relative to the parent to be set.
         super().__init__(**kwargs)
         self.initial_pose_relative_to_parent = self._get_referenced_prim_pose_relative_to_parent(parent_asset)
         self.parent_asset = parent_asset
+        self.object_cfg = self.get_cfg()
```
alexmillane: remove, after moving this behaviour to `ObjectBase`

docs/pages/concepts/concept_assets_design.rst
alexmillane: So this design gets what we want (that object_cfg can be modified from the outside), but has one big issue: the cfg doesn't update with updates to the objects members.

For example, if someone calls (`set_initial_pose`), with this design, the initial pose will not be taken into the, for example, `RigidObjectCfg`. This complicates matters. We need to update the cfgs with any update to any parameter that goes into the cgs.

I think we need to:
* move to protecting our members (like scale etc, by marking them private with a leading `_` e.g. `self._scale`). Then we can be intentional about what we allow.
* Make modification functions update the cfg, for example `set_initial_pose()` is going to have to modify `self.object_cfg`.

isaaclab_arena/tests/test_object_configuration.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_object_configuration():
+    result = run_simulation_app_function(
+        _test_object_initial_pose_update,
+        headless=HEADLESS,
+    )
+    assert result, "Test failed"
+
+if __name__ == "__main__":
+    test_object_configuration()
```
alexmillane, line 42: Thanks for adding this!!

This is great!

isaaclab_arena/assets/object_base.py
```diff
@@ -43,7 +43,10 @@ def set_prim_path(self, prim_path: str) -> None:
     def get_prim_path(self) -> str:
         return self.prim_path
 
-    def get_cfgs(self) -> dict[str, Any]:
+    def get_object_cfg(self) -> dict[str, Any]:
+        return {self.name: self.object_cfg}
+
+    def init_object_cfg(self) -> RigidObjectCfg | ArticulationCfg | AssetBaseCfg:
```
alexmillane: Can we add the `_` to the front of this function to indicate that it's intended to be called internally?

PR #250, opened 2025-12-01
Move to public CI

alexmillane:
## Summary
Move our CI infra to public runners

## Detailed description
- As part of our open-source release, we can no longer run on internal infra.
- This MR moves our runners to public runners.
- I also took the chance to refactor and modularize the workflow file.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -193,6 +193,12 @@ TABLE OF CONTENTS
 
 .. toctree::
    :maxdepth: 1
+   :caption: Advanced
+
+   pages/advanced/private_omniverse
+
+.. toctree::
+   :maxdepth: 1
    :caption: Concepts
 
    pages/concepts/concept_overview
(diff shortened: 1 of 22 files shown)


docker/run_docker.sh
```diff
@@ -139,6 +139,8 @@ else
                     "-v" "./isaaclab_arena:${WORKDIR}/isaaclab_arena"
                     "-v" "./isaaclab_arena_g1:${WORKDIR}/isaaclab_arena_g1"
                     "-v" "./isaaclab_arena_gr00t:${WORKDIR}/isaaclab_arena_gr00t"
+                    "-v" "./docker:${WORKDIR}/docker"
+                    "-v" "./.github:${WORKDIR}/.github"
```
alexmillane, line 143: At some point these were excluded from the dev container. I'm not exactly sure why. However it makes it nicer for development if they're inside. So I'm putting them back in.

PR #253, opened 2025-12-02
Add/ground plane anbd light

viiik-inside:
## Summary
Add ground plane and light objects

### DIFF
--- a/isaaclab_arena/tests/test_asset_registry.py
+++ b/isaaclab_arena/tests/test_asset_registry.py
@@ -132,8 +132,7 @@ def test_all_assets_in_registry():
     assert result, "Test failed"
 
 
-def _test_multiple_lights_and_ground_plane(simulation_app):
-    import isaaclab.sim as sim_utils
+def _test_multi_light_in_scene(simulation_app):
     from pxr import UsdLux
 
     from isaaclab_arena.assets.asset_registry import AssetRegistry
@@ -172,16 +171,13 @@ def _test_multiple_lights_and_ground_plane(simulation_app):
     # We dont add lights from anywhere else in this scene.
     light_prims = [prim for prim in all_prims_in_stage if prim.IsA(UsdLux.DomeLight)]
     assert len(light_prims) == 1
-    # Check that there is only one ground plane in the stage
-    ground_plane_prims = [prim for prim in all_prims_in_stage if prim.IsA(sim_utils.GroundPlaneCfg)]
-    assert len(ground_plane_prims) == 1
     env.close()
     return True
 
 
-def test_multiple_lights_and_ground_plane():
+def test_multi_light_in_scene():
     result = run_simulation_app_function(
-        _test_multiple_lights_and_ground_plane,
+        _test_multi_light_in_scene,
         headless=HEADLESS,
     )
     assert result, "Test failed"
@@ -190,4 +186,4 @@ def test_multiple_lights_and_ground_plane():
 if __name__ == "__main__":
     test_default_assets_registered()
     test_all_assets_in_registry()
-    test_multiple_lights_and_ground_plane()
+    test_multi_light_in_scene()


isaaclab_arena/assets/object.py
```diff
@@ -26,7 +27,9 @@ def __init__(
         initial_pose: Pose | None = None,
         **kwargs,
     ):
-        assert name is not None and usd_path is not None
+        assert name is not None
```
alexmillane: Let's move this down to `Asset`.

Basically the only requirement for asset is that it has a name.

isaaclab_arena/assets/object.py
```diff
@@ -75,13 +78,25 @@ def _generate_articulation_cfg(self) -> ArticulationCfg:
 
     def _generate_base_cfg(self) -> AssetBaseCfg:
         assert self.object_type == ObjectType.BASE
+        with open_stage(self.usd_path) as stage:
+            if has_light(stage):
+                print("WARNING: Base object has lights, this may cause issues when using with multiple environments.")
```
alexmillane, line 82: Nice!

isaaclab_arena/assets/object_library.py
```diff
@@ -230,3 +232,35 @@ class BrownBox(LibraryObject):
...
+
+@register_asset
+class GroundPlane(LibraryObject):
+    """
+    A ground plane.
+    """
+
+    name = "ground_plane"
+    tags = ["ground_plane"]
+    default_prim_path = "/World/GroundPlane"
```
alexmillane, line 246: Let's put a comment here to draw attention to the fact that this is a global prim path, rather than per env.

isaaclab_arena/assets/object_library.py
```diff
@@ -230,3 +232,35 @@ class BrownBox(LibraryObject):
...
+
+@register_asset
+class DomeLight(LibraryObject):
+    """
+    A dome light.
+    """
+
+    name = "dome_light"
+    tags = ["light"]
+    default_prim_path = "/World/defaultDomeLight"
```
alexmillane: As above.

isaaclab_arena/assets/object_library.py
```diff
@@ -230,3 +232,35 @@ class BrownBox(LibraryObject):
...
+    object_type = ObjectType.SPAWNER
+    spawner_cfg = sim_utils.DomeLightCfg(color=(0.75, 0.75, 0.75), intensity=3000.0)
```
alexmillane: Suggestion to take `color` and `intensity` as parameters

alexmillane: Edit: Because we don't want to relect parameters, I'm thinking we should take a whole `DomeLightCfg` as a parameter and overwrite `spawner_cfg` if the input is not `None`.

isaaclab_arena/assets/object_library.py
```diff
@@ -230,3 +232,35 @@ class BrownBox(LibraryObject):
...
+class GroundPlane(LibraryObject):
+    """
+    A ground plane.
+    """
+
+    name = "ground_plane"
+    tags = ["ground_plane"]
+    default_prim_path = "/World/GroundPlane"
+    object_type = ObjectType.SPAWNER
+    spawner_cfg = GroundPlaneCfg()
```
alexmillane: See below. Thinking we should take an optional parameter `ground_plane_cfg`. See below.

isaaclab_arena/tests/test_asset_registry.py
```diff
@@ -70,6 +76,15 @@ def _test_all_assets_in_registry(simulation_app):
...
         objects_in_registry_names.append(asset.name)
+    # Add lights
+    for asset_cls in asset_registry.get_assets_by_tag("light"):
+        asset = asset_cls()
+        objects_in_registry.append(asset)
+        objects_in_registry_names.append(asset.name)
+    # Add ground plane
+    ground_plane = asset_registry.get_asset_by_name("ground_plane")()
+    objects_in_registry.append(ground_plane)
+    objects_in_registry_names.append(ground_plane.name)
```
alexmillane, line 87: Can we add a test which adds multiple lights to a scene and checks that only a single one turns up?

`scene = Scene(assets=[light_1, light_2])`

I feel like the key thing to support here is a single light in the scene, we should check it.

isaaclab_arena/utils/usd_helpers.py
```diff
@@ -34,6 +34,24 @@ def get_all_prims(
...
+        UsdLux.DistantLight,
+        UsdLux.DiskLight,
+    )
+    has_light = False
+    all_prims = get_all_prims(stage)
+    for prim in all_prims:
+        if any(prim.IsA(t) for t in LIGHT_TYPES):
+            has_light = True
+            break
+    return has_light
```
alexmillane, line 52: Nice!

isaaclab_arena/assets/object.py
viiik-inside: yeah

isaaclab_arena/assets/object_library.py
viiik-inside: yes

isaaclab_arena/assets/object_library.py
viiik-inside: So I just leave this as default? This was my intent that people can replace the whole spawner cfg. Infact I can just create light and leave it to the user to add the spawner cfg with whatever they want

isaaclab_arena/tests/test_asset_registry.py
viiik-inside: Hmmm, i think we will add multiple lights if we do this? We will just override the same prim if we use the same light

isaaclab_arena/assets/object_library.py
viiik-inside: done

PR #254, opened 2025-12-03
Add CPU option to run policy closedloop

xyao-nv:
## Summary
Closedloop GN1.5 observed low SR in multi-episode rollout, esp in parallel-env where more contacts are being introduced / more episodes are being observed.

### Static manip
- Issue: At the beginning of episode, hands have close-open motions in recorded trajectories. Given microwave joint is not stiff enough, small deviations during first few inferences cause the door closed by mistake. And this closed door is hard to pull with static GR1, causing it to fail the task.
[Screencast from 12-02-2025 03:16:36 PM.webm](https://github.com/user-attachments/assets/da06de60-8f01-47e7-ae26-a48e08cb523f)

- Fix:  
a. Shorten task_episode_length_s to introduce more frequent resets once the door is closed. Tradeoff is introducing more episodes.
b. Also tried with shorter `action_horizon` but ended up getting worse SR. My hypothesis is that it's hard for VLA to tell from visuals/states whether the door is closed to 0.2 (success) vs 0.21 (fail).

>         16 -- Metrics: {'success_rate': 0.605, 'door_moved_rate': 0.955, 'num_episodes': 200}
>         8 -- Metrics: {'success_rate': 0.225, 'door_moved_rate': 0.615, 'num_episodes': 200}
>         1 -- Metrics: {'success_rate': 0.0, 'door_moved_rate': 0.985, 'num_episodes': 200}

c. Switching to CPU PhyX does not solve above issues. So keep it on GPU for faster parallelization (in theory).
 
### Loco manip
- Issue: After each reset, the left arm tends to have fast motions and the box is tilted. Also observed significant penetration among fingers.
See 00:15 VS 0:30 for 5 parallel env closedloop in below video.
[Screencast from 11-25-2025 03:42:36 PM.webm](https://github.com/user-attachments/assets/c4934817-65fa-412f-a88c-af143d25d7c2)

- Fix: switch to CPU phyX, keep the policy on GPU
Arms open first and G1 starts moving, box is placed with expected pose. 
[Screencast from 12-02-2025 10:15:59 PM.webm](https://github.com/user-attachments/assets/4a02e6cd-7baf-441b-8c0f-7146051e5c9a)

### Minor fixes
Update doc on cmds & metrics.

### DIFF
--- a/docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
+++ b/docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
@@ -145,4 +145,5 @@ and the number of episodes is more than the single environment evaluation becaus
 
 .. note::
 
-   The policy was trained using CPU device physics, therefore the evaluation uses ``--device cpu`` to ensure physics reproducibility.
+   The policy was trained on datasets generated using CPU-based physics, therefore the evaluation uses ``--device cpu`` to ensure physics reproducibility.
+   If you have GPU-generated datasets, you can switch to using GPU-based physics for evaluation by providing the ``--device cuda`` flag.
--- a/isaaclab_arena_gr00t/g1_locomanip_gr00t_closedloop_config.yaml
+++ b/isaaclab_arena_gr00t/g1_locomanip_gr00t_closedloop_config.yaml
@@ -7,6 +7,7 @@
 # NOTE(alexmillane, 2025-10-31): The model path here aligns with the model used in the static manipulation tutorial.
 model_path: /models/isaaclab_arena/locomanipulation_tutorial/checkpoint-20000
 language_instruction: "Pick up the brown box from the shelf, and place it into the blue bin on the table located at the right of the shelf."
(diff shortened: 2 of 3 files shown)


docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
```diff
@@ -111,25 +111,28 @@ Test the policy in 5 parallel environments with visualization via the GUI run:
      --num_steps 1200 \
      --num_envs 5 \  # run the policy in 5 parallel environments
      --enable_cameras \
+     --device cpu \
```
alexmillane, line 114: Unfortunately I don't think we can do this. One of our selling points is that we're running on the GPU. Product wants to run evaluations on this.

I would prefer that we get a lower success rate and stay on the GPU.

xyao-nv: data is generated on CPU. I don't think it makes sense to eval this CPU-trained policy on GPU given what we observed on GPU-physics issue. Whether GR00T/other VLAs can handle this out-of-distribution data is another unrelated question.

PR #273, opened 2025-12-09
Refactor mimic_env_cfg building logic in arena_env_builder

rebeccazhang0707:
## Summary
Refactor `mimic_env_cfg` building logic in `arena_env_builder`.

## Detailed description
    - Originally we need to maintain a list of embodiment_names in each
task's MimicEnvCfg, given its single_arm or dual_arm. It is not
efficient and scalable.
- What has been changed?
    - creates a new enum class `MimicArmMode` to represent the arm mode for the mimic environment: can select from
["single_arm", "dual_arm", "left", "right"]
    - assigns a property of 'mimic_arm_mode' to embodiment_base
    - changes task.get_mimic_env_cfg() method's input from 'embodiment_name'
to 'mimic_arm_mode'
    - refactors the SubTaskConfigs configuration logic in each MimicEnvCfg
based on embodiment.mimic_arm_mode
    - all existing embodiments and tasks with MimicEnvCfg.


isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -19,6 +19,7 @@ class EmbodimentBase(Asset):
 
     name: str | None = None
     tags: list[str] = ["embodiment"]
+    arm_mode: Literal["single_arm", "left", "right"] = "single_arm"
```
alexmillane: Suggestion to use an `Enum` here.

One issue with the literal is that the phrase `Literal["single_arm", "left", "right"]` has to be repeatedly and correct writing in many places in the code. The `Enum` centralizes the definition.

isaaclab_arena/embodiments/gr1t2/gr1t2.py
```diff
@@ -83,6 +83,7 @@ class GR1T2EmbodimentBase(EmbodimentBase):
     """Embodiment for the GR1T2 robot."""
 
     name = "gr1"
+    arm_mode = "right"
```
alexmillane: The GR1T2 has two arms no?

I guess the reasoning here is that we only define the MimicCfg in some of our tasks for one of the arms of the GR1, and this is a way to get around this?

I'm wondering if we could add a `NOTE` here indicating the reason? What do you think?

isaaclab_arena/tasks/open_door_task.py
```diff
@@ -5,7 +5,7 @@
 
 import numpy as np
 from dataclasses import MISSING
-
+from typing import Literal
```
alexmillane: I think that the formatting is incorrect here.

This passed CI because I realize I broke the linting step in the conversion to the public. I'll fix that!

To fix the formatting, you can run `pre-commit run --all-files` once pre-commit is installed and initialized in the repo.

alexmillane (approved):
These look great. Thank you for doing this!

I have a couple of tiny comments.

isaaclab_arena/embodiments/gr1t2/gr1t2.py
peterd-NV: I don't think it's a good idea to leave this defined as a single arm mode since this is the generic gr1t2 embodiment file. The example task we have right now is for one arm but there's two key points here:

1. The GR1T2 can be (and will very shortly be) used with tasks involving both arms 
2. Mimic subtasks still always include both arms. For tasks only using one arm, the idle arm has a single subtask whose end signal is inferred by the Mimic annotation script.

@rebeccazhang0707 What are your thought's on updating this to be dual_arm similar to the G1?

peterd-NV (changes requested):

isaaclab_arena/embodiments/franka/franka.py
```diff
@@ -39,6 +39,7 @@ class FrankaEmbodiment(EmbodimentBase):
     """Embodiment for the Franka robot."""
 
     name = "franka"
+    arm_mode = "single_arm"
```
xyao-nv: Can you use a `Enum` class for `arm_mode`?

isaaclab_arena/tasks/g1_locomanip_pick_and_place_task.py
```diff
@@ -69,7 +69,7 @@ def get_events_cfg(self):
     def get_prompt(self):
         raise NotImplementedError("Function not implemented yet.")
 
-    def get_mimic_env_cfg(self, embodiment_name: str):
+    def get_mimic_env_cfg(self, arm_mode: Literal["single_arm", "left", "right"]):
```
xyao-nv: Can you add in doc, explaining when to use each mode? It's unclear to me in cases where the robot has two arms available, but the trajectory interacted with obj with only one arm.

isaaclab_arena/embodiments/embodiment_base.py
rebeccazhang0707: thanks Alex, I have revised the code according to your suggestion.

isaaclab_arena/embodiments/gr1t2/gr1t2.py
rebeccazhang0707: thanks @alexmillane and @peterd-NV for pointing this out. I have renamed the property from `arm_mode` to `mimic_arm_mode` to avoid misleading, which means it is designed for building `mimic_env_cfg` easily and I have added docs in embodiment design part to explain this property. And users can switch the `mimic_arm_mode` when they initialize a specific embodiment now. Do you think this makes sense? Any suggestions are highly appreciated. Thanks.

isaaclab_arena/tasks/open_door_task.py
rebeccazhang0707: Thanks Alex, I will remember to do the pre-commit before submitting a MR.

isaaclab_arena/embodiments/franka/franka.py
rebeccazhang0707: Thanks @xyao-nv for your suggestion. Have revised the code according to your suggestion. And I have put the `MimicArmMode` enum class in `embodiment/common/mimic_utils.py`. Could you help review and do you think this is a proper position for this class?

isaaclab_arena/tasks/g1_locomanip_pick_and_place_task.py
rebeccazhang0707: I have added docs to explain the `mimic_arm_mode` in the `Embodiment Design` part, could you help review it? Thanks.

xyao-nv (approved):

isaaclab_arena/embodiments/common/mimic_utils.py
```diff
@@ -30,3 +31,20 @@ def pose_from(obj_state) -> "torch.Tensor":
...
+        SINGLE_ARM: Single arm mode (the robot has only one arm).
+        DUAL_ARM: Dual arm mode (bimanual robot, task is performed with both arms in the demonstration).
+        LEFT: Left arm mode (bimanual robot, task is performed with the left arm in the demonstration, right arm is idle).
+        RIGHT: Right arm mode (bimanual robot, task is performed with the right arm in the demonstration, left arm is idle).
+    """
+
+    SINGLE_ARM = "single_arm"
+    DUAL_ARM = "dual_arm"
+    LEFT = "left"
+    RIGHT = "right"
```
alexmillane: Suggestion to move this to it's own file.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -19,10 +20,14 @@ class EmbodimentBase(Asset):
...
     tags: list[str] = ["embodiment"]
+    mimic_arm_mode: MimicArmMode | None = None
 
-    def __init__(self, enable_cameras: bool = False, initial_pose: Pose | None = None):
+    def __init__(
+        self, enable_cameras: bool = False, initial_pose: Pose | None = None, mimic_arm_mode: MimicArmMode | None = None
+    ):
         self.enable_cameras = enable_cameras
         self.initial_pose = initial_pose
+        self.mimic_arm_mode = mimic_arm_mode or type(self).mimic_arm_mode
```
alexmillane: The repeated use of the variable which overlaps with a static variable is a bit confusing. Suggestion to

static `mimic_arm_mode` -> `default_mimic_arm_mode`

Such that:

`self.mimic_arm_mode = mimic_arm_mode or type(self).mimic_arm_mode`

becomes

`self.mimic_arm_mode = mimic_arm_mode or self.default_mimic_arm_mode`

(I'm not sure why the additional `type` is needed in the current implementation).

isaaclab_arena/tasks/open_door_task.py
```diff
@@ -199,11 +200,12 @@ def __post_init__(self):
...
         )
-        if self.embodiment_name == "franka":
+        if self.arm_mode == "single_arm":
             self.subtask_configs["robot"] = subtask_configs
         # We need to add the left and right subtasks for GR1.
-        elif self.embodiment_name == "gr1_pink":
-            self.subtask_configs["right"] = subtask_configs
+        elif self.arm_mode in ["left", "right"]:
+            self.another_arm_mode = "left" if self.arm_mode == "right" else "right"
+            self.subtask_configs[self.arm_mode] = subtask_configs
```
alexmillane: Suggestion to use the `Enum` for comparisons.

`if self.arm_mode == ArmMode.SINGLE_ARM:`

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -219,11 +220,12 @@ def __post_init__(self):
                 apply_noise_during_interpolation=False,
             )
         )
-        if self.embodiment_name == "franka":
+        if self.arm_mode == "single_arm":
```
alexmillane: see above

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -219,11 +220,12 @@ def __post_init__(self):
...
             self.subtask_configs["robot"] = subtask_configs
         # We need to add the left and right subtasks for GR1.
-        elif self.embodiment_name == "gr1_pink":
-            self.subtask_configs["right"] = subtask_configs
+        elif self.arm_mode in ["left", "right"]:
+            self.another_arm_mode = "left" if self.arm_mode == "right" else "right"
```
alexmillane: Suggestion to add this as a method to the `ArmMode` `Enum`

Something like

```
class ArmMode(Enum):
...
   def get_other_arm(self: ArmMode):
       asset self in [ArmMode.LEFT, ArmMode.RIGHT]
       if self == ArmMode.LEFT:
             return ArmMode.RIGHT
       else:
             return ArmMode.LEFT
```

Then your code becomes:
```
self.another_arm_mode = self.arm_mode.get_other_arm()
```

alexmillane (approved):
It's looking great!

peterd-NV (approved):

PR #281, opened 2025-12-10
Package example_environments into isaaclab_arena_environments

xyao-nv:
## Summary
All example environments.py are repackaged into isaaclab_arena_environments

## Reason
In prep for multi-task evaluation, as we may introduce more example envs.


cfg_entry.yaml
```diff
@@ -0,0 +1,2145 @@
+viewer:
```
peterd-NV: Is this file required?

xyao-nv: Good catch. Removed it. Thx!

PR #285, opened 2025-12-12
Add factory_assembly_task and the relevant task examples

yami007007:
## Summary
Add a generic `assembly_task` (bring two objects into proximity) and corresponding task examples: peg insertion and gear meshing.

## Detailed description
- Why this change?
    - The assembly task is a common pattern for bringing two generic objects into proximity.
    - `peg_insert` and `gear_mesh` examples are added to demonstrate how to use this atomic task in practice.
- What was changed?
    - Added a new atomic task: `assembly_task`, plus a test script for this task.
    - Added two new examples: `peg_insert` and `gear_mesh`.

## Open issues
For the `peg_insert` task, the current peg and hole assets are not compatible with Pinocchio. Ideally, this configuration should be defined at the task environment level so that everyone uses the correct setup by default. However, an effective solution has not been found yet, so this is left to Vikram Ramasamy for follow-up (see the discussion on [slack](https://nvidia.slack.com/archives/C09Q0PD199V/p1765965147252899) ).

### Dataset Preparation
Please dowload the pre-recorded hdf5 files from [link](https://nvidia-my.sharepoint.com/:f:/p/weihuaz/IgCLFpxMxwKdSpcQZMwhLF02AXDh-IIG9HYixfm4t--eY3s?e=CK85Xn) and place them in the `datasets` folder.

#### Peg insertion task example
```
python isaaclab_arena/examples/policy_runner.py \
        --policy_type zero_action \
        --disable_pinocchio \
        --environment isaaclab_arena_environments.tabletop_peginsert_environment:PegInsertEnvironment \
         peg_insert
```

```
python isaaclab_arena/scripts/imitation_learning/record_demos.py \
        --dataset_file datasets/franka_peginsert.hdf5 \
        --disable_pinocchio \
        --num_demos 1 \
        --environment isaaclab_arena_environments.tabletop_peginsert_environment:PegInsertEnvironment \
         peg_insert \
         --teleop_device keyboard
```

Since AVP teleop support is not available for Franka, I used an old HDF5 dataset with AbsEEF + binary gripper. Before running `replay_demos.py`, please update `franka.py` as follows, then you can use the original HDF5 dataset for demonstration replay.
1. Set `use_relative_mode` from True to False.
2. Set `scale` from 0.5 to 1.

> 
> Original implementation of arm_action in `franka.py`.
> 
>     arm_action: ActionTermCfg = DifferentialInverseKinematicsActionCfg(
>         asset_name="robot",
>         joint_names=["panda_joint.*"],
>         body_name="panda_hand",
>         controller=DifferentialIKControllerCfg(command_type="pose", use_relative_mode=True, ik_method="dls"),
>         scale=0.5,
>         body_offset=DifferentialInverseKinematicsActionCfg.OffsetCfg(pos=[0.0, 0.0, 0.107]),
>     )
> 

```
python isaaclab_arena/scripts/imitation_learning/replay_demos.py \
        --dataset_file datasets/franka_peginsert_10demos.hdf5 \
        --disable_pinocchio \
        --environment isaaclab_arena_environments.tabletop_peginsert_environment:PegInsertEnvironment \
        peg_insert
```
#### Gear meshing task example
```
python isaaclab_arena/examples/policy_runner.py \
        --policy_type zero_action \
        --environment isaaclab_arena_environments.tabletop_gearmesh_environment:GearMeshEnvironment \
         gear_mesh
```

```
python isaaclab_arena/scripts/imitation_learning/record_demos.py \
        --dataset_file datasets/franka_gearmesh.hdf5 \
        --num_demos 1 \
        --environment isaaclab_arena_environments.tabletop_gearmesh_environment:GearMeshEnvironment \
         gear_mesh \
         --teleop_device keyboard
```

For the same reason as above, please update the `arm_action` in `franka.py` accordingly.    
```
python isaaclab_arena/scripts/imitation_learning/replay_demos.py \
        --dataset_file datasets/franka_gearmesh_6demos.hdf5 \
        --environment isaaclab_arena_environments.tabletop_gearmesh_environment:GearMeshEnvironment \
        gear_mesh
```

### DIFF
--- a/isaaclab_arena/assets/object_library.py
+++ b/isaaclab_arena/assets/object_library.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
--- a/isaaclab_arena/assets/object_utils.py
+++ b/isaaclab_arena/assets/object_utils.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
--- a/isaaclab_arena/embodiments/franka/franka.py
+++ b/isaaclab_arena/embodiments/franka/franka.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
--- a/isaaclab_arena/tasks/assembly_task.py
+++ b/isaaclab_arena/tasks/assembly_task.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
--- a/isaaclab_arena/tasks/events.py
+++ b/isaaclab_arena/tasks/events.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
--- a/isaaclab_arena_environments/mdp/__init__.py
+++ b/isaaclab_arena_environments/mdp/__init__.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
--- a/isaaclab_arena_environments/tabletop_gearmesh_environment.py
+++ b/isaaclab_arena_environments/tabletop_gearmesh_environment.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
--- a/isaaclab_arena_environments/tabletop_peginsert_environment.py
+++ b/isaaclab_arena_environments/tabletop_peginsert_environment.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
--- a/isaaclab_arena/tests/test_assembly_task.py
+++ b/isaaclab_arena/tests/test_assembly_task.py
@@ -1,12 +1,13 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
 
 
 import gymnasium as gym
-import pytest
 import torch
+
+import pytest
 
 from isaaclab_arena.tests.utils.subprocess import run_simulation_app_function
 
@@ -389,7 +390,10 @@ def _test_gear_mesh_initialization(simulation_app) -> bool:
 
 # Test functions that will be called by pytest
 @pytest.mark.skip(
-    reason="Requires enable_pinocchio=False. Run separately: pytest -sv isaaclab_arena/tests/test_assembly_task.py::test_peg_insert_assembly_single"
+    reason=(
+        "Requires enable_pinocchio=False. Run separately: pytest -sv"
+        " isaaclab_arena/tests/test_assembly_task.py::test_peg_insert_assembly_single"
+    )
 )
 def test_peg_insert_assembly_single():
     result = run_simulation_app_function(_test_peg_insert_assembly_single, headless=HEADLESS, enable_pinocchio=False)
@@ -402,7 +406,10 @@ def test_gear_mesh_assembly_single():
 
 
 @pytest.mark.skip(
-    reason="Requires enable_pinocchio=False. Run separately: pytest -sv isaaclab_arena/tests/test_assembly_task.py::test_peg_insert_assembly_multi"
+    reason=(
+        "Requires enable_pinocchio=False. Run separately: pytest -sv"
+        " isaaclab_arena/tests/test_assembly_task.py::test_peg_insert_assembly_multi"
+    )
 )
 def test_peg_insert_assembly_multi():
     result = run_simulation_app_function(_test_peg_insert_assembly_multi, headless=HEADLESS, enable_pinocchio=False)
@@ -415,7 +422,10 @@ def test_gear_mesh_assembly_multi():
 
 
 @pytest.mark.skip(
-    reason="Requires enable_pinocchio=False. Run separately: pytest -sv isaaclab_arena/tests/test_assembly_task.py::test_peg_insert_initialization"
+    reason=(
+        "Requires enable_pinocchio=False. Run separately: pytest -sv"
+        " isaaclab_arena/tests/test_assembly_task.py::test_peg_insert_initialization"
+    )
 )
 def test_peg_insert_initialization():
     """
--- a/isaaclab_arena_environments/mdp/env_callbacks.py
+++ b/isaaclab_arena_environments/mdp/env_callbacks.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
--- a/isaaclab_arena_environments/mdp/robot_configs.py
+++ b/isaaclab_arena_environments/mdp/robot_configs.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0


.gitignore
```diff
@@ -35,7 +35,10 @@
 *.app
 
 # IDE stuff
-*.vscode
+**/.vscode/
```
cvolkcvolk: Generally VS Code config is usually user-specific.
If the intention is to share only workspace-level configs (extensions, launch, tasks), it would be safer to explicitly whitelist those files instead of allowing the entire .vscode/ directory. 
Here contributors don’t know which files are allowed and if I understand it correctly any file under .vscode/ toplevel is allowed to be committed.

Id suggest to point out the files explicitley eg something like
```
.vscode/*
!.vscode/tasks.json
...
```

isaaclab_arena/assets/object_library.py
```diff
@@ -278,3 +284,129 @@ def __init__(
...
+    object_type = ObjectType.ARTICULATION
+    scale = (3.0, 3.0, 3.0)
+    mass = 0.05
+    rigid_props = RIGID_BODY_PROPS_HIGH_PRECISION
+
+
+@register_asset
+class SmallGear(FactoryObject):
+    """
+    A small reference gear for gear mesh task.
```
cvolkcvolk: Nit, this could be used by other tasks as well not just for "gear mesh task"?  Suggest to update the comment

isaaclab_arena/assets/background_library.py
```diff
@@ -116,3 +116,19 @@ class GalileoLocomanipBackground(LibraryBackground):
 
     def __init__(self):
         super().__init__()
+
+
+@register_asset
+class FactoryTableBackground(LibraryBackground):
```
viiik-inside: Lets maybe call this the SeattleLabTable :)

isaaclab_arena/assets/object_library.py
```diff
@@ -278,3 +284,129 @@ def __init__(
...
+    """
+
+    # Class attributes that must be overridden by subclasses
+    mass: float
+    rigid_props: RigidBodyPropertiesCfg
+
+    def __init__(self, prim_path: str | None = None, initial_pose: Pose | None = None):
+        super().__init__(prim_path=prim_path, initial_pose=initial_pose)
+
+        # Create factory-specific articulation configuration
```
viiik-inside: I understand that this is being done so one can set other properties to the articulation. I think we should fix this directly where we generally create the articulations so we dont have extra functions.

isaaclab_arena/tasks/factory_assembly_task.py
```diff
@@ -0,0 +1,214 @@
...
+from isaaclab_arena.assets.asset import Asset
+from isaaclab_arena.metrics.metric_base import MetricBase
+from isaaclab_arena.metrics.object_moved import ObjectMovedRateMetric
+from isaaclab_arena.metrics.success_rate import SuccessRateMetric
+from isaaclab_arena.tasks.task_base import TaskBase
+from isaaclab_arena.tasks.terminations import objects_in_proximity
+from isaaclab_arena.terms.events import set_object_pose
+from isaaclab_arena.utils.cameras import get_viewer_cfg_look_at_object
+
+FRANKA_PANDA_FACTORY_HIGH_PD_CFG = FRANKA_PANDA_HIGH_PD_CFG.copy()
```
viiik-inside: Do you think it would be nice to have this as a function ? Maybe we will also use this for other tasks?

isaaclab_arena/tasks/factory_assembly_task.py
```diff
@@ -0,0 +1,214 @@
...
+        assist_asset_list: list[Asset],
+        background_scene: Asset,
+        episode_length_s: float | None = None,
+    ):
+        super().__init__(episode_length_s=episode_length_s)
+        self.fixed_asset = fixed_asset
+        self.held_asset = held_asset
+        self.assist_asset_list = assist_asset_list
+        self.background_scene = background_scene
+        self.scene_config = InteractiveSceneCfg(num_envs=1, env_spacing=3.0, replicate_physics=False)
```
viiik-inside: Do we need this here? If yes, maybe lets make them parameters

isaaclab_arena/tasks/factory_assembly_task.py
```diff
@@ -0,0 +1,214 @@
...
+    def get_termination_cfg(self):
+        return self.termination_cfg
+
+    def _make_termination_cfg(self):
+        success = TerminationTermCfg(
+            func=objects_in_proximity,
+            params={
+                "object_cfg": SceneEntityCfg(self.held_asset.name),
+                "target_object_cfg": SceneEntityCfg(self.fixed_asset.name),
+                "max_x_separation": 0.020,  # Tolerance for assembly alignment
```
viiik-inside: Would be nice to have these as parameters. Maybe a dataclass ?

isaaclab_arena_environments/mdp/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""This sub-module contains the functions that are specific to the environment."""
+
+from isaaclab.envs.mdp import *  # noqa: F401, F403
```
viiik-inside, line 9: Is there a reason why this folder is called mdp? Just curious

isaaclab_arena_environments/tabletop_gearmesh_environment.py
```diff
@@ -0,0 +1,143 @@
...
+        class EventCfgGearMesh:
+            """Configuration for events."""
+
+            reset_all = EventTerm(func=mdp.reset_scene_to_default, mode="reset", params={"reset_joint_targets": True})
+
+            randomize_gear_positions = EventTerm(
+                func=randomize_object_serials_pose,
+                mode="reset",
+                params={
+                    "pose_range": {"x": (0.25, 0.6), "y": (-0.20, 0.20), "z": (0.0, 0.0), "yaw": (-1.0, 1.0)},
```
viiik-inside: We can use our pose dataclass here

isaaclab_arena_environments/tabletop_gearmesh_environment.py
```diff
@@ -0,0 +1,143 @@
...
+        from isaaclab_arena.assets.object_library import GearBase, LargeGear, Light, MediumGear, SmallGear
+        from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
+        from isaaclab_arena.scene.scene import Scene
+        from isaaclab_arena.tasks.factory_assembly_task import FRANKA_PANDA_FACTORY_HIGH_PD_CFG, FactoryAssemblyTask
+        from isaaclab_arena.utils.pose import Pose
+        from isaaclab_arena_environments import mdp
+        from isaaclab_arena_environments.mdp.events import randomize_object_serials_pose
+
+        @configclass
+        class EventCfgGearMesh:
```
viiik-inside: Can this be in the task class?

isaaclab_arena_environments/tabletop_peginsert_environment.py
```diff
@@ -0,0 +1,120 @@
...
+        from isaaclab_arena.assets.object_library import Hole, Light, Peg
+        from isaaclab_arena.embodiments.franka.franka import FrankaEmbodiment
+        from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
+        from isaaclab_arena.scene.scene import Scene
+        from isaaclab_arena.tasks.factory_assembly_task import FRANKA_PANDA_FACTORY_HIGH_PD_CFG, FactoryAssemblyTask
+        from isaaclab_arena.utils.pose import Pose
+        from isaaclab_arena_environments import mdp
+
+        @configclass
+        class EventCfgPegInsert:
```
viiik-inside: If it has to be here, then we can add it outside this function, on top of the file maybe?

isaaclab_arena/assets/background_library.py
```diff
@@ -116,3 +116,19 @@ class GalileoLocomanipBackground(LibraryBackground):
...
+    """
+
+    name = "factory_table"
+    tags = ["background"]
+    usd_path = f"{ISAAC_NUCLEUS_DIR}/Props/Mounts/SeattleLabTable/table_instanceable.usd"
+    initial_pose = Pose(position_xyz=(0.55, 0.0, 0.0), rotation_wxyz=(0.707, 0, 0, 0.707))
+    object_min_z = -0.1
+
+    def __init__(self):
+        super().__init__(scale=(1.0, 1.0, 1.0))
```
alexmillane: Why are we passing the scale of 1.0 here? I would have thought that the default scale is 1.0? Can we remove this?

isaaclab_arena/assets/object_utils.py
```diff
@@ -59,3 +62,79 @@ def detect_object_type(usd_path: str | None = None, stage: Usd.Stage | None = No
...
+
+def create_factory_articulation_cfg(
+    prim_path: str,
+    usd_path: str,
+    scale: tuple[float, float, float],
+    mass: float,
+    rigid_props: sim_utils.RigidBodyPropertiesCfg = RIGID_BODY_PROPS_HIGH_PRECISION,
+    contact_offset: float = 0.005,
+    rest_offset: float = 0.0,
+) -> ArticulationCfg:
```
alexmillane: To echo what @viiik-inside said above, we need to expose this functionality through our `Object` class. I would suggest that we make a proposal next week, formulated as an MR against this one.

isaaclab_arena_environments/mdp/events.py
```diff
@@ -0,0 +1,69 @@
...
+            asset.write_root_pose_to_sim(
+                torch.cat([positions, orientations], dim=-1), env_ids=torch.tensor([cur_env], device=env.device)
+            )
+            asset.write_root_velocity_to_sim(
+                torch.zeros(1, 6, device=env.device), env_ids=torch.tensor([cur_env], device=env.device)
+            )
+
+            if (
+                i == 0
+            ):  # set the "small_gear" and "large_gear" positions according the position of fixed_asset('gear_base')
```
alexmillane: This seems to make some assumptions about the order of the assets in the list. Is that correct? I would guess that it makes the assumption that the "fixed_asset" appears first in the list. Could we enforce this assumption using the function parameters? For example we could do something like:

```
def randomize_object_serials_pose(
    env: ManagerBasedEnv,
    env_ids: torch.Tensor,
    fixed_asset_cfg: SceneEntityCfg,
    asset_cfgs: list[SceneEntityCfg],
    min_separation: float = 0.0,
    pose_range: dict[str, tuple[float, float]] = {},
    max_sample_tries: int = 5000,
    relative_asset_cfgs: SceneEntityCfg | None = None,
):
```

Such that the fixed_asset is passed separately. That way we remove assumptions about the order of assets in the list.

isaaclab_arena_environments/mdp/events.py
```diff
@@ -0,0 +1,69 @@
...
+            pose_tensor = torch.tensor([pose_list[i]], device=env.device)
+            positions = pose_tensor[:, 0:3] + env.scene.env_origins[cur_env, 0:3]
+            orientations = math_utils.quat_from_euler_xyz(pose_tensor[:, 3], pose_tensor[:, 4], pose_tensor[:, 5])
+
+            asset.write_root_pose_to_sim(
+                torch.cat([positions, orientations], dim=-1), env_ids=torch.tensor([cur_env], device=env.device)
+            )
+            asset.write_root_velocity_to_sim(
+                torch.zeros(1, 6, device=env.device), env_ids=torch.tensor([cur_env], device=env.device)
+            )
```
alexmillane, line 67: The code to get an asset and set it's pose and velocity as zero is repeated in this file. Here and below.

Suggestion to move it out to another function in this same file. Something like:

```
def set_pose_and_zero_velocity(env, asset_cfg, positions, orientations):
```

This will make this function more easily understood.

isaaclab_arena_environments/mdp/events.py
```diff
@@ -0,0 +1,69 @@
...
+
+def randomize_object_serials_pose(
+    env: ManagerBasedEnv,
+    env_ids: torch.Tensor,
+    asset_cfgs: list[SceneEntityCfg],
+    min_separation: float = 0.0,
+    pose_range: dict[str, tuple[float, float]] = {},
+    max_sample_tries: int = 5000,
+    relative_asset_cfgs: SceneEntityCfg | None = None,
+):
```
alexmillane: Suggestion to rename this function to something that indicates what it does. It's more specific than just randomizing poses. It randomizes poses, and then does something special with the relative assets.

Perhaps a docstring would help?

isaaclab_arena_environments/mdp/events.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+
+            if (
+                i == 0
+            ):  # set the "small_gear" and "large_gear" positions according the position of fixed_asset('gear_base')
+                for j in range(len(relative_asset_cfgs)):
+                    rel_asset_cfg = relative_asset_cfgs[j]
+                    rel_asset = env.scene[rel_asset_cfg.name]
+                    rel_asset.write_root_pose_to_sim(
+                        torch.cat([positions, orientations], dim=-1), env_ids=torch.tensor([cur_env], device=env.device)
```
alexmillane, line 83: Are these the same positions and orientations used above? I'm not sure I understand that?

isaaclab_arena_environments/tabletop_gearmesh_environment.py
```diff
@@ -0,0 +1,143 @@
...
+        scene = Scene(assets=[background, gear_base, medium_gear, small_gear, large_gear, light])
+
+        # Create gear mesh task
+        task = FactoryAssemblyTask(
+            fixed_asset=gear_base,
+            held_asset=medium_gear,
+            assist_asset_list=[small_gear, large_gear],
+            background_scene=background,
+        )
+        task.events_cfg = EventCfgGearMesh()
```
alexmillane: We replace the event_cfg in the underlying task. Do we ever use the original event_cfg defined in `FactoryAssemblyTask`?

isaaclab_arena/tasks/factory_assembly_task.py
```diff
@@ -0,0 +1,214 @@
...
+FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_shoulder"].stiffness = 150.0
+FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_shoulder"].damping = 30.0
+FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_forearm"].stiffness = 150.0
+FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_forearm"].damping = 30.0
+FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_hand"].stiffness = 150.0
+FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_hand"].damping = 30.0
+FRANKA_PANDA_FACTORY_HIGH_PD_CFG.init_state.pos = (0.0, 0.0, 0.0)  # for factory assembly task
+
+
+class FactoryAssemblyTask(TaskBase):
```
alexmillane: `FactoryAssemblyTask` -> `AssemblyTask`

There's nothing specific to factories here.

isaaclab_arena/tasks/factory_assembly_task.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    Factory assembly task where an object needs to be assembled with a base object, like peg insert, gear mesh, etc.
+    """
```
alexmillane: Suggestion to expand this docstring. From what I can tell, the goal in this task is to bring two objects into proximity. There are some additional (distractor?) objects in the scene (which are not involved in the task but are there to make things harder?).

If possible, could you expand the docstring to make that clear?

isaaclab_arena_environments/tabletop_gearmesh_environment.py
```diff
@@ -0,0 +1,143 @@
...
+            randomize_gear_positions = EventTerm(
+                func=randomize_object_serials_pose,
+                mode="reset",
+                params={
+                    "pose_range": {"x": (0.25, 0.6), "y": (-0.20, 0.20), "z": (0.0, 0.0), "yaw": (-1.0, 1.0)},
+                    "min_separation": 0.18,
+                    "asset_cfgs": [SceneEntityCfg("gear_base"), SceneEntityCfg("medium_gear")],
+                    "relative_asset_cfgs": [SceneEntityCfg("small_gear"), SceneEntityCfg("large_gear")],
+                },
+            )
```
alexmillane: Could we think of a way of incorperating this `EventCfg` in the `FactoryAssemblyTask`.

For example, the `FactoryAssemblyTask` could take an additional parameter `randomization_mode` where we select among a few different options for randomizing the assets.

What do you think? Is that possible?

isaaclab_arena_environments/tabletop_peginsert_environment.py
```diff
@@ -0,0 +1,120 @@
...
+                    "pose_range": {
+                        "x": (0.2, 0.6),
+                        "y": (-0.20, 0.20),
+                        "z": (0.0, 0.0),
+                        "yaw": (-1.0, 1.0),
+                    },
+                    "min_separation": 0.1,
+                    "asset_cfgs": [SceneEntityCfg("peg"), SceneEntityCfg("hole")],
+                },
+            )
```
alexmillane: Same as my comment above: what do you think about incorperating this into `FactoryAssemblyTask` with an option for reset mode.

isaaclab_arena/tasks/factory_assembly_task.py
yami007007: Hi @viiik-inside, what do you mean by “having this as a function”? Do you mean wrapping all of this into a standalone function?
This is a reference implementation of [franka.py](https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab_assets/isaaclab_assets/robots/franka.py#L77-L83) in Isaac Lab. I've put these configs in `franka.py` of Arena embodiments. I'd like to know your suggestions on this.

```python
from isaaclab_assets.robots.franka import FRANKA_PANDA_HIGH_PD_CFG
FRANKA_PANDA_FACTORY_HIGH_PD_CFG = FRANKA_PANDA_HIGH_PD_CFG.copy()
FRANKA_PANDA_FACTORY_HIGH_PD_CFG.spawn.activate_contact_sensors = True
FRANKA_PANDA_FACTORY_HIGH_PD_CFG.spawn.rigid_props.disable_gravity = True
FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_shoulder"].stiffness = 150.0
FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_shoulder"].damping = 30.0
FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_forearm"].stiffness = 150.0
FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_forearm"].damping = 30.0
FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_hand"].stiffness = 150.0
FRANKA_PANDA_FACTORY_HIGH_PD_CFG.actuators["panda_hand"].damping = 30.0
FRANKA_PANDA_FACTORY_HIGH_PD_CFG.init_state.pos = (0.0, 0.0, 0.0)  # for factory assembly task```

File: isaaclab_arena_environments/mdp/__init__.py (hunk 8 above)
Comment (yami007007) [reply isaaclab_arena_environments/mdp/__init__.py:9]: This is following the convention used in IsaacLab [mdp](https://github.com/isaac-sim/IsaacLab/tree/main/source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/stack/mdp).  The events, observations, and terminations for the task example are placed in the `mdp` folder.

File: isaaclab_arena/assets/object_utils.py (hunk 13 above)
Comment (yami007007) [reply isaaclab_arena/assets/object_utils.py]: Modify the code based on the new MR.

File: isaaclab_arena_environments/tabletop_gearmesh_environment.py (hunk 21 above)
Comment (yami007007) [reply isaaclab_arena_environments/tabletop_gearmesh_environment.py]: Good suggestion, thanks. I’ve updated the code accordingly.

File: isaaclab_arena/assets/object_utils.py [hunk 23]
```diff
@@ -59,3 +61,38 @@ def detect_object_type(usd_path: str | None = None, stage: Usd.Stage | None = No
...
+    enable_gyroscopic_forces=True,
+    solver_position_iteration_count=32,
+    solver_velocity_iteration_count=32,
+    max_contact_impulse=1e32,
+)
+
+ASSEMBLY_ARTICULATION_INIT_STATE = ArticulationCfg.InitialStateCfg(
+    joint_pos={},
+    joint_vel={},
+)
```
alexmillane: What's the idea here? I guess these empty dicts differ from what is there by default?

A couple of suggestions:
- Add a comment to clarify what the purpose of this initial state cfg is,
- Generalize the name. I presume that this could be used outside of an assembly setting? Maybe something like `EMPTY_ARTICULATION_INIT_STATE_CFG`? Would that make sense?

isaaclab_arena/assets/object_utils.py
```diff
@@ -59,3 +61,38 @@ def detect_object_type(usd_path: str | None = None, stage: Usd.Stage | None = No
...
+RIGID_BODY_PROPS_STANDARD = sim_utils.RigidBodyPropertiesCfg(
+    disable_gravity=False,
+    max_depenetration_velocity=5.0,
+    linear_damping=0.0,
+    angular_damping=0.0,
+    max_linear_velocity=1000.0,
+    max_angular_velocity=3666.0,
+    enable_gyroscopic_forces=True,
+    solver_position_iteration_count=32,
+    solver_velocity_iteration_count=32,
```
alexmillane: I find this name a little confusing. I would say that these settings are not standard, because I assume leaving the settings at their default would be "standard". What do you think about `RIGID_BODY_PROPS_MEDIUM_PRECISION`?

isaaclab_arena/embodiments/franka/franka.py
```diff
@@ -34,6 +37,17 @@
...
+FRANKA_PANDA_ASSEMBLY_HIGH_PD_CFG = FRANKA_PANDA_HIGH_PD_CFG.copy()
+FRANKA_PANDA_ASSEMBLY_HIGH_PD_CFG.spawn.activate_contact_sensors = True
+FRANKA_PANDA_ASSEMBLY_HIGH_PD_CFG.spawn.rigid_props.disable_gravity = True
+FRANKA_PANDA_ASSEMBLY_HIGH_PD_CFG.actuators["panda_shoulder"].stiffness = 150.0
+FRANKA_PANDA_ASSEMBLY_HIGH_PD_CFG.actuators["panda_shoulder"].damping = 30.0
+FRANKA_PANDA_ASSEMBLY_HIGH_PD_CFG.actuators["panda_forearm"].stiffness = 150.0
+FRANKA_PANDA_ASSEMBLY_HIGH_PD_CFG.actuators["panda_forearm"].damping = 30.0
+FRANKA_PANDA_ASSEMBLY_HIGH_PD_CFG.actuators["panda_hand"].stiffness = 150.0
+FRANKA_PANDA_ASSEMBLY_HIGH_PD_CFG.actuators["panda_hand"].damping = 30.0
+FRANKA_PANDA_ASSEMBLY_HIGH_PD_CFG.init_state.pos = (0.0, 0.0, 0.0)
```
alexmillane: As I understand it, this code is not used in this file. It is only used in the `isaaclab_arena_environments` package. Could we move it into a (subfolder) in the `isaaclab_arena_environments` package?

isaaclab_arena/tasks/assembly_task.py
```diff
@@ -0,0 +1,191 @@
...
+    def __init__(
+        self,
+        fixed_asset: Asset,
+        held_asset: Asset,
+        auxiliary_asset_list: list[Asset],
+        background_scene: Asset,
+        episode_length_s: float | None = None,
+        termination_cfg: TerminationConfig = TerminationConfig(
+            max_x_separation=0.020, max_y_separation=0.020, max_z_separation=0.020
+        ),
```
alexmillane: Suggestion to just use a `tuple` here.

Introduction of another class `TerminationConfig`, is kind of confusing because we have several classes called similar things in this file: `TerminationTermCfg`, `TerminationsCfg`

isaaclab_arena/tasks/events.py
```diff
@@ -0,0 +1,78 @@
...
+    randomization_mode: int = 0,
+):
+    """
+    Randomize object poses and update the poses of related assets accordingly.
+
+    Args:
+        randomization_mode (int):
+            0: Randomize the poses of the fixed and held assets.
+            1: Randomize the poses of the fixed, held, and auxiliary assets. The poses
+            of the auxiliary assets are set relative to the pose of the fixed asset.
```
alexmillane: `randomization_mode: Literal["held_and_fixed_only", "held_fixed_and_auxiliary"]`

isaaclab_arena/tasks/events.py
```diff
@@ -0,0 +1,78 @@
...
+                # set the poses of the auxiliary assets according to the pose of the fixed asset
+                for j in range(len(auxiliary_asset_cfgs)):
+                    rel_asset_cfg = auxiliary_asset_cfgs[j]
+                    rel_asset = env.scene[rel_asset_cfg.name]
+                    rel_asset.write_root_pose_to_sim(
+                        torch.cat([positions, orientations], dim=-1), env_ids=torch.tensor([cur_env], device=env.device)
+                    )
+                    rel_asset.write_root_velocity_to_sim(
+                        torch.zeros(1, 6, device=env.device), env_ids=torch.tensor([cur_env], device=env.device)
+                    )
```
alexmillane, line 87: This seems strange to me. It would appear that set the pose of the fixed asset and the auxilliary assets to the same value. Is that intended?

If it is intended, could we add a comment describing why we do this? It's a bit confusing for the reader

isaaclab_arena_environments/mdp/__init__.py
alexmillane: Fair enough. We're trying not to use `mdp` in arena, because I find the naming non-obvious for users who haven't used Isaac Lab before. What do you think about changing the folder name to `helpers` (or something similar)?

.gitignore
```diff
@@ -35,7 +35,9 @@
 *.app
 
 # IDE stuff
-*.vscode
+.vscode/*
+!.vscode/launch.json
+!.vscode/settings.json
```
alexmillane: Is this change intended?

isaaclab_arena/assets/object_utils.py
yami007007: Yes, great point — I’ve updated it with this more general name and a comment.
As you mentioned, yes, this is actually a patch for the IsaacLab factory assets (peg, hole, gear mesh, etc.). These assets are configured as articulations without any joints.
To work around this, I set the corresponding values to {} in the upper-layer code, since only a few assembly assets are affected. Otherwise, asset initialization fails with an error. The error looks like:
For more details, please see the internal Slack [discussion](https://nvidia.enterprise.slack.com/lists/T04SYRAP3/F09QAM7SGR1?record_id=Rec0A0JDD7EH5).
> self.cfg.init_state.joint_pos: {'.*': 0.0}
> self.joint_names: [] 
> Exception caught in SimulationAppContext: ValueError: Not all regular expressions are matched! Please check that the regular expressions are correct: 
>         .*: []
> Available strings: []

isaaclab_arena/tasks/events.py
yami007007: Yes, this is intended to set the poses of the auxiliary assets based on the fixed asset (for example, the small and large gears are placed on the gear base so the operator only needs to place the medium gear).
There is already a comment above, and I can extend it a bit to make this clearer. 
`# set the poses of the auxiliary assets according to the pose of the fixed asset `

.gitignore
yami007007: For the `.gitignore`, yes, this is intentional. Previously, `launch.json` and `settings.json` were included, so I thought we should allow developers to submit some shared changes in addition to user‑specific settings.
However, after checking the `main` and `release/0.1.1` branches, it looks like there have already been some adjustments to these files. I’ll revert the `.gitignore` to default and leave it to the core team to decide how we should handle this. Thanks.
```
!.vscode/launch.json
!.vscode/settings.json
```

isaaclab_arena_environments/mdp/__init__.py
yami007007: For `mdp`, I’d suggest keeping the same folder structure as in IsaacLab.
Since most Arena users are likely already using IsaacLab, using `mdp` to organize observations, actions, rewards, terminations, and resets is a good way to stay consistent.
Just a suggestion, and I’m happy to update it once we reach a conclusion.

alexmillane: OK let's merge this as is. Let's see what we think about the naming as we progress. Maybe we change it. I understand the motivation for remaining consistent with Isaac Lab. I just also remember being confused when I started using Isaac Lab by this naming.

isaaclab_arena/assets/object_utils.py
alexmillane: Thank you for updating this!

isaaclab_arena/tasks/events.py
alexmillane: What is confusing is that this is that the auxillary assets are not set to some (constant) offset from the fixed asset. _They're set to **exactly** the same pose_. For most assets that will result in an inconsistent scene, because it will generate a scene where objects have intersecting geometry, which will explode when the simulation starts. I would guess what's going on (it's not clear from the comment) is that the particular assets being used here have base frames defined such that an offset of 0 from the base asset generates some valid arrangement of the assets. This makes this function really only valid for a particular set of assets. Probably the assembly gears we're using. In general, assets passed to this function will generate an invalid scene that explodes at start up.

Could you expand the docstring to indicate that we only really expect this function to work with the assembly gears where 0 offset from the assets generates a valid scene.

alexmillane (approved):
Thank you for adding this, and for going back and forth on this with me.

I have one last request for expanding one docstring. However, I've approved it. So merge at will.

isaaclab_arena/tasks/events.py
yami007007: I see your point. 
Yes, 0 works for the factory assembly task, but for other assets customers will need to adjust it themselves. 
Let me expand the docstring a bit so that customers know how to adjust or extend it for their own assets.

PR #286, opened 2025-12-12
Feature/teleop design

viiik-inside:
## Summary
Simplify device registry and add a retargeter registry

### DIFF
--- a/docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
+++ b/docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
@@ -97,13 +97,13 @@ post-trained policy, the quality of the dataset, and number of steps in the eval
 
       .. code-block:: text
 
-         Metrics: {'success_rate': 0.8823529411764706, 'door_moved_rate': 1.0, 'num_episodes': 17}
+         Metrics: {'success_rate': 0.8823529411764706, 'revolute_joint_moved_rate': 1.0, 'num_episodes': 17}
 
    .. tab:: Low Hardware Requirements
 
       .. code-block:: text
 
-         Metrics: {'success_rate': 1.0, 'door_moved_rate': 1.0, 'num_episodes': 19}
+         Metrics: {'success_rate': 1.0, 'revolute_joint_moved_rate': 1.0, 'num_episodes': 19}
 
 
 Step 2: Run Parallel environments Evaluation
@@ -138,7 +138,7 @@ than the single environment evaluation because of the parallel evaluation.
 
 .. code-block:: text
 
-   Metrics: {'success_rate': 0.605, 'door_moved_rate': 0.955, 'num_episodes': 200}
+   Metrics: {'success_rate': 0.605, 'revolute_joint_moved_rate': 0.955, 'num_episodes': 200}
 
 .. note::
 
--- a/isaaclab_arena/affordances/openable.py
+++ b/isaaclab_arena/affordances/openable.py
@@ -15,11 +15,14 @@ from isaaclab_arena.utils.joint_utils import get_normalized_joint_position, set_
 class Openable(AffordanceBase):
     """Interface for openable objects."""
 
-    def __init__(self, openable_joint_name: str, openable_open_threshold: float = 0.5, **kwargs):
+    def __init__(self, openable_joint_name: str, openable_threshold: float = 0.5, **kwargs):
         super().__init__(**kwargs)
         # TODO(alexmillane, 2025.08.26): We probably want to be able to define the polarity of the joint.
         self.openable_joint_name = openable_joint_name
-        self.openable_open_threshold = openable_open_threshold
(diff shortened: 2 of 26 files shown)


isaaclab_arena/assets/asset_registry.py
```diff
@@ -112,6 +124,25 @@ def get_device_by_name(self, name: str) -> type["TeleopDeviceBase"]:
         ensure_assets_registered()
         return self.get_component_by_name(name)
 
+    def get_teleop_device_cfg(self, device: type["TeleopDeviceBase"], embodiment: object | None = None):
+        from isaaclab.devices.device_base import DevicesCfg
+
+        retargeter_registry = RetargeterRegistry()
+        retargeter = retargeter_registry.get_component_by_name((device.name, embodiment.name))()
```
alexmillane: The function signature says that embodiment can be `None`, however, we unconditionally access `embodiment.name`

I feel like one of two things is true:
- embodiment can't be `None`, in which case we should remove the default, or
- we should only look up a retargetter if embodiment is not `None`.

isaaclab_arena/assets/asset_registry.py
```diff
@@ -112,6 +124,25 @@ def get_device_by_name(self, name: str) -> type["TeleopDeviceBase"]:
...
+        retargeter_cfg = retargeter.get_retargeter_cfg(embodiment, sim_device=device.sim_device)
+        retargeters = [retargeter_cfg] if retargeter_cfg is not None else []
+        device_cfg = device.get_device_cfg(retargeters=retargeters, embodiment=embodiment)
```
alexmillane, line 136: I guess we have this strange line because `DeviceCfg` expects a list of retargetters.

Given that (it appears) we always have a list of length 0 or 1 for the list of retargetters, should we change the interface such that `get_device_cfg` accepts a single retargetter or `None`?

alexmillane: **Edit:** I see the reason below. The issue is that Isaac Lab expects a list. We're fine to leave this here.

isaaclab_arena/assets/asset_registry.py
```diff
@@ -19,41 +19,53 @@ class Registry(metaclass=SingletonMeta):
     def __init__(self):
         self._components = {}
 
-    def register(self, component: Any):
+    def register(self, component: Any, key: str | tuple[str, str] | None = None):
```
alexmillane: I know we discussed this IRL, and I suggested this interface with `tuple[str, str]`. Having looked at the code, it makes our (general) `Registry` object kind of ugly.

What do you think about going back to what you said restoring the interface here to allow only `str` keys, and doing the `tuple[str, str] -> str` conversion in the `DeviceRegistry`.

Sorry.

isaaclab_arena/assets/retargeter_library.py
```diff
@@ -0,0 +1,72 @@
...
+# limitations under the License.
+
+from isaaclab.devices.openxr.retargeters import GR1T2RetargeterCfg
+from isaaclab.devices.retargeter_base import RetargeterCfg
+
+from isaaclab_arena.assets.register import register_retargeter
+
+
+@register_retargeter
+class GR1T2PinkOpenXRRetargeter:
```
alexmillane: Suggestion to add a base class that enforces a common interface.

In this case

```
class RetargetterBase(ABC):
    device: str | None = None
    embodiment: str | None = None

    @abstractmethod
    def get_retargeter_cfg(...):
        raise NotImplementedError

```

etc.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -82,7 +83,10 @@ def compose_manager_cfg(self) -> IsaacLabArenaManagerBasedRLEnvCfg:
         actions_cfg = self.arena_env.embodiment.get_action_cfg()
         xr_cfg = self.arena_env.embodiment.get_xr_cfg()
         if self.arena_env.teleop_device is not None:
-            teleop_device_cfg = self.arena_env.teleop_device.get_teleop_device_cfg(embodiment=self.arena_env.embodiment)
+            device_registry = DeviceRegistry()
+            teleop_device_cfg = device_registry.get_teleop_device_cfg(
+                self.arena_env.teleop_device, self.arena_env.embodiment
+            )
```
alexmillane, line 89: Clean!

alexmillane (approved):
LGTM. I left a few comments.

isaaclab_arena/assets/asset_registry.py
viiik-inside: okay

isaaclab_arena/assets/asset_registry.py
viiik-inside: fair

isaaclab_arena/assets/retargeter_library.py
viiik-inside: yup

PR #287, opened 2025-12-12
Basic RL workflow example

viiik-inside:
## Summary
One RL training environment.

NOTE: Moved up from draft as the RL pipeline is running. Some changes to the code is imminent at this state, but it is a very good reference point.

### DIFF
--- a/isaaclab_arena/scripts/reinforcement_learning/cli_args.py
+++ b/isaaclab_arena/scripts/reinforcement_learning/cli_args.py
@@ -21,10 +21,6 @@ def add_rsl_rl_args(parser: argparse.ArgumentParser):
     """
     # create a new argument group
     arg_group = parser.add_argument_group("rsl_rl", description="Arguments for RSL-RL agent.")
-    # -- experiment arguments
-    arg_group.add_argument(
-        "--experiment_name", type=str, default=None, help="Name of the experiment folder where logs will be stored."
-    )
     arg_group.add_argument("--run_name", type=str, default=None, help="Run name suffix to the log directory.")
     # -- load arguments
     arg_group.add_argument("--resume", action="store_true", default=False, help="Whether to resume from a checkpoint.")
--- a/isaaclab_arena/tasks/lift_object_task.py
+++ b/isaaclab_arena/tasks/lift_object_task.py
@@ -162,7 +162,7 @@ class LiftObjectObservationsCfg:
             )
             object_position = ObsTerm(
                 func=observations.object_position_in_frame,
-                params={"robot_cfg": SceneEntityCfg(robot_name), "object_cfg": SceneEntityCfg(lift_object.name)},
+                params={"root_frame_cfg": SceneEntityCfg(robot_name), "object_cfg": SceneEntityCfg(lift_object.name)},
             )
 
             def __post_init__(self):


isaaclab_arena/embodiments/franka/franka.py
```diff
@@ -62,6 +68,12 @@ def _update_scene_cfg_with_robot_initial_pose(self, scene_config: Any, pose: Pos
         scene_config.stand.init_state.rot = pose.rotation_wxyz
         return scene_config
 
+    def get_ee_frame_name(self) -> str:
+        return "ee_frame"
+
+    def get_command_body_name(self) -> str:
+        return self.action_config.arm_action.body_name
```
alexmillane, line 75: Can we add a small docstring here. It's not clear what "command_bocy_frame" means.

isaaclab_arena/embodiments/franka/franka.py
```diff
@@ -184,6 +199,16 @@ class FrankaEventCfg:
...
 
 
+@configclass
+class FrankaRewardsCfg:
+    """Reward specifications for the MDP."""
+
+    action_rate = RewardTermCfg(func=mdp_isaac_lab.action_rate_l2, weight=-0.0001)
+    joint_vel = RewardTermCfg(
+        func=mdp_isaac_lab.joint_vel_l2, weight=-0.0001, params={"asset_cfg": SceneEntityCfg("robot")}
+    )
```
alexmillane, line 206: Out of interest do you think that these rewards are so general that they'll be used anywhere franka is used for RL?

isaaclab_arena/embodiments/gr1t2/gr1t2.py
```diff
@@ -370,36 +374,39 @@ def __post_init__(self):
...
-            self.enable_corruption = False
-            self.concatenate_terms = False
-
-    # observation groups
-    policy: PolicyCfg = PolicyCfg()
+    policy: ObsGroup = MISSING
+
+    def __init__(self, concatenate_terms: bool = False):
+        @configclass
+        class PolicyCfg(ObsGroup):
```
alexmillane: This way of doing things makes this class confusing to read and breaks with the structure elsewhere. I would rather that we keep the class the way it was, which is in line with elsewhere in IsaacLab, and then set the variable after construction

```
class GR1T2EmbodimentBase(EmbodimentBase):
    def __init__(concatenate_terms: bool = False):
        self.observation_config = GR1T2ObservationsCfg()
        self.observation_config.concatenate_terms = concatenate_terms
...
```

Maybe another question is whether we should be reflecting this option in the embodiment constructor. Another option would be to set this option in the RL script where it's needed. Something like

```
embodiment = GR1T2EmbodimentBase()
embodiment.observation_config.concatenate_terms = True
```

As you and I agree on, I'm a bit worried about reflecting and ever increasing list of IsaacLab options.

isaaclab_arena/embodiments/franka/franka.py
```diff
@@ -144,22 +156,25 @@ class FrankaActionsCfg:
...
-        """Observations for policy group with state values."""
+    policy: ObsGroup = MISSING
 
-        actions = ObsTerm(func=mdp_isaac_lab.last_action)
-        joint_pos = ObsTerm(func=mdp_isaac_lab.joint_pos_rel)
-        joint_vel = ObsTerm(func=mdp_isaac_lab.joint_vel_rel)
-        eef_pos = ObsTerm(func=ee_frame_pos)
-        eef_quat = ObsTerm(func=ee_frame_quat)
-        gripper_pos = ObsTerm(func=gripper_pos)
+    def __init__(self, concatenate_terms: bool = False):
```
alexmillane: See comment about this in the `GR1T2ObservationsCfg`.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -23,10 +23,15 @@ class EmbodimentBase(Asset):
...
-        self, enable_cameras: bool = False, initial_pose: Pose | None = None, mimic_arm_mode: MimicArmMode | None = None
+        self,
+        enable_cameras: bool = False,
+        initial_pose: Pose | None = None,
+        concatenate_observation_terms: bool = False,
+        mimic_arm_mode: MimicArmMode | None = None,
     ):
         self.enable_cameras = enable_cameras
         self.initial_pose = initial_pose
+        self.concatenate_observation_terms = concatenate_observation_terms
```
alexmillane, line 34: I'm not sure that we're using this anywhere?

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -109,3 +114,13 @@ def modify_env_cfg(self, env_cfg: IsaacLabArenaManagerBasedRLEnvCfg) -> IsaacLab
...
+
+    def get_embodiment_name_in_scene(self) -> str:
+        return "robot"
+
+    def get_ee_frame_name(self) -> str:
+        # In case of multiple ee frames one can use self.mimic_arm_mode to get the correct ee frame name
+        return ""
+
+    def get_command_body_name(self) -> str:
+        return ""
```
alexmillane, line 123: I'm still quite unsure about what this means. How does this work in the case of mutliple end effectors? What about in the locomanipulation case where we're potentially controlling both arms and base? What the meaning of command body name in that instance?

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -109,3 +114,13 @@ def modify_env_cfg(self, env_cfg: IsaacLabArenaManagerBasedRLEnvCfg) -> IsaacLab
 
     def get_mimic_arm_mode(self) -> MimicArmMode:
         return self.mimic_arm_mode
+
+    def get_embodiment_name_in_scene(self) -> str:
+        return "robot"
+
+    def get_ee_frame_name(self) -> str:
+        # In case of multiple ee frames one can use self.mimic_arm_mode to get the correct ee frame name
+        return ""
```
alexmillane: The comment seems to indicate that this function should take an argument of type `MimicArmMode`

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -27,11 +27,12 @@
 class ArenaEnvBuilder:
     """Compose IsaacLab Arena → IsaacLab configs"""
 
-    DEFAULT_SCENE_CFG = InteractiveSceneCfg(num_envs=4096, env_spacing=30.0, replicate_physics=False)
-
     def __init__(self, arena_env: IsaacLabArenaEnvironment, args: argparse.Namespace):
         self.arena_env = arena_env
         self.args = args
+        self.DEFAULT_SCENE_CFG = InteractiveSceneCfg(
```
alexmillane: I'm wondering why this is all caps. I'm thinking let's change the name `DEFAULT_SCENE_CFG` -> `interactive_scene_cfg`. At this point it's not really the default any more. It can be modified, and then is directly used.

Thanks for putting this as a class member to get people more flexability.

isaaclab_arena/policy/rl_policy/base_rsl_rl_policy.py
```diff
@@ -0,0 +1,62 @@
...
+from isaaclab_rl.rsl_rl import RslRlOnPolicyRunnerCfg, RslRlPpoActorCriticCfg, RslRlPpoAlgorithmCfg
+
+
+@configclass
+class RLPolicyCfg(RslRlOnPolicyRunnerCfg):
+
+    num_steps_per_env: int = 24
+    max_iterations: int = 4000
+    save_interval: int = 200
+    experiment_name: str = "franka_lift"
```
alexmillane: Shall we change this to something more generic? Or perhaps `MISSING`, to enforce that it is overridden?

isaaclab_arena/policy/rl_policy/base_rsl_rl_policy.py
```diff
@@ -0,0 +1,62 @@
...
+    obs_groups = field(
+        default_factory=lambda: {
+            "policy": ["policy"],
+            "critic": ["policy"],
+        }
+    )
+    policy: RslRlPpoActorCriticCfg = field(default_factory=lambda: RslRlPpoActorCriticCfg())
+    algorithm: RslRlPpoAlgorithmCfg = field(default_factory=lambda: RslRlPpoAlgorithmCfg())
```
alexmillane: I think the `lambda` is not required here if I'm not mistaken:

```
class RLPolicyCfg(RslRlOnPolicyRunnerCfg):
...
    policy: RslRlPpoActorCriticCfg = field(default_factory=RslRlPpoActorCriticCfg)

```

isaaclab_arena/scripts/reinforcement_learning/cli_args.py
```diff
@@ -0,0 +1,110 @@
...
+    arg_group.add_argument("--resume", action="store_true", default=False, help="Whether to resume from a checkpoint.")
+    arg_group.add_argument("--load_run", type=str, default=None, help="Name of the run folder to resume from.")
+    arg_group.add_argument("--checkpoint", type=str, default=None, help="Checkpoint file to resume from.")
+    # -- logger arguments
+    arg_group.add_argument(
+        "--logger", type=str, default=None, choices={"wandb", "tensorboard", "neptune"}, help="Logger module to use."
+    )
+    arg_group.add_argument(
+        "--log_project_name", type=str, default=None, help="Name of the logging project when using wandb or neptune."
+    )
```
alexmillane: I presume these can't be imported from anywhere because they're in an IsaacLab script?

isaaclab_arena/scripts/reinforcement_learning/cli_args.py
```diff
@@ -0,0 +1,110 @@
...
+        parser: The parser to add the arguments to.
+    """
+    arg_group = parser.add_argument_group("rsl_rl_policy", description="Arguments for RSL-RL policy.")
+    arg_group.add_argument("--num_steps_per_env", type=int, default=24, help="Number of steps per environment.")
+    arg_group.add_argument("--max_iterations", type=int, default=4000, help="Maximum number of iterations.")
+    arg_group.add_argument("--save_interval", type=int, default=200, help="Save interval.")
+    arg_group.add_argument(
+        "--experiment_name",
+        type=str,
+        default="franka_lift",
```
alexmillane: make something a bit more generic? May "experiment"?

isaaclab_arena/tasks/observations/observations.py
```diff
@@ -0,0 +1,33 @@
...
+    """Observation the position of the object in the world frame."""
+    object = env.scene[asset_cfg.name]
+    return object.data.root_pos_w
+
+
+def object_position_in_robot_root_frame(
+    env: ManagerBasedRLEnv,
+    robot_cfg: SceneEntityCfg = SceneEntityCfg("robot"),
+    object_cfg: SceneEntityCfg = SceneEntityCfg("object"),
+) -> torch.Tensor:
```
alexmillane: I would say that this deserves a more generic name.

This get the position of object A in the frame of object B no? That's a very general function, there's nothing in here that's specific to a robot and a manipulated object.

Maybe we could call this `object_position_in_frame` and rename the variables accordingly?

isaaclab_arena/tasks/rewards/lift_object_rewards.py
```diff
@@ -0,0 +1,41 @@
...
+    robot: RigidObject = env.scene[robot_cfg.name]
+    object: RigidObject = env.scene[object_cfg.name]
+    command = env.command_manager.get_command(command_name)
+    # compute the desired position in the world frame
+    des_pos_b = command[:, :3]
+    des_pos_w, _ = combine_frame_transforms(robot.data.root_pos_w, robot.data.root_quat_w, des_pos_b)
+    # distance of the end-effector to the object: (num_envs,)
+    distance = torch.norm(des_pos_w - object.data.root_pos_w, dim=1)
+    # rewarded if the object is lifted above the threshold
+    return (object.data.root_pos_w[:, 2] > minimal_height) * (1 - torch.tanh(distance / std))
```
alexmillane, line 41: Are these importable from IsaacLab? Or is this something we've developed?

isaaclab_arena/tasks/rewards/rewards.py
```diff
@@ -0,0 +1,31 @@
...
+    object: RigidObject = env.scene[object_cfg.name]
+    ee_frame: FrameTransformer = env.scene[ee_frame_cfg.name]
+    # Target object position: (num_envs, 3)
+    object_pos_w = object.data.root_pos_w
+    # End-effector position: (num_envs, 3)
+    ee_w = ee_frame.data.target_pos_w[..., 0, :]
+    # Distance of the end-effector to the object: (num_envs,)
+    object_ee_distance = torch.norm(object_pos_w - ee_w, dim=1)
+
+    return 1 - torch.tanh(object_ee_distance / std)
```
alexmillane, line 31: As above, is this something original or can this be imported.

isaaclab_arena/tasks/lift_object_task.py
```diff
@@ -0,0 +1,238 @@
...
+    def __init__(self, lift_object: Asset):
+        self.reset_lift_object_pose = EventTermCfg(
+            func=mdp_isaac_lab.reset_root_state_uniform,
+            mode="reset",
+            params={
+                "pose_range": {
+                    "x": (-0.1, 0.1),
+                    "y": (-0.25, 0.25),
+                    "z": (0.0, 0.0),
+                },
```
alexmillane, line 91: Personally I find having these values in here unmodifiable a bit... weird. What do you think about being able to specify this through the task constructor?

We should then add a helper object for pose ranges (next to pose.py in the code):

```
@dataclass
class PoseRange:
    position_xyz_min: tuple[float, float, float] = (0.0, 0.0, 0.0)
    position_xyz_max: tuple[float, float, float] = (0.0, 0.0, 0.0)
    rotation_rpy_min: tuple[float, float, float] = (0.0, 0.0, 0.0)
    rotation_rpy_max: tuple[float, float, float] = (0.0, 0.0, 0.0)

   def to_dict(self):
       return {
            x: (position_xyz_min[0], position_xyz_max[0])

...     
```

Then

```
def __init__(
        self,
        lift_object: Asset,
        background_scene: Asset,
        minimum_height_to_lift: float = 0.04,
        episode_length_s: float = 5.0,
        reset_pose_range: PoseRange = PoseRange()
    ):
```

Hopefully, when @cvolkcvolk finishes the object placement stuff, we can do something even better.

isaaclab_arena/tasks/lift_object_task.py
```diff
@@ -0,0 +1,238 @@
...
+            robot_name=self.embodiment.get_embodiment_name_in_scene(),
+            ee_frame_name=self.embodiment.get_ee_frame_name(),
+        )
+
+    def get_commands_cfg(self):
+        return LiftObjectCommandsCfg(
+            asset_name=self.embodiment.get_embodiment_name_in_scene(),
+            body_name=self.embodiment.get_command_body_name(),
+            lift_object=self.lift_object,
+        )
```
alexmillane: What do you think about following our "maximum modifyability" policy that we pursue elsewhere, where we build configs in the constructor, such that a user can modify after the fact.

isaaclab_arena/tasks/terminations.py
```diff
@@ -75,3 +75,13 @@ def objects_in_proximity(
...
+
+
+def object_lifted(
+    env: ManagerBasedRLEnv,
+    object_cfg: SceneEntityCfg,
+    maximum_height: float,
+) -> torch.Tensor:
+    """Determine if the object is lifted above a certain height."""
+    object: RigidObject = env.scene[object_cfg.name]
+    return object.data.root_pos_w[:, 2] > maximum_height
```
alexmillane: -> `object_above()`

means that this function isn't specific to lifting.

isaaclab_arena_environments/lift_object_environment.py
```diff
@@ -0,0 +1,76 @@
...
+
+        return isaaclab_arena_environment
+
+    @staticmethod
+    def add_cli_args(parser: argparse.ArgumentParser) -> None:
+        parser.add_argument("--object", type=str, default="dex_cube")
+        # NOTE(alexmillane, 2025.09.04): We need a teleop device argument in order
+        # to be used in the record_demos.py script.
+        parser.add_argument("--teleop_device", type=str, default=None)
+        # Note (xinjieyao, 2025.10.06): Add the embodiment argument for PINK IK EEF control or Joint positional control
```
alexmillane: remove?

isaaclab_arena/embodiments/embodiment_base.py
viiik-inside: We use it in the observation terms for all the robots

isaaclab_arena/embodiments/embodiment_base.py
viiik-inside: fair enough.

isaaclab_arena/embodiments/embodiment_base.py
viiik-inside: It is basically the part of the robot which you want to command so it can do a particular task. In terms of locomanipulation case I would imagine you command the torso to go near a goal. This is very unclear just in its definition. There is a possibility that you command multiple bodies in the same robot. Maybe making this a list might be an idea

isaaclab_arena/embodiments/franka/franka.py
viiik-inside: I think so yes, one by intuition that these rewards just makes the robot smooth. 

I went through lab and this has been used in every training franka has been used.

isaaclab_arena/embodiments/gr1t2/gr1t2.py
viiik-inside: I am gonna go with keeping it in the embodiment constructor but will get back to this point as a discussion or another PR soon after

viiik-inside: This at the end became a bit more complicated. But ill still keep it as such with the change

isaaclab_arena/environments/arena_env_builder.py
viiik-inside: fair

isaaclab_arena/policy/rl_policy/base_rsl_rl_policy.py
viiik-inside: changed it to more generic name

isaaclab_arena/policy/rl_policy/base_rsl_rl_policy.py
viiik-inside: yeah, its not needed.

isaaclab_arena/scripts/reinforcement_learning/cli_args.py
viiik-inside: yup

isaaclab_arena/tasks/lift_object_task.py
viiik-inside: makes sense

isaaclab_arena/tasks/lift_object_task.py
viiik-inside: fair point, done

isaaclab_arena/tasks/rewards/lift_object_rewards.py
viiik-inside: They are importable but Im scared to import them

isaaclab_arena/tasks/observations/observations.py
viiik-inside: done

isaaclab_arena/tasks/rewards/rewards.py
viiik-inside: can be imported but I would leave it here like this.

viiik-inside: Also shows were one can put their rewards etc.

isaaclab_arena/tasks/terminations.py
viiik-inside: yeah

isaaclab_arena/policy/rl_policy/generic_policy.json
```diff
@@ -0,0 +1,28 @@
...
+      "num_mini_batches": 4,
+      "learning_rate": 0.0001,
+      "schedule": "adaptive",
+      "gamma": 0.98,
+      "lam": 0.95,
+      "desired_kl": 0.01,
+      "max_grad_norm": 1.0
+    },
+    "obs_groups": {
+      "policy": ["policy", "task_obs"],
```
viiik-inside, line 25: see here

PR #289, opened 2025-12-12
Add SequentialTaskBase class

peterd-NV:
## Summary
This PR adds initial support for composite sequential tasks via the SequentialTaskBase class. The SequentialTaskBase class takes a list of atomic tasks (TaskBase) and automatically assembles them into a composite task with unified termination/event configs. 

Adds:

1. SequentialTaskBase class
2. Test case to validate class methods
3. Test case with example task (sequential open door task) to validate unified success check and events
4. Two new functions in `isaac_arena/utils/configclass.py` to perform config transformation and duplicate checking

### DIFF
--- a/isaaclab_arena/assets/object_set.py
+++ b/isaaclab_arena/assets/object_set.py
@@ -1,0 +1,87 @@
+# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import isaaclab.sim as sim_utils
+from isaaclab.assets import RigidObjectCfg
+
+from isaaclab_arena.assets.object import Object
+from isaaclab_arena.assets.object_base import ObjectBase, ObjectType
+from isaaclab_arena.assets.object_utils import detect_object_type
+from isaaclab_arena.utils.pose import Pose
+
+
+class RigidObjectSet(Object):
+    """
+    A set of rigid objects.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        objects: list[Object],
+        prim_path: str | None = None,
+        scale: tuple[float, float, float] = (1.0, 1.0, 1.0),
+        random_choice: bool = False,
+        initial_pose: Pose | None = None,
+        **kwargs,
+    ):
+        """
+        Args:
+            name: The name of the object set.
+            objects: The list of objects to be included in the object set.
+            prim_path: The prim path of the object set. Note that for all environments, the object set
+                prim path must be the same.
+            scale: The scale of the object set. Note all objects can only have the same scale, if
+                different scales are needed, considering scaling the object USD file.
+            random_choice: Whether to randomly choose an object from the object set to spawn in
+                each environment. If False, object is spawned based on the order of objects in the list.
+            initial_pose: The initial pose of the object from this object set.
+        """
+        if not self._are_all_objects_type_rigid(objects):
+            raise ValueError(f"Object set {name} must contain only rigid objects.")
+
+        self.object_usd_paths = [object.usd_path for object in objects]
+        self.random_choice = random_choice
+
+        # Set default prim_path if not provided
+        if prim_path is None:
+            prim_path = f"{{ENV_REGEX_NS}}/{name}"
+
+        super().__init__(
+            name=name,
+            object_type=ObjectType.RIGID,
+            usd_path="",
+            prim_path=prim_path,
+            scale=scale,
+            initial_pose=initial_pose,
+            **kwargs,
+        )
+
+    def _are_all_objects_type_rigid(self, objects: list[ObjectBase]) -> bool:
+        if objects is None or len(objects) == 0:
+            raise ValueError(f"Object set {self.name} must contain at least 1 object.")
+        return all(detect_object_type(usd_path=object.usd_path) == ObjectType.RIGID for object in objects)
+
+    def _generate_rigid_cfg(self) -> RigidObjectCfg:
+        assert self.object_type == ObjectType.RIGID
+        object_cfg = RigidObjectCfg(
+            prim_path=self.prim_path,
+            spawn=sim_utils.MultiUsdFileCfg(
+                usd_path=self.object_usd_paths,
+                random_choice=self.random_choice,
+                activate_contact_sensors=True,
+            ),
+        )
+        object_cfg = self._add_initial_pose_to_cfg(object_cfg)
+        return object_cfg
+
+    def _generate_articulation_cfg(self):
+        raise NotImplementedError("Articulation configuration is not supported for object sets")
+
+    def _generate_base_cfg(self):
+        raise NotImplementedError("Base configuration is not supported for object sets")
+
+    def _generate_spawner_cfg(self):
+        raise NotImplementedError("Spawner configuration is not supported for object sets")
--- a/isaaclab_arena/scene/scene.py
+++ b/isaaclab_arena/scene/scene.py
@@ -14,6 +14,7 @@ from pxr import Gf, Usd, UsdGeom
 from isaaclab_arena.assets.asset import Asset
 from isaaclab_arena.assets.object import Object
 from isaaclab_arena.assets.object_base import ObjectType
+from isaaclab_arena.assets.object_set import RigidObjectSet
 from isaaclab_arena.environments.isaaclab_arena_manager_based_env import IsaacLabArenaManagerBasedRLEnvCfg
 from isaaclab_arena.utils.configclass import make_configclass
 from isaaclab_arena.utils.phyx_utils import add_contact_report
@@ -23,8 +24,8 @@ AssetCfg = Union[AssetBaseCfg, RigidObjectCfg, ArticulationCfg, ContactSensorCfg
 
 class Scene:
 
-    def __init__(self, assets: list[Asset] | None = None):
-        self.assets: dict[str, Asset] = {}
+    def __init__(self, assets: list[Asset, RigidObjectSet] | None = None):
+        self.assets: dict[str, Asset | RigidObjectSet] = {}
         # We add these here so a user can override them if they want.
         self.observation_cfg = None
         self.events_cfg = None
@@ -35,11 +36,23 @@ class Scene:
         if assets is not None:
             self.add_assets(assets)
 
-    def add_asset(self, asset: Asset):
-        assert asset.name is not None, "Asset with the same name already exists"
+    def add_asset(self, asset: Asset | RigidObjectSet):
+        """Add an asset to the scene.
+
+        Args:
+            asset: An Asset instance or a dictionary of Assets. If a dictionary is provided,
+                   the keys will be used as the names of the assets and the values will be the list of assets.
+        """
+        if not isinstance(asset, Asset | RigidObjectSet):
+            raise ValueError(f"Invalid asset type: {type(asset)}")
+
+        if asset.name is None:
+            print("Asset name is None. Skipping asset.")
+            return
+        # if name already exists, overwrite
         self.assets[asset.name] = asset
 
-    def add_assets(self, assets: list[Asset]):
+    def add_assets(self, assets: list[Asset | RigidObjectSet]):
         for asset in assets:
             self.add_asset(asset)
 
--- a/isaaclab_arena/tests/test_object_set.py
+++ b/isaaclab_arena/tests/test_object_set.py
@@ -1,0 +1,288 @@
+# Copyright (c) 2025, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from isaaclab_arena.tests.utils.subprocess import run_simulation_app_function
+
+HEADLESS = True
+NUM_ENVS = 3
+OBJECT_SET_1_PRIM_PATH = "/World/envs/env_.*/ObjectSet_1"
+OBJECT_SET_2_PRIM_PATH = "/World/envs/env_.*/ObjectSet_2"
+
+
+def _test_empty_object_set(simulation_app):
+    from isaaclab_arena.assets.object_set import RigidObjectSet
+
+    try:
+        RigidObjectSet(name="empty_object_set", objects=[])
+    except Exception:
+        return True
+    return False
+
+
+def _test_articulation_object_set(simulation_app):
+    from isaaclab_arena.assets.asset_registry import AssetRegistry
+    from isaaclab_arena.assets.object_set import RigidObjectSet
+
+    asset_registry = AssetRegistry()
+    microwave = asset_registry.get_asset_by_name("microwave")()
+    try:
+        RigidObjectSet(name="articulation_object_set", objects=[microwave])
+    except Exception:
+        return True
+    return False
+
+
+def _test_single_object_in_one_object_set(simulation_app):
+    from isaacsim.core.utils.stage import get_current_stage
+
+    from isaaclab_arena.assets.asset_registry import AssetRegistry
+    from isaaclab_arena.assets.object_reference import ObjectReference
+    from isaaclab_arena.assets.object_set import RigidObjectSet
+    from isaaclab_arena.cli.isaaclab_arena_cli import get_isaaclab_arena_cli_parser
+    from isaaclab_arena.environments.arena_env_builder import ArenaEnvBuilder
+    from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
+    from isaaclab_arena.scene.scene import Scene
+    from isaaclab_arena.tasks.pick_and_place_task import PickAndPlaceTask
+    from isaaclab_arena.utils.pose import Pose
+    from isaaclab_arena.utils.usd_helpers import get_asset_usd_path_from_prim_path
+
+    asset_registry = AssetRegistry()
+    background = asset_registry.get_asset_by_name("kitchen")()
+    embodiment = asset_registry.get_asset_by_name("franka")()
+    cracker_box = asset_registry.get_asset_by_name("cracker_box")()
+    destination_location = ObjectReference(
+        name="destination_location",
+        prim_path="{ENV_REGEX_NS}/kitchen/Cabinet_B_02",
+        parent_asset=background,
+    )
+    obj_set = RigidObjectSet(
+        name="single_object_set", objects=[cracker_box, cracker_box], prim_path=OBJECT_SET_1_PRIM_PATH
+    )
+    obj_set.set_initial_pose(Pose(position_xyz=(0.1, 0.0, 0.1), rotation_wxyz=(1.0, 0.0, 0.0, 0.0)))
+    scene = Scene(assets=[background, obj_set])
+    isaaclab_arena_environment = IsaacLabArenaEnvironment(
+        name="single_object_set_test",
+        embodiment=embodiment,
+        scene=scene,
+        task=PickAndPlaceTask(
+            pick_up_object=obj_set, destination_location=destination_location, background_scene=background
+        ),
+        teleop_device=None,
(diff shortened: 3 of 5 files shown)


isaaclab_arena/tests/test_sequential_open_door.py
```diff
@@ -0,0 +1,243 @@
...
+            self,
+            openable_object,
+            subtasks,
+            episode_length_s=None,
+        ):
+            super().__init__(subtasks=subtasks, episode_length_s=episode_length_s)
+            self.openable_object = openable_object
+
+        def get_metrics(self):
+            return []
```
xyao-nv, line 37: Is the metrics something you are bringing in for future MRs? If so, can you leave a TODO in the base class?

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+            if result:
+                env._subtask_success_state[env_idx][current_subtask_idx] = True
+                if current_subtask_idx < len(task_instance.subtasks) - 1:
+                    env._current_subtask_idx[env_idx] += 1
+
+        # Compute composite task success state for each env
+        per_env_success = [all(env_successes) for env_successes in env._subtask_success_state]
+        success_tensor = torch.tensor(per_env_success, dtype=torch.bool, device=env.device)
+
+        env.extras["subtask_success_state"] = copy.copy(env._subtask_success_state)
```
xyao-nv, line 90: nice

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+@configclass
+class TerminationsCfg:
+    success: TerminationTermCfg = MISSING
+
+
+class SequentialTaskBase(TaskBase):
+    """
+    A base class for tasks composed sequentially from multiple subtasks.
+    The sequential task takes a list of TaskBase instances (subtasks),
+    and automatically collects configs to form a composite task.
```
alexmillane: Suggestion to expand the definition of a sequential task. I guess "sequential" is a specific form of composite, in that the atomic tasks must be completed in order, one after another. Once completed, a sub-tasks can then go back to being not-true, and the sequential task still completes. This is in contrast to something like a composite task where atomic tasks can be completed in any order, or a task where all atomic tasks must be true at once.

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+        # Rename the fields with suffix
+        for field in fields:
+            new_name = f"{field.name}{suffix}"
+            value = getattr(cfg_instance, field.name)
+            new_fields.append((new_name, field.type, value))
+            field_values[new_name] = value
+
+        # Create a new configclass with renamed fields
+        new_cfg_class = make_configclass(type(cfg_instance).__name__, new_fields)
+        return new_cfg_class(**field_values)
```
alexmillane: Suggestion to move this to `isaaclab_arena.utils.configclass`. We have a few other utilities in that file for modifying config classes.

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+            value = getattr(cfg_instance, field.name)
+            new_fields.append((field.name, field.type, value))
+            field_values[field.name] = value
+
+        if not new_fields:
+            return None
+
+        # Create a new configclass without the excluded fields
+        new_cfg_class = make_configclass(type(cfg_instance).__name__, new_fields)
+        return new_cfg_class(**field_values)
```
alexmillane: As above, suggestion to move this to `isaaclab_arena.utils.configclass`.

Additionally, the two functions above have a very similar form. They loop through the fields, build a new config class by performing some mutation on each field. I wonder if we could write them in terms of a function `transform_config_class` which takes a `Callable`, loops through the fields, calling the callable on each and adding the resulting field to a new `configclass` if the return is not None.

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+        return events_cfg
+
+    def make_sequential_task_termination_cfg(self) -> configclass:
+        "Make composite success check termination term."
+        success = TerminationTermCfg(
+            func=self.sequential_task_success_func,
+            params={
+                "task_instance": self,
+            },
+        )
```
alexmillane, line 171: I find this construction of a static method which takes `self` a bit unexpected. Arguably the defining characteristic of a static method is that it doesn't take `self`.

Is it possible that we make `sequential_task_success_func` non-static, and then do

```
from functools import partial
...

        success = TerminationTermCfg(
            func=partial(SequentialTaskBase.sequential_task_success_func, self)
        )
```

We can then simplify `SequentialTaskBase.sequential_task_success_func`, as it will have the form of a normal method, with accecss to self etc.

Maybe there's something I'm missing?

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+        # Initialize each env's current subtask index (state machine) to 0
+        if not hasattr(env, "_current_subtask_idx"):
+            env._current_subtask_idx = [0 for _ in range(env.num_envs)]
+        else:
+            for env_id in env_ids:
+                env._current_subtask_idx[env_id] = 0
+
+    def get_scene_cfg(self) -> configclass:
+        "Make combined scene cfg from all subtasks."
+        scene_cfg = combine_configclass_instances("SceneCfg", *(subtask.get_scene_cfg() for subtask in self.subtasks))
```
alexmillane: I wonder if we should add something in `isaaclab_arena.utils.configclass` to detect if there are repeated config class members. Here we could check and warn (before proceeding).

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def make_sequential_task_events_cfg(self) -> configclass:
+        "Make event to reset subtask success state."
+        reset_subtask_success_state = EventTermCfg(
+            func=self.reset_subtask_success_state,
+            mode="reset",
+            params={
+                "task_instance": self,
+            },
+        )
```
alexmillane, line 139: See above. I'd like to suggest to use partial to pass a method, if possible.

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+            renamed_cfg = self._add_suffix_to_configclass_fields(subtask_events_cfg, f"_subtask_{i}")
+            if renamed_cfg is not None:
+                renamed_events_cfgs.append(renamed_cfg)
+
+        # Add reset subtask success state event to the combined events cfgs
+        events_cfg = combine_configclass_instances(
+            "EventsCfg", *renamed_events_cfgs, self.make_sequential_task_events_cfg()
+        )
+
+        return events_cfg
```
alexmillane, line 162: Awesome!

isaaclab_arena/tests/test_sequential_open_door.py
```diff
@@ -0,0 +1,243 @@
...
+
+    class SequentialOpenDoorTask(SequentialTaskBase):
+        def __init__(
+            self,
+            openable_object,
+            subtasks,
+            episode_length_s=None,
+        ):
+            super().__init__(subtasks=subtasks, episode_length_s=episode_length_s)
+            self.openable_object = openable_object
```
alexmillane: I'm wondering why definition of this class is required.

This class just extends `SequentialTaskBase` through storage of `openable_object`. However, from what I can tell this member is never accessed (because each of the input subtasks thenmselves store a reference to the openable object). Suggestion to eliminate this class all together and just do

```
task = SequentialTaskBase([subtask_1, subtask_2])
```

... Perhaps this actually indicates one tiny issue. In our codebase so far `MyClassBase` always indicates pure-virtual class (which can't be instantiated). In this case, `SequentialTaskBase` is expected to actually be used. Perhaps we should rename `SequentialTaskBase` -> `SequentialTask`. Wdyt?

isaaclab_arena/tests/test_sequential_open_door.py
```diff
@@ -0,0 +1,243 @@
...
+        embodiment=FrankaEmbodiment(),
+        scene=scene,
+        task=SequentialOpenDoorTask(microwave, [subtask_1, subtask_2]),
+    )
+
+    env_builder = ArenaEnvBuilder(isaaclab_arena_environment, args_cli)
+    name, cfg = env_builder.build_registered()
+    if remove_reset_door_state_event:
+        cfg.events.reset_door_state_subtask_0 = None
+        cfg.events.reset_door_state_subtask_1 = None
```
alexmillane, line 87: Is this because we don't want the environment to reset before the test can check the condition? Suggestion to add a comment indicating why we zero these out.

isaaclab_arena/tests/test_sequential_open_door.py
```diff
@@ -0,0 +1,243 @@
...
+        step_zeros_and_call(env, NUM_STEPS, assert_composite_task_incomplete)
+        print("Opening microwave (completing subtask 0)")
+        microwave.open(env, env_ids=None)
+        step_zeros_and_call(env, NUM_STEPS, assert_composite_task_incomplete)
+        print("Closing microwave (starting subtask 1)")
+        microwave.close(env, env_ids=None)
+        step_zeros_and_call(env, NUM_STEPS, assert_composite_task_incomplete)
+        print("Opening microwave (completing subtask 1)")
+        microwave.open(env, env_ids=None)
+        step_zeros_and_call(env, NUM_STEPS, assert_composite_task_complete)
```
alexmillane, line 187: Hmmm I'm not quite sure why this works. I (niavely) would have thought, in this case where both sub tasks are the same thing, that opening the door once would complete subtask 1, and then one step later complete subtask 2. This test seems to indicate we need to close the door before re-opening it and completing the second subtask. Reading the code, I can't tell why that would be the case. Maybe you could let me know what I'm missing?

Unless there's a good reason, what do you think about making the task a bit more prototypical and separating the sub-tasks a bit more. One easy way of doing that would be that you have two microwaves, and you have to open the door on `mircowave_1` and then on `microwave_2`

isaaclab_arena/tests/test_sequential_open_door.py
```diff
@@ -0,0 +1,243 @@
...
+        _test_sequential_open_door_microwave_reset_condition,
+        headless=HEADLESS,
+    )
+    assert result, f"Test {_test_sequential_open_door_microwave_reset_condition.__name__} failed"
+
+
+if __name__ == "__main__":
+    test_sequential_open_door_microwave()
+    test_sequential_open_door_microwave_multiple_envs()
+    test_sequential_open_door_microwave_reset_condition()
```
alexmillane, line 394: If we move to my suggestion above, what do you think about adding a test to check that the ordering works. I.e. the microwaves are opened in the wrong order, 2-open, 2-close then 1-open, say, and we check that the task doesn't complete (and maybe then check that it _does_ complete when we do 2-open again. What do you think about that?

isaaclab_arena/tests/test_sequential_task_base.py
```diff
@@ -0,0 +1,131 @@
...
+    result = run_simulation_app_function(
+        _test_remove_configclass_fields,
+        headless=HEADLESS,
+    )
+    assert result, f"Test {_test_remove_configclass_fields.__name__} failed"
+
+
+if __name__ == "__main__":
+    test_add_suffix_to_configclass_fields()
+    test_remove_configclass_fields()
```
alexmillane: Great!

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+    def sequential_task_success_func(
+        env,
+        task_instance: "SequentialTaskBase",
+    ) -> torch.Tensor:
+        "Sequential task compositesuccess function."
+        # Check success of current subtask for each env
+        for env_idx in range(env.num_envs):
+            current_subtask_idx = env._current_subtask_idx[env_idx]
+            current_subtask_success_func = (
+                task_instance.subtasks[current_subtask_idx].get_termination_cfg().success.func
```
alexmillane: I'm wondering if going through the `termination_cfg` should be required. What do you think about changing the definition of `TaskBase` to enforce the implementation of a method `is_success(env)`?

That would a) add a bit more structure to what a `TaskBase` is, and b) bypass the constructing the termination here.

This line would simplify as:

```
current_subtask_success_func = task_instance.subtasks[current_subtask_idx].is_success
```

We could also simplify `TaskBase` by simply constructing the `TerminationCfg.success` from in the function `TaskBase.is_success`.

alexmillane: **Edit:** having looked through this MR, and thought about this comment a bit, I think if we do this, we should do it in another MR. This would touch all the tasks, including the sequential tasks. Let's ignore this comment for now, and proceed as is.

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+
+        if not new_fields:
+            return None
+
+        # Create a new configclass without the excluded fields
+        new_cfg_class = make_configclass(type(cfg_instance).__name__, new_fields)
+        return new_cfg_class(**field_values)
+
+    @staticmethod
+    def sequential_task_success_func(
```
alexmillane, line 62: Really cool!

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -0,0 +1,202 @@
...
+class SequentialTaskEventsCfg:
+    reset_subtask_success_state: EventTermCfg = MISSING
+
+
+@configclass
+class TerminationsCfg:
+    success: TerminationTermCfg = MISSING
+
+
+class SequentialTaskBase(TaskBase):
```
alexmillane, line 32: Great work! Really cool~

isaaclab_arena/tasks/sequential_task_base.py
peterd-NV: Good point, I've expanded the description with a more detailed definition to what "sequential" here means

isaaclab_arena/tasks/sequential_task_base.py
peterd-NV: See comment below on move to unified configclass transform func.

isaaclab_arena/tasks/sequential_task_base.py
peterd-NV: Agree with that. Moving to to a generic "transform_config_class" function would be cleaner and more reusable by other parts of the code down the line. I've added the following:

- A `transform_configclass_instance` function inside of `utils/configclass.py`. This function performs the "loop" seen in the original functions and takes a Callable to perform the transform
- Added simple Callables in the SequentialTaskBase class to perform the transforms
- Updated SequentialTaskBase class tests to use this format

isaaclab_arena/tasks/sequential_task_base.py
peterd-NV: Sounds good

isaaclab_arena/tasks/sequential_task_base.py
peterd-NV: Make's sense, I've added a function to utils.configclass to scan and return any duplicates: `check_configclass_field_duplicates`

If duplicates are found it'll give users a warning showing the names of them.

isaaclab_arena/tests/test_sequential_open_door.py
peterd-NV: Yes this is a good observation. For this particular task it works because I've set it up to use a single step (NUM_STEPS = 1) so that it gives a chance to close the door before moving onto the next subtask. So what happens is something like:

1. open microwave door and step (subtask 0 is marked as complete)
2. close microwave door and step (subtask 1 is still False here as the env only stepped once)
3. open microwave door and step (subtask 1 is now True as the env was stepped again)

isaaclab_arena/tasks/sequential_task_base.py
peterd-NV: Mmm I think setting the success/event functions to be non-static and use "self" would present an issue with other parts of Lab. Namely in some scripts like data recording, robomimic replay, mimic etc, they extract and use these functions directly to take full control of the env. For those, they depend on having the termination/event functions follow the standard function signature for Lab which is something like `func(env, params...)`. If we define these as normal class methods then it would cause an issue with a lot of downstream scripts.

In the current setup, the "self" that is passed to this function is essentially a copy of the as Lab makes a deep copy of all configs before instantiating the managers. However, I do agree that it's really not ideal to be passing "self" to a static method. 

I've made some changes which preserves these functions as staticmethods to keep the standard Lab function signature, but changed the argument to be the list of subtasks instead of an instance of SequentialTaskBase so that it's not directly using a copy of "self". What do you think of this solution?

isaaclab_arena/tasks/sequential_task_base.py
peterd-NV: See comment above on the blocker for doing this and an alternative proposal.

isaaclab_arena/tests/test_sequential_open_door.py
peterd-NV: The storage of `openable_object` can be removed (my bad that was a left over from some previous experimentation) but I think this class is still required as we need to instantiate the get metric/prompt/mimic functions to be "something" otherwise the Arena compiler will throw an error. This is because these functions in TaskBase are set to `raise NotImplementedError("Function not implemented yet.")`.

Because of that I don't think the SequentialTaskBase class can ever be used "as is" so I would lean towards keeping the current naming.

isaaclab_arena/tests/test_sequential_open_door.py
peterd-NV: Yes, that is correct. I've added a comment to explain it.

isaaclab_arena/tests/test_sequential_open_door.py
peterd-NV: good note, I've added a todo in the base class

isaaclab_arena/tests/test_sequential_open_door.py
peterd-NV: I agree that adding a second microwave will make things easier to understand. I've modified the test to use two separate microwaves and also set NUM_STEPS = 10 as we can now avoid doing what I mentioned above.

isaaclab_arena/tests/test_sequential_open_door.py
peterd-NV: Good idea, now that we have two microwaves I've added an "out of order" test to check that the task doesn't complete when done out of order but then reaches a success if done in order immediately afterwards.

isaaclab_arena/tasks/sequential_task_base.py
alexmillane: Thank you for the explanation. I hadn't considered that Lab expects a certain signature.

With that said, I feel that the using `partial` should result in the correct signature no?

Maybe I'm mistake but I thought that:

```
class MyClass:
    def my_method(self, env, param)

my_class = MyClass()
my_partial = partial(my_class.my_method, self)
```

results in `my_partial` having the signature of `func(env, param)`. Perhaps I'm mistaken.

With all of that said. Let's merge this all is. It's not that important.

isaaclab_arena/tests/test_sequential_open_door.py
alexmillane: Great peter! Thanks for that. Looks a lot easier to follow.

isaaclab_arena/utils/configclass.py
```diff
@@ -201,3 +201,71 @@ def new_post_init(self):
...
+    # Apply transformation
+    transformed_fields = transform(field_tuples)
+
+    if not transformed_fields:
+        return None
+
+    # Create a new configclass with transformed fields
+    field_values = {name: value for name, _, value in transformed_fields}
+    new_cfg_class = make_configclass(type(cfg_instance).__name__, transformed_fields)
+    return new_cfg_class(**field_values)
```
alexmillane, line 271: Cool!

alexmillane (approved):
LGTM!

PR #290, opened 2025-12-12
Add RigidObjectSet class to enable multi-object spawning

xyao-nv:
## Summary
`RigidObjectSet` inherited from `Object` to enable users provide a list of assets, and sim app spawn each `env_id` with one obj from this set.

## Detailed description
- Introduced `RigidObjectSet(Obejct)`  class for handle rigid body object set construction
- The order of each obj in the set to load in each env_id could be configured as following func args order, or being random.
- Introduced `--object_set` in `kitchen_pick_and_place.py` cli to allow spawning for each env_id
- Added tests for empty/single/multi object sets & checker each env_id's usd is referenced in expected sequence

## TODO
- Pipe clean & verify other task-centered obj metrics/ attributes access (Done in test)
- Introduce this concept in other sample envs & multi-task eval

## Note
- Naming to `set` instead of `collection` is to differentiate what  [`RigidBodyCollection`](https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab/isaaclab/assets/rigid_object_collection/rigid_object_collection_data.py) from IssacLab provides. In our use case, we need to spawn 1 obj from N objs, where `RigidBodyCollection` API is to spawn all N objs for each id.
- MultiAssetSpawnerCfg for articulated objs will be tricky (/buggy) as PhyX APIs require it has the same joint prim path. It puts too much constraints on what could be added into the set
- [MultiAssetSpawnerCfg](https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab/isaaclab/sim/spawners/wrappers/wrappers_cfg.py#L16) for rigid objs require the same type of collision meshes, as written in Lab's doc.

image

### DIFF
--- a/isaaclab_arena/tests/test_object_set.py
+++ b/isaaclab_arena/tests/test_object_set.py
@@ -104,6 +104,7 @@ def _test_multi_objects_in_one_object_set(simulation_app):
     from isaacsim.core.utils.stage import get_current_stage
 
     from isaaclab_arena.assets.asset_registry import AssetRegistry
+    from isaaclab_arena.assets.object_reference import ObjectReference
     from isaaclab_arena.assets.object_set import RigidObjectSet
     from isaaclab_arena.cli.isaaclab_arena_cli import get_isaaclab_arena_cli_parser
     from isaaclab_arena.environments.arena_env_builder import ArenaEnvBuilder
@@ -111,7 +112,6 @@ def _test_multi_objects_in_one_object_set(simulation_app):
     from isaaclab_arena.scene.scene import Scene
     from isaaclab_arena.tasks.pick_and_place_task import PickAndPlaceTask
     from isaaclab_arena.utils.usd_helpers import get_asset_usd_path_from_prim_path
-    from isaaclab_arena.assets.object_reference import ObjectReference
 
     asset_registry = AssetRegistry()
     background = asset_registry.get_asset_by_name("kitchen")()
@@ -218,14 +218,20 @@ def _test_multi_object_sets(simulation_app):
                 prim_path=OBJECT_SET_2_PRIM_PATH.replace(".*", str(i)), stage=get_current_stage()
             )
 
-            assert path_1 is not None, "Path_1 from Prim Path " + OBJECT_SET_1_PRIM_PATH.replace(".*", str(i)) + " is None"
-            assert path_2 is not None, "Path_2 from Prim Path " + OBJECT_SET_2_PRIM_PATH.replace(".*", str(i)) + " is None"
+            assert path_1 is not None, (
+                "Path_1 from Prim Path " + OBJECT_SET_1_PRIM_PATH.replace(".*", str(i)) + " is None"
+            )
+            assert path_2 is not None, (
+                "Path_2 from Prim Path " + OBJECT_SET_2_PRIM_PATH.replace(".*", str(i)) + " is None"
+            )
             if i % 2 == 0:
                 assert "cracker_box.usd" in path_1, "Path_1 does not contain cracker_box.usd for env index " + str(i)
                 assert "sugar_box.usd" in path_2, "Path_2 does not contain sugar_box.usd for env index " + str(i)
             else:
                 assert "sugar_box.usd" in path_1, "Path_1 does not contain sugar_box.usd for env index " + str(i)
-                assert "mustard_bottle.usd" in path_2, "Path_2 does not contain mustard_bottle.usd for env index " + str(i)
+                assert (
+                    "mustard_bottle.usd" in path_2
+                ), "Path_2 does not contain mustard_bottle.usd for env index " + str(i)
     except Exception as e:
         print(f"Error: {e}")
         return False


isaaclab_arena/assets/object_library.py
```diff
@@ -40,6 +41,22 @@ def __init__(self, prim_path: str | None = None, initial_pose: Pose | None = Non
...
 
+class LibraryObjectSet(RigidObjectSet):
+    """
+    A set of objects.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        objects: list[LibraryObject],
```
alexmillane: Hmm I'm wondering why we have a `LibraryObjectSet`.

A `LibraryObject` is different to an `Object` in that it's parameters are static (i.e. compile time in c++ land).

A `LibraryObjectSet` is dynamic though, so I'm not quite sure why we need it.

Perhaps I'm missing something.

isaaclab_arena/assets/object_library.py
```diff
@@ -40,6 +41,22 @@ def __init__(self, prim_path: str | None = None, initial_pose: Pose | None = Non
...
+        prim_path: str | None = None,
+        initial_pose: Pose | None = None,
+        **kwargs,
+    ):
+        super().__init__(name=name, objects=objects, prim_path=prim_path, initial_pose=initial_pose, **kwargs)
+
+
 # TODO(peterd, 2025.11.05): Update all OV drive paths to use {ISAACLAB_NUCLEUS_DIR}
 # alias prior to public release once assets are synced to S3
```
alexmillane: Not part of your MR, but I just noticed this. @peterd-NV all our assets are now at least public, however I notice that some use `https://omniverse-content-production.s3-us-west-2.amazonaws.com/Assets/` and some use `{ISAACLAB_NUCLEUS_DIR}/Arena/assets/`. Am I correct that `ISAACLAB_NUCLEUS_DIR` translates to `https://omniverse-content-production.s3-us-west-2.amazonaws.com/`. Is there some opportunity for unification here? (I guess we can at least remove the comment now)?

isaaclab_arena/assets/object_set.py
```diff
@@ -0,0 +1,70 @@
...
+    A set of rigid objects.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        objects: list[Object],
+        prim_path: str | None = "/World/envs/env_.*/Object",
+        scale: tuple[float, float, float] = (1.0, 1.0, 1.0),
+        random_choice: bool = False,
```
alexmillane, line 26: I guess if this is true, each `env` gets a random object? What happens if it's false? I guess potentially objects are assigned sequentially or something?

I think that because this isn't immediately obvious, let's add a small description in the docstring of the class :).

isaaclab_arena/assets/object_set.py
```diff
@@ -0,0 +1,70 @@
...
+class RigidObjectSet(Object):
+    """
+    A set of rigid objects.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        objects: list[Object],
+        prim_path: str | None = "/World/envs/env_.*/Object",
```
alexmillane: This prim path looks a little unexpected to me. Usually we have `{ENV_REGEX_NS}/` appearing.

Perhaps there's something special about the prim paths of the multi-object things?

If this is intended, suggestion to explain in the doc string :).

isaaclab_arena/assets/object_set.py
```diff
@@ -0,0 +1,70 @@
...
+        scale: tuple[float, float, float] = (1.0, 1.0, 1.0),
+        random_choice: bool = False,
+        initial_pose: Pose | None = None,
+        **kwargs,
+    ):
+        if not self._examine_objects_type_are_rigid(objects):
+            raise ValueError(f"Object set {name} must contain at least 1 rigid object.")
```
alexmillane: This comment seems slightly incorrect according to my reading. We require that _all_ objects are rigid, not just one.

Suggestion to update the name: `_examine_objects_type_are_rigid` to `_are_all_objects_type_rigid`.

isaaclab_arena/assets/object_set.py
```diff
@@ -0,0 +1,70 @@
...
+
+        self.object_usd_paths = [object.usd_path for object in objects]
+        self.random_choice = random_choice
+
+        super().__init__(
+            name=name,
+            object_type=ObjectType.RIGID,
```
alexmillane, line 54: Do you have sense if `RigidObjectSet` conforms to the interfaces that a rigid object offers.

Looking at the code, I think those are:
- That you can get a contact sensor for it.
- That retrieving the pose of it works.

isaaclab_arena/scene/scene.py
```diff
@@ -23,8 +23,8 @@
 
 class Scene:
 
-    def __init__(self, assets: list[Asset] | None = None):
-        self.assets: dict[str, Asset] = {}
+    def __init__(self, assets: list[Asset, dict[str, list[Asset]]] | None = None):
+        self.assets: dict[str, Asset | list[Asset]] = {}
```
alexmillane: I wonder if we should introduce another type here. From my understanding, the requirement for a `dict[str, list[Asset]]` here is that we need to assign a name to the asset collection as a whole. The individual assets can just use their `asset.name`, but this doesn't work in the case of an asset collection, as we need a single name to refer to the whole collection.

What about

```
class AssetSelection:
    name: str
    assets: list[Asset]
```

Which would make the interface for the `Scene`:

```
def __init__(self, assets: list[Asset |  AssetSelection]
```

To me that would be a little easier to parse for a user.

What do you think? Is there something that I'm missing?

isaaclab_arena/scene/scene.py
```diff
@@ -35,8 +35,26 @@ def __init__(self, assets: list[Asset] | None = None):
...
+        return True
+
+    def add_asset(self, asset: Asset | dict[str, list[Asset]]):
+        """Add an asset to the scene.
+
+        Args:
+            asset: An Asset instance or a dictionary of Assets. If a dictionary is provided,
+                   the keys will be used as the names of the assets and the values will be the list of assets.
+        """
+        self._check_asset_name_is_valid(asset.name)
```
alexmillane: The return value of `_check_asset_name_is_valid` is ignored.

However I like that there are two different exceptions raised depending on the type of wrong name...

I think in this case maybe move the check code into `add_asset` and kill the separate function.

isaaclab_arena/scene/scene.py
alexmillane: **Edit:** It seems below that you're actually passing `LibraryObjectSet` to `Scene` so I think the above is just a case of incorrect type hints? Am I correct?

isaaclab_arena/tests/test_object_set.py
```diff
@@ -0,0 +1,244 @@
...
+    from isaaclab_arena.utils.usd_helpers import get_asset_usd_path_from_prim_path
+
+    asset_registry = AssetRegistry()
+    background = asset_registry.get_asset_by_name("packing_table")()
+    embodiment = asset_registry.get_asset_by_name("franka")()
+    cracker_box = asset_registry.get_asset_by_name("cracker_box")()
+
+    obj_set = LibraryObjectSet(
+        name="single_object_set", objects=[cracker_box, cracker_box], prim_path=OBJECT_SET_1_PRIM_PATH
+    )
```
alexmillane: Suggestion that we call this `ObjectSet`, rather than `LibraryObjectSet`.

isaaclab_arena/tests/test_object_set.py
```diff
@@ -0,0 +1,244 @@
...
+        assert path_2 is not None, (
+            "Path_2 from Prim Path " + OBJECT_SET_2_PRIM_PATH + " is None for env index " + str(i)
+        )
+        if i % 2 == 0:
+            assert "cracker_box.usd" in path_1, "Path_1 does not contain cracker_box.usd for env index " + str(i)
+            assert "sugar_box.usd" in path_2, "Path_2 does not contain sugar_box.usd for env index " + str(i)
+        else:
+            assert "sugar_box.usd" in path_1, "Path_1 does not contain sugar_box.usd for env index " + str(i)
+            assert "mustard_bottle.usd" in path_2, "Path_2 does not contain mustard_bottle.usd for env index " + str(i)
+    return True
```
alexmillane: Very cool!! 

Great functionality and great tests!

isaaclab_arena/tests/test_object_set.py
```diff
@@ -0,0 +1,244 @@
...
+    )
+    assert result, f"Test {_test_multi_object_sets.__name__} failed"
+
+
+if __name__ == "__main__":
+    test_empty_object_set()
+    test_articulation_object_set()
+    test_single_object_in_one_object_set()
+    test_multi_objects_in_one_object_set()
+    test_multi_object_sets()
```
alexmillane, line 288: Thank you for writing such thorough test. This is really high quality code.

isaaclab_arena_environments/kitchen_pick_and_place_environment.py
```diff
@@ -49,8 +45,22 @@ def get_env(self, args_cli: argparse.Namespace):  # -> IsaacLabArenaEnvironment:
...
+            objects = []
+            for obj in args_cli.object_set:
+                obj_from_set = self.asset_registry.get_asset_by_name(obj)()
+                objects.append(obj_from_set)
+            object_set = LibraryObjectSet(name="object_set", objects=objects)
+            object_set.set_initial_pose(Pose(position_xyz=(0.4, 0.2, 0.1), rotation_wxyz=(1.0, 0.0, 0.0, 0.0)))
+            scene = Scene(assets=[background, pick_up_object, destination_location, object_set])
+
+        else:
+            scene = Scene(assets=[background, pick_up_object, destination_location])
```
alexmillane, line 63: Very cool!

isaaclab_arena/assets/object_library.py
viiik-inside: `ISAACLAB_NUCLEUS_DIR` translates to `https://omniverse-content-production.s3-us-west-2.amazonaws.com/Assets/Isaac/5.1/Isaac/IsaacLab`

This is how these are defined

ISAAC_NUCLEUS_DIR = f"{nucleus_utils.get_assets_root_path()}/Isaac"
"""Path to the `Isaac` directory on the NVIDIA Nucleus Server."""

ISAACLAB_NUCLEUS_DIR = f"{ISAAC_NUCLEUS_DIR}/IsaacLab"
"""Path to the `Isaac/IsaacLab` directory on the NVIDIA Nucleus Server."""

Maybe one can use nucleus_utils

isaaclab_arena/assets/object_library.py
viiik-inside: Fair point. We probably would not need it, unless we have predefined libraryobject sets.

isaaclab_arena/assets/object_library.py
peterd-NV: I took a quick look and seems like all the ones using `https://omniverse-content-production.s3-us-west-2.amazonaws.com/Assets/Isaac/4.5/Isaac/Props/YCB` are for the YCB objects and point to the Isaac 4.5 S3 folder. Now that Sim 5.1 is public those files are also present in the 5.1 S3 folder so we can unify and replace those paths with something like `f{ISAAC_NUCLEUS_DIR}/Props/YCB...`

I'll open a separate MR for that.

isaaclab_arena/assets/object_set.py
xyao-nv: It's not intended and I was blindly coping stuffs over and forgot to change this. Thanks for pointing it out!

isaaclab_arena/assets/object_library.py
xyao-nv: Thanks for the suggestion. Removed it and consolidated the interface to only `RigidObjectSet`

isaaclab_arena/assets/object_library.py
xyao-nv: Thanks for clarifications!

isaaclab_arena/assets/object_set.py
xyao-nv: Added in the code as docstring for more details.

isaaclab_arena/assets/object_set.py
xyao-nv: I was mixing two checks in one call. Refactored it. Good suggestion. thx

isaaclab_arena/assets/object_set.py
xyao-nv: I believe it shall, as it returns the rigidbodycfg ultimately. Add the test for retrieving the pose and getting/accessing contact sensor.

isaaclab_arena/scene/scene.py
xyao-nv: Get rid of `AssetSelection` as we can just use `RigidObjectSet` to face the scene interaction.

isaaclab_arena/scene/scene.py
xyao-nv: yeah moved the check name is not None into add_asset and removed this func.

isaaclab_arena/tests/test_object_set.py
xyao-nv: I stick to `RigidObjectSet` to be more explicit. In case users provides microwave + box into this args.

isaaclab_arena/tests/test_object_set.py
xyao-nv: IssacLab does not have a test and I am just afraid changes from them breaking us.

xyao-nv (commented):
The Second Go-Around.
Changes involved,
- Stick to `RigidObjectSet` only for interfacing with `Scene`
- Stick to `ENV_REGEX_NS` for prim path
- Add tests for contact sensor and pose retrieval

isaaclab_arena/assets/object_set.py
alexmillane: Thanks!

isaaclab_arena/scene/scene.py
alexmillane: Sounds good! I think you forgot to change the signature of `__init__` (I notice it is changed in `add_asset` where it looks good!).

Thanks for doing this. Looks clean!

isaaclab_arena/scene/scene.py
```diff
@@ -35,13 +36,25 @@ def __init__(self, assets: list[Asset] | None = None):
...
         self.assets[asset.name] = asset
 
-    def add_assets(self, assets: list[Asset]):
+    def add_assets(self, assets: list[Asset | RigidObjectSet]):
         for asset in assets:
-            self.add_asset(asset)
+            if isinstance(asset, Asset | RigidObjectSet):
+                self.add_asset(asset)
+            else:
+                raise ValueError(f"Invalid asset type: {type(asset)}")
```
alexmillane: I'm think let's move this check to `add_asset`, that way it applies if the user calls `add_asset` or `add_assets`. Right now we only check in the case of adding multiple assets.

isaaclab_arena/tests/test_object_set.py
alexmillane: Agreed. Great!

isaaclab_arena/tests/test_object_set.py
```diff
@@ -0,0 +1,254 @@
...
+            prim_path=OBJECT_SET_1_PRIM_PATH.replace(".*", str(i)), stage=get_current_stage()
+        )
+        assert path is not None, "Path is None"
+        assert "cracker_box.usd" in path, "Path does not contain cracker_box.usd"
+        assert obj_set.get_initial_pose() is not None, "Initial pose is None"
+
+    assert env.scene[obj_set.name].data.root_pose_w is not None, "Root pose is None"
+    assert (
+        env.scene.sensors["pick_up_object_contact_sensor"].data.force_matrix_w is not None
+    ), "Contact sensor data is None"
```
alexmillane: Maybe this is obvious, but does this work in the case of different objects in the environment. Could we copy these lines into the test below?

isaaclab_arena/scene/scene.py
xyao-nv: good point. thx

isaaclab_arena/tests/test_object_set.py
xyao-nv: yeah added the test.

PR #299, opened 2025-12-18
Add texture randomizations to objects

xyao-nv:
## Summary
Users may want to render different textures for same objects in parallel envs, especially for USDs they don't maintain/have source access. This is a step towards heterogeneous parallel envs, i.e. same obj, diff textures.

## Detailed description
- There are obvious drawbacks from adding all kinds of rendering styles directly into USD files, and selecting those variants using USDFileCfg variants params. E.g. USDs will get bigger and bigger. And maybe there are USDs you don't maintain.
- As a consumer of those USDs, it creates the material (default/rand) on the fly, and wraps the assets with those material variants. Following [guidances](https://docs.isaacsim.omniverse.nvidia.com/5.1.0/replicator_tutorials/tutorial_replicator_isaac_randomizers.html#randomizing-textures) from IsaacSim
- It is only texture here, aka rendering related. PhyX related properties will be tricky to do, like scaling due to the collider.
- Users specify the object names to add those variants, and it fetches the prim paths for all those objects in policy_runner.py

E.g. kitchen cracker box
Screenshot from 2025-12-17 23-56-06

Screenshot from 2025-12-18 00-29-03

Screenshot from 2025-12-18 00-30-07

### DIFF
--- a/isaaclab_arena/examples/policy_runner.py
+++ b/isaaclab_arena/examples/policy_runner.py
@@ -28,20 +28,25 @@ def main():
         # Build scene
         arena_builder = get_arena_builder_from_cli(args_cli)
         env = arena_builder.make_registered()
-        from isaaclab_arena.utils.usd_helpers import apply_render_variants_to_scene
-        from isaaclab.sim.utils import get_current_stage
-        # convert to list of prim paths by speicfy the env_index
-        expanded_paths = [
-            f"{env.env_ns.format(i)}/cracker_box"
-            for i in range(args_cli.num_envs)
-        ]
-        apply_render_variants_to_scene(env, prim_paths=expanded_paths, stage=get_current_stage())
 
         if args_cli.seed is not None:
             env.seed(args_cli.seed)
             torch.manual_seed(args_cli.seed)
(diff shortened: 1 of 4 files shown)


isaaclab_arena/examples/policy_runner.py
```diff
@@ -35,6 +35,19 @@ def main():
...
+            from isaaclab.sim.utils import get_current_stage
+
+            from isaaclab_arena.utils.usd_helpers import randomize_objects_texture
+
+            randomize_objects_texture(
+                object_names=args_cli.randomize_object_texture_names,
+                num_envs=args_cli.num_envs,
+                env_ns=env.scene.env_ns,
+                stage=get_current_stage(),
+            )
```
alexmillane, line 49: In my opinion, we need to keep variations (and other feature requests) out of the policy_runner.

We should aim for an interface where objects can have texture variations applied. Something like:

```
apple.apply_color_variations()
```

That way, the policy runner doesn't need to worry about this stuff; it's all hidden away in the object definition and the policy runner just runs the policy.

PR #305, opened 2025-12-18
Add task <adjust_pose> and corresponding example <adjust_dexcube_pose>

lancel-nv:
## Summary
Adjusting object's pose is a classic and basic task in many complicated tasks, like pick-and-place, assembly, and manipulation. This PR introduces an `AdjustPoseTask` as a reusable task component and provides a concrete example environment (`AdjustCubePoseEnvironment`) that demonstrates how to use it.

The arena framework needed a fundamental task type for object pose adjustment that can be used as a building block for more complex manipulation tasks. Many robotic manipulation scenarios require moving an object to a specific position and/or orientation, so having a standardized task definition helps maintain consistency and reusability.

### What has been changed?

1. New Task:
- Added `AdjustPoseTask` in `isaaclab_arena/tasks/adjust_pose_task.py` - a configurable task that defines success criteria based on:
  - Position constraints: Optional bounding box defined by `x_range`, `y_range`, and `z_range`
  - Orientation constraints: Target quaternion with angular tolerance (in radians)

2. New Termination Logic:
- Added `adjust_pose_task_termination` function in `isaaclab_arena/tasks/terminations.py` that evaluates whether the object has reached the target pose

3. New Assets:
- Added `DexCube` object to the object library
- Added `FactoryTableBackground` to the background library

4. New Example Environment:
- Added `AdjustCubePoseEnvironment` in `isaaclab_arena_environments/adjust_cube_pose_environment.py` - a complete example showing how to configure and use the task with a Franka robot manipulating a cube

5. Tests:
- Added comprehensive tests in `isaaclab_arena/tests/test_adjust_cube_pose.py`:
  - Tests initial state (cube should not be in success state)
  - Tests success state (cube moved to target pose)
  - Tests multiple environments in parallel

- New capability: Users can now create object pose adjustment tasks with flexible success criteria (position zone + orientation tolerance)
- Reusable components: The `AdjustPoseTask` can be reused for various manipulation scenarios
- New assets: `DexCube` and `FactoryTableBackground` are now available in the asset library for other environments

### DIFF
--- a/isaaclab_arena/assets/object_library.py
+++ b/isaaclab_arena/assets/object_library.py
@@ -276,21 +276,6 @@ class Mug(LibraryObject, Placeable):
 
 
 @register_asset
-class DexCube(LibraryObject):
-    """
-    A dex cube.
-    """
-
-    name = "dex_cube"
-    tags = ["object"]
-    usd_path = f"{ISAAC_NUCLEUS_DIR}/Props/Blocks/DexCube/dex_cube_instanceable.usd"
-    scale = (1.0, 1.0, 1.0)
-
-    def __init__(self, prim_path: str | None = None, initial_pose: Pose | None = None):
-        super().__init__(prim_path=prim_path, initial_pose=initial_pose)
-
-
-@register_asset
 class GroundPlane(LibraryObject):
     """
     A ground plane.


docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
```diff
@@ -138,7 +138,7 @@ than the single environment evaluation because of the parallel evaluation.
 
 .. code-block:: text
 
-   Metrics: {'success_rate': 0.605, 'revolute_joint_moved_rate': 0.955, 'num_episodes': 200}
+   Metrics: {'success_rate': 0.605, 'door_moved_rate': 0.955, 'num_episodes': 200}
```
alexmillane: I think that you didn't intend to change this no?

I think what is happening here is that you need to rebase your MR on the top of main. There will likely be some conflicts that you'll need to sort out.

isaaclab_arena/tasks/adjust_pose_task.py
```diff
@@ -0,0 +1,115 @@
...
+from isaaclab_arena.metrics.metric_base import MetricBase
+from isaaclab_arena.metrics.object_moved import ObjectMovedRateMetric
+from isaaclab_arena.metrics.success_rate import SuccessRateMetric
+from isaaclab_arena.tasks.task_base import TaskBase
+from isaaclab_arena.tasks.terminations import adjust_pose_task_termination
+from isaaclab_arena.terms.events import set_object_pose
+from isaaclab_arena.utils.cameras import get_viewer_cfg_look_at_object
+
+
+class AdjustPoseTask(TaskBase):
```
alexmillane: "adjust" implies a relative pose target (to me). What do you think about renaming this to `GoalPoseTask`

isaaclab_arena/tasks/terminations.py
```diff
@@ -77,11 +78,78 @@ def objects_in_proximity(
...
+    if "y_range" in zone_cfg:
+        y_min, y_max = zone_cfg["y_range"]
+        in_y = (object_root_pos_w[:, 1] >= y_min) & (object_root_pos_w[:, 1] <= y_max)
+        success &= in_y
+
+    # Z Axis Check
+    if "z_range" in zone_cfg:
+        z_min, z_max = zone_cfg["z_range"]
+        in_z = (object_root_pos_w[:, 2] >= z_min) & (object_root_pos_w[:, 2] <= z_max)
+        success &= in_z
```
alexmillane: Can we express these lines as a loop over axis?

```
for idx, name in enumerate(["x_range", "y_range", "z_range]):
    range_min, range_max = zone_cfg[name]
    in = (object_root_pos_w[:, idx] >= range_min) & (object_root_pos_w[:, idx] <= range_max)
    ...
```

isaaclab_arena/tests/test_adjust_cube_pose.py
```diff
@@ -0,0 +1,259 @@
...
+        _test_adjust_cube_pose_multiple_envs,
+        headless=HEADLESS,
+    )
+    assert result, f"Test {_test_adjust_cube_pose_multiple_envs.__name__} failed"
+
+
+if __name__ == "__main__":
+    test_adjust_cube_pose_initial_state()
+    test_adjust_cube_pose_success()
+    test_adjust_cube_pose_multiple_envs()
```
alexmillane: Great! Thanks for adding such thorough tests!

isaaclab_arena_environments/cube_goal_pose_environment.py
```diff
@@ -0,0 +1,108 @@
...
+
+            reset_all = EventTerm(func=reset_scene_to_default, mode="reset", params={"reset_joint_targets": True})
+
+            init_franka_arm_pose = EventTerm(
+                func=franka_stack_events.set_default_joint_pose,
+                mode="reset",
+                params={
+                    "default_pose": [0.0444, -0.1894, -0.1107, -2.5148, 0.0044, 2.3775, 0.6952, 0.0400, 0.0400],
+                },
+            )
```
alexmillane: Is the idea here that you would like a different initial pose than the franka comes with?

What do you think about adding an optional argument to the franke for its initial joint state?

Then we could have that option available in all environments.

isaaclab_arena/tasks/goal_pose_task.py
```diff
@@ -0,0 +1,115 @@
...
+                        "x_range": [min, max],  # meters, optional
+                        "y_range": [min, max],  # meters, optional
+                        "z_range": [min, max]   # meters, optional
+                    },
+                    "orientation": {
+                        "target": [w, x, y, z],   # target quaternion
+                        "tolerance_rad": 0.1      # angular tolerance in radians
+                    }
+                }
+        """
```
alexmillane, line 45: One issue with passing these parameters as a dict is that it's not clear from the function signature what's expected in the dict. Suggestion to pass these all as parameters:

```
 def __init__(
      self,
      object: Asset,
      object_thresholds: dict | None = None,
      episode_length_s: float | None = None,
      target_x_range: tuple[float, float] | None = None,
      target_y_range: tuple[float, float] | None = None,
      target_z_range: tuple[float, float] | None = None,
      target_orientation_wxyz: tuple[float, float, float, float] | None = None,
      target_orientation_tolerance_rad: float | None = None
  ):
```

Then the user knows exactly what the task expects and the their types.

What we really need is a `PoseRange` object, which encapsulates these things into an object. @viiik-inside and I were thinking about adding such an object (because this has come up elsewhere). That would dramatically simplify the signature here:

```
 def __init__(
      self,
      object: Asset,
      target: PoseRange
      object_thresholds: dict | None = None,
      episode_length_s: float | None = None,
  ):
```

But maybe we can move to that in a later MR.

alexmillane (approved):
Thank you for adding this. I left a few comments. Once you've addressed them, merge away!

docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
lancel-nv: Solved. Will use rebase instead of merge. Thanks for your advice.

isaaclab_arena/tasks/adjust_pose_task.py
lancel-nv: Renamed task to: "GoalPoseTask". Renamed example to "CubeGoalPoseEnvironment"

isaaclab_arena/tasks/terminations.py
lancel-nv: Done

isaaclab_arena_environments/cube_goal_pose_environment.py
lancel-nv: I noticed that the current default pose often leads to singularities during teleoperation. The pose I used here is a ['proven' configuration from Isaac Lab](https://github.com/isaac-sim/IsaacLab/blob/9d17b524536b341da09987ef765a0ad0501c64ca/source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/stack/config/franka/stack_joint_pos_env_cfg.py#L35) that avoids these issues. 

I think adding an optional argument is a good idea which can be used in various environment and be adjusted more elegantly.

lancel-nv: I will submit another PR for that. Thanks

isaaclab_arena/tasks/goal_pose_task.py
lancel-nv: I see. I just update code as you suggested. And I agree on that create a PoseRange object would be helpful in many tasks and it can keep code neat.

isaaclab_arena_environments/cube_goal_pose_environment.py
lancel-nv: "What do you think about adding an optional argument to the franke for its initial joint state?"
---
I still added it here, not in another PR.. Because it's trivial if I only update it for Franka and it's difficult to make it compatible for all embodiments. In that case, we may need to re-write the reset-joint-pose function, while currently such functions are defined diversely.

PR #308, opened 2025-12-22
Add add ons to set configurations

viiik-inside:
## Summary
Add cfgs

### DIFF
--- a/isaaclab_arena/assets/background.py
+++ b/isaaclab_arena/assets/background.py
@@ -3,8 +3,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Any
-
 from isaaclab_arena.assets.object import Object
 from isaaclab_arena.assets.object_base import ObjectType
 from isaaclab_arena.utils.pose import Pose
@@ -22,8 +20,6 @@ class Background(Object):
         object_min_z: float,
         prim_path: str | None = None,
         initial_pose: Pose | None = None,
-        spawn_cfg_addon: dict[str, Any] = {},
-        asset_cfg_addon: dict[str, Any] = {},
         **kwargs
     ):
         super().__init__(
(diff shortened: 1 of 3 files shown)


isaaclab_arena/assets/background.py
```diff
@@ -29,6 +33,8 @@ def __init__(
             prim_path=prim_path,
             # Backgrounds don't have physics (at the moment)
             object_type=ObjectType.BASE,
+            spawn_cfg_addon=spawn_cfg_addon,
+            asset_cfg_addon=asset_cfg_addon,
```
alexmillane: You can remove these from here and just have them passed along in `kawgs` to keep the list of parameters short.

isaaclab_arena/assets/object.py
```diff
@@ -82,7 +92,8 @@ def _generate_base_cfg(self) -> AssetBaseCfg:
                 print("WARNING: Base object has lights, this may cause issues when using with multiple environments.")
         object_cfg = AssetBaseCfg(
             prim_path="{ENV_REGEX_NS}/" + self.name,
-            spawn=UsdFileCfg(usd_path=self.usd_path, scale=self.scale),
+            spawn=UsdFileCfg(usd_path=self.usd_path, scale=self.scale, **self.spawn_cfg_addon),
```
alexmillane: mega nit: suggestion to break this over multiple lines as we have in the other `UsdFileCfg`s.

isaaclab_arena/assets/object_library.py
```diff
@@ -36,6 +40,8 @@ def __init__(self, prim_path: str | None = None, initial_pose: Pose | None = Non
             object_type=self.object_type,
             scale=self.scale,
             initial_pose=initial_pose,
+            spawn_cfg_addon=self.spawn_cfg_addon,
+            asset_cfg_addon=self.asset_cfg_addon,
             **kwargs,
         )
```
alexmillane, line 47: Suggestion to add a test for these. You could just add a test which creates a Library object which has a non-default mass or something?

PR #310, opened 2025-12-30
[FEATURE] remote policy with server/client

kanghui0204:
Hi Team,

I added a remote server/client mechanism for IsaacLab-Arena that allows the IsaacLab environment and the policy model environment to run separately. Alex and I wrote a design document([design doc](https://docs.google.com/document/d/1uNcYswolxPD4PpHgvCapYuqwZ2QhzSnlJB9KRMFMQbQ/edit?tab=t.0)) for this feature, which includes our discussions about the design. Below is an overview of this PR.

### 1. Functionality
Previously, the IsaacLab-Arena pipeline ran entirely in a single process:

**IsaacLab env** → **Env policy class** (e.g., Gr00tClosedloopPolicy) → **local policy model** (e.g., Gr00tPolicy), all in one environment.

This PR adds remote policy/server–client support and decouples the env policy class from the local policy model. The new pipeline becomes:

**Client**: **IsaacLab env** → **Env policy class** (e.g., ActionChunkingClientPolicy)

**Server**: **VLA/VLN policy model** (e.g., Gr00tRemoteServerSidePolicy)

The client and server exchange observations and actions via sockets.

The server/client implementation lives in `isaaclab_arena/remote_policy` inside IsaacLab-Arena. Users can copy this directory and import it on the server side directly, without needing to install it as a separate package.

### 2. How to use it
On the client side, you still use `isaaclab_arena/examples/policy_runner.py`, but you now pass a few extra arguments. Example:
```
/isaac-sim/python.sh isaaclab_arena/examples/policy_runner.py \
  --policy_type isaaclab_arena.policy.action_chunking_client.ActionChunkingClientSidePolicy \
  --remote_host 127.0.0.1 \
  --remote_port 5555 \
  --remote_api_token API_TOKEN_123 \
  --remote_timeout_ms 5000 \
  --num_steps 2000 \
  --num_envs 10 \
  --enable_cameras \
  --headless \
  --remote_kill_on_exit \
  gr1_open_microwave \
  --embodiment gr1_joint
```
On the server side, there is a Python entry point `isaaclab_arena.remote_policy.remote_policy_server_runner.py`. You specify host/port/etc., select the policy type, and provide the policy config file:

```
python isaaclab_arena/remote_policy/remote_policy_server_runner.py
   --host 0.0.0.0 
   --port 5555
   --api_token API_TOKEN_123
   --timeout_ms 5000
   --policy_type isaaclab_arena_gr00t.policy.gr00t_remote_policy.Gr00tRemoteServerSidePolicy 
   --policy_config_yaml_path /workspace/isaaclab_arena_gr00t/policy/config/gr1_manip_gr00t_closedloop_config.yaml
```
### 3. Current example
Right now there is a working example for GR00T. On a given machine, the steps are:

**Start the server**
```
 bash docker/run_gr00t_server.sh   -d /home/scratch.hkang_sw/Isaac/dataset/models   --host 0.0.0.0   --port 5555   --api_token API_TOKEN_123   --timeout_ms 5000   --policy_type isaaclab_arena_gr00t.policy.gr00t_remote_policy.Gr00tRemoteServerSidePolicy   --policy_config_yaml_path /workspace/isaaclab_arena_gr00t/policy/config/gr1_manip_gr00t_closedloop_config.yaml
```
This uses the following defaults inside the script:

- host: 0.0.0.0
- port: 5555
- api_token: API_TOKEN_123
- timeout_ms: 5000
- policy_type: isaaclab_arena_gr00t.policy.gr00t_remote_policy.Gr00tRemoteServerSidePolicy 
- policy_config_yaml_path: /workspace/isaaclab_arena_gr00t/gr1_manip_gr00t_closedloop_config.yaml

**Start the client**
Set up the IsaacLab Docker container following
[IsaacLab Arena docker documentation](https://isaac-sim.github.io/IsaacLab-Arena/release/0.1.1/pages/quickstart/docker_containers.html
)(no GR00T installation is needed inside this container).

Inside the container, run:
```
/isaac-sim/python.sh isaaclab_arena/examples/policy_runner.py \
  --policy_type isaaclab_arena.policy.action_chunking_client.ActionChunkingClientSidePolicy \
  --remote_host 127.0.0.1 \
  --remote_port 5555 \
  --remote_api_token API_TOKEN_123 \
  --remote_timeout_ms 5000 \
  --num_steps 2000 \
  --num_envs 10 \
  --enable_cameras \
  --headless \
  --remote_kill_on_exit \
  gr1_open_microwave \
  --embodiment gr1_joint

```
With this setup, you can obtain results for the GR1 open microwave task using a remote GR00T policy server.
and I can get the result:
`{'success_rate': 0.58, 'revolute_joint_moved_rate': 0.95, 'num_episodes': 200}
`
### 4. What remains to be done? Future work
At the moment, documentation for this feature has not been updated. I’d like the team to review this PR first and confirm that the usage and interface look reasonable.

Planned future work includes:

1. Support for the new VLN task.
2. Improved communication efficiency. Currently, the server and client communicate over sockets. We can explore more efficient transports and/or observation compression to reduce communication latency.


docker/setup/install_gr00t_deps_wo_isaac.sh
```diff
@@ -0,0 +1,44 @@
...
+echo "Installing Isaac-GR00T with dependency group: $GROOT_DEPS_GROUP"
+python -m pip install --no-build-isolation --use-pep517 \
+    -e "${WORKDIR}/submodules/Isaac-GR00T/[$GROOT_DEPS_GROUP]"
+
+# Install flash-attn (optional, keep same version as Arena Dockerfile)
+echo "Installing flash-attn..."
+python -m pip install --no-build-isolation --use-pep517 flash-attn==2.7.1.post4 || \
+    echo "flash-attn install failed, continue without it"
+
+echo "GR00T dependencies installation completed successfully"
```
alexmillane: Thanks for putting this together.

This script seems to differ only very slightly from the `install_gr00t_deps.sh`. Is the only difference the `python` command used? I.e. `python` vs `/isaac-sim/python.sh`.

Could we combine these two scripts into a single script that takes an argument(s)?

docker/run_gr00t_server.sh
```diff
@@ -0,0 +1,96 @@
...
+
+# REQUIRED: host models directory (must be set by user)
+if [[ -z "${MODELS_DIR:-}" ]]; then
+  echo "ERROR: MODELS_DIR is not set."
+  echo "Please export MODELS_DIR to your host models directory, e.g.:"
+  echo "  export MODELS_DIR=/path/to/your/models"
+  echo "Then run:"
+  echo "  bash ./docker/run_gr00t_server.sh"
+  exit 1
+fi
```
alexmillane: Suggestion to used the same thing used in `run_docker.sh`:

Models are by default expected on the host at `$HOME/models` but this can be changed with the `-d` flag to the script.

See [default](https://github.com/isaac-sim/IsaacLab-Arena/blob/release/0.1.1/docker/run_docker.sh#L13) and the [optional override](https://github.com/isaac-sim/IsaacLab-Arena/blob/release/0.1.1/docker/run_docker.sh#L30)

docker/run_gr00t_server.sh
```diff
@@ -0,0 +1,96 @@
...
+  --name gr00t_policy_server_container \
+  -v "${MODELS_DIR}":/models \
+  gr00t_policy_server:latest \
+  --host "${HOST}" \
+  --port "${PORT}" \
+  --api_token "${API_TOKEN}" \
+  --timeout_ms "${TIMEOUT_MS}" \
+  --policy_type "${POLICY_TYPE}" \
+  --policy_config_yaml_path "${POLICY_CONFIG_YAML_PATH}"
+
```
alexmillane, line 183: Not strictly related to this MR, but I'm wondering if we should move all the gr00t related stuff. I.e. all the gr00t docker stuff and the gr00t related tests out of the core framework and into the `isaaclab_arena_gr00t` package. Actually this is basically certainly a good idea. Ideally the core framework would make no mention of gr00t.

It would probably make sense to do this _before_ merging this MR. So we can get everything gr00t related in the right package, before expanding what we can do there.

isaaclab_arena/examples/policy_runner_cli.py
```diff
@@ -137,9 +178,18 @@ def create_policy(args: argparse.Namespace) -> tuple[PolicyBase, int]:
             num_steps = policy.get_trajectory_length(policy.get_trajectory_index())
 
     elif args.policy_type == "gr00t_closedloop":
+        from pathlib import Path
```
alexmillane: Suggestion to move to the top of the file. 

We have some imports not at the top of the file if they require special dependencies that are conditionally not required. But prefer to put imports at the top.

isaaclab_arena/examples/policy_runner_cli.py
```diff
@@ -82,6 +82,45 @@ def add_gr00t_closedloop_arguments(parser: argparse.ArgumentParser) -> None:
...
+        type=int,
+        default=5555,
+        help="Remote policy server port (used when --policy_deployment remote).",
+    )
+    remote_group.add_argument(
+        "--remote_api_token",
+        type=str,
+        default=None,
+        help="Optional API token for remote policy server.",
+    )
```
alexmillane: Why is an API token (optionally) required? Suggestion to expand the help string here.

isaaclab_arena/examples/policy_runner.py
```diff
@@ -62,6 +62,17 @@ def main():
         metrics = compute_metrics(env)
         print(f"Metrics: {metrics}")
 
+        # NOTE(huikang, 2025-12-30)Explicitly clean up the remote policy client / server.
+        # Do NOT rely on a __del__ destructor in policy for this, since destructors are
+        # triggered implicitly and their execution time (or even whether they run)
+        # is not guaranteed, which makes resource cleanup unreliable.
+        if getattr(policy, "is_remote", False) and policy.is_remote:
+            kill_on_exit = getattr(args_cli, "remote_kill_on_exit", False)
```
alexmillane: Do we need this line? From what I can see our flags always contain `remote_kill_on_exit`

isaaclab_arena/examples/policy_runner.py
```diff
@@ -62,6 +62,17 @@ def main():
         metrics = compute_metrics(env)
         print(f"Metrics: {metrics}")
 
+        # NOTE(huikang, 2025-12-30)Explicitly clean up the remote policy client / server.
+        # Do NOT rely on a __del__ destructor in policy for this, since destructors are
+        # triggered implicitly and their execution time (or even whether they run)
+        # is not guaranteed, which makes resource cleanup unreliable.
+        if getattr(policy, "is_remote", False) and policy.is_remote:
```
alexmillane: Our `PolicyBase` enforces that all policies have the property `is_remote`. Suggestion to remove the `getattr` here - it hurts readability and will always be true.

isaaclab_arena/examples/policy_runner_cli.py
```diff
@@ -137,9 +178,18 @@ def create_policy(args: argparse.Namespace) -> tuple[PolicyBase, int]:
...
+        remote_config = None
+        if args.policy_deployment == "remote":
+            remote_config = RemotePolicyConfig(
+                host=args.remote_host,
+                port=args.remote_port,
+                api_token=args.remote_api_token,
+                timeout_ms=args.remote_timeout_ms,
+            )
+
+        policy = Gr00tClosedloopPolicy(policy_config_yaml_path=Path(args.policy_config_yaml_path), num_envs=args.num_envs, device=args.policy_device,policy_deployment=args.policy_deployment, remote_config=remote_config)
```
alexmillane: Something seems sub-optimal about this, although it is only partially related to your MR. 

We're still having to mention "gr00t" inside the core framework. Ideally, the core framework would be independent of any policy type.

This is more related to my comment above, that we should get policy specifics out of the core framework... Maybe let's speak about this topic. It's only partially related to the policy-server stuff. My fear though is that it gets harder to extract policy-specifics from the core framework once we add this (the policy-server) change.

isaaclab_arena/policy/policy_base.py
```diff
@@ -3,17 +3,56 @@
...
+from typing import Optional
+
+from isaaclab_arena.remote_policy.utils import RemotePolicyConfig
+from isaaclab_arena.remote_policy.policy_client import PolicyClient
 
 class PolicyBase(ABC):
-    def __init__(self):
+    def __init__(
+        self,
+        *,
```
alexmillane: Suggestion to remove this. We're not enforcing keyword arguments anywhere else. Or was there a specific reason for this?

isaaclab_arena/policy/policy_base.py
```diff
@@ -3,17 +3,56 @@
...
+        policy_deployment: str = "local",           # "local" or "remote"
```
alexmillane: Suggestion to use an `Enum`

isaaclab_arena/policy/policy_base.py
```diff
@@ -3,17 +3,56 @@
...
+        remote_config: Optional[RemotePolicyConfig] = None,
```
alexmillane: `remote_config: RemotePolicyConfig | None = None`

is the syntax we're using elsewhere (I believe it's considered "more modern")

isaaclab_arena/remote_policy/model_policy.py
```diff
@@ -0,0 +1,51 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Any, Dict
+
+
+class ModelPolicy(ABC):
```
alexmillane: Suggestion to rename to `ServerSidePolicy`?

isaaclab_arena/remote_policy/policy_client.py
```diff
@@ -0,0 +1,94 @@
...
+    # ------------------------------------------------------------------ #
+    # Public API
+    # ------------------------------------------------------------------ #
+
+    def ping(self) -> bool:
+        """Check if the server is reachable."""
+        try:
+            self.call_endpoint("ping", requires_input=False)
+            return True
+        except Exception:
```
alexmillane: Suggestion to print a warning message.

isaaclab_arena/remote_policy/policy_client.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def get_action(
+        self,
+        observation: Dict[str, Any],
+        options: Optional[Dict[str, Any]] = None,
+    ) -> Dict[str, Any]:
+        """Send policy_observations and get back policy action dict."""
+        payload: Dict[str, Any] = {"observation": observation}
+        # if options is not None:
+        #     payload["options"] = options
```
alexmillane: Intentional? `options` is currently unused.

isaaclab_arena/remote_policy/policy_client.py
```diff
@@ -0,0 +1,94 @@
...
+    
+        resp = self.call_endpoint(
+            endpoint="get_action",
+            data=payload,
+            requires_input=True,
+        )
+        if isinstance(resp, dict) and "action" in resp:
+            return resp["action"]
+        return resp
```
alexmillane: Why have this optional unwrapping? Can we assume one form of the response? Either the action dict, or a dict containing the action dict? Is there a reason to accept both?

isaaclab_arena_gr00t/policy_config.py
```diff
@@ -179,7 +179,14 @@ def __post_init__(self):
...
         ).exists(), f"state_joints_config_path does not exist: {self.state_joints_config_path}"
-        assert Path(self.model_path).exists(), f"model_path does not exist: {self.model_path}"
+        if not Path(self.model_path).exists():
+            warnings.warn(
+                "[GR00TConfig] model_path does not exist: "
+                f"{self.model_path}. No model checkpoint was found. "
+                "If this is the client side of a remote policy, this warning can be ignored. "
+                "However, if you are running a local policy or the server side of a remote policy, "
+                "the program will fail to load the model and cannot run correctly."
+            )
```
alexmillane: Suggestion to remove this warning altogether. From the message the warning will certainly fire in cases where nothing is wrong. That's pretty confusing for the user, I feel, so I'd suggest that we remove this check here. Perhaps we could move the check to a more appropriate place? For example in the local policy where we know we need a valid model path.

isaaclab_arena/remote_policy/remote_policy_server_runner.py
```diff
@@ -0,0 +1,104 @@
...
+from pathlib import Path
+from typing import Type
+
+from isaaclab_arena.remote_policy.model_policy import ModelPolicy
+from isaaclab_arena.remote_policy.policy_server import PolicyServer
+
+POLICY_REGISTRY: dict[str, str] = {
+    # policy_type: "module_path:ClassName"
+    "gr00t_closedloop": "isaaclab_arena_gr00t.gr00t_remote_policy:Gr00tRemoteModelPolicy",
+}
```
alexmillane: I'm thinking we should generalize and use our registry class used for Assets, Devices, and Retargetters for Policies too.

But I agree with the idea here: let's have a policy registry to allow users to register policies, to free the core code from directly depending on policy code. I think that this will be critical to getting policy (currently gr00t) code out of the core framework.

isaaclab_arena/remote_policy/remote_policy_config.py
alexmillane, line 1: Suggestion to rename to `remote_policy_config.py`

alexmillane (commented):
Thank you for putting this together.

Most of my comments are minor syntactic things that we can easily address.

I have one major comment on the design. At the moment, the server-client split point leaves half of gr00t policy running on the client and half (the model itself) running on the server. To me it would make sense to move the split point such that the whole policy is launched on the server. That way, in the remote inference case, we remove all gr00t details from the client side. It also simplifies the server side - the server just launches the same policy (called `Gr00tClosedloopPolicy`) that is launched in the local case.

This diagram explains (a simplified version) of the current design and the alternative that I'm proposing.

Image

docker/setup/install_gr00t_deps_wo_isaac.sh
kanghui0204: I remove this to `install_gr00t_deps.sh `and add a arguement on` install_gr00t_deps.s`

docker/run_gr00t_server.sh
kanghui0204: please check again ，I remove this part and add similar code as run_docker.sh

isaaclab_arena/examples/policy_runner_cli.py
kanghui0204: done

isaaclab_arena/examples/policy_runner_cli.py
kanghui0204: I add a detail description ，please check

isaaclab_arena/examples/policy_runner.py
kanghui0204: yes ， I refactor this part of code ，please check

isaaclab_arena/examples/policy_runner.py
kanghui0204: done

isaaclab_arena/policy/policy_base.py
kanghui0204: remove it

isaaclab_arena/policy/policy_base.py
kanghui0204: done ，please check

isaaclab_arena/policy/policy_base.py
kanghui0204: done

isaaclab_arena/remote_policy/model_policy.py
kanghui0204: done

isaaclab_arena/remote_policy/policy_client.py
kanghui0204: done

isaaclab_arena/remote_policy/policy_client.py
kanghui0204: remove options，right now it is not used. When I was developing this, I considered that if some special information needed to be transmitted, an extra options field might be necessary. Since it is not needed at the moment, I have removed it.

isaaclab_arena/remote_policy/policy_client.py
kanghui0204: also removed

isaaclab_arena_gr00t/policy_config.py
kanghui0204: OK ，I remove it now，we can add it once we have fully separated the policy and the client.

isaaclab_arena/remote_policy/remote_policy_server_runner.py
kanghui0204: I add a registry class in remote_policy folder .please check

isaaclab_arena/remote_policy/remote_policy_config.py
kanghui0204: done

docker/run_gr00t_server.sh
alexmillane: I have now made the change I suggested above. All gr00t-related code lives in `isaaclab_arena_gr00t`

isaaclab_arena/remote_policy/remote_policy_server_runner.py
alexmillane: Thank you for doing that. Unfortunately we've double-done the work :smiling_face_with_tear: . I added a new registry as part of moving the gr00t code out of the core package in #316 . I would suggest that you use my registry (which utilizes the same machinery we use for registering assets etc.)

kanghui0204:
> Thank you for putting this together.
> 
> Most of my comments are minor syntactic things that we can easily address.
> 
> I have one major comment on the design. At the moment, the server-client split point leaves half of gr00t policy running on the client and half (the model itself) running on the server. To me it would make sense to move the split point such that the whole policy is launched on the server. That way, in the remote inference case, we remove all gr00t details from the client side. It also simplifies the server side - the server just launches the same policy (called `Gr00tClosedloopPolicy`) that is launched in the local case.
> 
> This diagram explains (a simplified version) of the current design and the alternative that I'm proposing.
> 
> Image

Hi @alexmillane , I’ve reintroduced the `ActionChunkingClientPolicy` class to separate the client and server. The client now uses `ActionChunkingClientPolicy` to record action chunks and no longer needs to import anything from the model. During initialization, however, the server needs to send the client information such as the **action dimension, action chunk length, and observation keys**. This allows `ActionChunkingClientPolicy` to initialize the correct tensors and select the appropriate observations to send, without converting all observations to NumPy and sending everything to the server. In addition, in the interaction between client and server, the **task_description** must be set on the client side; once it is set, the client sends the **task_description** to the server so the server can store it. Could you help review the current implementation? If it looks good, I can clean up and organize the code.

kanghui0204:
image
for now the workflow is like this picture.

isaaclab_arena_gr00t/policy/gr00t_closedloop_policy.py
```diff
@@ -53,19 +61,14 @@ def __init__(self, policy_config_yaml_path: Path, num_envs: int = 1, device: str
 
         self.action_dim = len(self.robot_action_joints_config)
         if self.task_mode == TaskMode.G1_LOCOMANIPULATION:
-            # GR00T outputs are used for WBC inputs dim. So adding WBC commands to the action dim.
-            # WBC commands: navigate_command, base_height_command, torso_orientation_rpy_command
```
xyao-nv, line 131: Can you keep those comments?

isaaclab_arena_gr00t/policy/gr00t_remote_policy.py
```diff
@@ -0,0 +1,251 @@
...
+from isaaclab_arena_gr00t.utils.image_conversion import resize_frames_with_padding
+from isaaclab_arena_gr00t.utils.io_utils import create_config_from_yaml, load_robot_joints_config_from_yaml
+from isaaclab_arena_gr00t.utils.joints_conversion import (
+    remap_policy_joints_to_sim_joints,
+    remap_sim_joints_to_policy_joints,
+)
+from isaaclab_arena_gr00t.utils.robot_joints import JointsAbsPosition
+
+
+class Gr00tRemoteServerSidePolicy(ServerSidePolicy):
```
xyao-nv, line 44: A lot of Gr00t core codes from gr00t_closedloop_policy.py are duplicated in this class. Can you modify this class such that if there is any changes for the GR00T core logics (input prep, model forward to get predictions, and postprocess), there will be only 1 file needed to modify.

isaaclab_arena/policy/action_chunking_client.py
```diff
@@ -0,0 +1,208 @@
...
+        self.current_action_chunk[env_ids] = 0.0
+        self.current_action_index[env_ids] = -1
+        self.env_requires_new_action_chunk[env_ids] = True
+
+        if env_ids is None:
+            env_ids = torch.arange(self.num_envs, device=self.device, dtype=torch.long)
+
+        self.current_action_chunk[env_ids] = 0.0
+        self.current_action_index[env_ids] = -1
+        self.env_requires_new_action_chunk[env_ids] = True
```
xyao-nv: Can you remove duplication?

isaaclab_arena/remote_policy/remote_policy_server_runner.py
```diff
@@ -0,0 +1,76 @@
...
+import argparse
+from pathlib import Path
+from typing import Type
+from importlib import import_module
+
+#from isaaclab_arena.remote_policy.server_side_policy import ServerSidePolicy
+from isaaclab_arena.remote_policy.policy_server import PolicyServer
+
+
+def get_policy_cls(policy_type: str) -> type["ServerSidePolicy"]:
```
xyao-nv: Is the function copied? Is there any one to have one version of the func being kept in the repo?

isaaclab_arena/policy/action_chunking_client.py
```diff
@@ -0,0 +1,208 @@
...
+    packed: Dict[str, Any] = {}
+    for key_path in key_paths:
+        value = _get_nested(observation, key_path)
+        # torch.Tensor -> numpy
+        if isinstance(value, torch.Tensor):
+            value = value.detach().cpu().numpy()
+        packed[key_path] = value
+    return packed
+
+class ActionChunkingClientPolicy(PolicyBase):
```
xyao-nv: I wonder if it makes sense to have `ActionChunkingClientPolicy` inherited `ClientPolicy,` `ClientPolicy` inheriates `PolicyBase`.
ActionChunking is a specific post processing method used by major VLAs, there are others like avg smoothing method too. The simplest postprocessing could just be passing through, aka doing nothing for single-obs-single-action.

xyao-nv:
Thanks for updating your commits! One major concern is from code maintenance and future dev. Esp for GR00T core logics in terms of preprocess obs, get predictions, and postprocess predictions, I observed there are a significant amount of code duplication for those, partially because of the server-client setup. 

This is very risky and not ideal at all. When we have two versions of the same logic with slight modifications, we double the surface area for bugs and make future updates twice as risky.

To address this, we should aim for Single Source of Truth (SSOT) by abstracting the common logic and injecting the "slight modifications" as parameters or strategies.

Other, I prefer having a sample test setup as well. As our repo is OSS, every single commit is public and community will suffer from poorly-tested features. Also it prevents your added feature are broken by us, mostly me, when upgrading to GN1.6.

isaaclab_arena_gr00t/policy/gr00t_closedloop_policy.py
kanghui0204: done

isaaclab_arena/policy/action_chunking_client.py
kanghui0204: done

isaaclab_arena/policy/action_chunking_client.py
kanghui0204: Hi @xyao-nv , Let me check if I understand you correctly: you are suggesting a hierarchy like `PolicyBase` -> `ClientPolicy` -> post-processing policies (`ActionChunking`, avg smoothing, and others?). In this design, `ClientPolicy` would handle the networking part, and each post-processing policy would focus on how to handle the actions, right?

Could you share some code examples for `avg smoothing and other post-processing `methods? I’d like to take a look, because in the current workflow the server reads the config and sends it to the client, and my current `get_init_info` only transmits the action chunk length, which might not be sufficient for other post-processing strategies.
​

kanghui0204:
> As you are thinking of retiring `gr00t_closedloop_policy` and switching to the remote in one MR, it's better to keep in mind there will be a series of changes coming,
> 
> * by early next week, I hope I can finish upgrading everything to GN1.6
> * by early Feb, we need everything need to show for GTC DLI Lab ready, and we are showing a couple E2E GR00T tasks eval.
> * by end of this week, I am derisking the arena + gr00t deps setup on the DLI Lab dev env. I won't have any bandwidth to enable this awesome client-server arch for that DLI Lab in late Jan, given it probably wont be ready in a week and we have the list of todos.
> 
> So I would prefer, if you can do a 2-phase approach, adding client-server with policy core logics consolidation & tests & doc in one MR, and retiring our previous imple in another MR. Does it make sense? Or am I asking too much?

OK, I agree with you. I can split it into two phases.

kanghui0204:
> Re this server-client workflow (the server loading the policy config, the init config being sent from server to client, the task text being sent from client to server, and how observations are transmitted), it makes sense to me esp setting task description and init cfg. I would like to request refactoring for a better design on the ActionChunkingPolicyClass.
> 
> One more question, it looks to me your imple in`remote_policy` shall be pretty generic, is there any difference when adding sever supports to VLN vs. VLA? Is there any specific funcs VLN needs?

Actually, the current action post-processing in navila is somewhat similar to action chunking, but instead of generating many actions, it generates a single action that is executed multiple steps in the Isaac environment, so with my current implementation this can be adjusted on the server side.

kanghui0204:
Hi @alexmillane @xyao-nv , I’ve redesigned the server/client APIs and code. In this new API, we can decouple the client and server code, and it also supports extending other action post-processing methods. Could you help review it?

docker/setup/install_gr00t_deps.sh
```diff
@@ -1,41 +1,80 @@
...
-echo "CUDA_HOME=$CUDA_HOME"
-echo "PATH=$PATH"
-echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
-echo "TORCH_CUDA_ARCH_LIST=$TORCH_CUDA_ARCH_LIST"
+  echo "CUDA environment variables set:"
+  echo "CUDA_HOME=$CUDA_HOME"
+  echo "PATH=$PATH"
+  echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
+  echo "TORCH_CUDA_ARCH_LIST=$TORCH_CUDA_ARCH_LIST"
+fi
```
alexmillane, line 55: I think we can get rid of some duplication here. Actually I'm not sure I see any difference between then the branches of the conditionals.

For example:
- `CUDA_HOME=/usr/local/cuda-12.8` can be replaced with `CUDA_HOME=/usr/local/cuda` (which is a symlink to `/usr/local/cuda-12.8` that always exists.
- `export LD_LIBRARY_PATH=/usr/local/cuda-12.8/lib64:${LD_LIBRARY_PATH:-}` can be replaced with `export LD_LIBRARY_PATH={CUDA_HOME}/lib64:${LD_LIBRARY_PATH:-}`
- The `echo` statements are the same between both branches and can be moved outside the conditional statements.
- etc.

After these modifications these two branches of the conditional look the same to me? What was the idea here?

docker/Dockerfile.gr00t_server
```diff
@@ -0,0 +1,40 @@
...
+
+COPY isaaclab_arena/remote_policy ${WORKDIR}/isaaclab_arena/remote_policy
+COPY isaaclab_arena_gr00t ${WORKDIR}/isaaclab_arena_gr00t
+COPY isaaclab_arena_g1 ${WORKDIR}/isaaclab_arena_g1
+
+RUN pip install --no-cache-dir pyzmq msgpack
+
+ENV PYTHONPATH=${WORKDIR}
+
+ENTRYPOINT ["python", "-u", "-m", "isaaclab_arena.remote_policy.remote_policy_server_runner"]
```
alexmillane, line 40: Cool!

docker/run_gr00t_server.sh
```diff
@@ -0,0 +1,131 @@
...
+# Model directory on the host.
+# By default, use $HOME/models, but this can be overridden
+# by the MODELS_DIR environment variable or the -d / --models_dir flag.
+MODELS_DIR="${MODELS_DIR:-$HOME/models}"
+
+# Other parameters (can also be overridden via environment variables)
+HOST="${HOST:-0.0.0.0}"
+PORT="${PORT:-5555}"
+API_TOKEN="${API_TOKEN:-API_TOKEN_123}"
```
alexmillane: Not sure I understand this default. Shall we make that the default is no API token checking? Wdyt?

isaaclab_arena/policy/action_chunking_client.py
```diff
@@ -0,0 +1,205 @@
...
+    ) -> None:
+        super().__init__(config=config, remote_config=remote_config)
+
+        if self.action_mode is not ActionMode.CHUNK:
+            raise ValueError(
+                f"ActionChunkingClientSidePolicy expects CHUNK mode, "
+                f"got {self.action_mode.value!r}."
+            )
+
+        proto: ChunkingActionProtocol = self.protocol  # type: ignore
```
alexmillane: Appears unused?

isaaclab_arena/policy/action_chunking_client.py
```diff
@@ -0,0 +1,205 @@
...
+            default="cuda",
+            help="Device to use for the policy-related operations.",
+        )
+        return parser
+
+    @staticmethod
+    def from_args(args: argparse.Namespace) -> "ActionChunkingClientSidePolicy":
+        """Create an ActionChunkingClientSidePolicy from CLI arguments."""
+        remote_config = ClientSidePolicy.build_remote_config_from_args(args)
+        dummy_config: Any = None  # Replace with real config dataclass if needed.
```
alexmillane: Is this intended? I'm wondering what the purpose of this variable is, and how a user could make use of it?

isaaclab_arena/policy/action_chunking_client.py
```diff
@@ -0,0 +1,205 @@
...
+            raise ValueError(
+                f"Expected at least {protocol.action_chunk_length} "
+                f"actions per chunk, got {raw_chunk.shape[1]}"
+            )
+        if raw_chunk.shape[2] != protocol.action_dim:
+            raise ValueError(
+                f"Expected action_dim {protocol.action_dim}, got {raw_chunk.shape[2]}"
+            )
+
+        return raw_chunk[:, : protocol.action_chunk_length, :]
```
alexmillane: I'm wondering what's going on with the `action_chunk_length`. Perhaps niavely I would have expected to get back an exactly-correctly-sized chunk.

In the dimesions check, we seem to allow for a smaller chunk... Although we then slice up to the chunk length. Wouldn't that generate an error if the chunk was smaller?

What's the thinking here?

isaaclab_arena/policy/client_side_policy.py
```diff
@@ -0,0 +1,246 @@
...
+        if self.requested_action_mode is ActionMode.NONE:
+            raise NotImplementedError(
+                f"{self.__class__.__name__} must set 'requested_action_mode', "
+                f"can't be {ActionMode.NONE}."
+            )
+        if self.protocol_cls is None:
+            raise NotImplementedError(
+                f"{self.__class__.__name__} must set 'protocol_cls' "
+                f"to a concrete ActionProtocol subclass."
+            )
```
alexmillane: Suggestion to move these two parameters (`requested_action_mode` and `protocol_cls`) to constructor arguments without defaults. That is a more typical method for communicating "this parameter is required for initialization of this class".

isaaclab_arena/policy/client_side_policy.py
```diff
@@ -0,0 +1,246 @@
...
+
+    @staticmethod
+    def build_remote_config_from_args(args: argparse.Namespace) -> RemotePolicyConfig:
+        """Helper to construct RemotePolicyConfig from CLI arguments."""
+        if getattr(args, "remote_host", None) is None:
+            raise ValueError(
+                "Remote policy requires a remote policy server. "
+                "Please set --remote_host (and optionally --remote_port, "
+                "--remote_api_token, --remote_timeout_ms)."
+            )
```
alexmillane: Suggestion to make the arugment `required=True` and remove this check.

isaaclab_arena/policy/client_side_policy.py
```diff
@@ -0,0 +1,246 @@
...
+        return RemotePolicyConfig(
+            host=args.remote_host,
+            port=args.remote_port,
+            api_token=args.remote_api_token,
+            timeout_ms=args.remote_timeout_ms,
+        )
```
alexmillane, line 202: This function appears to be duplicated above? Suggestion to remove this duplicate.

isaaclab_arena/policy/client_side_policy.py
```diff
@@ -0,0 +1,246 @@
...
+        group = parser.add_argument_group(
+            "Remote Policy",
+            "Arguments for connecting to a remote policy server.",
+        )
+        group.add_argument(
+            "--remote_host",
+            type=str,
+            default=None,
+            help="Remote policy server host.",
+        )
```
alexmillane, line 164: See above, suggestion to remove default and make this `required`.

isaaclab_arena/policy/policy_base.py
```diff
@@ -86,6 +86,11 @@ def length(self) -> int | None:
         """Get the length of the policy (for dataset-driven policies)."""
         pass
 
+    @property
+    def is_remote(self) -> bool:
+        """Check if the policy is builted on remote."""
```
alexmillane: -> "Check if policy is run remotely"

isaaclab_arena/remote_policy/action_protocol.py
```diff
@@ -0,0 +1,78 @@
...
+
+class ActionMode(str, Enum):
+    """Action output mode of a policy.
+
+    Currently only CHUNK is used.
+    Other modes can be added later if needed.
+    """
+
+    NONE = "none"
+    CHUNK = "chunk"
```
alexmillane, line 21: From what I can tell `ActionMode.NONE` always results in an error. One of the advantages of using an `Enum` over say a `string` is that the values are constrained to being valid. The usage of `NONE` here breaks that.

I would suggest to remove `NONE` (unless I'm incorrect and there is a place where its usage is valid).

isaaclab_arena/remote_policy/action_protocol.py
```diff
@@ -0,0 +1,78 @@
...
+class ActionProtocol(ABC):
+    """Base handshake/config for a policy's action output.
+
+    - Encapsulates the ActionMode.
+    - Holds common fields (action_dim, observation_keys).
+    - Subclasses add mode-specific fields (e.g. chunk_length).
+    """
+
+    # Subclasses must override this.
+    MODE: ActionMode = ActionMode.NONE
```
alexmillane: `MISSING`

alexmillane: Perhaps we could also add a `__post_init__` check that this was actually set.

isaaclab_arena/remote_policy/message_serializer.py
```diff
@@ -0,0 +1,119 @@
...
+
+        # generic binary blob
+        if "__blob_class__" in obj:
+            return {
+                "mime": obj.get("mime"),
+                "data": obj.get("as_bytes"),
+            }
+
+        # other tagged types can be added here
+        return obj
```
alexmillane: Suggestion to expand the docstring. I dont understand the behaviour when obj is niether a `__ndarray_class__` or `__bloc_class__`. In those instances we just return the `obj` back. Is the `obj` then decoded by some other method?

isaaclab_arena/remote_policy/policy_client.py
```diff
@@ -0,0 +1,134 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Dict, Optional
+
+import zmq
+
+from .message_serializer import MessageSerializer
+from .remote_policy_config import RemotePolicyConfig
```
alexmillane: Is there a reason to use relative imports here? Perhaps this is because of the only partial install we're doing of `isaaclab_arena` in the server case?

I would recommend against the partial imports. I suggest we instead install the full (python) package (even though it will likely not pass tests due to missing dependencies on the server).

isaaclab_arena/policy/action_chunking_client.py
alexmillane, line 1: I'm wondering about the split of files between `isaaclab_arena/policy` and `isaaclab_arena/remote_policy`. There are a couple of files, which seem related to remote inference, that are in this folder.

Could we move them to the `remote_policy` folder? Wdyt?

isaaclab_arena/remote_policy/policy_server.py
```diff
@@ -0,0 +1,214 @@
...
+        self._policy = policy
+        self._running = True
+        self._context = zmq.Context()
+        self._socket = self._context.socket(zmq.REP)
+        self._socket.setsockopt(zmq.RCVTIMEO, timeout_ms)
+        bind_addr = f"tcp://{host}:{port}"
+        print(f"[PolicyServer] binding on {bind_addr}")
+        self._socket.bind(bind_addr)
+        self._api_token = api_token
+        self._serializer = MessageSerializer
```
alexmillane: nit: storing a class with static methods like this appears a bit.. strange. It's a nit, but I'd suggest just calling `MessageSerializer.serialize(msg)`, because that clearly communicates that these are static methods.

isaaclab_arena/remote_policy/policy_server.py
```diff
@@ -0,0 +1,214 @@
...
+                if not isinstance(request, dict):
+                    raise TypeError(f"Expected dict request, got {type(request)!r}")
+
+                print(f"[PolicyServer] request keys: {list(request.keys())}")
+
+                if not self._validate_token(request):
+                    self._socket.send(
+                        self._serializer.to_bytes({"error": "Unauthorized: invalid api_token"})
+                    )
+                    continue
```
alexmillane, line 153: We're you inspired by this API token design from somewhere else?

What's the intention? Because we transmit the API token unencrypted we don't get any security, so I'm wondering what the intention is?

isaaclab_arena/remote_policy/policy_server.py
```diff
@@ -0,0 +1,214 @@
...
+
+                endpoint = request.get("endpoint", "get_action")
```
alexmillane, line 155: nit: we have a default here. I would have thought that all requests _must_ include an endpoint. I would suggest enforcing that by not having a default. (and probably `assert "endpoint" in request`).

docker/setup/install_gr00t_deps.sh
kanghui0204:
```
44 # Set CUDA environment variables for GR00T installation
45 export CUDA_HOME=/usr/local/cuda-12.8
46 export PATH=/usr/local/cuda-12.8/bin:${PATH}
47 export LD_LIBRARY_PATH=/usr/local/cuda-12.8/lib64:${LD_LIBRARY_PATH:-}
48 export TORCH_CUDA_ARCH_LIST=8.0+PTX
49
50 echo "CUDA environment variables set:"
51 echo "CUDA_HOME=$CUDA_HOME"
52 echo "PATH=$PATH"
53 echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
54 echo "TORCH_CUDA_ARCH_LIST=$TORCH_CUDA_ARCH_LIST"
```

Actually, this part of the code comes from the existing GR00T installation process in IsaacLab Arena. It’s indeed a bit different from the standard setup (it uses `/usr/local/cuda-12.8`), so I didn’t modify it. Because of this special setup, I wasn’t sure if it was required for a specific reason, so I kept a separate branch.

isaaclab_arena/policy/action_chunking_client.py
kanghui0204: policybase requires a config to be passed in, but since action chunking doesn’t use it, I pass in None.

isaaclab_arena/policy/action_chunking_client.py
kanghui0204: removed

isaaclab_arena/policy/action_chunking_client.py
kanghui0204: This should be the logic inside the Gr00tCloseLoopPolicy. I guess the chunk_length might sometimes be longer than the configured value, but on both the server and client side we can use an exact chunk length instead.

isaaclab_arena/policy/client_side_policy.py
kanghui0204: I’ve modified the __init__ of ClientSidePolicy so that it now takes protocol_cls as an argument. ActionMode is a class attribute of protocol_cls, which lets us bind protocol_cls and ActionMode together. Please double-check this part of the code later.

isaaclab_arena/policy/client_side_policy.py
kanghui0204: removed

isaaclab_arena/policy/client_side_policy.py
kanghui0204: removed

isaaclab_arena/policy/client_side_policy.py
kanghui0204: removed

isaaclab_arena/policy/policy_base.py
kanghui0204: Done

isaaclab_arena/remote_policy/action_protocol.py
kanghui0204: removed

isaaclab_arena/remote_policy/action_protocol.py
kanghui0204: added

isaaclab_arena/remote_policy/message_serializer.py
kanghui0204: I’ve expanded the _decode_custom docstring to clarify this. _decode_custom is registered as the object_hook for msgpack.unpackb and only converts dicts tagged by _encode_custom back into higher-level types (e.g. numpy arrays or blobs).
Untagged dicts and all other value types are fully handled by msgpack’s default decoder, so in those cases _decode_custom simply returns obj as-is without any additional processing.
At the moment we only have custom handling for numpy arrays and generic blobs. If we introduce additional custom types in the future, we’ll add new tagged cases here.

isaaclab_arena/remote_policy/policy_client.py
kanghui0204: add absolute path

isaaclab_arena/policy/action_chunking_client.py
kanghui0204: Do you mean the two files **action_chunking_client.py** **and client_side_policy.py**? I actually see the corresponding `ClientSidePolicy` and `ActionChunkingClientSidePolicy` (together with `ReplayActionPolicy`) as client-side policies, so I placed them in the policy folder.

There is also another concern: both `ClientSidePolicy` and `ActionChunkingClientSidePolicy` inherit from `PolicyBase`. If we move them into the **remote_policy** folder, the remote policy code would need to import `PolicyBase` from the **policy package**, which introduces an unnecessary dependency. What do you think?

isaaclab_arena/remote_policy/policy_server.py
kanghui0204: Done

isaaclab_arena/remote_policy/policy_server.py
kanghui0204: The API token is not intended to provide real security. The goal is just to avoid accidental connections or mixed-up processes on the same network (i.e. a lightweight guardrail), not to protect against an active attacker, since the token is sent in plaintext.

isaaclab_arena/remote_policy/policy_server.py
kanghui0204: Done

docker/setup/install_gr00t_deps.sh
alexmillane: Yea, I don't see the difference between the branches of this if/else condition. The only difference is `/usr/local/cuda-12.8` vs `/usr/local/cuda`. On all systems I've worked on, these two are symlinked to one another. So I assume we can just use `/usr/local/cuda`.

But in the interests of getting this in, I'm happy to fix this after merging.

isaaclab_arena/policy/client_side_policy.py
alexmillane: looks good to me.

isaaclab_arena/remote_policy/message_serializer.py
alexmillane: That makes sense to me. Thank you for clarifying

isaaclab_arena/policy/action_chunking_client.py
alexmillane: I see your point. There are dependencies between these two folders. What do you think about just making `remote_policy` a subfolder of `policy`?

What do you think about that?

isaaclab_arena/remote_policy/policy_server.py
alexmillane: Ok let's leave it as is.

isaaclab_arena/policy/action_chunking_client.py
kanghui0204: Hi  @alexmillane, `remote_policy` does not depend on anything inside the `policy` folder.
In contrast, the __init__ in the `policy` folder does `from .replay_action_policy import *`, and `replay_action_policy` imports IsaacLab code.

When the server side starts, it needs to copy `remote_policy` into its environment.
If it instead copies the whole `policy` folder, it will also bring in other code, which can cause the program to crash.

For this reason, I feel that `remote_policy` should not be placed inside `policy`. What do you think?

PR #314, opened 2026-01-05
Refactor: CLI flags into policies

alexmillane:
## Summary
Moves the policy-related CLI flags from being centrally-defined, to travelling with the policy to which they apply.

## Note
- This is the first step towards moving GR00T code out of the core package.

## Detailed description
- Changes the definition of the base class to require defining flags, and construction from flags.
- Updates the run policy script to create a policy based on the passed string, and only add+parse policy relavent flags.


xyao-nv:
Thanks for refactoring that!

xyao-nv (approved):

PR #315, opened 2026-01-06
Add TurnKnob atomic task

xyao-nv:
### Summary
Introduce turning knob into atomic task library, together with a stand mixer asset curated from [LW open source repo](https://github.com/LightwheelAI/Lightwheel-simready-asset).

### Detailed description
- Stand Mixer object is added into OV server downloaded from LW's simready assets (open source, non commercial use)
- Introduce Turnable as affordance which has a linear mapping between revolute joint rotation angles and discrete value(e.g. temperature, power, time...)
- Add turn knob task defined as turning from init level to a target level. Details regarding mapping continuous range to discrete levels can be referred in docstrings.
- 
image

https://github.com/user-attachments/assets/e782ddf2-ab77-45c9-b160-5e4230f7aec0

### TODO
- Add task-specific metrics


isaaclab_arena/affordances/turnable.py
```diff
@@ -0,0 +1,147 @@
...
+            num_levels: Total number of discrete active levels (must be >= 1)
+            **kwargs: Additional arguments passed to AffordanceBase
+        """
+        super().__init__(**kwargs)
+        self.turnable_joint_name = turnable_joint_name
+
+        assert min_level_angle >= 0.0, "min_level_angle must be non-negative"
+        assert min_level_angle < max_level_angle, "min_level_angle must be less than max_level_angle"
+        self.min_level_angle = min_level_angle * math.pi / 180.0
+        self.max_level_angle = max_level_angle * math.pi / 180.0
```
alexmillane: suggestion to suffix with unit: `self.min_level_angle_rad`

isaaclab_arena/affordances/turnable.py
```diff
@@ -0,0 +1,147 @@
...
+            - Level 0: First active level at min_level_angle
+            - Level X-1: Last active level at max_level_angle
+            - target_level = integer in [-1, X-1]
+            - min_level_angle = joint angle (degrees) at level 0 (start of active range)
+            - max_level_angle = joint angle (degrees) at level X-1 (end of active range)
+            - Intermediate levels are linearly interpolated between min and max angles
+    """
+
+    def __init__(
+        self, turnable_joint_name: str, min_level_angle: float, max_level_angle: float, num_levels: int, **kwargs
```
alexmillane: Suggestion to suffix with units: `min_level_angle_deg`

isaaclab_arena/affordances/turnable.py
```diff
@@ -0,0 +1,147 @@
...
+            asset_cfg = SceneEntityCfg(self.name)
+        asset_cfg = self._add_joint_name_to_scene_entity_cfg(asset_cfg)
+
+        if target_level == -1:
+            # Dead zone: set to 0°
+            theta = 0.0
+        else:
+            # Active range: map level [0, num_levels-1] to angle [min_level_angle, max_level_angle]
+            step_size = (self.max_level_angle - self.min_level_angle) / (self.num_levels - 1)
+            theta = self.min_level_angle + step_size * target_level
```
alexmillane: To me it might make more sense to set it in the middle of the level. With this, you could set it to some level, then read it back and get the level below due to a rounding change, or a small motion due to physX.

Perhaps there's a good reason to have it this way?

isaaclab_arena/affordances/turnable.py
```diff
@@ -0,0 +1,147 @@
...
+
+        set_unnormalized_joint_position(env, asset_cfg, theta, env_ids)
+
+    def is_at_level(
+        self, env: ManagerBasedEnv, asset_cfg: SceneEntityCfg | None = None, target_level: int = -1
+    ) -> torch.Tensor:
+        """Check if the object is at the given level (in all the environments)."""
```
alexmillane: Consider rewriting the docstring slightly. This description makes it sound like you'd get back `True` if the asset reaches the target level in all envs (simultaneously).

isaaclab_arena/affordances/turnable.py
```diff
@@ -0,0 +1,147 @@
...
+        if asset_cfg is None:
+            asset_cfg = SceneEntityCfg(self.name)
+        asset_cfg = self._add_joint_name_to_scene_entity_cfg(asset_cfg)
+        current_level = self.get_turning_level(env, asset_cfg)
+        return torch.abs(current_level - target_level) <= 1e-6
```
alexmillane: Should this be an integer comparison? (And if not, should we make the level and integer? We're rounding it anyway.)

isaaclab_arena/assets/object_library.py
```diff
@@ -175,6 +176,36 @@ def __init__(self, prim_path: str | None = None, initial_pose: Pose | None = Non
         )
 
 
+@register_asset
+class StandMixer(LibraryObject, Turnable):
+    """
+    Encapsulates the pick-up object config for a pick-and-place environment.
```
alexmillane: Update the docstring

isaaclab_arena/assets/object_library.py
```diff
@@ -175,6 +176,36 @@ def __init__(self, prim_path: str | None = None, initial_pose: Pose | None = Non
...
+    """
+
+    name = "stand_mixer"
+    tags = ["object", "turnable"]
+    usd_path = (
+        "omniverse://isaac-dev.ov.nvidia.com/Isaac/IsaacLab/Arena/assets/object_library/StandMixer013/StandMixer013.usd"
+    )
```
alexmillane: If I'm not mistaken, this will fail in CI. Our CI only has access to public assets (which reflects what a public user has access to).

In your description, you say that this asset is from LightWheel. Could we retrieve this asset through the sdk, as we've done for the other LightWheel Assets?

If we plan on hosting ourselves (which I would be in favour of not doing), we need to name the asset `lightwheel_stand_mixer` (see [attribution guide](https://github.com/LightwheelAI/Lightwheel-simready-asset?tab=readme-ov-file#license))

isaaclab_arena/tasks/turn_knob_task.py
```diff
@@ -0,0 +1,107 @@
...
+
+class TurnKnobTask(TaskBase):
+    def __init__(
+        self,
+        turnable_object: Turnable,
+        target_level: int,
+        reset_level: int = -1,
+        episode_length_s: float | None = None,
+        task_description: str | None = None,
+    ):
```
alexmillane, line 30: Cool!

isaaclab_arena/utils/joint_utils.py
```diff
@@ -19,34 +20,72 @@ def unnormalize_value(value: float, min_value: float, max_value: float) -> float
...
+    joint_min, joint_max = get_joint_position_limits_from_articulation(articulation, joint_index)
     if joint_min < 0.0:
         target_joint_position = 1 - target_joint_position
     target_joint_position_unnormlized = unnormalize_value(target_joint_position, joint_min, joint_max)
-    articulation.write_joint_position_to_sim(
-        torch.tensor([[target_joint_position_unnormlized]]).to(env.device),
-        torch.tensor([joint_index]).to(env.device),
-        env_ids=env_ids.to(env.device) if env_ids is not None else None,
-    )
+    set_unnormalized_joint_position(env, asset_cfg, target_joint_position_unnormlized, env_ids)
```
alexmillane, line 91: Thanks for modularizing this file!

alexmillane (approved):
Looks great!

Just a few small comments. The only major thing is the (private) non-commercial asset. I think we should try to retieve that through the LightWheel sdk.

isaaclab_arena/affordances/turnable.py
xyao-nv: done

isaaclab_arena/affordances/turnable.py
xyao-nv: done

isaaclab_arena/assets/object_library.py
xyao-nv: done

isaaclab_arena/assets/object_library.py
xyao-nv: thx for the info! I was wondering why CI cannot fetch it. Lemme see if LW is hosted somehwere. If not, then i need to think of some ways.

isaaclab_arena/affordances/turnable.py
xyao-nv: great point of handling 0.5 oscillation. So I changed from round to floor in getter, and set it to mid point in setter.

isaaclab_arena/affordances/turnable.py
xyao-nv: done

isaaclab_arena/affordances/turnable.py
xyao-nv: done

PR #316, opened 2026-01-06
Refactor: move remaining gr00t-related code out of core

alexmillane:
## Summary
Moves all remaining gr00t-related code out of the `isaaclab_arena` package and into `isaaclab_arena_gr00t`.

## Detailed description
- We want the core framework to be gr00t independent. This change achieves that.
- Add `PolicyRegistry` to allow registration of policies via name.
- Adds the ability to request a policy via-module path, to be able to test policies not registered with the registry (gr00t will now be one of these policies)
- Moves the gr00t-related policy tests to `isaaclab_arena_gr00t`

## Note
- Running a gr00t evaluation gets slightly more verbose to run, because the `run_policy.py` script doesn't know about gr00t, so we need to specify which policy to run via it's module path.

Previously:

```
python isaaclab_arena/examples/policy_runner.py \
  --policy_type gr00t_closedloop \
  --policy_config_yaml_path isaaclab_arena_gr00t/gr1_manip_gr00t_closedloop_config.yaml \
  --enable_cameras \
  gr1_open_microwave \
  --embodiment gr1_joint
```

becomes

```
python isaaclab_arena/examples/policy_runner.py \
  --policy_type isaaclab_arena_gr00t.gr00t_closedloop_policy.Gr00tClosedloopPolicy \
  --policy_config_yaml_path isaaclab_arena_gr00t/gr1_manip_gr00t_closedloop_config.yaml \
  --enable_cameras \
  gr1_open_microwave \
  --embodiment gr1_joint
```

i.e. `gr00t_closedloop` -> `isaaclab_arena_gr00t.gr00t_closedloop_policy.Gr00tClosedloopPolicy`

I can't see a good way of avoiding this right now.

### DIFF
--- a/isaaclab_arena/examples/policy_runner.py
+++ b/isaaclab_arena/examples/policy_runner.py
@@ -8,12 +8,15 @@ import random
 import torch
 import tqdm
 from importlib import import_module
+from typing import TYPE_CHECKING
 
 from isaaclab_arena.cli.isaaclab_arena_cli import get_isaaclab_arena_cli_parser
 from isaaclab_arena.examples.policy_runner_cli import add_policy_runner_arguments
-from isaaclab_arena.policy.policy_base import PolicyBase
 from isaaclab_arena.utils.isaaclab_utils.simulation_app import SimulationAppContext
 from isaaclab_arena_environments.cli import get_arena_builder_from_cli, get_isaaclab_arena_environments_cli_parser
+
+if TYPE_CHECKING:
+    from isaaclab_arena.policy.policy_base import PolicyBase
 
 
 def get_policy_cls(policy_type: str) -> type["PolicyBase"]:


isaaclab_arena/examples/policy_runner.py
```diff
@@ -7,14 +7,39 @@
...
+
+    """
+    from isaaclab_arena.assets.asset_registry import PolicyRegistry
+
+    policy_registry = PolicyRegistry()
+    if policy_registry.is_registered(policy_type):
+        return policy_registry.get_policy(policy_type)
+    else:
+        print(f"Policy {policy_type} is not registered. Dynamically importing from path: {policy_type}")
+        assert "." in policy_type, "Unregistered policy_type must be a path to a policy class"
```
cvolkcvolk: Nit, Maybe a more direct assert error message like:  
"policy_type must be a dotted Python import path of the form 'module.submodule.ClassName', got: {policy_type}"

cvolkcvolk (approved):

isaaclab_arena/examples/policy_runner.py
alexmillane: Thanks. I think you're right there :)

xyao-nv (approved):
Thanks for refactoring!
I'd def prefer in the future, a better reference to`isaaclab_arena_gr00t.gr00t_closedloop_policy.Gr00tClosedloopPolicy`.

PR #321, opened 2026-01-07
Darrelldai/no docker

DarrellDai:
## Summary
Add setup for developing without docker

## Detailed description
- For people like developing without docker
- Make it accessible to Pycharm debugger

### DIFF
--- a/isaaclab_arena.sh
+++ b/isaaclab_arena.sh
@@ -109,6 +109,96 @@ extract_pip_uninstall_command() {
     echo ${pip_uninstall_command}
 }
 
+# setup uv environment for Isaac Lab Arena
+setup_uv_env() {
+    # get environment name from input
+    local env_name="$1"
+    local python_path="$2"
+
+    # check uv is installed
+    if ! command -v uv &>/dev/null; then
+        echo "[ERROR] uv could not be found. Please install uv and try again."
+        echo "[ERROR] uv can be installed here:"
+        echo "[ERROR] https://docs.astral.sh/uv/getting-started/installation/"
+        exit 1
+    fi
+
+    # check if _isaac_sim symlink exists and isaacsim-rl is not installed via pip
+    if [ ! -L "${ISAACLAB_PATH}/_isaac_sim" ] && ! python -m pip list | grep -q 'isaacsim-rl'; then
+        echo -e "[WARNING] _isaac_sim symlink not found at ${ISAACLAB_PATH}/_isaac_sim"
+        echo -e "\tThis warning can be ignored if you plan to install Isaac Sim via pip."
+        echo -e "\tIf you are using a binary installation of Isaac Sim, please ensure the symlink is created before setting up the conda environment."
+    fi
+
+    # check if the environment exists
+    local env_path="${ISAACLAB_ARENA_PATH}/${env_name}"
+    if [ ! -d "${env_path}" ]; then
+        echo -e "[INFO] Creating uv environment named '${env_name}'..."
+        uv venv --clear --python "${python_path}" "${env_path}"
+        # Install pip so isaaclab.sh can use python -m pip
+        echo -e "[INFO] Installing pip in uv environment..."
+        uv pip install --python "${env_path}/bin/python" pip
+    else
+        echo "[INFO] uv environment '${env_name}' already exists."
+    fi
+
+    # define root path for activation hooks
+    local isaaclab_root="${ISAACLAB_ARENA_PATH}"
+
+    # cache current paths for later
+    cache_pythonpath=$PYTHONPATH
+    cache_ld_library_path=$LD_LIBRARY_PATH
+
+    # ensure activate file exists
+    touch "${env_path}/bin/activate"
+
+     # add variables to environment during activation
+    cat >> "${env_path}/bin/activate" <<EOF
+export ISAACLAB_PATH="${ISAACLAB_PATH}"
+alias isaaclab="${ISAACLAB_PATH}/isaaclab.sh"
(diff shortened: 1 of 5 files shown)


docker/run_docker.sh
```diff
@@ -136,12 +136,19 @@ else
                     "--runtime=nvidia"
                     "--gpus=all"
                     "-v" ".:${WORKDIR}"
+                    "-v" "./isaaclab_arena:${WORKDIR}/isaaclab_arena"
+                    "-v" "./isaaclab_arena_g1:${WORKDIR}/isaaclab_arena_g1"
+                    "-v" "./isaaclab_arena_gr00t:${WORKDIR}/isaaclab_arena_gr00t"
+                    "-v" "./submodules/IsaacLab:${WORKDIR}/submodules/IsaacLab"
```
alexmillane, line 142: I think these are covered by the line above which maps the folder at a higher level. We recently added this higher level mapping, so perhaps these lines are no longer needed.

docs/index.rst
```diff
@@ -2,8 +2,9 @@
 =================================
 
 .. warning::
-   This is the latest version of IsaacLab Arena. It contains the newest features but may not be fully tested yet.
-   For the tested version, please refer to the `release/0.1.1 branch <https://isaac-sim.github.io/IsaacLab-Arena/release/0.1.1/index.html>`_.
+   This is the release/0.1.1 version of Isaac Lab Arena. If you are looking for the latest development version,
+   please refer to the `main branch <https://isaac-sim.github.io/IsaacLab-Arena/main/index.html>`_.
+   This comes with the latest features but is not yet fully tested.
```
alexmillane, line 7: This change seems in error.

I checked and this file has a merge conflict with `main`. Suggestion to rebase on main. There should be no changes on this file.

isaaclab_arena_environments/gr1_open_microwave_environment.py
```diff
@@ -62,7 +62,7 @@ def get_env(self, args_cli: argparse.Namespace):  # -> IsaacLabArenaEnvironment:
             name=self.name,
             embodiment=embodiment,
             scene=scene,
-            task=OpenDoorTask(microwave, openness_threshold=0.8, reset_openness=0.2, episode_length_s=2.0),
+            task=OpenDoorTask(microwave, openness_threshold=0.8, reset_openness=0.3, episode_length_s=2.0),
```
alexmillane, line 65: Intended change? This doesn't appear to be related to the subject of the MR.

isaaclab_arena.sh
```diff
@@ -0,0 +1,294 @@
+# Path to your Isaac Sim installation (e.g., ~/.local/share/ov/pkg/isaac-sim-5.0.0)
+export ISAAC_SIM_PATH="$HOME/isaacsim"
```
alexmillane, line 2: I can't see this variable used anywhere. Suggestion to remove?

isaaclab_arena.sh
```diff
@@ -0,0 +1,294 @@
...
+            else
+                framework_name=$2
+                shift # past argument
+            fi
+            # Install IsaacLab
+            $ISAACLAB_PATH/isaaclab.sh -i $framework_name
+            # Install Isaac-GR00T
+            ${pip_command} -e "${ISAACLAB_ARENA_PATH}/submodules/Isaac-GR00T"
+            # Install IsaacLab Arena
+            ${pip_command} -e .
```
alexmillane, line 249: This assumes that all dependencies not listed in setup.py are available in the environment this script is being run in, I suppose? Is that correct?

alexmillane: _Edit: I notice you've added some additional dependencies to our `setup.py`. Nice!_

Perhaps with those additions, we have everything we need on some platforms?

I would suggest that we set up a test that runs this script in a ubuntu22 docker, and we're left with a valid Arena installation (i.e. an installation which passes the tests).

setup.py
```diff
@@ -15,6 +15,16 @@
...
+    install_requires=[
+        "pytest",
+        "jupyter",
+        "typing_extensions",
+        "onnxruntime",
+        "qpsolvers==4.8.1",
+        "vuer[all]",
+        "lightwheel-sdk",
+        "huggingface-hub[cli]",
+    ],
```
alexmillane, line 27: thanks!!

alexmillane (commented):
Looks good!

I have a few minor comments. My only significant request is that we add a test which runs the install script inside a ubuntu22 container and checks that we're left with a valid Arena installation.

Thanks for doing this!

isaaclab_arena_environments/gr1_open_microwave_environment.py
DarrellDai: I think this is a unintended change, thanks for pointing out

docs/index.rst
DarrellDai: Yes

isaaclab_arena.sh
DarrellDai: Yes. Forgot to delete this one

docker/run_docker.sh
DarrellDai: This part is for a different purpose, I'll seperate them

isaaclab_arena.sh
DarrellDai: I've tested it on Ubuntu 24, and it works. Do you think it's still necessary to test it in a docker for 22?

DarrellDai:
I just made a clearer PR, which is only for installation outside docker and pycharm debugging. This PR will be closed. https://github.com/isaac-sim/IsaacLab-Arena/pull/327

PR #323, opened 2026-01-07
Add sequential task example: put into microwave and close door

qianl-nv:
## Summary
Add an example environment for sequential task: pick&place object into microwave and then close the door

## Detailed description
- This is a simple sequential task environment.
- It combines the PickAndPlace and CloseDoor atomic tasks to demonstrate how to use the sequential task API.
- This example is verified with teleop record/replay only. Mimc datagen/policy finetuning has not be verified yet.
image
image


isaaclab_arena_environments/franka_put_and_close_door_environment.py
```diff
@@ -0,0 +1,138 @@
...
+        # Create a sequential task wrapper class
+        class SequentialPutAndCloseDoorTask(SequentialTaskBase):
+            def __init__(self, subtasks, episode_length_s=None):
+                super().__init__(subtasks=subtasks, episode_length_s=episode_length_s)
+
+            def get_metrics(self):
+                return []
+
+            def get_mimic_env_cfg(self, arm_mode):
+                return None
```
alexmillane, line 113: Are these required? If so, I wonder if we can move these to defaults in `SequentialTaskBase` such that users needn't implement empty functions if they don't have anything specific to add.

isaaclab_arena_environments/franka_put_and_close_door_environment.py
```diff
@@ -0,0 +1,138 @@
...
+            pick_up_object=pick_object,
+            destination_location=destination_ref,
+            background_scene=background,
+            task_description=task_description_pick,
+        )
+
+        # Create a sequential task wrapper class
+        class SequentialPutAndCloseDoorTask(SequentialTaskBase):
+            def __init__(self, subtasks, episode_length_s=None):
+                super().__init__(subtasks=subtasks, episode_length_s=episode_length_s)
```
alexmillane, line 107: This is more a question for @peterd-NV, but one thing I don't understand is why we need to define a new class here. The class is empty.

It would make more sense to me to do:

```
put_and_close_door_task = SequentialTaskBase(subtasks=subtasks, episode_length_s=90)
```

That way we get out of defining a new class in the middle of a function like this. To me the above seems more parsimonious.

(This comment shouldn't prevent merging this MR by the way, it's more intended as a general design question).

alexmillane (approved):
Looks great! Thank you for doing this!

isaaclab_arena_environments/franka_put_and_close_door_environment.py
qianl-nv: yes those are required atm, the default would raise NotImplementedError (abstractmethod) from TaskBase. SequentialTaskBase does not implement default for these.
Is there a particular reason we are enforcing atomic tasks (TaskBase) to implement get_metrics/get_mimic_env_cgf? If not we should just update TaskBase (instead of SequentialTaskBase) with the empty defaults.

isaaclab_arena_environments/franka_put_and_close_door_environment.py
qianl-nv: If we move the empty default for get_metrics/get_mimic_env_cgf to the base comment then we should be able to remove the new class definition and use your suggestion

isaaclab_arena_environments/franka_put_and_close_door_environment.py
qianl-nv: Since this is dependent on the resolution for the comment below regarding SequentialTaskBase, we'll address it in the follow up PR that updates SequentialTaskBase potentially

PR #327, opened 2026-01-08
Feature/install outside docker pycharm debug

DarrellDai:
## Summary
- Add installation scripts outside docker
- Add Pycharm debug support

## Detailed description
- For those who don't like use a docker for more flexible development
- For those who prefer using Pycharm for debugging

#### Create Virtual Environment:
- uv
`isaaclab_arena.sh -u [env_name]`
- conda
`isaaclab_arena.sh -c [env_name]`
#### Install Isaac Lab Arena:
`isaaclab_arena.sh -i [framework_name]`
#### Add Debugger Config (First Time):
`isaaclab_arena.sh -y`
-  It will add project settings, and create a run/debug configuration example with all required environment variables
#### Complete Debugger Config :
`isaaclab_arena.sh -y`
- Each time you create a run/debug configuration, this command will help you add all required environment variables.
- Note that, brefore running this command, you need to toggle "Store as project file" when you create a run/debug configuration, and restart Pycharm.


isaaclab_arena.sh
viiik-inside, line 1: This looks amazing!! Thanks for doing this. I have a small request if you have time :) 

Can this be split into multiple files. Like isaaclab_arena.sh contains only the CLI and then calls other funtions which can be under scripts called uv.sh, pycharm.sh, install.sh, python.sh etc. It would look much nicer and the user will see less code that way.

DarrellDai: Thanks. I just pushed, please check

viiik-inside: Amazing

PR #328, opened 2026-01-08
Add Differentiable Object Placement with Semantic Relations

cvolkcvolk:
## Summary

Introduces the relations module - a differentiable solver for placing objects in 3D space using semantic spatial relationships. Users define high-level constraints like On(table) or NextTo(box), and the solver computes optimal positions via gradient descent.

```
coffee_machine.add_relation(On(table, clearance_m=0.02))
mug.add_relation(On(table, clearance_m=0.01))
mug.add_relation(NextTo(coffee_machine, side=Side.RIGHT, distance_m=0.05))

placer = ObjectPlacer(params=ObjectPlacerParams())
result = placer.place(objects=[table, mug, coffee_machine], anchor_object=table)
```

- `Relations` (currently `On()` & `NextTo(right)` implemented, are pure data containers holding relationship
  metadata (parent, side, distance) without loss computation logic.
- `LossStrategy` classes (e.g., NextToLossStrategy) handle the actual loss
  computation, separating relation from lower level loss computatoin.
- `RelationSolver` optimization using a strategy registry mapping the Relations to respective LossStrategies
- `ObjectPlacer` facing the user API and handling 1) random initialization within configurable bounds, 2) Running the gradient-based solver, 3) Retry logic for failed placements 4) Applying solved positions back to objects

### DIFF
--- a/isaaclab_arena/assets/object.py
+++ b/isaaclab_arena/assets/object.py
@@ -8,13 +8,14 @@ from typing import Any
 from isaaclab.assets import ArticulationCfg, AssetBaseCfg, RigidObjectCfg
 from isaaclab.managers import EventTermCfg, SceneEntityCfg
 from isaaclab.sim.spawners.from_files.from_files_cfg import UsdFileCfg
+from isaaclab_tasks.manager_based.manipulation.stack.mdp.franka_stack_events import randomize_object_pose
 
 from isaaclab_arena.assets.object_base import ObjectBase, ObjectType
 from isaaclab_arena.assets.object_utils import detect_object_type
 from isaaclab_arena.relations.relations import Relation
 from isaaclab_arena.terms.events import set_object_pose
 from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox
-from isaaclab_arena.utils.pose import Pose
+from isaaclab_arena.utils.pose import Pose, PoseRange
 from isaaclab_arena.utils.usd_helpers import compute_bounding_box_from_usd, has_light, open_stage
 
 
@@ -90,12 +91,19 @@ class Object(ObjectBase):
             self.bounding_box = compute_bounding_box_from_usd(self.usd_path, self.scale)
         return self.bounding_box.get_corners_at(pos)
 
-    def set_initial_pose(self, pose: Pose) -> None:
+    def set_initial_pose(self, pose: Pose | PoseRange) -> None:
+        """Set the initial pose of the object.
+
+        Args:
+            pose: The pose to set. Can be a single pose or a pose range.
+                  In the case of a PoseRange, the object will be reset
+                  to a random pose within the range on environment reset.
+        """
         self.initial_pose = pose
         self.object_cfg = self._add_initial_pose_to_cfg(self.object_cfg)
         self.event_cfg = self._update_initial_pose_event_cfg(self.event_cfg)
 
-    def get_initial_pose(self) -> Pose | None:
+    def get_initial_pose(self) -> Pose | PoseRange | None:
         return self.initial_pose
 
     def is_initial_pose_set(self) -> bool:
@@ -172,8 +180,12 @@ class Object(ObjectBase):
     ) -> RigidObjectCfg | ArticulationCfg | AssetBaseCfg:
         # Optionally specify initial pose
         if self.initial_pose is not None:
-            object_cfg.init_state.pos = self.initial_pose.position_xyz
-            object_cfg.init_state.rot = self.initial_pose.rotation_wxyz
+            if isinstance(self.initial_pose, Pose):
+                initial_pose = self.initial_pose
+            elif isinstance(self.initial_pose, PoseRange):
+                initial_pose = self.initial_pose.get_midpoint()
+            object_cfg.init_state.pos = initial_pose.position_xyz
+            object_cfg.init_state.rot = initial_pose.rotation_wxyz
         return object_cfg
 
     def _requires_reset_pose_event(self) -> bool:
@@ -185,24 +197,55 @@ class Object(ObjectBase):
 
     def _init_event_cfg(self) -> EventTermCfg | None:
         if self._requires_reset_pose_event():
-            return EventTermCfg(
-                func=set_object_pose,
-                mode="reset",
-                params={
-                    "pose": self.initial_pose,
-                    "asset_cfg": SceneEntityCfg(self.name),
-                },
-            )
+            # Two possible event types:
+            # - initial pose is a Pose - reset to a single pose
+            # - initial pose is a PoseRange - reset to a random pose within the range
+            if isinstance(self.initial_pose, Pose):
+                return EventTermCfg(
+                    func=set_object_pose,
+                    mode="reset",
+                    params={
+                        "pose": self.initial_pose,
+                        "asset_cfg": SceneEntityCfg(self.name),
+                    },
+                )
+            elif isinstance(self.initial_pose, PoseRange):
+                return EventTermCfg(
+                    func=randomize_object_pose,
+                    mode="reset",
+                    params={
+                        "pose_range": self.initial_pose.to_dict(),
+                        "asset_cfgs": [SceneEntityCfg(self.name)],
+                    },
+                )
+            else:
+                raise ValueError(f"Initial pose {self.initial_pose} is not a Pose or PoseRange")
         else:
             return None
+
+    def _needs_reinit_of_event_cfg(self):
+        # If there is no event cfg, needs to be reinitialized
+        if self.event_cfg is None:
+            return True
+        # Here we check if the event cfg is for the correct pose type.
+        # If not, needs to be reinitialized.
+        if (isinstance(self.initial_pose, Pose) and ("pose" not in self.event_cfg.params)) or (
+            isinstance(self.initial_pose, PoseRange) and ("pose_range" not in self.event_cfg.params)
+        ):
+            return True
+        return False
 
     def _update_initial_pose_event_cfg(self, event_cfg: EventTermCfg | None) -> EventTermCfg | None:
         if self._requires_reset_pose_event():
             # Create an event cfg if one does not yet exist
-            if event_cfg is None:
+            if self._needs_reinit_of_event_cfg():
                 event_cfg = self._init_event_cfg()
-            # Add the initial pose to the event cfg
-            event_cfg.params["pose"] = self.initial_pose
+            if isinstance(self.initial_pose, Pose):
+                event_cfg.params["pose"] = self.initial_pose
+            elif isinstance(self.initial_pose, PoseRange):
+                event_cfg.params["pose_range"] = self.initial_pose.to_dict()
+            else:
+                raise ValueError(f"Initial pose {self.initial_pose} is not a Pose or PoseRange")
         else:
             event_cfg = None
         return event_cfg
--- a/docs/pages/example_workflows/locomanipulation/step_1_environment_setup.rst
+++ b/docs/pages/example_workflows/locomanipulation/step_1_environment_setup.rst
@@ -26,7 +26,7 @@ Environment Description
                from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
                from isaaclab_arena.scene.scene import Scene
                from isaaclab_arena.tasks.g1_locomanip_pick_and_place_task import G1LocomanipPickAndPlaceTask
-               from isaaclab_arena.utils.pose import Pose
+               from isaaclab_arena.utils.pose import Pose, PoseRange
 
                background = self.asset_registry.get_asset_by_name("galileo_locomanip")()
                pick_up_object = self.asset_registry.get_asset_by_name(args_cli.object)()
@@ -35,10 +35,13 @@ Environment Description
 
                teleop_device = self.device_registry.get_device_by_name(args_cli.teleop_device)()
 
+               XY_RANGE_M = 0.025
                pick_up_object.set_initial_pose(
-                   Pose(
-                       position_xyz=(0.5785, 0.18, 0.0707),
-                       rotation_wxyz=(0.0, 0.0, 1.0, 0.0),
+                   PoseRange(
+                       position_xyz_min=(0.5785 - XY_RANGE_M, 0.18 - XY_RANGE_M, 0.0707),
+                       position_xyz_max=(0.5785 + XY_RANGE_M, 0.18 + XY_RANGE_M, 0.0707),
+                       rpy_min=(0.0, 0.0, 0.0),
+                       rpy_max=(0.0, 0.0, 0.0),
                    )
                )
                blue_sorting_bin.set_initial_pose(
@@ -86,10 +89,13 @@ See :doc:`../../concepts/concept_assets_design` for details on asset architectur
 
 .. code-block:: python
 
+   XY_RANGE_M = 0.025
    pick_up_object.set_initial_pose(
-       Pose(
-           position_xyz=(0.5785, 0.18, 0.0707),
-           rotation_wxyz=(0.0, 0.0, 1.0, 0.0),
+       PoseRange(
+           position_xyz_min=(0.5785 - XY_RANGE_M, 0.18 - XY_RANGE_M, 0.0707),
+           position_xyz_max=(0.5785 + XY_RANGE_M, 0.18 + XY_RANGE_M, 0.0707),
+           rpy_min=(0.0, 0.0, 0.0),
+           rpy_max=(0.0, 0.0, 0.0),
        )
    )
    blue_sorting_bin.set_initial_pose(
--- a/isaaclab_arena/tasks/g1_locomanip_pick_and_place_task.py
+++ b/isaaclab_arena/tasks/g1_locomanip_pick_and_place_task.py
@@ -4,16 +4,13 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import numpy as np
-import torch
 from dataclasses import MISSING
 
 import isaaclab.envs.mdp as mdp_isaac_lab
 from isaaclab.envs.common import ViewerCfg
 from isaaclab.envs.mimic_env_cfg import MimicEnvCfg, SubTaskConfig
-from isaaclab.managers import EventTermCfg, SceneEntityCfg, TerminationTermCfg
+from isaaclab.managers import SceneEntityCfg, TerminationTermCfg
 from isaaclab.utils import configclass
-from isaaclab.utils.math import euler_xyz_from_quat
-from isaaclab_tasks.manager_based.manipulation.stack.mdp import franka_stack_events
 
 from isaaclab_arena.assets.asset import Asset
 from isaaclab_arena.embodiments.common.arm_mode import ArmMode
@@ -43,6 +40,7 @@ class G1LocomanipPickAndPlaceTask(TaskBase):
             if task_description is None
             else task_description
         )
+        self.events_cfg = None
 
     def get_scene_cfg(self):
         pass
@@ -71,7 +69,7 @@ class G1LocomanipPickAndPlaceTask(TaskBase):
         )
 
     def get_events_cfg(self):
-        return EventsCfg(pick_up_object=self.pick_up_object)
+        return self.events_cfg
 
     def get_mimic_env_cfg(self, arm_mode: ArmMode):
         return G1LocomanipPickPlaceMimicEnvCfg()
@@ -95,39 +93,6 @@ class TerminationsCfg:
     success: TerminationTermCfg = MISSING
 
     object_dropped: TerminationTermCfg = MISSING
-
-
-@configclass
-class EventsCfg:
-    """Configuration for Pick and Place."""
-
-    reset_pick_up_object_pose: EventTermCfg = MISSING
-
-    def __init__(self, pick_up_object: Asset):
-        initial_pose = pick_up_object.get_initial_pose()
-        if initial_pose is not None:
-            roll, pitch, yaw = euler_xyz_from_quat(torch.tensor(initial_pose.rotation_wxyz).reshape(1, 4))
-            self.reset_pick_up_object_pose = EventTermCfg(
-                func=franka_stack_events.randomize_object_pose,
-                mode="reset",
-                params={
-                    "pose_range": {
-                        "x": (initial_pose.position_xyz[0] - 0.025, initial_pose.position_xyz[0] + 0.025),
-                        "y": (initial_pose.position_xyz[1] - 0.025, initial_pose.position_xyz[1] + 0.025),
-                        "z": (initial_pose.position_xyz[2], initial_pose.position_xyz[2]),
-                        "roll": (roll, roll),
-                        "pitch": (pitch, pitch),
-                        "yaw": (yaw, yaw),
-                    },
-                    "asset_cfgs": [SceneEntityCfg(pick_up_object.name)],
-                },
-            )
-        else:
-            print(
-                f"Pick up object {pick_up_object.name} has no initial pose. Not setting reset pick up object pose"
-                " event."
-            )
-            self.reset_pick_up_object_pose = None
 
 
 @configclass
--- a/isaaclab_arena/tasks/lift_object_task.py
+++ b/isaaclab_arena/tasks/lift_object_task.py
@@ -8,7 +8,7 @@ from dataclasses import MISSING
 
 import isaaclab.envs.mdp as mdp_isaac_lab
 from isaaclab.envs.common import ViewerCfg
-from isaaclab.managers import CommandTermCfg, EventTermCfg
+from isaaclab.managers import CommandTermCfg
 from isaaclab.managers import ObservationGroupCfg as ObsGroup
 from isaaclab.managers import ObservationTermCfg as ObsTerm
 from isaaclab.managers import RewardTermCfg, SceneEntityCfg, TerminationTermCfg
@@ -38,7 +38,7 @@ class LiftObjectTask(TaskBase):
         self.background_scene = background_scene
         self.minimum_height_to_lift = minimum_height_to_lift
         self.scene_config = None
-        self.events_cfg = LiftObjectEventsCfg(lift_object=self.lift_object, reset_pose_range=reset_pose_range.to_dict())
+        self.events_cfg = None
         self.termination_cfg = self.make_termination_cfg()
 
     def get_scene_cfg(self):
@@ -74,28 +74,6 @@ class LiftObjectTask(TaskBase):
 
 
 @configclass
-class LiftObjectEventsCfg:
-    """Configuration for Lift Object."""
-
-    reset_lift_object_pose: EventTermCfg = MISSING
-
-    def __init__(self, lift_object: Asset, reset_pose_range: dict[str, tuple[float, float]]):
-        self.reset_lift_object_pose = EventTermCfg(
-            func=mdp_isaac_lab.reset_root_state_uniform,
-            mode="reset",
-            params={
-                "pose_range": {
-                    "x": reset_pose_range["x"],
-                    "y": reset_pose_range["y"],
-                    "z": reset_pose_range["z"],
-                },
-                "velocity_range": {},
-                "asset_cfg": SceneEntityCfg(lift_object.name),
-            },
-        )
-
-
-@configclass
 class LiftObjectTerminationsCfg:
     """Termination terms for the Lift Object task."""
 
@@ -111,14 +89,12 @@ class LiftObjectTaskRL(LiftObjectTask):
         embodiment: EmbodimentBase,
         minimum_height_to_lift: float = 0.04,
         episode_length_s: float = 5.0,
-        reset_pose_range: PoseRange = PoseRange(),
     ):
         super().__init__(
             lift_object=lift_object,
             background_scene=background_scene,
             minimum_height_to_lift=minimum_height_to_lift,
             episode_length_s=episode_length_s,
-            reset_pose_range=reset_pose_range,
         )
         self.embodiment = embodiment
         self.observation_cfg = LiftObjectObservationsCfg(
@@ -180,6 +156,8 @@ class LiftObjectCommandsCfg:
 
     def __init__(self, asset_name: str, body_name: str, lift_object: Asset):
         initial_pose = lift_object.get_initial_pose()
+        if isinstance(initial_pose, PoseRange):
+            initial_pose = initial_pose.get_midpoint()
         self.object_pose = mdp_isaac_lab.UniformPoseCommandCfg(
             asset_name=asset_name,
             body_name=body_name,
--- a/isaaclab_arena/tasks/pick_and_place_task.py
+++ b/isaaclab_arena/tasks/pick_and_place_task.py
@@ -9,7 +9,7 @@ from dataclasses import MISSING
 import isaaclab.envs.mdp as mdp_isaac_lab
 from isaaclab.envs.common import ViewerCfg
 from isaaclab.envs.mimic_env_cfg import MimicEnvCfg, SubTaskConfig
-from isaaclab.managers import EventTermCfg, SceneEntityCfg, TerminationTermCfg
+from isaaclab.managers import SceneEntityCfg, TerminationTermCfg
 from isaaclab.sensors.contact_sensor.contact_sensor_cfg import ContactSensorCfg
 from isaaclab.utils import configclass
 
@@ -21,7 +21,6 @@ from isaaclab_arena.metrics.success_rate import SuccessRateMetric
 from isaaclab_arena.tasks.task_base import TaskBase
 from isaaclab_arena.tasks.terminations import object_on_destination
 from isaaclab_arena.utils.cameras import get_viewer_cfg_look_at_object
-from isaaclab_arena.utils.pose import PoseRange
 
 
 class PickAndPlaceTask(TaskBase):
@@ -34,7 +33,6 @@ class PickAndPlaceTask(TaskBase):
         destination_object: Asset | None = None,
         episode_length_s: float | None = None,
         task_description: str | None = None,
-        reset_pose_range: PoseRange = PoseRange(),
     ):
         super().__init__(episode_length_s=episode_length_s)
         self.pick_up_object = pick_up_object
@@ -46,9 +44,7 @@ class PickAndPlaceTask(TaskBase):
                 contact_against_prim_paths=[self.destination_location.get_prim_path()],
             ),
         )
-        self.events_cfg = PickPlaceEventsCfg(
-            pick_up_object=self.pick_up_object, reset_pose_range=reset_pose_range.to_dict()
-        )
+        self.events_cfg = None
         self.termination_cfg = self.make_termination_cfg()
         self.task_description = (
             f"Pick up the {pick_up_object.name}, and place it into the {destination_location.name}"
@@ -120,28 +116,6 @@ class TerminationsCfg:
     success: TerminationTermCfg = MISSING
 
     object_dropped: TerminationTermCfg = MISSING
-
-
-@configclass
-class PickPlaceEventsCfg:
-    """Configuration for Pick Up Object."""
-
-    reset_pick_up_object_pose: EventTermCfg = MISSING
-
-    def __init__(self, pick_up_object: Asset, reset_pose_range: dict[str, tuple[float, float]]):
-        self.reset_pick_up_object_pose = EventTermCfg(
-            func=mdp_isaac_lab.reset_root_state_uniform,
-            mode="reset",
-            params={
-                "pose_range": {
-                    "x": reset_pose_range["x"],
-                    "y": reset_pose_range["y"],
-                    "z": reset_pose_range["z"],
-                },
-                "velocity_range": {},
-                "asset_cfg": SceneEntityCfg(pick_up_object.name),
-            },
-        )
 
 
 @configclass
--- a/isaaclab_arena/tests/test_object_pose_randomization.py
+++ b/isaaclab_arena/tests/test_object_pose_randomization.py
@@ -1,0 +1,113 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+import tqdm
+
+from isaaclab_arena.tests.utils.subprocess import run_simulation_app_function
+
+NUM_RESETS = 10
+NUM_STEPS_PER_RESET = 10
+HEADLESS = True
+
+
+def _test_object_pose_randomization(simulation_app):
+    """Test that object poses are randomized within the specified range."""
+
+    from isaaclab_arena.assets.asset_registry import AssetRegistry
+    from isaaclab_arena.cli.isaaclab_arena_cli import get_isaaclab_arena_cli_parser
+    from isaaclab_arena.environments.arena_env_builder import ArenaEnvBuilder
+    from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
+    from isaaclab_arena.scene.scene import Scene
+    from isaaclab_arena.tasks.dummy_task import DummyTask
+    from isaaclab_arena.utils.pose import PoseRange
+
+    asset_registry = AssetRegistry()
+
+    background = asset_registry.get_asset_by_name("kitchen")()
+    embodiment = asset_registry.get_asset_by_name("franka")()
+    cracker_box = asset_registry.get_asset_by_name("cracker_box")()
+
+    pose_range = PoseRange(
+        position_xyz_min=(0.4 - 0.08, 0.0 - 0.08, 0.1),
+        position_xyz_max=(0.4 + 0.08, 0.0 + 0.08, 0.1),
+        rpy_min=(0.0, 0.0, 0.0),
+        rpy_max=(0.0, 0.0, 0.0),
+    )
+    cracker_box.set_initial_pose(pose_range)
+
+    scene = Scene(assets=[background, cracker_box])
+    isaaclab_arena_environment = IsaacLabArenaEnvironment(
+        name="reference_object_test",
+        embodiment=embodiment,
+        scene=scene,
+        task=DummyTask(),
+        teleop_device=None,
+    )
+
+    args_cli = get_isaaclab_arena_cli_parser().parse_args([])
+    env_builder = ArenaEnvBuilder(isaaclab_arena_environment, args_cli)
+    env = env_builder.make_registered()
+    env.reset()
+
+    try:
+        pose_per_reset = []
+        for _ in tqdm.tqdm(range(NUM_RESETS)):
+            with torch.inference_mode():
+                env.reset()
+
+            # Run some zero actions.
+            for _ in range(NUM_STEPS_PER_RESET):
+                with torch.inference_mode():
+                    actions = torch.zeros(env.action_space.shape, device=env.unwrapped.device)
+                    env.step(actions)
+
+            pose = cracker_box.get_object_pose(env)
+            print(f"pose: {pose}")
+            pose_per_reset.append(pose)
+
+        print(f"pose_per_reset: {pose_per_reset}")
+
+        # Verify that poses vary within the expected range
+        min_x = min([pose[:, 0].item() for pose in pose_per_reset])
+        max_x = max([pose[:, 0].item() for pose in pose_per_reset])
+        min_y = min([pose[:, 1].item() for pose in pose_per_reset])
+        max_y = max([pose[:, 1].item() for pose in pose_per_reset])
+
+        print(f"min_x: {min_x}, max_x: {max_x}")
+        print(f"min_y: {min_y}, max_y: {max_y}")
+
+        # Verify that there is variation in the poses
+        assert min_x < max_x, f"No variation in x position: min_x={min_x}, max_x={max_x}"
+        assert min_y < max_y, f"No variation in y position: min_y={min_y}, max_y={max_y}"
+
+        # Verify that poses are within the specified range
+        assert min_x >= pose_range.position_xyz_min[0], f"min_x {min_x} < expected {pose_range.position_xyz_min[0]}"
+        assert max_x <= pose_range.position_xyz_max[0], f"max_x {max_x} > expected {pose_range.position_xyz_max[0]}"
+        assert min_y >= pose_range.position_xyz_min[1], f"min_y {min_y} < expected {pose_range.position_xyz_min[1]}"
+        assert max_y <= pose_range.position_xyz_max[1], f"max_y {max_y} > expected {pose_range.position_xyz_max[1]}"
+
+        print("All assertions passed!")
+
+    except Exception as e:
+        print(f"Error: {e}")
+        return False
+
+    finally:
+        env.close()
+
+    return True
+
+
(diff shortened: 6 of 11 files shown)


isaaclab_arena/assets/dummy_object.py
```diff
@@ -0,0 +1,50 @@
...
+
+class DummyObject:
+    """
+    Encapsulates the pick-up object config for a pick-and-place environment.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        bounding_box: BoundingBox,
```
cvolkcvolk: bounding_box in constructor only for the DummyObject. The IsaacArena Object would compute it.

isaaclab_arena/assets/dummy_object.py
```diff
@@ -0,0 +1,50 @@
...
+import torch
+
+from isaaclab_arena.utils.bounding_box import BoundingBox
+from isaaclab_arena.utils.pose import Pose
+from isaaclab_arena.utils.relations import Relation
+
+
+class DummyObject:
+    """
+    Encapsulates the pick-up object config for a pick-and-place environment.
```
xyao-nv: forgot to update the comment?

isaaclab_arena/utils/bounding_box.py
```diff
@@ -0,0 +1,185 @@
...
+#        raise ValueError(f"Failed to open USD file: {usd_path}")
+#
+#    # Get the default prim (or pseudo root if no default prim)
+#    default_prim = stage.GetDefaultPrim()
+#    if not default_prim:
+#        default_prim = stage.GetPseudoRoot()
+#
+#    # Compute the bounding box using USD's built-in functionality
+#    # This computes the bounding box in the local space of the prim
+#    bbox_cache = UsdGeom.BBoxCache(Usd.TimeCode.Default(), includedPurposes=[UsdGeom.Tokens.default_])
```
xyao-nv: Do you know if this bbox is calculated using visibility/rendering mesh, or exact source geometry collision mesh, or the simplified phyX mesh?

One scenario I hope it wont run into is, loss get it right for rendering, but physics thinks it's wrong -- collision mesh used by phyX are not being used in the loss func.

isaaclab_arena/utils/bounding_box.py
```diff
@@ -0,0 +1,185 @@
...
+
+    # Create pose with random position and identity rotation (w=1, x=0, y=0, z=0)
+    pose = Pose(position_xyz=tuple(random_position.tolist()), rotation_wxyz=(1.0, 0.0, 0.0, 0.0))
+
+    return pose
+
+
+# Note(cvolk): Commented out as this imports IsaacSim dependencies that require Simulation app to be running.
+
+# def compute_bounding_box_from_usd(
```
xyao-nv: wanna move this to usd_helpers.py? so pxr import is safe

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -0,0 +1,114 @@
...
+        # 3. Distance loss: child should be at target distance from parent
+        assert distance >= 0.0, f"NextTo distance must be non-negative, got {distance}"
+
+        lower_bound = parent_pos[0] + parent_bbox.size[0] / 2 + distance
+        upper_bound = lower_bound + child_bbox.size[0]
+        # Lower band takes into account the objects boundaries. For now we don't additionally add a collision checker/loss.
+        # Maybe we need in the future but its cheaper to do it implicitly by the optimizer.
+        next_to_distance_loss = linear_band_loss(child_pos[0], lower_bound, upper_bound, slope=self.slope)
+
+        total_loss = right_side_loss + top_bottom_band_loss + next_to_distance_loss
```
xyao-nv, line 142: Does it worth to have linearly weighted loss term here? Say users may want to cover every possible position on parent's right side, meaning hitting all feasible position along the frontier curve. Yeah I simplify the problem into a convex one and may overcomplicate it into a hard problem to solve. WDYT

isaaclab_arena/relations/relation_solver.py
```diff
@@ -0,0 +1,410 @@
...
+import numpy as np
+import torch
+from matplotlib.patches import Rectangle
+
+from isaaclab_arena.assets.dummy_object import DummyObject
+from isaaclab_arena.utils.loss_strategies import LossStrategy, NextToLossStrategy
+from isaaclab_arena.utils.relations import NextTo, Relation
+
+
+class RelationSolver:
```
xyao-nv, line 20: I am not sure if example folder is the right one to park this important solver. How about put it under a relation folder as I foresee this solver will get more and more code snippets.

isaaclab_arena/examples/relation_solver.py
```diff
@@ -0,0 +1,410 @@
...
+                - Object names mapped to final (x, y, z) positions
+                - '_loss_history': List of loss values during optimization
+        """
+        # Initialize positions from objects. All objects must have an initial pose at this stage.
+        # TODO(cvolk) Should this be the Solver's responsibility to randomly initialize objects positions or should the caller do it?
+        all_positions = self._get_positions_from_objects(objects)
+
+        # Identify fixed and optimizable objects based on anchor_objects
+        anchor_set = set(self.anchor_objects)
+        fixed_mask = torch.tensor([obj in anchor_set for obj in objects])
```
xyao-nv: maybe you want to move them to gpus/specified device explicitly, when more objs or future parallel envs are added

isaaclab_arena/examples/relation_solver.py
```diff
@@ -0,0 +1,410 @@
...
+            convergence_threshold: Stop when loss falls below this value.
+            verbose: Print optimization progress.
+            anchor_objects: List of fixed reference objects that won't be optimized.
+            strategies: Optional custom strategies to override defaults.
+        """
+        self.max_iters = max_iters
+        self.lr = lr
+        self.convergence_threshold = convergence_threshold
+        self.verbose = verbose
+        self.anchor_objects = anchor_objects or []
```
xyao-nv: given you have assert >0 following, is init to [] necessary?

isaaclab_arena/assets/object.py
```diff
@@ -43,31 +44,40 @@ def __init__(
...
+        assert self.usd_path is not None
+        if self.bounding_box is None:
+            self.bounding_box = compute_bounding_box_from_usd(self.usd_path, self.scale)
+        return self.bounding_box
+    
+    def get_corners_aabb_axis_aligned(self, pos: torch.Tensor) -> torch.Tensor:
+        assert self.usd_path is not None
+        if self.bounding_box is None:
+            self.bounding_box = compute_bounding_box_from_usd(self.usd_path, self.scale)
+        return self.bounding_box.get_corners_aabb_axis_aligned(pos)
```
alexmillane: Suggestion to instead give the bounding box type a method for getting its corners.

Also, something is a bit weird about `get_corners_aabb_axis_aligned`. The box is axis-aligned. So expanding the `aabb` acronym we have `get_corners_axis_aligned_bounding_box_axis_aligned`. So there's some redundancy there.

If this is just a naming thing, no worries, that's easily fixed, but I'm wondering if we're mixing oriented bounding boxes and axis-aligned bounding boxes. I.e. we're saying axis aligned when we mean oriented. I know I was guilty of that when we were working on the whiteboard.

What's your opinion here?

isaaclab_arena/examples/relation_solver.py
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    def __init__(
+        self,
+        max_iters: int = 1000,
+        lr: float = 0.01,
+        convergence_threshold: float = 1e-4,
+        verbose: bool = True,
+        anchor_objects: list[DummyObject] | None = None,
+        strategies: dict[type[Relation], LossStrategy] | None = None,
```
alexmillane: Nit: I feel like we may end up with a few configuration parameters for this guy, so at some point we may move to:

```
@dataclass
class RalationSolverParams
```

isaaclab_arena/examples/relation_solver.py
```diff
@@ -0,0 +1,410 @@
...
+            verbose: Print optimization progress.
+            anchor_objects: List of fixed reference objects that won't be optimized.
+            strategies: Optional custom strategies to override defaults.
+        """
+        self.max_iters = max_iters
+        self.lr = lr
+        self.convergence_threshold = convergence_threshold
+        self.verbose = verbose
+        self.anchor_objects = anchor_objects or []
+        assert len(self.anchor_objects) > 0, "At least one anchor object is required"
```
alexmillane: Do you forsee a use of multiple anchor objects?

isaaclab_arena/examples/relation_solver.py
```diff
@@ -0,0 +1,410 @@
...
+                if parent_pos is None:
+                    raise ValueError(f"Parent {relation.parent.name} not found in objects list")
+
+                # Get strategy for this relation type
+                strategy = self._get_strategy(relation)
+
+                # Compute loss using the strategy
+                loss = strategy.compute_loss(
+                    relation=relation,
+                    child_pos=positions[i],
```
alexmillane: nit: This works. I had to double take on the why `child_pos=positions[i]` though.

Suggestion

```
child_pos = obj_to_pos.get(obj)
parent_pos = obj_to_pos.get(relation.parent)

...

loss = strategy.compute_loss(
    relation=relation,
    child_pos=child_pos,
    parent_pos=parent_pos,

...

```

isaaclab_arena/relations/relation_solver.py
```diff
@@ -0,0 +1,410 @@
...
+            objects: List of DummyObject instances
+
+        Returns:
+            Dictionary with:
+                - Object names mapped to final (x, y, z) positions
+                - '_loss_history': List of loss values during optimization
+        """
+        # Initialize positions from objects. All objects must have an initial pose at this stage.
+        # TODO(cvolk) Should this be the Solver's responsibility to randomly initialize objects positions or should the caller do it?
+        all_positions = self._get_positions_from_objects(objects)
```
alexmillane: Good question. Probably for now let's just try what you have

In the long run, I feel we may need to re-optimize a few times with different initializations. This would mean the optimizer would have to handle it. (Maybe the optimizer could take care of initialization in cases where the object has no initial position? We could basically use initial position to mean pre-optimization position, in the case where a relation affects the object.)

For initialization, I'm thinking that each object initializes to some random location within the oriented bounded box of the object that it's "on".

isaaclab_arena/examples/relation_solver.py
```diff
@@ -0,0 +1,410 @@
...
+        anchor_objects: list[DummyObject] | None = None,
+        strategies: dict[type[Relation], LossStrategy] | None = None,
+    ):
+        """
+        Args:
+            max_iters: Maximum optimization iterations.
+            lr: Learning rate for Adam optimizer.
+            convergence_threshold: Stop when loss falls below this value.
+            verbose: Print optimization progress.
+            anchor_objects: List of fixed reference objects that won't be optimized.
```
alexmillane: Is there a reason to take `anchor_objects` in the constructor and `objects` in the solve method. To me, the motivation to take the `objects` in the solve method is clear - the solver could be used for solving multiple different problems. However following that reasoning through, the anchor objects (which are also problem specific) should also be taken in the solve method rather than the constructor.

isaaclab_arena/relations/relation_solver.py
```diff
@@ -0,0 +1,410 @@
...
+            positions: Tensor of shape (N, 3) with current positions
+            objects: List of DummyObject instances
+
+        Returns:
+            Total loss tensor
+        """
+        total_loss = torch.tensor(0.0)
+
+        # Create position mapping for parent lookup
+        obj_to_pos = {obj: positions[i] for i, obj in enumerate(objects)}
```
alexmillane: I love the design of this class. Really clean. Really understandable.

I have one small suggestion. There are a few things that could be better encapsulated.

One example might be this mapping between objects and their row-position in the position matrix... Actually... is there any reason to form the full matrix of positions (as opposed to having the two matrices for optimized and un-optimized positions)? I feel like you're composing the full matrix just to pass it to this function, which then unpacks the matrices into their rows during per-relation loss computation.

I would suggest encapsulating the state of the optimization in an object that wraps all of this into an object:

```
class RelationSolverState:

  def __init__(objects, archor_objects):
    self.object_position_tensor = ...
    self.anchor_object_position_tensor = ...
    
  def get_position(object) -> torch.Tensor:
  ...

```

Then you can just pass this object around and forget about forming the full position-state matrix.

isaaclab_arena/utils/bounding_box.py
```diff
@@ -0,0 +1,185 @@
...
+            torch.tensor([x - width / 2, y - depth / 2, z - height / 2]),  # Bottom-front-left
+            torch.tensor([x + width / 2, y - depth / 2, z - height / 2]),  # Bottom-front-right
+            torch.tensor([x + width / 2, y + depth / 2, z - height / 2]),  # Bottom-back-right
+            torch.tensor([x - width / 2, y + depth / 2, z - height / 2]),  # Bottom-back-left
+            # Top face (z_max)
+            torch.tensor([x - width / 2, y - depth / 2, z + height / 2]),  # Top-front-left
+            torch.tensor([x + width / 2, y - depth / 2, z + height / 2]),  # Top-front-right
+            torch.tensor([x + width / 2, y + depth / 2, z + height / 2]),  # Top-back-right
+            torch.tensor([x - width / 2, y + depth / 2, z + height / 2]),  # Top-back-left
+        ])
```
alexmillane: Could formulate this as a call to a method that gets the corners in the object frame and then adds the center. Perhaps a nested call to the method itself.

```
def get_corners_aabb_axis_aligned(self, pos: torch.Tensor | None) -> torch.Tensor:
     if pos is None:
         corners = ...
         return corners
     else:
         corners_in_object_frame = self.get_corners_aabb_axis_aligned(pos=None)
         return corners + pos
```

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        Args:
+            slope: Gradient magnitude for linear loss (default: 10.0).
+                   Loss increases by `slope` per meter of violation.
+        """
+        self.slope = slope
+
+    def compute_loss(
+        self,
+        relation: "NextTo",
```
alexmillane, line 71: Potentially this is meant to be:

```
type["NextTo"]
```

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        side = relation.side
+        distance = relation.distance_m
+
+        if side != "right":
+            # TODO(cvolk): Implement support for other sides and make generic.
+            raise NotImplementedError(f"Side '{side}' not yet implemented, only 'right' is supported")
+
+        # 1. Half-plane loss: child must be to the right of parent's right edge
+        parent_right_bound = parent_pos[0] + parent_bbox.size[0] / 2
```
alexmillane: So "right" here is in the world frame. In other words, +x in the world frame.

We talked about this in person and I'm not sure where we got to. To me, that seems a bit unintuitive. Right, to me, means right in the frame of the parent object.

With that said, because we're using **Axis Aligned** bounding boxes, maybe "right in the object frame" is actually poorly defined - objects don't have orientations.

I think we should move to oriented bounding boxes eventually. But let's get everything going with axis align bounding boxes first, in which case this is fine.

isaaclab_arena/utils/loss_strategies.py
```diff
@@ -0,0 +1,114 @@
...
+        top_bottom_band_loss = linear_band_loss(child_pos[1], parent_top_bound, parent_bottom_bound, slope=self.slope)
+
+        # 3. Distance loss: child should be at target distance from parent
+        assert distance >= 0.0, f"NextTo distance must be non-negative, got {distance}"
+
+        lower_bound = parent_pos[0] + parent_bbox.size[0] / 2 + distance
+        upper_bound = lower_bound + child_bbox.size[0]
+        # Lower band takes into account the objects boundaries. For now we don't additionally add a collision checker/loss.
+        # Maybe we need in the future but its cheaper to do it implicitly by the optimizer.
+        next_to_distance_loss = linear_band_loss(child_pos[0], lower_bound, upper_bound, slope=self.slope)
```
alexmillane: Not sure how this is a band. Niavely I would have thought that the distance would specify a single distance rather than a range. Either a constraint on the separator distance, or the distance between centers. what's your thinking here?

isaaclab_arena/utils/relations.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Note: Loss computation is handled by NextToLossStrategy in loss_strategies.py.
+    """
+
+    def __init__(
+        self,
+        parent: "DummyObject",
+        relation_loss_weight: float = 1.0,
+        distance_m: float = 0.05,
+        side: str = "right",
```
alexmillane: `Enum`

isaaclab_arena/utils/relations.py
```diff
@@ -0,0 +1,54 @@
...
+            parent: The parent asset that this object should be placed next to.
+            relation_loss_weight: Weight for the relationship loss function.
+            distance_m: Target distance from parent's boundary in meters (default: 5cm).
+            side: Which side to place object: "front" (-Y), "back" (+Y),
+                  "left" (-X), or "right" (+X).
+        """
+        super().__init__(parent, relation_loss_weight)
+        assert distance_m >= 0.0, f"Distance must be non-negative, got {distance_m}"
+        self.distance_m = distance_m
+        assert side in ["front", "back", "left", "right"]
```
alexmillane: Enum should obviate this.

isaaclab_arena/utils/bounding_box.py
cvolkcvolk: Mh good point Im not sure. I will look into it.

isaaclab_arena/utils/bounding_box.py
cvolkcvolk: Good idea. Will do.

isaaclab_arena/relations/relation_loss_strategies.py
cvolkcvolk: We could extend this to be able to set scaling factors for each respective loss term. For example distance = None could get rid of that constraint altogether and enable the use case you described

isaaclab_arena/relations/relation_solver.py
cvolkcvolk: Will move it out of the example folder. I initially put my first draft scripts there, and this is just an artifact from earlier on

isaaclab_arena/assets/object.py
cvolkcvolk: Agree this is some unnecessary redundancy on the BoundingBox side. This BoundingBox class only handles axis aligned ones, so its just a naming thing. Maybe we just rename the whole class to `class AxisAlignedBoundingBox` to make it clear. 
On the `object` side which exposes it, I'd stick with `object. get_corners_aabb**_axis_aligned()**` to make it obvioius that its not an oriented bounding box when working with objects.

isaaclab_arena/examples/relation_solver.py
cvolkcvolk: Agree, Good idea

isaaclab_arena/relations/relation_solver.py
cvolkcvolk: Ill leave it out of the Solver for now.
Later we could wrap the Solver again into a class like `ObjectPlacer` that holds 
1) A position intializer 
2) The RelationSolver 
3) A feasability checker 
such that it could respin the solver multiple times without exposing it to the user.

isaaclab_arena/examples/relation_solver.py
cvolkcvolk: You are right. That makes actually more sense. 
Instead of 
```
relation_solver = RelationSolver(anchor_objects = ...)
relation_solver.solve(objects)
```
it would indeed be cleaner to just go 
```
relation_solver = RelationSolver()
relation_solver.solve(objects, anchor_objects = ...)
```

isaaclab_arena/relations/relation_solver.py
cvolkcvolk: Aboslute genius. That would make it super clean. Im looking into it

isaaclab_arena/examples/relation_solver.py
cvolkcvolk: Agree

isaaclab_arena/utils/loss_strategies.py
cvolkcvolk: We can change it to use a single distance minimum instead of a band. I think I had some randomization in mind that a band gives... However the randomization should be handled more explicitely in anyway.

isaaclab_arena/examples/relation_solver.py
cvolkcvolk: I thought about fixing multiple objects on the table for certain tasks. But agree, its rather an avanced feature. I downgraded to only accept a single object for now.

isaaclab_arena/relations/relation_loss_strategies.py
cvolkcvolk: Apparently `"NextTo"` is correct here. `type["NextTo"]` would mean "the class itself" rather than an instance.
We're accessing relation.side, relation.distance_m, etc. which are instance attributes

isaaclab_arena/utils/loss_strategies.py
cvolkcvolk: Introduced a `single_point_linear_loss()` to place the object at the exact distance. Makes more sense

isaaclab_arena/utils/bounding_box.py
cvolkcvolk: Yep, that should work. Updated

isaaclab_arena/utils/isaac_sim_debug_draw.py
```diff
@@ -0,0 +1,141 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
cvolkcvolk, line 1: Tested and works to draw bb into IsaacSim, Can take it out of this MR as its currently not used anymore

isaaclab_arena/embodiments/null_embodiment_new.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
cvolkcvolk: I will move it out to another MR

isaaclab_arena/assets/dummy_object.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
cvolkcvolk, line 1: I could move this out into another MR

isaaclab_arena_examples/relations/relation_solver_visualization_notebook.py
```diff
@@ -0,0 +1,276 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
cvolkcvolk, line 1: Note: This file contains lots of AI visualization code. Not a priority to go through this. Can Take it out into another MR.

isaaclab_arena_examples/relations/relation_solver_visualizer.py
```diff
@@ -0,0 +1,536 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
cvolkcvolk, line 1: Note: Same here. This file contains lots of AI visualization code. Not a priority to go through this. Can Take it out into another MR.

isaaclab_arena/assets/object.py
```diff
@@ -46,8 +49,46 @@ def __init__(
         self.reset_pose = True
         self.spawn_cfg_addon = spawn_cfg_addon
         self.asset_cfg_addon = asset_cfg_addon
+        self.bounding_box = None
         self.object_cfg = self._init_object_cfg()
         self.event_cfg = self._init_event_cfg()
+        self.relations = []
```
alexmillane, line 56: should this be `self.relations = relations` (or delete the constructor argument).

isaaclab_arena_examples/relations/dummy_object_placer_notebook.py
alexmillane, line 1: I'm think we should move these examples out of the core framework. Maybe we should create a new package `isaaclab_arena_examples` or something. This is similar to the `isaaclab` which separates the core framework from any environments and anything runnable.

isaaclab_arena_examples/relations/dummy_object_placer_notebook.py
```diff
@@ -0,0 +1,72 @@
...
+# ^^^ Suppress type errors for DummyObject → Object (duck typing works at runtime)
+
+# Example notebook demonstrating the ObjectPlacer class not using any IsaacSim dependencies.
+
+# %%
+# Install plotly dependency if missing
+import subprocess
+import sys
+
+subprocess.check_call([sys.executable, "-m", "pip", "install", "-q", "tenacity"])
```
alexmillane: I can't see this being used?

isaaclab_arena_examples/relations/isaac_sim_object_placer_notebook.py
```diff
@@ -0,0 +1,84 @@
...
+cracker_box.add_relation(On(office_table, clearance_m=0.02))
+cracker_box.add_relation(NextTo(coffee_machine, side=Side.RIGHT, distance_m=0.15))
+mug.add_relation(On(coffee_machine, clearance_m=0.02))
+
+# Place objects using ObjectPlacer (handles random init, solving, and applying positions)
+placer = ObjectPlacer()
+result = placer.place(
+    objects=[office_table, coffee_machine, cracker_box, mug],
+    anchor_object=office_table,
+)
```
alexmillane: We discussed this on the phone. I reckon eventually we want all of this called inside the compiler so the user doesn't see it/have to explicitly do it.

With that said, given the size of the code here, we could do it in a follow up MR.

isaaclab_arena_examples/relations/relation_solver_visualization_notebook.py
alexmillane: If we move it out of the core framework, I think we can have some less-tidy scripts.

One thing, though, is that these example scripts tend to break quickly. So it would be good to have a smoke test for them -  basically a test that just runs them with `--dont_visualize` and checks that they don't crash.

isaaclab_arena_examples/relations/relation_solver_visualizer.py
alexmillane: Let's move it out of the core package, but just merge it, if it's useful

isaaclab_arena/relations/loss_primitives.py
```diff
@@ -0,0 +1,128 @@
...
+    Examples:
+        >>> # Penalize if x > 0.5 (should be to the left of boundary)
+        >>> loss = single_boundary_linear_loss(x, 0.5, slope=10.0, penalty_side='greater')
+
+        >>> # Penalize if x < 0.5 (should be to the right of boundary)
+        >>> loss = single_boundary_linear_loss(x, 0.5, slope=10.0, penalty_side='less')
+    """
+    # Ensure value is a tensor
+    if not isinstance(value, torch.Tensor):
+        value = torch.tensor(value, dtype=torch.float32)
```
alexmillane: According to the type hint, `value` must be a tensor.

Is there cases where we want to accept non-tensors? As the docstring points out, this is required for gradient flow. If not lets swap this for an assert.

isaaclab_arena/relations/object_placer.py
```diff
@@ -0,0 +1,202 @@
...
+from typing import TYPE_CHECKING
+
+from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
+from isaaclab_arena.relations.placement_result import PlacementResult
+from isaaclab_arena.relations.relation_solver import RelationSolver
+from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox, get_random_pose_within_bounding_box
+from isaaclab_arena.utils.pose import Pose
+
+# TYPE_CHECKING: Import Object for type hints without runtime Isaac Sim dependency.
+# At runtime, duck typing allows Object to work as well.
```
alexmillane: ? suggestion to remove this comment.

isaaclab_arena/relations/object_placer.py
```diff
@@ -0,0 +1,202 @@
...
+
+        Raises:
+            ValueError: If anchor_object has no initial_pose set.
+        """
+        # Validate anchor has initial pose
+        if anchor_object.initial_pose is None:
+            raise ValueError(
+                f"anchor_object '{anchor_object.name}' must have an initial_pose set. "
+                "Call anchor_object.set_initial_pose(...) before placing."
+            )
```
alexmillane: I feel like we should come to a policy on `assert` versus `if condition; raise`. I.e. we should prefer one by default in the code base (but be fine if we need to use the other).

Personally, I feel that we should default to `assert`, and use `raise` where the error is somehow semi-expected and recoverable.

isaaclab_arena/relations/object_placer.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Save RNG state and set seed if provided (for reproducibility without affecting Isaac Sim)
+        rng_state = None
+        if self.params.placement_seed is not None:
+            rng_state = torch.get_rng_state()
+            torch.manual_seed(self.params.placement_seed)
+
+        # Determine bounds for random position initialization from the anchor object
+        init_bounds = self._get_init_bounds(anchor_object)
```
alexmillane, line 68: Nice approach.

I could imagine that soon (in the future) we may want to do something a bit... smarter/tighter(?) for initialization such that we don't get most solutions piling up on the boundaries.

Perhaps using an oriented bounding box might help.

isaaclab_arena/relations/object_placer.py
```diff
@@ -0,0 +1,202 @@
...
+                print(f"Attempt {attempt + 1}/{self.params.max_placement_attempts}: loss = {loss:.6f}")
+
+            # Track best result
+            if loss < best_loss:
+                best_loss = loss
+                best_positions = positions
+
+            # Check if placement is valid
+            if self._validate_placement(positions):
+                # NOTE(cvolk): Not implemented yet.
```
alexmillane: What's not implemented here? It appears implemented?

isaaclab_arena/relations/object_placer.py
```diff
@@ -0,0 +1,202 @@
...
+                center[0] - half_size[0],
+                center[1] - half_size[1],
+                center[2] - half_size[2],
+            ),
+            max_point=(
+                center[0] + half_size[0],
+                center[1] + half_size[1],
+                center[2] + half_size[2],
+            ),
+        )
```
alexmillane, line 142: This seems a bit awkward to use. The user basically has to know something about the scene to configure the initialization. Seems like a nice(r) solution might be to: initialize all objects randomly within the bounding box of the their parent from their `On` relation. (is it true to say that all objects other than the root have an on relation?). I feel like that might give us solutions which are less likely to pile up on boundaries.

But we can do that in a future MR.

isaaclab_arena/relations/object_placer.py
alexmillane: Ah I see below. `_validate_placement` always returns true at the moment. So I guess the placement always has exactly one attempt?

isaaclab_arena/relations/placement_result.py
```diff
@@ -0,0 +1,31 @@
...
+if TYPE_CHECKING:
+    from isaaclab_arena.assets.object import Object
+
+
+@dataclass
+class PlacementResult:
+    """Result of an ObjectPlacer.place() call."""
+
+    success: bool
+    """Whether placement succeeded (loss < threshold within max_attempts)."""
```
alexmillane: I think that this is incorrect atm. success is true if the placement passes the validation checks (which don't include loss).

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -0,0 +1,242 @@
...
+
+from isaaclab_arena.relations.relations import Side
+
+
+class RelationLossStrategy(ABC):
+    """Abstract base class for relation loss computation strategies.
+
+    Loss strategies compute constraints using world-space extents:
+        world_min = position + bbox.min_point
+        world_max = position + bbox.max_point
```
alexmillane: This seems wrong. I guess the `RelationLossStrategy` defines how an Relation maps to a Loss.

isaaclab_arena/relations/relation_loss_strategies.py
alexmillane: OK! I guess I've been doing that wrong then!

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -0,0 +1,242 @@
...
+
+        Returns:
+            Weighted loss tensor.
+        """
+        # Compute parent world-space extents
+        parent_x_min = parent_pos[0] + parent_bbox.min_point[0]
+        parent_x_max = parent_pos[0] + parent_bbox.max_point[0]
+        parent_y_min = parent_pos[1] + parent_bbox.min_point[1]
+        parent_y_max = parent_pos[1] + parent_bbox.max_point[1]
+        parent_z_max = parent_pos[2] + parent_bbox.max_point[2]  # Top surface
```
alexmillane, line 192: I'm seeing repeated computations of this form. I feel like we should add a method to our aabb to translate it:

```
aabb_W = aabb_O.translate(parent_pos)
```

Where `W` is the world frame and `O` is the object frame.

isaaclab_arena/relations/relation_solver.py
```diff
@@ -0,0 +1,254 @@
...
+import torch
+from typing import TYPE_CHECKING
+
+from isaaclab_arena.relations.relation_loss_strategies import NextToLossStrategy, OnLossStrategy, RelationLossStrategy
+from isaaclab_arena.relations.relation_solver_params import RelationSolverParams
+from isaaclab_arena.relations.relation_solver_state import RelationSolverState
+from isaaclab_arena.relations.relations import NextTo, On, Relation
+
+# TYPE_CHECKING: Import Object for type hints without runtime Isaac Sim dependency.
+# At runtime, duck typing allows DummyObject to work as well.
```
alexmillane: This pattern is all over the codebase. Suggestion to just delete these comments. If we were consistent with adding these comments they'd end up everywhere.

isaaclab_arena/relations/relation_solver.py
```diff
@@ -0,0 +1,254 @@
...
+            params: Solver configuration parameters. If None, uses defaults.
+        """
+        # Use provided params or defaults
+        self.params = params or RelationSolverParams()
+
+        # Merge user strategies with defaults (user overrides take precedence)
+        self._strategies: dict[type[Relation], RelationLossStrategy] = {
+            **self.DEFAULT_STRATEGIES,
+            **self.params.strategies,
+        }
```
alexmillane: Suggestion to move the default strategies into the `RelationSolverParams`. That way they're captured with the other default values. Also a user can introspect the default strategies through the parameter struct.

isaaclab_arena/relations/relation_solver.py
```diff
@@ -0,0 +1,254 @@
...
+        """Solve for optimal positions of all objects.
+
+        Args:
+            objects: List of DummyObject instances (may include anchor_object).
+            anchor_object: Fixed reference object that won't be optimized.
+
+        Returns:
+            Dictionary mapping object instances to final (x, y, z) positions.
+        """
+        self._anchor_object = anchor_object  # Storing for debug_losses()
```
alexmillane: nit: Suggestion just to require that `debug_losses` requires to be passed an anchor object (right now it's optional). I feel like adding between-solves state like this is asking to be bitten down the track, for a very marginal amount on convenience 😅

isaaclab_arena/relations/relation_solver.py
```diff
@@ -0,0 +1,254 @@
...
+        # Optimization loop
+        loss_history = []
+        position_history = []  # Track positions for visualization
+
+        for iter in range(self.params.max_iters):
+            optimizer.zero_grad()
+
+            # Save position snapshot (every 10 iterations to save memory)
+            if iter % 10 == 0:
+                position_history.append(state.get_all_positions_snapshot())
```
alexmillane: Potentially you want to put the saving of debug state behind a flag (and maybe put this constant as a named constant `SAVE_SOLVER_POSITION_HISTORY_EVERY_N_STEPS=10` as a class-level(?) constant?

isaaclab_arena/relations/relation_solver.py
```diff
@@ -0,0 +1,254 @@
...
+
+    @property
+    def last_loss_history(self) -> list[float]:
+        """Loss values from the most recent solve() call."""
+        return getattr(self, "_last_loss_history", [])
+
+    @property
+    def last_position_history(self) -> list:
+        """Position snapshots from the most recent solve() call."""
+        return getattr(self, "_last_position_history", [])
```
alexmillane: Suggestion to just initialize these things in the `__init__` to avoid these `geattrs`.

isaaclab_arena/relations/relation_solver_params.py
```diff
@@ -0,0 +1,29 @@
...
+    """Configuration parameters for RelationSolver."""
+
+    max_iters: int = 1000
+    """Maximum optimization iterations."""
+
+    lr: float = 0.01
+    """Learning rate for Adam optimizer."""
+
+    convergence_threshold: float = 1e-4
+    """Stop when loss falls below this value."""
```
alexmillane, line 31: Right now, all of our losses could estimate their degree of violation in meters, correct?

It would be great if we could specify a convergence condition in meters. Something like (in natural language) "you're converged when all relations within 0.01 meters of being satisfied. That would be much more understandable metric for convergence. It would also be helpful when debugging a failed solve because you could say "the solve failed because NextTo constrain between object X and object Y was violated by 0.05 meters".

I don't forsee (without too much thinking) any constraint whose violation would not be measurable in meters.

What do you think about this idea?

(Of course, I think that this should go in the future work bin, given the current deadlines).

isaaclab_arena/relations/relation_solver_state.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Extract initial positions from all objects
+        positions = []
+        for obj in objects:
+            pose = obj.get_initial_pose()
+            assert pose is not None, f"Pose is None for {obj.name}"
+            positions.append(torch.tensor(pose.position_xyz, dtype=torch.float32))
+
+        # Separate anchor from optimizable positions
+        self._anchor_idx = self._obj_to_idx[anchor_object]
```
alexmillane, line 51: This line is surprising to me. I guess what's happening is that we expect that `objects` contains `anchor_object` right? So the `anchor_object` appears twice in the arguments to this object. Somehow, I didn't expect that. Perhaps we should `assert` that in this constructor.

isaaclab_arena/relations/relation_solver_state.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Build object-to-index mapping
+        self._obj_to_idx: dict[Object, int] = {obj: i for i, obj in enumerate(objects)}
+
+        # Extract initial positions from all objects
+        positions = []
+        for obj in objects:
+            pose = obj.get_initial_pose()
+            assert pose is not None, f"Pose is None for {obj.name}"
+            positions.append(torch.tensor(pose.position_xyz, dtype=torch.float32))
```
alexmillane: This was somewhat unexpected to me. I guess that the expectation is that `ObjectPlacer` calls `set_initial_position()` on each object before it gets to the solver.

I feel like we're abusing `Object.initial_pose` a bit to mean:
- the pose at simulation start up (its intended use), and
- the pose at the start of optimization, which will be overwritten at the end (basically a back channel communication for the solve initialization).

What do you think about moving to explicitly passing the initialization positions to the solve via a `dict[Object, tuple(float, float, float)]`

isaaclab_arena/relations/relation_solver_state.py
alexmillane: having looked back at `ObjectPlacer` the docstring say that `object` "may" contain the `anchor_object`. I would say that at this level it must(?) contain the `achor_object`.

isaaclab_arena/relations/relation_solver_state.py
```diff
@@ -0,0 +1,107 @@
...
+        opt_idx = self._optimizable_indices.index(idx)
+        return self._optimizable_positions[opt_idx]
+
+    def get_all_positions_snapshot(self) -> list[list[float]]:
+        """Get detached copy of all positions for history tracking.
+
+        Returns:
+            List of [x, y, z] positions for each object (in original order).
+        """
+        return [self.get_position(obj).detach().tolist() for obj in self._objects]
```
alexmillane: super nit: elsewhere were using a `tuple` to represent (x,y,z), (such as the function below) here we use a list. Suggestion:

```
return [tuple(self.get_position(obj).detach().tolist()) for obj in self._objects]
```

isaaclab_arena/utils/isaac_sim_debug_draw.py
alexmillane: You could also leave it in if you think it's useful. (although given that it will probably break when we upgrade IsaacSim it may make sense to add a smoke test to it).

alexmillane (approved):
Looks great! Thanks for doing this.

Most of the comments are very small. Some of the suggestions should be moved to future work. This MR is already quite big so I suggest that we get it in and improve things later.

isaaclab_arena/assets/object.py
cvolkcvolk: Good catch

cvolkcvolk: Interestingly the IsaacLab runs a validation function recursively to check all nested objects for missing  fields.  With multiple objects each having relations pointing to the same parents, this thing nukes itself. 

I tooke the relations out of the constructor for now and initializing an emtpy list as we don't use the constructor for relations anyway.

isaaclab_arena/relations/loss_primitives.py
cvolkcvolk: Added an assert. We need a torch tensor and allowing anything else would be a welcome sign for bugs.

isaaclab_arena/relations/object_placer.py
cvolkcvolk: Removed

isaaclab_arena/relations/object_placer.py
cvolkcvolk: I thought the same. And in this case I chose `raise` over `assert` because initial_pose of the anchor needs to be set by the user. The raise is imo a soft/ recoverable state that the user can fix by providing the initial_position and run again. 
Its not a program bug that this should never happen when the code is correct.

But I feel this is always a bit of a personal decision. I would be fine with just asserting.

isaaclab_arena/relations/object_placer.py
cvolkcvolk: Noted. Thats a good idea

isaaclab_arena/relations/object_placer.py
cvolkcvolk: Yes currently it only runs once. Will implement this check soon so added the loop already

isaaclab_arena/relations/object_placer.py
cvolkcvolk: > is it true to say that all objects other than the root have an on relation?)

Not necessarily. An object could have only a NextTo relation (for example Bin NextTo table) while all other items on the table are holding an On relation.

But we could maybe come up with a per-relation initialization

isaaclab_arena/relations/placement_result.py
cvolkcvolk: Good catch. It does only geometric checks (or will, once implemented), not loss threshold checks. I updated the docstring

isaaclab_arena/relations/relation_loss_strategies.py
cvolkcvolk: Agree. Should not mention implementation details here.

isaaclab_arena/relations/relation_loss_strategies.py
cvolkcvolk: Noted to adress in a follow up MR.

isaaclab_arena/relations/relation_solver.py
cvolkcvolk: True. Removed.

isaaclab_arena/relations/relation_solver.py
cvolkcvolk: Done

isaaclab_arena/relations/relation_solver.py
cvolkcvolk: I would say its not a nit anymore. Its a valid concern :sweat_smile: 
Updated.

isaaclab_arena/relations/relation_solver_params.py
cvolkcvolk: That's a great idea. I was already tweaking around with the indeed not very useful loss threshold as it differs from scene to scene.

cvolkcvolk: Noted to adress in a follow up.

isaaclab_arena/relations/relation_solver_state.py
cvolkcvolk: Fully agree on the overuse of object.initial_pose. The data flow is indeed a bit of a mess this way. 
I refactored the classes in a way you suggested that `RelationSolver.solve()` requires initial_positions parameter. `RelationSolverState.__init__()` now takes initial_positions as dict as well and `ObjectPlacer._generate_initial_positions()` returns a dict instead of mutating objects. This should make it hopefully explicit.

isaaclab_arena/relations/relation_solver_state.py
cvolkcvolk: At this stage it must contain the anchor_object. The docstring saying it might contain the anchor_object is not correct. I'm adding an assert to make this assumption explicit

isaaclab_arena/utils/isaac_sim_debug_draw.py
cvolkcvolk: Added a smoke test

isaaclab_arena_examples/relations/dummy_object_placer_notebook.py
cvolkcvolk: Put all of those into a new package `isaaclab_arena_examples/relations`. Do we want to move our other examples there as well (`example_env_notebook.py`, `compile_env_notebook.py` ?)

isaaclab_arena_examples/relations/relation_solver_visualization_notebook.py
cvolkcvolk: Done

isaaclab_arena/examples/relation_solver.py
cvolkcvolk: Marked for future multi env upgrade

PR #332, opened 2026-01-13
Refactor Policy args with a dataclass

xyao-nv:
## Summary
Enables users to specify policy related configs thru cli args parser (policy runner) or a dict (multi-task eval). 

## Detailed description
- Before this change, configs for policy can only be specified using cli args
- Multi-task eval expects user to submit eval jobs thru json dict.
- Instead of json dict -> cli args list -> parser, it allows two paths, using the follow orders
1. Json dict -> PolicyConfigClass -> Policy.from_dict()
2. Json dict -> cli args list -> args parser -> Policy.from_args()
Depending on the availability of  policy's method.

### DIFF
--- a/isaaclab_arena/policy/policy_base.py
+++ b/isaaclab_arena/policy/policy_base.py
@@ -22,10 +22,11 @@ class PolicyBase(ABC):
     # Optional: Subclasses can define this to enable from_dict()
     config_class: type | None = None
 
-    def __init__(self):
+    def __init__(self, config: Any):
         """
         Base class for policies.
         """
+        self.config = config
 
     @classmethod
     def from_dict(cls, config_dict: dict[str, Any]) -> "PolicyBase":
--- a/isaaclab_arena/policy/replay_action_policy.py
+++ b/isaaclab_arena/policy/replay_action_policy.py
@@ -84,8 +84,7 @@ class ReplayActionPolicy(PolicyBase):
         Args:
             config: ReplayActionPolicyArgs configuration dataclass
         """
-        super().__init__()
-        self.config = config
+        super().__init__(config)
         self.episode_name = config.episode_name
         self.dataset_file_handler = HDF5DatasetFileHandler()
         self.dataset_file_handler.open(config.replay_file_path)
--- a/isaaclab_arena/policy/zero_action_policy.py
+++ b/isaaclab_arena/policy/zero_action_policy.py
@@ -44,15 +44,14 @@ class ZeroActionPolicy(PolicyBase):
     # enable from_dict() from policy_base.PolicyBase
     config_class = ZeroActionPolicyArgs
 
-    def __init__(self, config: ZeroActionPolicyArgs = None):
+    def __init__(self, config: ZeroActionPolicyArgs):
         """
         Initialize ZeroActionPolicy.
 
         Args:
             config: ZeroActionPolicyArgs configuration dataclass (optional, not used)
         """
-        super().__init__()
-        self.config = config if config is not None else ZeroActionPolicyArgs()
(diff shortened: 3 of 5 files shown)


isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -42,16 +42,23 @@ def load_env(arena_env_args: list[str], job_name: str):
...
-    policy_args_parser = get_isaaclab_arena_cli_parser()
-    policy_added_args_parser = policy_cls.add_args_to_parser(policy_args_parser)
-    # only for policy related arguments
-    policy_args = policy_added_args_parser.parse_args(job.policy_args)
-    policy = policy_cls.from_args(policy_args)
+    # Use direct from_dict if the policy class has config_class defined
+    if hasattr(policy_cls, "config_class") and policy_cls.config_class is not None:
+        # Use the inherited from_dict() method from PolicyBase
+        policy = policy_cls.from_dict(job.policy_config_dict)
+    else:
```
viiik-inside, line 57: A really good PR!! What do you think about removing the cli way completely? I think its nicer by just having the ConfigDataclass

isaaclab_arena/policy/policy_base.py
```diff
@@ -8,14 +8,50 @@
...
+    Base class for policies.
+
+    Subclasses should define a `config_class` class variable pointing to their configuration dataclass
+    to enable configuration from dictionaries via the from_dict() method.
+    """
+
+    # Optional: Subclasses can define this to enable from_dict()
+    config_class: type | None = None
+
     def __init__(self):
```
viiik-inside: i think it would be a good idea to add config_class to the init here

isaaclab_arena/policy/replay_action_policy.py
```diff
@@ -22,24 +74,33 @@ class ReplayActionPolicy(PolicyBase):
...
+    def __init__(self, config: ReplayActionPolicyArgs):
+        """
+        Initialize ReplayActionPolicy from a configuration dataclass.
+
+        Args:
+            config: ReplayActionPolicyArgs configuration dataclass
+        """
         super().__init__()
-        self.episode_name = episode_name
+        self.config = config
```
viiik-inside: can remove this line as super will take care of it

isaaclab_arena/policy/policy_base.py
xyao-nv: I don't agree. config_class is a class var, if adding to the init(), it's essentially becoming a instance var, which shall be the actual config. When defining the PolicyClass, its config_class shall already be defined. Each instance shall just specify the config, not config_class.

isaaclab_arena/policy/replay_action_policy.py
xyao-nv: I am not sure which you are referring to. config and config_class are different, config is the data field which its type is config_class.

isaaclab_arena/policy/policy_base.py
viiik-inside: Got confused. I meant the adding config to the init here as its used in every other child class.

viiik-inside: This is because the from_dict method calls the class constructor directly and it will fail if the subclass does not explicitely define config.

isaaclab_arena/policy/replay_action_policy.py
viiik-inside: Commented above

isaaclab_arena/policy/policy_base.py
xyao-nv: that makes sense. good catch! done

PR #333, opened 2026-01-14
Evaluation of a RSL RL policy

viiik-inside:
## Summary
RSL RL policy evaluation class


isaaclab_arena/metrics/lift_success.py
```diff
@@ -0,0 +1,102 @@
...
+        return self.name, lift_success
+
+
+@configclass
+class LiftSuccessRecorderCfg(RecorderTermCfg):
+    """Configuration for the lift success recorder."""
+
+    class_type: type[RecorderTerm] = LiftSuccessRecorder
+    name: str = "lift_success"
+    params: dict = {}  # Will be set by metric
```
alexmillane: Suggestion to just put the two parameters directly in the `@dataclass`. That way we get types, defaults, and they're in the constructor etc etc.

isaaclab_arena/tasks/lift_object_task.py
```diff
@@ -64,7 +65,12 @@ def get_mimic_env_cfg(self, embodiment_name: str):
         raise NotImplementedError("Function not implemented yet.")
 
     def get_metrics(self) -> list[MetricBase]:
-        return []
+        return [
+            LiftSuccessMetric(
+                minimum_height=self.minimum_height_to_lift,
+                object_name=self.lift_object.name,
+            )
+        ]
```
alexmillane: Suggestion also to return `SuccessRateMetric()`. Actually, all tasks should return that. Not sure of a good way of enforcing that.

alexmillane: Edit: looking at the code closely, this task doesn't define a success term. This is a (apparently unenforced) assumption in Arena.

What's the reason there? Is there anything preventing us from defining a success term? Right now the `LiftObjectTask` wont really work in an IL setting.

The second question is, if we define success, do we need `LiftSuccessMetric`. I.e. is `SuccessRateMetric` the same thing? Potentially not because here success may also entail reaching an `(x,y,z)` goal, rather than just `z>min_height`. If that's the case, `LiftSuccessMetric` does indeed provide more information.

alexmillane (commented):
Looks great! Thanks for doing this!

I have some questions about the success term in the `LiftObjectTask` which I think that we need to sort out.

Suggestion also to add a test. Could just run the the `policy_runner` with the default configs that we have and check that it doesn't crash.

isaaclab_arena/metrics/lift_success.py
viiik-inside: yeah, makes sense. Was just unsure about this term. Just added it so our metrics can print something

isaaclab_arena/tasks/lift_object_task.py
viiik-inside: its a bit tricky in this case.

1. Lift Obejct task actually uses the command term to get its goal location. Its not a certain height but a certain position in space. In the RL training when it reaches the position it waits until time out to reset.

2. As the goal location comes from the command term, it starts in the RL problem and couldnt be included in the termination config which is in the IL problem.

3. This particular problem is ill defined in the IL setting as such. We need an observation term and convert the problem into setting a random goal pose as well.

viiik-inside: I am gonna write a note here. When using a command manager, the terminations dont have a success term. Which makes sense, but how do we combine them in our setting is gonna be a different question. I would say we would need to override the termination cfg by removing the success term in the RL setting.I will implement that

viiik-inside: I have made some changes here and i think its the neatest way forward to solve this issue. Ill try to explain below.

IL Base class needs success
RL class should not have success during training but needs success during policy evaluation(by our script)
RL env sets goal pose from the command manager, which the IL base class does not have access to.

Solution, have a smarter success condition. Which will return False during rl training, enforced with a flag. Will get the goal position from the command manager during rl_evaluation. Will take in the goal pose directly for IL training if command manager does not exist.

isaaclab_arena/tasks/lift_object_task.py
```diff
@@ -47,24 +74,44 @@ def get_scene_cfg(self):
     def get_termination_cfg(self):
         return self.termination_cfg
 
-    def make_termination_cfg(self):
+    def get_events_cfg(self):
+        return self.events_cfg
+
+    def make_termination_cfg(self, rl_training: bool = False, use_command_goal: bool = False):
```
alexmillane: Is this behavior is general to all RLTasks? That the RL task termination is always (or at least usually) the IL task termination minus the success condition?

Can we push this to a function or a base class? Maybe the simplest is a function `remove_success_condition(super.get_termination_cfg())`. Actually I think that @peterd-NV did this for sequential tasks - he had to remove the success conditions from subtasks when forming the composite.

isaaclab_arena/tasks/lift_object_task.py
```diff
@@ -31,12 +33,37 @@ def __init__(
...
     ):
+        """Initialize the Lift Object task.
+
+        Args:
+            lift_object: The object to lift.
+            background_scene: The background scene (table, etc.).
+            episode_length_s: Episode length in seconds.
+            goal_position_delta_xyz: Goal position delta [dx, dy, dz] relative to initial pose (m).
+                Default: (0, 0, 0.3) = 30cm above initial position.
+            goal_orientation_wxyz: Target orientation [w, x, y, z]. Default: (1, 0, 0, 0) = identity.
```
alexmillane: I think this might be an error. I can't see a goal orientation in the argument list.

isaaclab_arena/tasks/lift_object_task.py
```diff
@@ -31,12 +33,37 @@ def __init__(
...
+            goal_position_tolerance: Position tolerance for success (m).
+        """
         super().__init__(episode_length_s=episode_length_s)
         self.lift_object = lift_object
         self.background_scene = background_scene
-        self.minimum_height_to_lift = minimum_height_to_lift
```
alexmillane, line 39: Suggestion to also remove from the argument list.

isaaclab_arena/tasks/lift_object_task.py
```diff
@@ -89,13 +141,55 @@ def __init__(
...
+        Args:
+            lift_object: The object to lift.
+            background_scene: The background scene (table, etc.).
+            embodiment: The robot embodiment.
+            minimum_height_to_lift: Minimum height to consider the object lifted (m).
+            episode_length_s: Episode length in seconds.
+            target_x_delta: Target range deltas for x [min_delta, max_delta] relative to initial pose (m).
+            target_y_delta: Target range deltas for y [min_delta, max_delta] relative to initial pose (m).
+            target_z_delta: Target range deltas for z [min_delta, max_delta] relative to initial pose (m).
+            rl_training_mode: If True, disables success termination. Set to False for evaluation.
```
alexmillane, line 155: Can we use the underlying (IL) task (`LiftObjectTask`) for evaluation, rather than passing this flag?

isaaclab_arena/tasks/lift_object_task.py
```diff
@@ -89,13 +141,55 @@ def __init__(
...
+            initial_pose.position_xyz[0] + target_x_delta[1],
+        )
+        self.target_y_range = (
+            initial_pose.position_xyz[1] + target_y_delta[0],
+            initial_pose.position_xyz[1] + target_y_delta[1],
+        )
+        self.target_z_range = (
+            initial_pose.position_xyz[2] + target_z_delta[0],
+            initial_pose.position_xyz[2] + target_z_delta[1],
+        )
```
alexmillane, line 177: Is there a reason we use a different definition of success between IL and RL? In IL the success region is a sphere, in RL it's a rectangular prism. If there's not a strong reason for this, I'd suggest we align on one.

isaaclab_arena/tasks/lift_object_task.py
```diff
@@ -112,6 +209,14 @@ def __init__(
             ee_frame_name=self.embodiment.get_ee_frame_name(self.embodiment.get_arm_mode()),
         )
 
+        # Override termination config with RL training mode
+        # Training: rl_training_mode=True -> no success termination
+        # Evaluation: rl_training_mode=False -> success from command manager
+        self.termination_cfg = super().make_termination_cfg(
+            rl_training=self.rl_training_mode,
```
alexmillane: As mentioned above, suggestion to just create the IL task termination and then remove the success conditions. That way details of the RL task (a child class) don't bleed into the IL task (a parent class).

isaaclab_arena/tasks/lift_object_task.py
viiik-inside: yup

isaaclab_arena/tasks/lift_object_task.py
viiik-inside: done

isaaclab_arena/tasks/lift_object_task.py
viiik-inside: This is the range for the command manager. The success condition is still a sphere.

PR #334, opened 2026-01-15
Upgrade to GR00T N1.6

xyao-nv:
## Summary
Update doc & code to consume GR00T N1.6 policy

## Detailed description
- Update submodule to GN1.6 with Arena-required fixes merged in GR00T repo ([list of fixes](https://github.com/NVIDIA/Isaac-GR00T/commits/main/?author=xyao-nv))
- Use customized deps pip install in dockerfile as GR00T requires py=3.10 in toml. Arena-GR00T docker can also run GN1.5, which I verified locally.
- Upload to HF gn1_6 branch ([gr1](https://huggingface.co/nvidia/GN1x-Tuned-Arena-GR1-Manipulation/tree/gn1_6), [g1](https://huggingface.co/nvidia/GN1x-Tuned-Arena-G1-Loco-Manipulation/tree/gn1_6)) for retrained ckpts using the v0.1-released data
- Modify policy interfaced code segments & doc
- Include OSMO docker build & job template to support multi-gpu post-training

## Todo
- Verify Arena-GR00T running on Blackwell. I have been verified GR00T itself can run on BW with those deps.

### DIFF
--- a/docker/setup/install_gr00t_deps.sh
+++ b/docker/setup/install_gr00t_deps.sh
@@ -26,7 +26,8 @@ sudo apt-get update && sudo apt-get install -y ffmpeg && rm -rf /var/lib/apt/lis
 # Install flash-attn immediately after torch (requires torch to be installed first)
 echo "Installing flash-attn 2.7.4.post1..." && \
 /isaac-sim/python.sh -m pip install --no-build-isolation --use-pep517 flash-attn==2.7.4.post1 && \
-# Install GR00T package without dependencies, as ignore its required py-3.10 version
+# Install GR00T package without dependencies. GR00T pyproject.toml specifies python 3.10, which conflicts with IsaacSim's python 3.11.
+# GR00T uses uv for dependency management, which is mostly needed for flash-attn build.
 echo "Installing Isaac-GR00T package (no deps)..." && \
 /isaac-sim/python.sh -m pip install --no-deps --ignore-requires-python -e ${WORKDIR}/submodules/Isaac-GR00T/ && \
 # Install GR00T main dependencies manually
@@ -76,8 +77,7 @@ echo "Installing GR00T main dependencies..."
 echo "Ensuring pytorch torchrun script is in PATH..."
 echo "export PATH=/isaac-sim/kit/python/bin:\$PATH" >> /etc/bash.bashrc
 
-echo "Removing pre-bundled torch to avoid conflicts..." && \
-rm -rf /isaac-sim/exts/omni.isaac.ml_archive/pip_prebundle/functorch* || true && \
+echo "Removing pre-bundled typing_extensions to avoid conflicts..." && \
 rm -rf /isaac-sim/exts/omni.isaac.ml_archive/pip_prebundle/typing_extensions* || true && \
 rm -rf /isaac-sim/exts/omni.pip.cloud/pip_prebundle/typing_extensions* || true
 
--- a/osmo/Dockerfile.gr00t_1_6
+++ b/osmo/Dockerfile.gr00t_1_6
@@ -45,25 +45,3 @@ RUN git clone https://huggingface.co/nvidia/GR00T-N1.6-3B /workspace/pretrained_
 RUN curl -LO https://github.com/filebrowser/filebrowser/releases/download/v2.32.0/linux-amd64-filebrowser.tar.gz
 RUN tar -xzvf linux-amd64-filebrowser.tar.gz
 RUN mv filebrowser /usr/bin/
-
-RUN apt-get install -y libegl1
-RUN mkdir -p /usr/share/glvnd/egl_vendor.d && \
-    cat >/usr/share/glvnd/egl_vendor.d/10_nvidia.json <<'EOF'
-{
-    "file_format_version" : "1.0.0",
-    "ICD" : {
-        "library_path" : "libEGL_nvidia.so.0"
-    }
-}
-EOF
-
-RUN mkdir -p /usr/share/vulkan/icd.d && \
-    cat >/usr/share/vulkan/icd.d/nvidia_icd.json <<'EOF'
-{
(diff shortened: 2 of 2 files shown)


docker/setup/install_gr00t_deps.sh
```diff
@@ -22,20 +21,64 @@ echo "TORCH_CUDA_ARCH_LIST=$TORCH_CUDA_ARCH_LIST"
...
-
+# Install torch first (force reinstall all dependencies to avoid prebundle version conflicts)
+# Torch 2.7.0 requested by GR00T is installed in isaacsim, skip here.
+# Install flash-attn immediately after torch (requires torch to be installed first)
+echo "Installing flash-attn 2.7.4.post1..." && \
+/isaac-sim/python.sh -m pip install --no-build-isolation --use-pep517 flash-attn==2.7.4.post1 && \
+# Install GR00T package without dependencies, as ignore its required py-3.10 version
+echo "Installing Isaac-GR00T package (no deps)..." && \
+/isaac-sim/python.sh -m pip install --no-deps --ignore-requires-python -e ${WORKDIR}/submodules/Isaac-GR00T/ && \
+# Install GR00T main dependencies manually
```
alexmillane, line 33: I guess we're doing this because we want different deps than those specified by their `pyproject.toml`?

If possible, could you expand the comment to indicate the reason here, and potentially list the packages we're need different version for? I guess that most of these packages are the same as specified in their `toml`.

docker/setup/install_gr00t_deps.sh
```diff
@@ -22,20 +21,64 @@ echo "TORCH_CUDA_ARCH_LIST=$TORCH_CUDA_ARCH_LIST"
...
+    peft==0.17.0 \
+    protobuf==3.20.3 \
+    onnx==1.17.0 \
+    tyro \
+    pytest && \
 # Ensure pytorch torchrun script is in PATH
 echo "Ensuring pytorch torchrun script is in PATH..."
 echo "export PATH=/isaac-sim/kit/python/bin:\$PATH" >> /etc/bash.bashrc
 
+echo "Removing pre-bundled torch to avoid conflicts..." && \
```
alexmillane: Is it fair to say that we're still using the Isaac Sim pre-bundled torch, but we're not using a few pre-bundled torch extensions: `functorch` and `typing_extensions`.

Suggestion to update the comment to reflect that?

docs/pages/example_workflows/locomanipulation/step_3_policy_training.rst
```diff
@@ -15,12 +15,9 @@ Once inside the container, set the dataset and models directories.
...
 
-.. note::
-    The GR00T N1.5 codebase does not support running on Blackwell architecture by default. There are
-    instructions `here <https://github.com/NVIDIA/Isaac-GR00T?tab=readme-ov-file#faq>`_ to building certain packages from source to support running on these architectures.
-    We have not tested these instructions, and therefore we do not recommend using
-    the **Base + GR00T** container for policy post-training and evaluation on
-    Blackwell architecture, like RTX 50 series, RTX Pro 6000 or DGX Spark.
+.. todo::
+
+   Verifying Arena-GR00T on Blackwell architecture.
```
alexmillane, line 20: This section is GR00T-only right? So I guess this section _has_ been verified no?

docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
```diff
@@ -41,6 +37,7 @@ pre-trained model checkpoint below:
    .. code-block:: bash
 
       hf download \
+         --revision gn1_6 \
```
alexmillane, line 40: Nice. Thank you for doing this. I guess the only checkpoints/datasets will be under another revision?

osmo/Dockerfile.gr00t_1_6
```diff
@@ -0,0 +1,69 @@
...
+RUN mkdir -p /usr/share/vulkan/icd.d && \
+    cat >/usr/share/vulkan/icd.d/nvidia_icd.json <<'EOF'
+{
+    "file_format_version": "1.0.0",
+    "ICD": {
+        "library_path": "libGLX_nvidia.so.0",
+        "api_version": "1.2.140"
+    }
+}
+EOF
```
alexmillane: What's the reason for changes to the graphics libs. Niavely I would assume that these were not required on OSMO.

docker/setup/install_gr00t_deps.sh
xyao-nv: Mainly because they use uv for deps management, which seems to help with flash-attn building. This extra deps can be avoided by manual installation. 
Updated the comments.

docker/setup/install_gr00t_deps.sh
xyao-nv: good catch. updated. thx

docs/pages/example_workflows/locomanipulation/step_3_policy_training.rst
xyao-nv: it's arena + gr00t. Gr00t itself on py3.10, which i have verified separately.

docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
xyao-nv: No need for datasets. It's only ckpts. Ideally it shall be under another HF 'model' but I don't want to be bothered by legal now and just push it to this public HF model with a public branch called gn1_6. For v0.2 release, I can update it to another HF model.

osmo/Dockerfile.gr00t_1_6
xyao-nv: True not on OSMO. And I should have not copied over from GR00T repo. It was added by Gear as sometimes they are partially installed and it's just adding missing files.

PR #337, opened 2026-01-20
Add mimic and metrics support to sequential composite tasks

peterd-NV:
## Summary
Add Mimic datagen and metrics support for sequential composite tasks.

For Mimic:

1. Add feature to automatically combine Mimic subtask configs from atomic tasks into single config for composite task
2. Add example of a mimic config for a sequential task (`franka_put_and_close_door_task.py`)
3. Update annotate_demos.py script to call the `success_term` after each step. This is required to make sure the sequential task's success state machine is properly updated after each step.

For Metrics:

1. Add support for `get_metrics` in sequential composite tasks.
2. Add sequential task specific success/subtask success metrics.
4. Update existing arena metrics recorder term naming to enable metric recorder terms to be renamed. This allows sequential composite tasks to rename metrics with a subtask suffix.

### DIFF
--- a/isaaclab_arena/affordances/affordance_base.py
+++ b/isaaclab_arena/affordances/affordance_base.py
@@ -3,16 +3,25 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from abc import ABC, abstractmethod
+from abc import ABC
 
 
 class AffordanceBase(ABC):
-    """Base class for affordances."""
+    """Base class for affordances.
 
-    @property
-    @abstractmethod
-    def name(self) -> str:
+    NOTE: Affordances must always be combined with an Asset class through multiple inheritance.
+    This ensures that affordances have access to the asset's name and other properties.
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
         # NOTE(alexmillane, 2025.09.19) Affordances always have be combined with
-        # an Asset which has a "name" property. By declaring this property
-        # abstract here, we enforce this.
-        pass
+        # an Asset which has a "name" property. We enforce this at runtime.
+        from isaaclab_arena.assets.asset import Asset
+
+        if not isinstance(self, Asset):
+            raise TypeError(
+                f"{self.__class__.__name__} must inherit from Asset. "
+                "Affordances must be combined with Asset through multiple inheritance. "
+                f"Example: class MyClass(Asset, {self.__class__.__bases__[0].__name__})."
+            )
--- a/isaaclab_arena/assets/asset.py
+++ b/isaaclab_arena/assets/asset.py
@@ -14,12 +14,6 @@ class Asset:
         # Calling super even though this is a base class to support
         # multiple inheritance of inheriting classes.
         super().__init__(**kwargs)
-        # self.name = name
         assert name is not None, "Name is required for all assets"
-        self._name = name
+        self.name = name
         self.tags = tags
-
-    # name is a read-only property
-    @property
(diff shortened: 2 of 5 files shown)


isaaclab_arena/scripts/imitation_learning/annotate_demos.py
```diff
@@ -320,6 +320,7 @@ def replay_episode(
                 continue
         action_tensor = torch.Tensor(action).reshape([1, action.shape[0]])
         env.step(torch.Tensor(action_tensor))
+        success_term.func(env, **success_term.params)
```
alexmillane, line 324: Is this intended? We appear not to use the result?

If it is intended, suggestion to add a comment.

isaaclab_arena/tasks/sequential_composite_tasks/franka_put_and_close_door_task.py
```diff
@@ -0,0 +1,63 @@
...
+        self.datagen_config.generation_keep_failed = False
+        self.datagen_config.generation_num_trials = 100
+        self.datagen_config.generation_select_src_per_subtask = False
+        self.datagen_config.generation_select_src_per_arm = False
+        self.datagen_config.generation_relative = False
+        self.datagen_config.generation_joint_pos = False
+        self.datagen_config.generation_transform_first_robot_pose = False
+        self.datagen_config.generation_interpolate_from_last_target_pose = True
+        self.datagen_config.max_num_failures = 25
+        self.datagen_config.seed = 1
```
alexmillane, line 63: These values seem the same everywhere across. Would it make sense to capture them in a default constant `MIMIC_DATAGEN_CONFIG_DEFAULTS = {...}`. That way, we could save repeating them across the code base. Wdyt?

If we were to do that, we could add a function that returns a default `*MimicEnvCfg` for a (generic) sequential task. And add this as the default implementation of `get_mimic_env_cfg`.

That way if a user doesn't want something special, they don't have to write any mimic configuration struct.

Wdyt?

isaaclab_arena/scripts/imitation_learning/annotate_demos.py
peterd-NV: Yes this is intended, its a consequence of the success term in sequential tasks being coupled with the "success state machine". Normally this isn't an issue since all termination terms are executed on each env step, but for the annotation script it zeros out all env terminations and manages them by itself. Therefore it needs this extra step here to manually step the success term. 

I'll add a comment to give context in the code.

isaaclab_arena/tasks/sequential_composite_tasks/franka_put_and_close_door_task.py
peterd-NV: I like that suggestion, we could use that clean up the other Mimic env configs as well.

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -116,6 +122,28 @@ class TerminationsCfg:
...
+            params={
+                "pose_range": {
+                    "x": reset_pose_range["x"],
+                    "y": reset_pose_range["y"],
+                    "z": reset_pose_range["z"],
+                },
+                "velocity_range": {},
+                "asset_cfg": SceneEntityCfg(pick_up_object.name),
+            },
+        )
```
alexmillane, line 144: In parallel, I have changed the way we do randomization. #340 .

I'll let you merge this MR as is, and then update this 😄

PR #342, opened 2026-01-22
Allow IsaacLabArenaEnvironment embodiment to be None

cvolkcvolk:
## Summary
Makes the `IsaacLabArenaEnvironment` embodiment optional for scenes where we don't need an embodiment. This is particularly helpful in developing, debugging RelationSolver code, or demonstration purpose. 
The `ArenaEnvBuilder` handles Non-values of the embodiment. 
Because IsaacLab's ActionManager requires a valid config object (not None), an `EmtpyActionCfg` @configclass is introduced.

### DIFF
--- a/isaaclab_arena/embodiments/no_embodiment.py
+++ b/isaaclab_arena/embodiments/no_embodiment.py
@@ -18,6 +18,8 @@ class EmptyActionsCfg:
 class NoEmbodiment(EmbodimentBase):
     """Null object for environments without an embodiment."""
 
+    name = "no_embodiment"
+
     def __init__(self):
         super().__init__()
         self.action_config = EmptyActionsCfg()


isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -48,41 +49,43 @@ def orchestrate(self) -> None:
...
+        if self.arena_env.embodiment is not None:
+            env_cfg = self.arena_env.embodiment.modify_env_cfg(env_cfg)
         env_cfg = self.arena_env.scene.modify_env_cfg(env_cfg)
         return env_cfg
 
     def compose_manager_cfg(self) -> IsaacLabArenaManagerBasedRLEnvCfg:
         """Return base ManagerBased cfg (scene+events+terminations+xr), no registration."""
 
         # Constructing the environment by combining inputs from the scene, embodiment, and task.
+        _embodiment = self.arena_env.embodiment
```
alexmillane: Suggestion to drop the leading `_`. This is a normal local variable, no need to have that (unless there was some specific reason I'm not seeing).

alexmillane: You could define 

```
class NoEmbodiment(EmbodimentBase):
    def __init__(self):
          self.action_cfg = EmptyActionsCfg()
```

Then in the env builder

```
embodiment = self.arena_env.embodiment if self.arena_env.embodiment else NoEmbodiment()
```

This would clean up the repeated checking in this function. All the functions `get_*_cfg` should naturally return `None`

cvolkcvolk: Done. Introduced a `NoEmbodiement`

PR #345, opened 2026-01-23
Relation solving works within ArenaEnvBuilder

cvolkcvolk:
## Summary

Change the API slightly to make the integration into ArenaEnvBuilder a bit cleaner:
Introduce `IsAnchor`. This implicitly marks an object as anchor and we don't need to pass this info along to the `ArenaEnvBuilder`. This way we can infer everything we need from the scene:

- The optimizable objects: All objects that hold a relation

- Anchor object: The respectively marked object.

Furthermore this allows us to just pass a list of all objects, including the anchor, to `ObjectPlacer.place(objects=objects)`. instead of duplicating the `anchor_object` as a seperate argument `ObjectPlacer.place(objects=objects, anchor_object=anchor_object)`

In the ObjectPlacer we now:
1) Enforce all objects passed to ObjectPlacer hold a relation and are meant to be optimized.
2) We indeed have an anchor object with a fixed position.

### DIFF
--- a/isaaclab_arena_examples/relations/relation_solver_visualization_notebook.py
+++ b/isaaclab_arena_examples/relations/relation_solver_visualization_notebook.py
@@ -68,7 +68,7 @@ def create_loss_heatmap_2d(
                     positions[obj] = obj.get_initial_pose().position_xyz
 
             # Create state and compute loss
-            state = RelationSolverState(all_objects, anchor_object, positions)
+            state = RelationSolverState(all_objects, positions)
             loss = solver._compute_total_loss(state)
             losses[i, j] = loss.item()
 
@@ -238,7 +238,7 @@ def run_visualization_demo():
             parent: parent_pos,
             child1: (x, parent_pos[1], parent_pos[2]),
         }
-        state = RelationSolverState(objects_child1, parent, positions)
+        state = RelationSolverState(objects_child1, positions)
         loss = solver._compute_total_loss(state)
         losses_x_child1.append(loss.item())
 
@@ -255,7 +255,7 @@ def run_visualization_demo():
             child1: child1_ideal_pos,
             child2: (x, parent_pos[1], parent_pos[2]),
         }
-        state = RelationSolverState(objects_child2, parent, positions)
+        state = RelationSolverState(objects_child2, positions)
         loss = solver._compute_total_loss(state)
         losses_x_child2.append(loss.item())
 


isaaclab_arena/cli/isaaclab_arena_cli.py
```diff
@@ -41,6 +42,19 @@ def add_isaac_lab_cli_args(parser: argparse.ArgumentParser) -> None:
...
+    """Add Isaac Lab Arena specific command line arguments to the given parser."""
+    arena_group = parser.add_argument_group(
+        "Isaac Lab Arena Arguments", "Arguments specific to Isaac Lab Arena framework"
+    )
+    arena_group.add_argument(
+        "--solve-relations",
+        action="store_true",
+        default=False,
+        help="Solve spatial relations in the environment.",
+    )
```
alexmillane, line 56: Am I reading this correctly that by default, we don't solve relation problem? That seems like a strange default to me. I would think that if a user has specified relations, we would solve them by default. What's the thinking here?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -41,6 +43,46 @@ def orchestrate(self) -> None:
...
+        """Get all objects from the scene that have relations.
+
+        Returns:
+            List of Object instances that have at least one relation.
+        """
+        objects_with_relations: list[Object] = []
+        for asset in self.arena_env.scene.assets.values():
+            # TODO(cvolk): Support ObjectSets.
+            if not isinstance(asset, Object):
+                continue
```
alexmillane, line 62: Potentially, shall we `assert` that if a relation is specified, the asset should be an `Object`? Such that if a user specifies a relation on a non-supported asset type they know that that's not going to work. Otherwise I guess the behaviour is going to be that the asset shows up at `(0,0,0)` and they're probably going to blame the solver.

isaaclab_arena/examples/compile_env_notebook.py
```diff
@@ -27,10 +28,13 @@
 background = asset_registry.get_asset_by_name("kitchen")()
 embodiment = asset_registry.get_asset_by_name("franka")()
 cracker_box = asset_registry.get_asset_by_name("cracker_box")()
+tomato_soup_can = asset_registry.get_asset_by_name("tomato_soup_can")()
 
 cracker_box.set_initial_pose(Pose(position_xyz=(0.4, 0.0, 0.1), rotation_wxyz=(1.0, 0.0, 0.0, 0.0)))
+cracker_box.add_relation(IsAnchor())
+tomato_soup_can.add_relation(On(cracker_box))
```
alexmillane, line 35: cool!

isaaclab_arena/relations/object_placer.py
```diff
@@ -41,20 +42,31 @@ def __init__(self, params: ObjectPlacerParams | None = None):
...
 
         Returns:
             PlacementResult with success status, positions, loss, and attempt count.
         """
+        # Validate all objects have at least one relation
+        for obj in objects:
+            assert obj.get_relations(), (
+                f"Object '{obj.name}' has no relations. All objects passed to place() must have "
+                "at least one relation (e.g., On(), NextTo(), or IsAnchor())."
+            )
```
alexmillane, line 60: Is this necessary? I could imagine a scene where I have a bunch of objects placed manually, by setting their pose explicitly, and then some objects placed using relations. Imagine a scene with a table and some chairs whose poses are set explicitly and some objects on the table described by relations.

Actually, that makes me wonder if we need an anchor at all. An anchor is just an object involved in a relation whose pose is not optimized. There could be multiple such objects, like in the example above. We could just say that all objects' poses are optimized, except for those with an explicit pose set. Objects who have an explicit pose set are what we're currently calling anchors (there could be multiple in the new formulation).

Maybe let's chat about this. I feel like mixing explicit setting of poses with relations is probably the right way to go, but let's think about how that aligns with current deadlines 😅. We could just go with this approach for now.

isaaclab_arena/cli/isaaclab_arena_cli.py
cvolkcvolk: Thinking was to further test it before its shipped by default. 
But valid point, for all cases where no Relations are used the solver is skipped anyway. It's enabled now by default.

isaaclab_arena/environments/arena_env_builder.py
cvolkcvolk: Done

isaaclab_arena/relations/object_placer.py
cvolkcvolk: Noted. As discussed offline, I will move this into a follow up MR.

PR #350, opened 2026-01-26
Support Quest teleop for G1 locomanipulation example

qianl-nv:
## Summary
Support Quest3 for loco-manipulation teleop and record/replay

## Detailed description
Support Quest3 for loco-manipulation teleop for g1_pink example. This allows proper data recording for the locomanipulation.
- What has been changed?
Adds openxr + g1_pink teleop support
Adds documention for locomanip data recording with Quest

### DIFF
--- a/docs/pages/example_workflows/locomanipulation/step_2_teleoperation.rst
+++ b/docs/pages/example_workflows/locomanipulation/step_2_teleoperation.rst
@@ -120,19 +120,6 @@ Step 4: Setup and Connect from Meta Quest 3
 * **Right joystick**: Squat (down), rotate torso (left/right).
 * **Controllers**: Move end-effector (EE) targets for the arms.
 
-.. hint::
-
-   Suggested sequence for the task:
-
-   #. Align your body with the robot.
-   #. Walk forward (left joystick forward).
-   #. Grab the box (controllers).
-   #. Walk backward (left joystick back).
-   #. Turn toward the bin (right joystick).
-   #. Walk forward to the bin.
-   #. Squat (right joystick down).
-   #. Place the box in the bin (controllers).
-
 
 Step 5: Record with Quest 3
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -155,6 +142,19 @@ Step 5: Record with Quest 3
         --teleop_device openxr
 
 #. Complete the task for each demo. Reset between demos. The script saves successful runs to the HDF5 file above.
+
+.. hint::
+
+   Suggested sequence for the task:
+
+   #. Align your body with the robot.
+   #. Walk forward (left joystick forward).
+   #. Grab the box (controllers).
+   #. Walk backward (left joystick back).
+   #. Turn toward the bin (right joystick).
+   #. Walk forward to the bin.
+   #. Squat (right joystick down).
+   #. Place the box in the bin (controllers).
 
 .. image:: ../../../images/g1_galileo_arena_box_pnp_locomanip.gif
    :align: center


docs/pages/example_workflows/locomanipulation/step_2_teleoperation.rst
```diff
@@ -0,0 +1,173 @@
...
+        --xr \
+        --device cpu \
+        --enable_pinocchio \
+        --dataset_file $DATASET_DIR/arena_g1_locomanipulation_dataset_recorded.hdf5 \
+        --num_demos 10 \
+        --num_success_steps 2 \
+        galileo_g1_locomanip_pick_and_place \
+        --teleop_device openxr
+
+#. Complete the task (pick the brown box, place it in the blue bin) for each demo. Reset between demos. The script saves successful runs to the HDF5 file above.
```
peterd-NV: Suggestion: It might be good to include the same GIF from the `locomanipulation/index.rst` here as well to show what the task should look like, and updating the description to something like (walk to the brown box, pick it up, turn and walk to the blue bin and place the box in it). 

In my experience VDR likes to see stuff like that :)

isaaclab_arena/scripts/imitation_learning/record_demos.py
```diff
@@ -450,8 +450,16 @@ def stop_recording_instance():
         while simulation_app.is_running():
             # Get keyboard command
             action = teleop_interface.advance()
+
             # Expand to batch dimension
             actions = action.repeat(env.num_envs, 1)
+            # Hack for G1 Pink WBC to transferm EE into robot base coordinates
```
peterd-NV, line 456: I wondering why this transformation from world frame to robot frame needs to happen in the action term versus being something done by the XR retargetters. Seems like it would make sense for the action from the teleop device to be already relative to the robot since the robot can be placed arbitrarily anywhere. 

Totally fine to have this as it is right now but just curious if you know why XR designed the retargetter for G1 like that.

isaaclab_arena/tests/test_g1_wbc_pink_preprocess_actions.py
```diff
@@ -0,0 +1,191 @@
...
+        robot_base_quat = term._asset.data.root_link_quat_w[0]
+
+        # World-frame wrist positions: base + offset (so base-frame offset is known)
+        left_offset = torch.tensor([1.0, 2.0, 3.0], device=device)
+        right_offset = torch.tensor([4.0, 5.0, 6.0], device=device)
+        left_pos_world = robot_base_pos + left_offset
+        right_pos_world = robot_base_pos + right_offset
+
+        actions = torch.zeros(1, action_dim, device=device)
+        actions[0, 2:5] = left_pos_world
```
peterd-NV: Would be good to use the action constants for the G1 defined [here](https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena_g1/g1_whole_body_controller/wbc_policy/policy/action_constants.py)

isaaclab_arena/tests/test_g1_wbc_pink_preprocess_actions.py
```diff
@@ -0,0 +1,191 @@
...
+        actions[0, 5:9] = torch.tensor([1.0, 0.0, 0.0, 0.0], device=device)
+        actions[0, 9:12] = right_pos_world
+        actions[0, 12:16] = torch.tensor([1.0, 0.0, 0.0, 0.0], device=device)
+
+        out = term.preprocess_actions(actions)
+
+        # Base frame position = world - base (in world), then rotated by base_inv => offset when base quat is identity
+        torch.testing.assert_close(out[0, 2:5], left_offset, atol=1e-4, rtol=0)
```
peterd-NV: Same here for action constants

isaaclab_arena/tests/test_g1_wbc_pink_preprocess_actions.py
```diff
@@ -0,0 +1,191 @@
...
+        num_envs = robot_base_pos.shape[0]
+
+        # Arbitrary world-frame wrist poses
+        left_pos_w = torch.tensor([[1.0, 0.0, 0.5]], device=device).expand(num_envs, 3)
+        left_quat_w = torch.tensor([[1.0, 0.0, 0.0, 0.0]], device=device).expand(num_envs, 4)
+        right_pos_w = torch.tensor([[0.0, 1.0, 0.5]], device=device).expand(num_envs, 3)
+        right_quat_w = torch.tensor([[1.0, 0.0, 0.0, 0.0]], device=device).expand(num_envs, 4)
+
+        actions = torch.zeros(num_envs, action_dim, device=device)
+        actions[:, 2:5] = left_pos_w
```
peterd-NV: Same here for action constants

isaaclab_arena_g1/g1_env/mdp/actions/g1_decoupled_wbc_pink_action.py
```diff
@@ -360,3 +367,35 @@ def reset(self, env_ids: Sequence[int] | None = None) -> None:
...
+
+        Returns:
+            The processed actions tensor (same shape as input).
+        """
+        actions = actions.clone()
+
+        robot_base_pos = self._asset.data.root_link_pos_w[:, :3]
+        robot_base_quat = self._asset.data.root_link_quat_w
+
+        left_wrist_pos_world = actions[:, 2:5]
```
peterd-NV: Same here for action constants

isaaclab_arena/scripts/imitation_learning/record_demos.py
rwiltz: We can do world space to some frame of reference attached to some part of the robot, but then that would require that the retargeter be aware of where the robot is w.r.t. the teleop "world origin/base space" coordinate frame. I felt that would introduce additional shared state between the teleop stack and the application and require a tighter integration.

The other option is to consider every EE target a "relative pose" based on the initial or calibrated position of the teleoperator. We opted against this for the Lab integration because it requires a manual calibration step, or for the teleoperator to be aware of their initial starting position relative to the initial position of the robot. This is currently how real robot teleop works since they do not have a shared origin.

Once Isaac Teleop is integrated here, we can take advantage of this setup if you all prefer since it's also shared by the real robot stack (with clutching of the inputs, etc). Another option once Isaac Teleop is integrated, you can specify which coordinate system to output transforms in and apply a per frame offset. So if you pass in the robot base link to scene origin transform, you'll get all the poses relative to that link.

rwiltz: FWIW, in lab this is done in an action term before passing the EE tasks to IK

qianl-nv: I've tried including this in the action term process_action before passing the IK, but then the recorded action is in world coordinate, and during reply this is sensitive to the record/replay delta of the robot position and leads to gradual but noticable motion drift during replay when replaying multiple episodes. See the video in this [thread](https://nvidia.slack.com/archives/C09NF6YR77C/p1769443950247779?thread_ts=1769057316.439839&cid=C09NF6YR77C)

docs/pages/example_workflows/locomanipulation/step_2_teleoperation.rst
qianl-nv: Great suggestion! updated the doc

isaaclab_arena/tests/test_g1_wbc_pink_preprocess_actions.py
qianl-nv: Good point! Done

isaaclab_arena/tests/test_g1_wbc_pink_preprocess_actions.py
qianl-nv: done

isaaclab_arena/tests/test_g1_wbc_pink_preprocess_actions.py
qianl-nv: done

isaaclab_arena_g1/g1_env/mdp/actions/g1_decoupled_wbc_pink_action.py
qianl-nv: done

PR #352, opened 2026-01-26
Add mimic data gen default params and test case for seq task

peterd-NV:
## Summary

1. Adds a set of default Mimic data gen config params and updates existing tasks to use them.
2. Adds a test for Mimic data generation on a sequential task - Franka Put & Close door task (adds a small dataset to support test).
3. Fixes obj orientation in franka put & close task to match what was there previously so older datasets still work with Mimic.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
     needs: [pre_commit]
 
     container:
-      image: nvcr.io/nvstaging/isaac-amr/isaaclab_arena:cuda_gr00t
+      image: nvcr.io/nvstaging/isaac-amr/isaaclab_arena:cuda_gr00t_gn16
       credentials:
         username: $oauthtoken
         password: ${{ env.NGC_API_KEY }}
@@ -233,4 +233,4 @@ jobs:
       - name: Build & push image to NGC (GR00T Image)
         run: |
           chmod +x ./docker/push_to_ngc* || true
-          ./docker/push_to_ngc.sh -p -g -t cuda_gr00t
+          ./docker/push_to_ngc.sh -p -g -t cuda_gr00t_gn16
--- a/docker/Dockerfile.isaaclab_arena
+++ b/docker/Dockerfile.isaaclab_arena
@@ -4,7 +4,6 @@ FROM ${BASE_IMAGE}
 
 # GR00T Policy Build Arguments, these are only used if INSTALL_GROOT is true
 ARG INSTALL_GROOT=false
-ARG GROOT_DEPS_GROUP=base
 
 ARG WORKDIR="/workspace"
 ENV WORKDIR=${WORKDIR}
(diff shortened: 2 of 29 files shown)


isaaclab_arena/tasks/sequential_composite_tasks/franka_put_and_close_door_task.py
```diff
@@ -33,31 +31,6 @@ def get_prompt(self) -> str:
...
-        self.datagen_config.generation_select_src_per_arm = False
-        self.datagen_config.generation_relative = False
-        self.datagen_config.generation_joint_pos = False
-        self.datagen_config.generation_transform_first_robot_pose = False
-        self.datagen_config.generation_interpolate_from_last_target_pose = True
-        self.datagen_config.max_num_failures = 25
-        self.datagen_config.seed = 1
+        cfg = super().get_mimic_env_cfg(arm_mode)
+        cfg.name = "franka_put_and_close_door_task"
+        return cfg
```
alexmillane, line 36: This is potentially a suggestion for a separate MR. But this class is now doing very little. The only things it is doing:
- setting the viewer_cfg
- changing `cfg.name`.

I feel like we could eliminate this class altogether by defining a default for the viewer config. A reasonable default could be:

```
    def get_viewer_cfg(self) -> ViewerCfg:
        return self.subtasks[0].get_viewer_cfg()
```

That is, the camera focuses (by default) on the first sub-task.

With this final default in place, we could move away from naming our sequential tasks and just go:

```
task = SequentialTaskBase([tasks=[pick_and_place_task, close_door_task])
```

This will prevent a proliferation of classes in this folder that decribe and name vairous combinations of atomic tasks. Wdyt?

isaaclab_arena/tests/test_sequential_task_mimic_data_generation.py
```diff
@@ -0,0 +1,75 @@
...
+        # example env
+        args.append("franka_put_and_close_door")
+        args.append("--embodiment")
+        args.append("franka")
+        run_subprocess(args)
+
+
+if __name__ == "__main__":
+    test_franka_put_and_close_door_mimic_data_generation_single_env()
+    test_franka_put_and_close_door_mimic_data_generation_multi_env()
```
alexmillane, line 75: Thanks for adding these!!

isaaclab_arena/tasks/sequential_composite_tasks/franka_put_and_close_door_task.py
peterd-NV: I had initially tried to remove this and move the task directly to the environment class and to something like 

`task = SequentialTaskBase([tasks=[pick_and_place_task, close_door_task])`

However, this causes an issue (which I think we've run into before) in that when Mimic is introduced to the sequential task config, it cannot be called before `app launcher` starts as there are dependencies on OV/omni-physics. 

Since the Arena compiler runs before app launcher is started, creating the task directly using SequentialTaskBase instead of making a separate class runs into this issue. I think one way of getting around this in the future is trying to remove the omni-physics dep that Mimic has so that it can be run prior to starting the app though I'm not 100% sure if that'll be feasible.

PR #355, opened 2026-01-27
Change string to compare as we now call a device openxr

viiik-inside:
## Summary
Dirty fix to get the new teleop device name working

### DIFF
--- a/isaaclab_arena/scripts/imitation_learning/record_demos.py
+++ b/isaaclab_arena/scripts/imitation_learning/record_demos.py
@@ -72,7 +72,8 @@ if args_cli.enable_pinocchio:
     # pinocchio is required by the Pink IK controllers and the GR1T2 retargeter
     import pinocchio  # noqa: F401
 
-    # Extremely hacky way to enable xr args. Please fix this.
+    # TODO(cvolk): XR mode is inferred from teleop device name via string matching.
+    # Ideally, AppLauncher or the device config would auto-detect XR requirements.
     if "openxr" in args_cli.teleop_device.lower():
         app_launcher_args["xr"] = True
 


isaaclab_arena/scripts/imitation_learning/record_demos.py
```diff
@@ -72,7 +72,8 @@
     # pinocchio is required by the Pink IK controllers and the GR1T2 retargeter
     import pinocchio  # noqa: F401
 
-    if "handtracking" in args_cli.teleop_device.lower():
+    # Extremely hacky way to enable xr args. Please fix this.
```
cvolkcvolk: Suggestion on the comment: "Extremely hacky... Please fix this" doesn't quite help future readers understand what needs fixing or why. If this needs a follow-up, maybe add a more actionable `TODO(someones_name)`

viiik-inside: dont know who is gonna todo this :)

viiik-inside: I think its more of a comment to work with the isaaclab team for us internally.

cvolkcvolk: You can add mine

cvolkcvolk: However as this file lives in our repo and is released through us I still suggest to change the comment with what should be fixed. 
Did not dig deep into it but potentially:
```
# TODO(cvolk): XR mode is inferred from teleop device name via string matching.
# Ideally, AppLauncher or the device config would auto-detect XR requirements.
```

viiik-inside: added a todo for you. Thanks

PR #357, opened 2026-01-27
NextTo relation handles all sides

cvolkcvolk:
## Summary
Extend the `NextToLossStrategy` to support all four placement sides (FRONT, BACK, LEFT, RIGHT) and fixes a bug where child objects could extend beyond the parent's extent. 

Misc: 
- Extended `isaac_sim_object_placer_notebook.py` & `dummy_object_placer_notebook.py` to demonstrate all four sides with a arrangement around a central object.

### DIFF
--- a/isaaclab_arena/embodiments/common/common.py
+++ b/isaaclab_arena/embodiments/common/common.py
@@ -1,0 +1,39 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from isaaclab.devices.openxr import XrCfg
+
+from isaaclab_arena.utils.pose import Pose
+
+
+def get_default_xr_cfg(initial_pose: Pose | None = None, xr_offset: Pose | None = None) -> XrCfg:
+    """
+    Get the default XR configuration for the robot.
+    Args:
+        initial_pose: The initial pose of the robot.
+        xr_offset: The offset of the XR device from the robot.
+
+    Returns:
+        The default XR configuration for the robot.
+    """
+    if initial_pose is None and xr_offset is None:
+        raise ValueError("initial_pose or xr_offset must be provided")
+    # If robot has an initial pose, compose it with the XR offset
+    if initial_pose is not None:
+        from isaaclab_arena.utils.pose import compose_poses
+
+        # Compose robot pose with XR offset: T_world_xr = T_world_robot * T_robot_xr
+        xr_pose_global = compose_poses(initial_pose, xr_offset)
+
+        return XrCfg(
+            anchor_pos=xr_pose_global.position_xyz,
+            anchor_rot=xr_pose_global.rotation_wxyz,
+        )
+    else:
+        # If no initial pose set, use the offset as global coordinates (robot at origin)
+        return XrCfg(
+            anchor_pos=xr_offset.position_xyz,
+            anchor_rot=xr_offset.rotation_wxyz,
+        )
--- a/isaaclab_arena/embodiments/g1/g1.py
+++ b/isaaclab_arena/embodiments/g1/g1.py
@@ -27,6 +27,7 @@ from isaaclab.utils.assets import ISAAC_NUCLEUS_DIR
 import isaaclab_arena.terms.transforms as transforms_terms
 from isaaclab_arena.assets.register import register_asset
 from isaaclab_arena.embodiments.common.arm_mode import ArmMode
+from isaaclab_arena.embodiments.common.common import get_default_xr_cfg
 from isaaclab_arena.embodiments.embodiment_base import EmbodimentBase
 from isaaclab_arena.utils.isaaclab_utils.resets import reset_all_articulation_joints
 from isaaclab_arena.utils.pose import Pose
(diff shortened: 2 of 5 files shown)


isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -20,6 +21,33 @@
...
+class SideConfig:
+    """Configuration for computing NextTo loss for a given side.
+
+    Attributes:
+        primary_axis: Index of the axis along which child is placed (0=X, 1=Y).
+        direction: +1 if child should be in positive direction from parent (RIGHT, BACK),
+                   -1 if child should be in negative direction (LEFT, FRONT).
+    """
+
+    primary_axis: int  # 0=X, 1=Y
```
xyao-nv: I wonder if Enum would be a better choice to establish those mappings easily.

isaaclab_arena_examples/relations/isaac_sim_object_placer_notebook.py
```diff
@@ -48,34 +46,60 @@ def run_isaac_sim_object_placer_demo(num_steps: int = 10000):
...
+        ground_plane,
+        light,
+        office_table,
+        cracker_box,
+        dex_cube,
+        sugar_box,
+        mustard_bottle,
+        tomato_soup_can,
+        mug,
+    ]
```
xyao-nv, line 98: Can't wait for the day when it can place all objs automatically from obj_library.py & LW SDK into a kitchen scene 😆

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -89,57 +117,69 @@ def compute_loss(
...
-        parent_y_max = parent_pos[1] + parent_bbox.max_point[1]  # Back edge
-
-        # 1. Half-plane loss: child must be to the right of parent's right edge
-        right_side_loss = single_boundary_linear_loss(
-            child_pos[0],
-            parent_x_max,
+        # Select parent edge and child offset based on direction
+        # direction +1 (RIGHT/BACK): child in positive direction, use parent max edge
+        # direction -1 (LEFT/FRONT): child in negative direction, use parent min edge
+        if cfg.direction == +1:
```
xyao-nv: Maybe it's me with bad memories, is it possible to not let users be bothered by the relationship between +1 and right/back while understanding this code?

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -89,57 +117,69 @@ def compute_loss(
...
-        # 3. Distance loss: child's left edge should be at distance from parent's right edge
-        assert distance >= 0.0, f"NextTo distance must be non-negative, got {distance}"
-        # child_x_min = child_pos[0] + child_bbox.min_point[0], so:
-        # target: child_pos[0] + child_bbox.min_point[0] = parent_x_max + distance
-        # target_child_pos_x = parent_x_max + distance - child_bbox.min_point[0]
-        target_x = parent_x_max + distance - child_bbox.min_point[0]
-        next_to_distance_loss = single_point_linear_loss(child_pos[0], target_x, slope=self.slope)
+        # 3. Distance loss: child edge at target distance from parent edge
+        # For direction +1: target = parent_max + distance - child_min
+        # For direction -1: target = parent_min - distance - child_max
```
xyao-nv, line 173: If you switch to Enum, then you can add a linear ops func to map 0/1 to +1/-1.

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -20,6 +21,33 @@
...
+    direction: int  # +1 or -1
+
+    @property
+    def band_axis(self) -> int:
+        """Perpendicular axis for band constraint."""
+        return 1 - self.primary_axis
+
+
+SIDE_CONFIGS: dict[Side, SideConfig] = {
+    Side.RIGHT: SideConfig(primary_axis=0, direction=+1),
```
alexmillane: Just to echo @xyao-nv 's comment. I think if you define Enums these might look a bit more readable

```
Side.RIGHT: SideConfig(primary_axis=Axis.X, direction=Direction.Positive),
```

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -20,6 +21,33 @@
...
+    """Configuration for computing NextTo loss for a given side.
+
+    Attributes:
+        primary_axis: Index of the axis along which child is placed (0=X, 1=Y).
+        direction: +1 if child should be in positive direction from parent (RIGHT, BACK),
+                   -1 if child should be in negative direction (LEFT, FRONT).
+    """
+
+    primary_axis: int  # 0=X, 1=Y
+    direction: int  # +1 or -1
```
alexmillane: see @xyao-nv 's comment above.

isaaclab_arena/relations/relation_loss_strategies.py
alexmillane: Potentially if we use the Enums as suggested above, this might look more understandable.

isaaclab_arena_examples/relations/isaac_sim_object_placer_notebook.py
alexmillane: That day is on the way!

isaaclab_arena/relations/relation_loss_strategies.py
cvolkcvolk: Done. Implemented an Axis and Direction Enum.

isaaclab_arena/relations/relation_loss_strategies.py
cvolkcvolk: That's a good idea!

isaaclab_arena/relations/relation_loss_strategies.py
cvolkcvolk: Thanks for the hint! Introduced an `Axis` and `Direction` enum with which the above config becomes much cleaner

PR #358, opened 2026-01-27
Unary AtPosition() strategy

cvolkcvolk:
## Summary

Introduce the `AtPosition` relation, a unary constraint that pins objects to specific world coordinates. When composing object placement constraints, the existing On relation controls z-positioning (placing on a surface) and constrains x/y to stay within the parent's bounds. However, there was no way to specify an exact x/y position on that surface. 

```
tomato_soup_can.add_relation(On(kitchen_counter, clearance_m=0.02))  # z from On
tomato_soup_can.add_relation(AtPosition(x=3.8, y=-0.32))  # exact x/y
```
`AtPosition` can be used along arbitary axes. If not specified as z in the example above, its just omitted in the loss calculation.

### DIFF
--- a/isaaclab_arena/relations/relation_loss_strategies.py
+++ b/isaaclab_arena/relations/relation_loss_strategies.py
@@ -5,6 +5,8 @@
 
 import torch
 from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from enum import IntEnum
 from typing import TYPE_CHECKING
 
 from isaaclab_arena.relations.loss_primitives import (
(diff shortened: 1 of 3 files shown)


isaaclab_arena/assets/dummy_object.py
alexmillane, line 1: I'm wondering whether we should move this file to `isaaclab_arena/tests`. Seems a bit specific to have in the core framework. It's basically "just" for testing the relation solver no?

cvolkcvolk: Mh the examples in `isaaclab_arena_examples/` would then import from `tests/`, which is a bit of a code smell no?
How about still keeping it in the core framework but in `isaaclab_arena/relations`?

cvolkcvolk: Merging the MR as it is. Noted down to follow up on this.

PR #360, opened 2026-01-27
Support multi-env & enable_cameras in multi-task eval

xyao-nv:
## Summary
Evaluation jobs may have different num of envs. Some may require camera sensors and some may not.` Eval_runner ` needs to handle those configurations with persistent app setup.

## Detailed description
- Fix tests with evaluation job status checking
- Activate `enable_camera` if any job requires cameras before Sim App starts
- Allow different number of envs per job

### DIFF
--- a/isaaclab_arena/tests/test_job_manager.py
+++ b/isaaclab_arena/tests/test_job_manager.py
@@ -33,9 +33,8 @@ def test_job_from_dict():
 def test_job_convert_args_dict_to_cli_args_list():
     """Test converting arguments dictionary to CLI args list."""
     # Test basic string arguments
-    args_dict = {"env": "test_env", "object": "box", "num_envs": "4"}
+    args_dict = {"environment": "test_env", "object": "box", "num_envs": "4", "headless": True, "enable_cameras": False}
     args_list = Job.convert_args_dict_to_cli_args_list(args_dict)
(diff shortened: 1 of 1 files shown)


isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -41,6 +42,24 @@ def load_env(arena_env_args: list[str], job_name: str):
...
+
+    Args:
+        eval_jobs_config: Dictionary containing job configurations
+        args_cli: CLI arguments namespace to modify
+    """
+    for job_dict in eval_jobs_config["jobs"]:
+        if "arena_env_args" in job_dict and job_dict["arena_env_args"].get("enable_cameras", False):
+            # will not override the enable_cameras flag if it is already set from cli
+            if not hasattr(args_cli, "enable_cameras") or not args_cli.enable_cameras:
+                args_cli.enable_cameras = True
```
alexmillane: This comment confused me a bit. If `args.enable_cameras` doesn't exist or is `Flase` we set it, effective overriding the command line. If it's already `True` we don't set it to `True` again (which is kind of a no-op anyway). I think we can just remove this comment.

xyao-nv: sorry about that. Updated the comments.

PR #365, opened 2026-01-28
Add user specifiable final subtask success state

peterd-NV:
## Summary
By default, sequential tasks track if a subtask was completed before moving onto the next subtask. At the end of the seq task, the current success state of prior subtasks can be false. This is useful for tasks where different subtask success states depend on the same object which may move.

This PR adds a feature to allow users to specify the precise subtask state that needs to be satisfied at the end of the seq task, allowing users to ensure that some/all prior subtasks are still True during the final success check.

An example of using this is added to `franka_put_and_close_door_task.py`

### DIFF
--- a/isaaclab_arena/assets/object.py
+++ b/isaaclab_arena/assets/object.py
@@ -13,12 +13,12 @@ from isaaclab_tasks.manager_based.manipulation.stack.mdp.franka_stack_events imp
 
 from isaaclab_arena.assets.object_base import ObjectBase, ObjectType
 from isaaclab_arena.assets.object_utils import detect_object_type
-from isaaclab_arena.relations.relations import AtPosition, Relation, RelationBase
+from isaaclab_arena.relations.relations import RelationBase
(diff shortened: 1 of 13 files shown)


isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -152,7 +164,15 @@ def sequential_task_success_func(
...
-        per_env_success = [all(env_successes) for env_successes in env._subtask_success_state]
+        if desired_subtask_success_state:
+            per_env_success = [
+                all([a == b for a, b in zip(env_successes, desired_subtask_success_state)])
+                for env_successes in env._subtask_success_state
+            ]
+        else:
+            per_env_success = [all(env_successes) for env_successes in env._subtask_success_state]
+
+        # per_env_success = [all(env_successes) for env_successes in env._subtask_success_state]
```
xyao-nv: rm

xyao-nv (approved):

PR #366, opened 2026-01-28
Support Objects with rigid bodies not at root

alexmillane:
## Summary
Adds Support Objects with rigid bodies not at root

## Detailed description
- Previously assets which have their rigid bodies not at the root of the object failed if a contact sensor was added.
- Now this is supported
- This is implemented by seaching the USD for the shallowest rigidi body in the object and adding the contact sensor there.
- This is particularly relavent for LightWheel assets.

## Not done
- Support for `ObjectReferences` and `RigidObjectSet`s is still remaining.

### DIFF
--- a/isaaclab_arena/utils/usd/rigid_bodies.py
+++ b/isaaclab_arena/utils/usd/rigid_bodies.py
@@ -75,8 +75,6 @@ def find_shallowest_rigid_body(usd_path: str, relative_to_root: bool = False) ->
             )
         shallowest_rigid_body = shallowest_rigid_bodies[0]
 
-    print("HERE")
-    print(f"shallowest_rigid_body: {shallowest_rigid_body}")
     if relative_to_root:
         assert shallowest_rigid_body[0] == "/", "We expect USD paths to start with a /"
         root_and_rest = shallowest_rigid_body.lstrip("/").split("/", 1)


isaaclab_arena/assets/object_library.py
```diff
@@ -540,3 +540,27 @@ class MediumGear(LibraryObject):
...
+    from lightwheel_sdk.loader import object_loader
+
+    name = "sweet_potato"
+    tags = ["object", "vegetable", "graspable"]
+    file_path, object_name, metadata = object_loader.acquire_by_registry(
+        registry_type="objects", file_name="SweetPotato005", file_type="USD"
+    )
+    usd_path = file_path
+    object_type = ObjectType.RIGID
+    scale = (1.3, 1.3, 1.3)
```
peterd-NV, line 561: I think @xyao-nv scaled this up to 1.5 for better graspability when training the GR00T model. Shall we unify and stick with 1.5 as the default?

xyao-nv: We can change it in later MRs once we show E2E is working

alexmillane: Yea I noticed that when I pulled this branch into the GTC branch and had a merge conflict. I took Xinjie's number there so it should be updated with the merge of the GTC branch :). Thank you for picking this up.

PR #371, opened 2026-02-02
Add Task Sorting and Corresponding Example

lancel-nv:
## Summary

Object sorting is a classic task in robotic manipulation, requiring the robot to pick multiple objects and place them at their corresponding destinations. This PR introduces a `SortMultiObjectTask` as a reusable task component and provides a concrete example environment (`TableTopSortCubesEnvironment`) that demonstrates how to use it.

The arena framework needed a fundamental task type for multi-object sorting that can be used as a building block for more complex manipulation tasks. Many robotic manipulation scenarios require moving multiple objects to their respective target positions (e.g., sorting by color), so having a standardized task definition helps maintain consistency and reusability.

### What has been changed?

**New Task:**
- Added `SortMultiObjectTask` in `isaaclab_arena/tasks/sorting_task.py` — a configurable multi-object sorting task that supports:
  - Multiple pick-up objects list with corresponding destination list
  - Contact sensor-based detection for object-destination contact
  - Configurable force threshold and velocity threshold for success criteria

**New Termination Logic:**
- Added `objects_on_destinations` function in `isaaclab_arena/tasks/terminations.py` that evaluates whether **ALL** objects in the list have reached their respective destinations
- Added `root_height_below_minimum_multi_objects` function for detecting failure conditions when any object falls below minimum height

**New Assets:**
- Added the following assets to the object library:
  - `RedCube`: A red cube block
  - `GreenCube`: A green cube block
  - `RedContainer`: A red container
  - `GreenContainer`: A green container

**New Example Environment:**
- Added `TableTopSortCubesEnvironment` in `isaaclab_arena_environments/sorting_environment.py` — a complete example showing how to configure and use the task with a Franka robot sorting colored cubes into their matching colored containers

**Tests:**
- Added comprehensive tests in `isaaclab_arena/tests/test_sorting_task.py`:
  - Tests initial state (cubes should not be in success state)
  - Tests success state (all cubes moved to their respective target containers)
  - Tests multiple environments in parallel

- **New capability**: Users can now create multi-object sorting tasks with flexible success criteria (contact force + velocity thresholds)
- **Reusable components**: The `SortMultiObjectTask` can be reused for various object sorting and classification scenarios
- **New assets**: RedCube, GreenCube, RedContainer, and GreenContainer are now available in the asset library for other environments
- **Multi-object support**: The framework now supports complex tasks that require coordinating multiple object manipulations

### DIFF
--- a/isaaclab_arena/tasks/terminations.py
+++ b/isaaclab_arena/tasks/terminations.py
@@ -8,6 +8,7 @@ import torch
 
 from isaaclab.assets import RigidObject
 from isaaclab.envs import ManagerBasedRLEnv
+from isaaclab.envs.mdp.terminations import root_height_below_minimum
 from isaaclab.managers import SceneEntityCfg
 from isaaclab.sensors.contact_sensor.contact_sensor import ContactSensor
 
@@ -54,25 +55,16 @@ def objects_on_destinations(
     Returns True only when ALL objects in the list satisfy the destination condition.
     See `object_on_destination` for details on the single-object logic.
     """
-    condition_met = torch.ones((env.num_envs), device=env.device)
+    condition_met = torch.ones((env.num_envs), device=env.device, dtype=torch.bool)
     for object_cfg, contact_sensor_cfg in zip(object_cfg_list, contact_sensor_cfg_list):
-        object: RigidObject = env.scene[object_cfg.name]
-        sensor: ContactSensor = env.scene[contact_sensor_cfg.name]
-
-        # See comments in `object_on_destination` for shape assumptions
-        assert sensor.data.force_matrix_w.shape[2] == 1
-        assert sensor.data.force_matrix_w.shape[1] == 1
-
-        force_matrix_norm = torch.norm(sensor.data.force_matrix_w.clone(), dim=-1).reshape(-1)
-        force_above_threshold = force_matrix_norm > force_threshold
-
-        velocity_w = object.data.root_lin_vel_w
-        velocity_w_norm = torch.norm(velocity_w, dim=-1)
-        velocity_below_threshold = velocity_w_norm < velocity_threshold
-
-        condition_met = torch.logical_and(
-            torch.logical_and(force_above_threshold, velocity_below_threshold), condition_met
+        single_condition = object_on_destination(
+            env=env,
+            object_cfg=object_cfg,
+            contact_sensor_cfg=contact_sensor_cfg,
+            force_threshold=force_threshold,
+            velocity_threshold=velocity_threshold,
         )
+        condition_met = torch.logical_and(condition_met, single_condition)
     return condition_met
 
 
@@ -245,31 +237,18 @@ def goal_pose_task_termination(
     return success
 
 
-def object_above(
-    env: ManagerBasedRLEnv,
-    object_cfg: SceneEntityCfg,
-    maximum_height: float,
-) -> torch.Tensor:
-    """Determine if the object is lifted above a certain height."""
-    object: RigidObject = env.scene[object_cfg.name]
-    return object.data.root_pos_w[:, 2] > maximum_height
-
-
 def root_height_below_minimum_multi_objects(
     env: ManagerBasedRLEnv, minimum_height: float, asset_cfg_list: list[SceneEntityCfg] = [SceneEntityCfg("robot")]
 ) -> torch.Tensor:
-    """Terminate when the asset's root height is below the minimum height.
+    """Terminate when any asset's root height is below the minimum height.
 
     Note:
         This is currently only supported for flat terrains, i.e. the minimum height is in the world frame.
     """
-    # extract the used quantities (to enable type-hinting)
-    outs = []
-    for asset_cfg in asset_cfg_list:
-        asset: RigidObject = env.scene[asset_cfg.name]
-        out = asset.data.root_pos_w[:, 2] < minimum_height
-        outs.append(out)
-
+    outs = [
+        root_height_below_minimum(env=env, minimum_height=minimum_height, asset_cfg=asset_cfg)
+        for asset_cfg in asset_cfg_list
+    ]
     outs_tensor = torch.stack(outs, dim=0)  # [X, N]
     terminated = outs_tensor.any(dim=0)  # [N], bool
     return terminated


isaaclab_arena/tasks/terminations.py
```diff
@@ -42,6 +42,40 @@ def object_on_destination(
...
+    velocity_threshold: float = 0.5,
+) -> torch.Tensor:
+    """Multi-object version of `object_on_destination`.
+
+    Returns True only when ALL objects in the list satisfy the destination condition.
+    See `object_on_destination` for details on the single-object logic.
+    """
+    condition_met = torch.ones((env.num_envs), device=env.device)
+    for object_cfg, contact_sensor_cfg in zip(object_cfg_list, contact_sensor_cfg_list):
+        object: RigidObject = env.scene[object_cfg.name]
```
cvolkcvolk: Could we reuse `object_on_destination()` in here to avoid duplicating its code?

isaaclab_arena/assets/object_library.py
```diff
@@ -564,3 +564,76 @@ def __init__(
...
+@register_asset
+class RedCube(LibraryObject):
+    """
+    A red cube.
+    """
+
+    name = "red_cube"
+    tags = ["object"]
+
+    # usd_path =f"{ISAAC_NUCLEUS_DIR}/Props/Blocks/red_block.usd" # not support, rigid body attribute need to be bind to root xform.
```
cvolkcvolk, line 578: Leftover comment?

isaaclab_arena/assets/object_library.py
```diff
@@ -564,3 +564,76 @@ def __init__(
...
+@register_asset
+class GreenCube(LibraryObject):
+    """
+    A green cube.
+    """
+
+    name = "green_cube"
+    tags = ["object"]
+
+    # usd_path = f"{ISAAC_NUCLEUS_DIR}/Props/Blocks/green_block.usd" # not support, rigid body attribute need to be bind to root xform.
```
cvolkcvolk, line 598: Leftover comment?

isaaclab_arena/tasks/sorting_task.py
```diff
@@ -0,0 +1,129 @@
...
+
+from isaaclab_arena.assets.asset import Asset
+from isaaclab_arena.metrics.metric_base import MetricBase
+from isaaclab_arena.metrics.success_rate import SuccessRateMetric
+from isaaclab_arena.tasks.task_base import TaskBase
+from isaaclab_arena.tasks.terminations import objects_on_destinations, root_height_below_minimum_multi_objects
+from isaaclab_arena.utils.cameras import get_viewer_cfg_look_at_object
+
+
+class SortMultiObjectTask(TaskBase):
```
cvolkcvolk, line 23: I feel like we could use a _Sequential_ task composed of multiple atomic `PickAndPlace` tasks for that behaviour. @peterd-NV what do you think? 

We would have an example how to compose a subtask in  `/isaac_arena/isaaclab_arena_environments/franka_put_and_close_door_environment.py`

isaaclab_arena/tasks/sorting_task.py
```diff
@@ -0,0 +1,129 @@
...
+            func=objects_on_destinations,
+            params={
+                "object_cfg_list": object_cfg_list,
+                "contact_sensor_cfg_list": contact_sensor_cfg_list,
+                "force_threshold": 1.0,
+                "velocity_threshold": 0.1,
+            },
+        )
+        object_dropped = TerminationTermCfg(
+            func=root_height_below_minimum_multi_objects,
```
cvolkcvolk, line 82: If we compose this _sorting_task_ as a sequential task we could reuse the single `mdp_isaac_lab.root_height_below_minimum` of each subtask (`PickAndPlaceTask`) individually

isaaclab_arena/tasks/sorting_task.py
peterd-NV: Might wanna note that if the ordering of how the objects get sorted doesn't matter, then the sequential task might not be the best fit since it assumes the subtasks to be completed in a specific order. For this task where the success depends on `object_on_destination` I think it would still work since any out of order sorted object will eventually have it's success triggered once all objects are placed, but this does go against the idea of a sequential task and mean the intermediate subtask success states won't have meaning.

isaaclab_arena/assets/object_library.py
lancel-nv: This is left here because there is a known bug that rigid body attributes can only bind to the root layer. So I adjusted the original assets from `ISAAC_NUCLEUS_DIR` and uploaded adjusted one to `ISAAC_NUCLEUS_STAGING_DIR`. One the bug is solved, we can use original assets.
Do you think if we should remove this comments?

isaaclab_arena/assets/object_library.py
lancel-nv: Same as above.

isaaclab_arena/tasks/sorting_task.py
lancel-nv: Yes, the order in which we place objects into containers doesn’t matter. So maybe we can use a more intuitive example — for instance, pulling out a drawer, putting something inside, and then closing it.

isaaclab_arena/tasks/terminations.py
lancel-nv: Updated to reuse ready-made functions.

isaaclab_arena/tasks/sorting_task.py
lancel-nv: As we are still discussing whether to use sequential tasks, I’ve temporarily updated the implementation to reuse the existing `mdp_isaac_lab.root_height_below_minimum` function.

isaaclab_arena/assets/object_library.py
alexmillane: Thank you for the description @Nyquist0. I think we should leave the commented out usd_path in. However, what do you think about adding a `TODO` with the contents of your reply above?

isaaclab_arena/tasks/sorting_task.py
```diff
@@ -0,0 +1,129 @@
...
+class SortMultiObjectTask(TaskBase):
+
+    def __init__(
+        self,
+        pick_up_object_list: list[Asset],
+        destination_location_list: list[Asset],
+        background_scene: Asset,
+        episode_length_s: float | None = None,
+    ):
+        super().__init__(episode_length_s=episode_length_s)
```
alexmillane, line 32: suggestion `assert len(pick_up_object_list) == len(distination_location_list)`

isaaclab_arena/tasks/sorting_task.py
```diff
@@ -0,0 +1,129 @@
...
+        assert len(pick_up_object_list) == len(destination_location_list)
+
+        pick_up_object_contact_sensor_list = []
+        for pick_up_object, destination_location in zip(pick_up_object_list, destination_location_list):
+            pick_up_object_contact_sensor_list.append(
+                pick_up_object.get_contact_sensor_cfg(contact_against_prim_paths=[destination_location.get_prim_path()])
+            )
+        self.pick_up_object_contact_sensor_list = pick_up_object_contact_sensor_list
+        self.contact_sensor_name_list = [
+            f"contact_sensor_{i}" for i in range(len(self.pick_up_object_contact_sensor_list))
```
alexmillane, line 46: shall we replace `i` with the name of the object, for example, i.e. `pick_up_object.name`

isaaclab_arena/tasks/sorting_task.py
```diff
@@ -0,0 +1,129 @@
...
+
+@configclass
+class SceneCfg:
+    """
+    Scene configuration for the pick and place task.
+    Note: only support <4 objects. Need to figure out a more flexible method, like __post_init__()
+    """
+
+    contact_sensor_0: ContactSensorCfg = MISSING
+    contact_sensor_1: ContactSensorCfg = MISSING
```
alexmillane, line 120: I would suggest that we solve this non-flexibility in this MR. 

Your issue, if I'm correct is that you'd like to make a `SceneCfg` with a variable number of members. This is made possible through `from isaaclab_arena.utils.configclass import make_configclass`.

For example we use is the `Scene` to make a `SceneCfg` from the variable number of objects in the scene. See [here](https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena/scene/scene.py#L59)

isaaclab_arena_environments/sorting_environment.py
```diff
@@ -0,0 +1,134 @@
...
+
+        if args_cli.teleop_device is not None:
+            teleop_device = self.device_registry.get_device_by_name(args_cli.teleop_device)()
+            # increase sensitivity for teleop device
+            teleop_device.pos_sensitivity = 0.25
+            teleop_device.rot_sensitivity = 0.5
+        else:
+            teleop_device = None
+
+        destination_location1 = self.asset_registry.get_asset_by_name(args_cli.destination[0])()
```
alexmillane, line 65: `destination_location_1`

isaaclab_arena_environments/sorting_environment.py
```diff
@@ -0,0 +1,134 @@
...
+            )
+        )
+
+        scene = Scene(
+            assets=[background, light, pick_up_object1, pick_up_object2, destination_location1, destination_location2]
+        )
+
+        task = SortMultiObjectTask(
+            [pick_up_object1, pick_up_object2], [destination_location1, destination_location2], background
+        )
```
alexmillane, line 103: Very cool! Thanks for doing this!

isaaclab_arena_environments/sorting_environment.py
```diff
@@ -0,0 +1,134 @@
...
+            scene=scene,
+            task=task,
+            teleop_device=teleop_device,
+        )
+        return isaaclab_arena_environment
+
+    @staticmethod
+    def add_cli_args(parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--object",
```
alexmillane, line 120: `--objects`

isaaclab_arena_environments/sorting_environment.py
```diff
@@ -0,0 +1,134 @@
...
+            nargs="*",
+            default=["red_cube", "green_cube"],
+            help="object list (example: --object red_cube green_cube)",
+        )
+        parser.add_argument(
+            "--destination",
```
alexmillane, line 126: `--destinations`

alexmillane (approved):
Thank you for doing this. This is a key task to have in the framework! I have a couple of small comments.

isaaclab_arena/tasks/sorting_task.py
lancel-nv: The `assert len(pick_up_object_list) == len(distination_location_list)` is on the line 37. Are you recommending to move it here?

isaaclab_arena_environments/sorting_environment.py
lancel-nv: May I ask if you mean that the naming is not standardized?
I agree on that it would be better to use a list, such as destination_location_list. However, I haven’t yet figured out how to automatically assign the correct position for each object—such as the container—so that they don’t overlap and the embodiment can reach all of them.

isaaclab_arena/tasks/sorting_task.py
lancel-nv: Thank you very much for this suggestion. It helps to solve the block well !!

isaaclab_arena_environments/sorting_environment.py
lancel-nv: Never mind. I got you...

PR #377, opened 2026-02-03
RandomAroundSolution marker for pose randomization around solver solutions

cvolkcvolk:
## Summary

Introduces `RandomAroundSolution`, a new marker that allows objects to be randomized around the solver's computed position. 
When the `ObjectPlacer` applies positions, objects with this marker receive a `PoseRange` instead of a fixed `Pose`, enabling position randomization at each environment reset.

This feature is introduced for the DLI demo to demonstrate domain randomization with the ObjectPlacer. This is more of a workaround for now to get randomization into the env_cfg without running the Solver again.

### DIFF
--- a/isaaclab_arena/relations/object_placer.py
+++ b/isaaclab_arena/relations/object_placer.py
@@ -13,7 +13,7 @@ from isaaclab_arena.relations.placement_result import PlacementResult
 from isaaclab_arena.relations.relation_solver import RelationSolver
 from isaaclab_arena.relations.relations import RandomAroundSolution, get_anchor_objects
 from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox, get_random_pose_within_bounding_box
-from isaaclab_arena.utils.pose import Pose, PoseRange
+from isaaclab_arena.utils.pose import Pose
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


isaaclab_arena/relations/object_placer.py
```diff
@@ -206,11 +206,57 @@ def _apply_positions(
...
+                        -random_marker.roll_half_rad,
+                        -random_marker.pitch_half_rad,
+                        -random_marker.yaw_half_rad,
+                    ),
+                    rpy_max=(
+                        random_marker.roll_half_rad,
+                        random_marker.pitch_half_rad,
+                        random_marker.yaw_half_rad,
+                    ),
+                )
```
alexmillane: I wonder if we could move this out of the solver somehow. Maybe the random marker could have a function for adding `def to_pose_range(pos: torch.tensor) -> PoseRange`? Just to keep the code terse in the `ObjectPlacer`?

alexmillane (approved):
Really clean. Thanks for this!

isaaclab_arena/relations/object_placer.py
cvolkcvolk: Good idea!

PR #381, opened 2026-02-04
Support variable contact sensor for Sorting task. 

lancel-nv:
## Summary
Support variable number of objects in sorting task

The original SortMultiObjectTask used a hardcoded SceneCfg with fixed fields (contact_sensor_0, contact_sensor_1, etc.), limiting support to fewer than 4 objects.

### What has been changed?

- Use make_configclass to dynamically generate SceneCfg, supporting variable number of contact sensors
- Improve contact sensor naming: from index-based (contact_sensor_0) to object name-based (contact_sensor_{object.name})
- Rename CLI arguments: --object → --objects, --destination → --destinations
- Add TODO comments for RedCube/GreenCube documenting a known bug where rigid body attributes must bind to root layer

SortMultiObjectTask now supports an arbitrary number of objects without hardcoded limitations.

### DIFF
--- a/isaaclab_arena_environments/sorting_environment.py
+++ b/isaaclab_arena_environments/sorting_environment.py
@@ -22,7 +22,9 @@ class TableTopSortCubesEnvironment(ExampleEnvironmentBase):
         from isaaclab_arena.tasks.sorting_task import SortMultiObjectTask
         from isaaclab_arena.utils.pose import Pose
 
-        assert len(args_cli.destinations) == len(args_cli.objects)
+        assert (
+            len(args_cli.destinations) == len(args_cli.objects) == 2
+        ), "Only 2 objects and 2 destinations are supported in this environment."
 
         # Add the asset registry from the arena migration package
         light = self.asset_registry.get_asset_by_name("light")()


isaaclab_arena_environments/sorting_environment.py
```diff
@@ -22,7 +22,7 @@ def get_env(self, args_cli: argparse.Namespace):
         from isaaclab_arena.tasks.sorting_task import SortMultiObjectTask
         from isaaclab_arena.utils.pose import Pose
 
-        assert len(args_cli.destination) == len(args_cli.object)
+        assert len(args_cli.destinations) == len(args_cli.objects)
```
alexmillane: The code below only works for 2 objects. Let's add an additional assert that `len(args_cli.destinations) == 2`

alexmillane (approved):
Thanks for cleaning this up! Looks great!

isaaclab_arena_environments/sorting_environment.py
lancel-nv: Sure. Do you think is there a way that we could support multiple objects and destinations more than two because I think the current blocker is to automatically arrange the objects position without overlap.

lancel-nv: Anyway, let's go ahead and merge this for now. We can explore a better solution in a follow-up MR if needed..

PR #385, opened 2026-02-04
Add GR1 DLI env and workflow

peterd-NV:
Adds the GR1 put and close door env/task used for the DLI course

### DIFF
--- a/docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
+++ b/docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
@@ -74,7 +74,7 @@ Test the policy in a single environment with visualization via the GUI run:
 
 .. code-block:: bash
 
-   python isaaclab_arena/examples/policy_runner.py \
+   python isaaclab_arena/evaluation/policy_runner.py \
      --policy_type isaaclab_arena_gr00t.policy.gr00t_closedloop_policy.Gr00tClosedloopPolicy \
      --policy_config_yaml_path isaaclab_arena_gr00t/policy/config/g1_locomanip_gr00t_closedloop_config.yaml \
      --num_steps 1200 \
@@ -102,7 +102,7 @@ Test the policy in 5 parallel environments with visualization via the GUI run:
 
 .. code-block:: bash
 
-   python isaaclab_arena/examples/policy_runner.py \
+   python isaaclab_arena/evaluation/policy_runner.py \
      --policy_type isaaclab_arena_gr00t.policy.gr00t_closedloop_policy.Gr00tClosedloopPolicy \
      --policy_config_yaml_path isaaclab_arena_gr00t/policy/config/g1_locomanip_gr00t_closedloop_config.yaml \
      --num_steps 1200 \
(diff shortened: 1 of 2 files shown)


isaaclab_arena_environments/gr1_put_and_close_door_environment.py
```diff
@@ -0,0 +1,228 @@
...
+                self.datagen_config.generation_keep_failed = False
+                self.datagen_config.generation_num_trials = 100
+                self.datagen_config.generation_select_src_per_subtask = False
+                self.datagen_config.generation_select_src_per_arm = False
+                self.datagen_config.generation_relative = False
+                self.datagen_config.generation_joint_pos = False
+                self.datagen_config.generation_transform_first_robot_pose = False
+                self.datagen_config.generation_interpolate_from_last_target_pose = True
+                self.datagen_config.max_num_failures = 25
+                self.datagen_config.seed = 1
```
alexmillane: @peterd-NV do we still need all of this? Do the defaults that we added not work for this task?

I'm wonder if it's possible to simplify?

isaaclab_arena_gr00t/lerobot/config/gr1_manip_ranch_bottle_config.yaml
alexmillane, line 1: @xyao-nv. Seems like we shouldn't merge this into main?

Is there some way we can get around having to create a new file each time we change the manipulated object? Could we specify some parameters via the CLI?

We can also leave this for later given the deadline.

isaaclab_arena_environments/gr1_put_and_close_door_environment.py
```diff
@@ -0,0 +1,228 @@
...
+        # Create refrigerator shelf reference (destination for pick and place)
+        refrigerator_shelf = ObjectReference(
+            name="refrigerator_shelf",
+            prim_path="{ENV_REGEX_NS}/lightwheel_robocasa_kitchen/fridge_main_group/Refrigerator034",
+            parent_asset=kitchen_background,
+        )
+
+        # Consider changing to other values for different objects, below is for ranch dressing bottle
+        z_position = 1.0082
+        yaw = 130.0
```
cvolkcvolk: This should be in radian, eg `yaw_rad`. 
The corresponding degree around z read from the IsaacSim Gui is `-111.55` to end up in the desired position you are placing the jug here. I will update to 
`yaw_rad = math.radians(-111.55)`

isaaclab_arena_environments/gr1_put_and_close_door_environment.py
peterd-NV: Whoops I missed this, thanks for catching. Yes, this can be simplified

isaaclab_arena_gr00t/lerobot/config/gr1_manip_ranch_bottle_config.yaml
xyao-nv: Let's leave this for the future. Doable but I don't want to do it now given other higher priority stuffs

PR #393, opened 2026-02-06
ObjectPlacement supports RigidObjectSets

cvolkcvolk:
## Summary

`ObjectPlacement` supports `RigidObjectSets`.

Upate the `put_item_in_fridge_and_close_door` example to use the ObjectPlacement API.

The assert to run object_sets is left in until a fix for the USD paths is in.

To run the ObjectSet in this example as a workaround:
- Use `NoTask()`
- Use the same USD asset:
` python isaaclab_arena/evaluation/policy_runner.py --num_envs 2 --policy_type zero_action put_item_in_fridge_and_close_door --object_set jug jug`

### DIFF
--- a/isaaclab_arena/assets/object_set.py
+++ b/isaaclab_arena/assets/object_set.py
@@ -11,7 +11,7 @@ from isaaclab_arena.assets.object import Object
 from isaaclab_arena.assets.object_base import ObjectBase, ObjectType
 from isaaclab_arena.assets.object_utils import detect_object_type
 from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox
-from isaaclab_arena.utils.pose import Pose, PoseRange
+from isaaclab_arena.utils.pose import Pose
 
 
 class RigidObjectSet(Object):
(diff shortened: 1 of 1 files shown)


isaaclab_arena/assets/object_set.py
```diff
@@ -60,6 +62,16 @@ def __init__(
             **kwargs,
         )
 
+    def set_initial_pose(self, pose: Pose | PoseRange) -> None:
+        """Set the initial pose of the object set."""
+        self.initial_pose = pose
+        self.object_cfg = self._add_initial_pose_to_cfg(self.object_cfg)
+        self.event_cfg = self._update_initial_pose_event_cfg(self.event_cfg)
```
alexmillane: In this different that the implementation in the parent class `Object`?

If not, can we just use the parent class implementation? If it is different, it might be worth adding a comment how (I can't tell from looking at it).

isaaclab_arena/assets/object_set.py
```diff
@@ -60,6 +62,16 @@ def __init__(
...
+
+    def get_bounding_box(self) -> AxisAlignedBoundingBox:
+        """Get the bounding box of the object set."""
+        return self.objects[0].get_bounding_box()
```
alexmillane: I feel like we should take the bounding box which is greatest in the z-dimension. That way we don't get items spawning inside their support surfaces? Of course, there can/will be other problems with this approach, but I think that "biggest in z" is the best we can do right now.

Wdyt?

alexmillane (approved):
LGTM. I have one comment.

isaaclab_arena/assets/object_set.py
cvolkcvolk: Looks redundant to me. Removed.

isaaclab_arena/assets/object_set.py
cvolkcvolk: Updated

PR #394, opened 2026-02-07
[FEATURE] remote policy with server/client

kanghui0204:
## Summary
old MR is [MR 310](https://github.com/isaac-sim/IsaacLab-Arena/pull/310)

Hi Team,

I added a remote server/client mechanism for IsaacLab-Arena that allows the IsaacLab environment and the policy model environment to run separately. Alex and I wrote a design document([design doc](https://docs.google.com/document/d/1uNcYswolxPD4PpHgvCapYuqwZ2QhzSnlJB9KRMFMQbQ/edit?tab=t.0)) for this feature, which includes our discussions about the design. Below is an overview of this PR.

### Functionality
Previously, the IsaacLab-Arena pipeline ran entirely in a single process:

**IsaacLab env** → **Env policy class** (e.g., Gr00tClosedloopPolicy) → **local policy model** (e.g., Gr00tPolicy), all in one environment.

This PR adds remote policy/server–client support and decouples the env policy class from the local policy model. The new pipeline becomes:

**Client**: **IsaacLab env** → **Env policy class** (e.g., ActionChunkingClientPolicy)

**Server**: **VLA/VLN policy model** (e.g., Gr00tRemoteServerSidePolicy)

The client and server exchange observations and actions via sockets.

The server/client implementation lives in `isaaclab_arena/remote_policy` inside IsaacLab-Arena. Users can copy this directory and import it on the server side directly, without needing to install it as a separate package.

### DIFF
--- a/docs/pages/quickstart/docker_containers.rst
+++ b/docs/pages/quickstart/docker_containers.rst
@@ -51,8 +51,8 @@ These directories are configurable through argument to the run docker script.
 For a full list of arguments see the ``run_docker.sh`` script at
 ``isaac_arena/docker/run_docker.sh``.
 
-Sever-Client Mode With GR00T (Optional)
--------------------------------------
+Server-Client Mode With GR00T (Optional)
+-----------------------------------------
 
 We also support running the simulation and the policy model in separate containers connected via a lightweight RPC protocol.
 This is useful when the policy environment has different dependencies or needs to run on a different machine.


kanghui0204:
Hi @alexmillane @xyao-nv ,could you please review the latest code? I’ve tested it repeatedly and also added some unit tests and documentation. If everything looks good to you, I think we can merge it now, because the Chinese New Year holidays are coming up and if we wait it might have to be postponed until March.

docker/setup/install_gr00t_deps.sh
```diff
@@ -1,42 +1,89 @@
...
+$PYTHON_CMD -m pip install --no-build-isolation --use-pep517 flash-attn==2.7.4.post1
+
+# Install Isaac-GR00T package itself without pulling its dependencies.
+# GR00T's pyproject.toml pins python=3.10, which conflicts with Isaac Sim's python 3.11,
+# so we ignore 'requires-python' and install dependencies manually.
+echo "Installing Isaac-GR00T package (no deps)..."
+$PYTHON_CMD -m pip install --no-deps --ignore-requires-python \
+    -e ${WORKDIR}/submodules/Isaac-GR00T/
+
+# Install GR00T main dependencies (part 1, with build isolation)
```
alexmillane: with build isolation -> without build isolation

docker/setup/install_gr00t_deps.sh
```diff
@@ -1,42 +1,89 @@
...
+##########################
+# CUDA environment setup
+##########################
+if [[ "$USE_SERVER_ENV" -eq 1 ]]; then
+  # Server environment: use system CUDA if available
+  if [ -d "/usr/local/cuda" ]; then
+      export CUDA_HOME=${CUDA_HOME:-/usr/local/cuda}
+      export PATH=${CUDA_HOME}/bin:${PATH}
+      export LD_LIBRARY_PATH=${CUDA_HOME}/lib64:${LD_LIBRARY_PATH:-}
+  fi
```
alexmillane: Given that this script is used by our docker, is there a situation where we end up without CUDA? In other words, is there a reason we have this `if` condition?

docker/setup/install_gr00t_deps.sh
```diff
@@ -1,42 +1,89 @@
...
+  # Isaac Sim Docker: fixed CUDA 12.8 toolchain
+  export CUDA_HOME=/usr/local/cuda-12.8
+  export PATH=/usr/local/cuda-12.8/bin:${PATH}
+  export LD_LIBRARY_PATH=/usr/local/cuda-12.8/lib64:${LD_LIBRARY_PATH:-}
+  export TORCH_CUDA_ARCH_LIST=8.0+PTX
+  echo "[ISAACSIM] CUDA_HOME=$CUDA_HOME"
+  echo "[ISAACSIM] PATH=$PATH"
+  echo "[ISAACSIM] LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
+  echo "[ISAACSIM] TORCH_CUDA_ARCH_LIST=$TORCH_CUDA_ARCH_LIST"
+fi
```
alexmillane: I believe this whole conditional can be avoided by setting

```
export CUDA_HOME=/usr/local/cuda
export PATH=${CUDA_HOME}/bin:${PATH}
export LD_LIBRARY_PATH=${CUDA_HOME}/lib64:${LD_LIBRARY_PATH:-}
export TORCH_CUDA_ARCH_LIST=8.0+PTX
```

both dockers, pytorch and isaac_arena_gr00t, have cuda installed at `/usr/local/cuda`. In pytorch this folder is a symlink to `/usr/local/cuda-12.5`, in the `isaac_arena_gr00t` the folder is a symlink to `/usr/local/cuda-12.8`

docs/pages/concepts/concept_remote_policies_design.rst
```diff
@@ -0,0 +1,241 @@
...
+
+- which observation entries are exchanged and how they are structured;
+- how actions produced by the server should be interpreted on the client
+  side (for example, one action per step, or sequences of actions),
+  without prescribing a specific model or task.
+
+In practice, you implement a ``ServerSidePolicy`` in the remote
+environment and a matching ``ClientSidePolicy`` inside Isaac Lab Arena.
+As long as they agree on an ``ActionProtocol``, the environments and
+evaluation scripts can remain unchanged.
```
alexmillane: I'm wondering if we can reword this. This makes it sound like we need both a `ServerSidePolicy` and a `ClientSidePolicy` per model. If I'm not mistake, the case is much better than that. We need a `ServerSidePolicy` per model, however the `ClientSidePolicy` is policy-independent. For example the `ActionChunkingClientSidePolicy`, in my opinion, is likely to work for many existing VLAs. In the case of those compatible policies, no further implementation is needed on the client side.

docs/pages/concepts/concept_remote_policies_design.rst
alexmillane, line 1: Thanks for adding such clear and comprehensive docs!

docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
```diff
@@ -140,3 +140,51 @@ and the number of episodes is more than the single environment evaluation becaus
...
+     --host 127.0.0.1 \
+     --port 5555 \
+     --policy_type isaaclab_arena_gr00t.policy.gr00t_remote_policy.Gr00tRemoteServerSidePolicy \
+     --policy_config_yaml_path isaaclab_arena_gr00t/policy/config/g1_locomanip_gr00t_closedloop_config.yaml \
+     --policy_device cuda
+
+If the models directory was already configured earlier (for example via
+the ``MODELS_DIR`` environment variable), the ``-m`` flag can be
+omitted. Otherwise, use ``-m`` to point to the directory that contains
+the GR00T model checkpoint.
```
alexmillane: You can ommit this paragraph. We're assuming all over this tutorial that the relavent environment variables have been set.

docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
```diff
@@ -141,3 +141,51 @@ than the single environment evaluation because of the parallel evaluation.
...
+   bash docker/run_gr00t_server.sh \
+     --host 127.0.0.1 \
+     --port 5555 \
+     --policy_type isaaclab_arena_gr00t.policy.gr00t_remote_policy.Gr00tRemoteServerSidePolicy \
+     --policy_config_yaml_path isaaclab_arena_gr00t/policy/config/gr1_manip_gr00t_closedloop_config.yaml
+
+If the models directory was already configured in previous steps (for
+example via the ``MODELS_DIR`` environment variable), the ``-m`` flag
+can be omitted. Otherwise, use ``-m`` to point to the directory that
+contains the GR00T model files on the host.
```
alexmillane: As above, you can ommit this.

docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
```diff
@@ -140,3 +140,51 @@ and the number of episodes is more than the single environment evaluation becaus
...
+     --enable_cameras \
+     --device cpu \
+     --remote_kill_on_exit \
+     galileo_g1_locomanip_pick_and_place \
+     --object brown_box \
+     --embodiment g1_wbc_joint
+
+In this configuration, the environment and evaluation logic run inside
+Isaac Lab Arena, while GR00T inference runs in the separate server
+process, connected through the remote policy interface.
```
alexmillane: Very cool!!

Out of interest, I presume that this policy(client) can be run without the gr00t docker?

docs/pages/concepts/concept_remote_policies_design.rst
alexmillane: What do you think about adding a diagram of the design. I think that this would go a long way to explaining the structure.

isaaclab_arena/policy/client_side_policy.py
```diff
@@ -0,0 +1,203 @@
...
+
+    Subclasses:
+      - Must implement get_action().
+    """
+
+    def __init__(self, config: Any, remote_config: RemotePolicyConfig, protocol_cls: type[ActionProtocol]) -> None:
+        super().__init__(config=config)
+
+        if protocol_cls is None:
+            raise ValueError("protocol_cls is required.")
```
alexmillane: You can ommit this check. The constructor requires it to be passed, so the check that the user hasn't passed `None` is superfluous.

isaaclab_arena/remote_policy/action_protocol.py
```diff
@@ -0,0 +1,86 @@
...
+
+@dataclass
+class ChunkingActionProtocol(ActionProtocol):
+    """ActionProtocol for CHUNK mode."""
+
+    MODE: ClassVar[ActionMode] = ActionMode.CHUNK
+
+    # Mode-specific field.
+    action_chunk_length: int = 0
+    action_horizon: int = 0
```
alexmillane: Could we describe, in a comment, the difference between these two variables? They're quite similar. I'm not sure I understand the difference 😅

alexmillane: Could we omit the defaults (` = 0`) to ensure that a user passes values?

isaaclab_arena_gr00t/policy/gr00t_remote_policy.py
alexmillane, line 1: What is the difference here between `gr00t_remote_policy` and `gr00t_closed_loop_policy`.

Perhaps niavely I wouldn't expect any difference in the gr00t policy if it's running on a server or running locally.

What are the differences?

docker/setup/install_gr00t_deps.sh
kanghui0204: done

docker/setup/install_gr00t_deps.sh
kanghui0204: remove

docker/setup/install_gr00t_deps.sh
kanghui0204: done

docs/pages/concepts/concept_remote_policies_design.rst
kanghui0204: I change the doc to 

"In this configuration, the environment and evaluation logic run inside
Isaac Lab Arena, while GR00T inference runs in the separate server
process, connected through the remote policy interface. The client-side
policy only depends on Isaac Lab Arena; it does not require GR00T code
or GR00T-specific Python dependencies to be installed in the Arena
container."

How do you think

docs/pages/concepts/concept_remote_policies_design.rst
kanghui0204: add a diagram

docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
kanghui0204: done

docs/pages/example_workflows/static_manipulation/step_5_evaluation.rst
kanghui0204: done

docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
kanghui0204: yes , without gr00t docker , you can build docker with `bash docker/run_docker.sh` without -g , and it can run。

isaaclab_arena/policy/client_side_policy.py
kanghui0204: done

isaaclab_arena/remote_policy/action_protocol.py
kanghui0204: My understanding is that `action_horizon` is the length of the action
sequence produced by the model for one forward pass, while
`action_chunk_length` is the length that the client-side post-processing
actually consumes from each chunk at a time. In the current GN1.6
configuration these two values are the same, but GN1.6 has already
split them conceptually and we expect future use cases with asynchronous
execution where the model’s planning horizon and the client’s chunk size
may differ. I’ve updated the comments accordingly and removed the `= 0`
defaults so both fields have to be provided explicitly.

isaaclab_arena_gr00t/policy/gr00t_remote_policy.py
kanghui0204: Functionally, `Gr00tClosedLoopPolicy` and the combination
`Gr00tRemoteServerSidePolicy` + `ActionChunkingClientSidePolicy` are
equivalent:

docs/pages/quickstart/docker_containers.rst
```diff
@@ -50,3 +50,52 @@ These directories are configurable through argument to the run docker script.
...
+Remote policies and GR00T
+-------------------------
+
+Remote policy workflows (for example the GR1 Open Microwave Door Task)
+are split into two containers:
+
+- The **Base** Isaac Lab Arena container, started via
+  ``docker/run_docker.sh``. This container does not need to install
+  GR00T when you run policies in remote mode.
+- A separate **GR00T policy server** container, started via
```
xyao-nv: Mark it as optional. Add gr000t with docker option.

docker/setup/install_gr00t_deps.sh
```diff
@@ -1,42 +1,76 @@
...
+##########################
+# Python dependencies
+##########################
+
+# Note:
+# - Torch 2.7.0 is pre-installed inside Isaac Sim, so we do NOT install torch here.
+# - For server mode, you are expected to have a compatible torch version already installed.
+
+echo "Installing flash-attn 2.7.4.post1..."
+$PYTHON_CMD -m pip install --no-build-isolation --use-pep517 flash-attn==2.7.4.post1
```
xyao-nv, line 55: only $PYTHON_CMD is changed

isaaclab_arena_gr00t/policy/gr00t_core.py
```diff
@@ -0,0 +1,207 @@
...
+    )
+
+
+# --------------------------------------------------------------------------- #
+# Core SSOT logic (numpy-based, modality-config driven)
+# --------------------------------------------------------------------------- #
+
+
+def build_gr00t_policy_inputs_np(
+    rgb_np: np.ndarray,  # (N, H, W, C)
```
xyao-nv: I dont like this. Multi-rgb are expected

xyao-nv: I don't like all those standalone func calls. Need to think

isaaclab_arena_gr00t/policy/gr00t_remote_policy.py
```diff
@@ -0,0 +1,212 @@
...
+        self._task_description = task_description
+        return {"status": "ok"}
+
+    def get_action(
+        self, observation: dict[str, Any], options: dict[str, Any] | None = None
+    ) -> tuple[dict[str, Any], dict[str, Any]]:
+        camera_name = self.policy_config.pov_cam_name_sim
+
+        # 1) Shared numpy-based preprocessing
+        policy_observations = self._build_policy_observations(observation, camera_name)
```
xyao-nv: client-server is np, one-docker is torch. Need to do a explicit data convert func

docker/setup/install_gr00t_deps.sh
kanghui0204: Thanks for the review! You're right that $PYTHON_CMD is the primary difference for the Python install commands. However the --server flag also controls a few other things beyond that:
- $SUDO: The server environment may not be running as root, so we need sudo for apt-get. Inside the Isaac Sim Docker container we're always root, so sudo is not needed.
- Environment finalization (lines 115-124): The torchrun PATH fix and typing_extensions cleanup are Isaac Sim-specific and should be skipped in server mode.
 The CUDA setup has already been simplified per Alex's suggestion — both environments now use /usr/local/cuda (a symlink), removing the previous conditional block.

isaaclab_arena_gr00t/policy/gr00t_core.py
kanghui0204: Thanks for the feedback! I actually think the current standalone-function approach works better than a class-based one for our use case. The reason is that the remote policy and the local policy each use a different subset of these functions — if we bundle them into a single class, that class would end up being partially used by each side, which feels awkward and violates the single-responsibility principle.

Standalone functions are also stateless and easier to compose — each caller picks exactly the functions it needs without carrying unused baggage.

That said, I'm open to revisiting this in a follow-up PR. In particular, once we have a clearer picture of whether the local (single-container) path is still needed long-term, we can reconsider the abstraction. Happy to discuss further!

isaaclab_arena_gr00t/policy/gr00t_remote_policy.py
kanghui0204: I have some changes for this

docs/pages/quickstart/docker_containers.rst
kanghui0204: I have some changes for this

isaaclab_arena_gr00t/policy/gr00t_core.py
xyao-nv: Sorry for leaving my self review todos on it. Local path is still needed for v0.2 release at least and I'm assuming we will having both options available for some time.

xyao-nv:
Changes I applied so far,
1. Refactored policy core and action chunking related funcs such that they consume the same set of base funcs.
2. Bug fixes to inference pipeline regarding chunking reset logics, and task description setting in remote.
3. Verified locally for our 3 examples, achieve similar metrics between remote & local policy inference.
4. Refactored obs/action packing logics such that data conversion between numpy/torch is modular, and reusable by others.

In general, it allows remote & local has a single source of truth for GR00T policy inference related modules, easy for maintenance and bug fixes.

@kanghui0204 Up to this point, Im done code refactoring and once CI passes, I'll merge it. Thanks for your contribution!
Also thanks to @alexmillane for offering great reviews!

PR #398, opened 2026-02-09
Add configurable object scale for library objects (Rigid Objs)

xyao-nv:
## Summary
Allow object scale to be set when instantiating library objects so users can override default scales without changing asset code. Only add it to Rigid Objs.

## Detailed description
- Users may need configure different scales than what's provided asset library (need it when building a droid multi-objs PnP [example](https://github.com/isaac-sim/IsaacLab-Arena/pull/395)). 
- Add optional` scale: tuple[float, float, float] | None = None` to `LibraryObject.__init__`. If provided, it overrides the class default; otherwise the class scale is used.
- Add and forward the scale parameter in all subclasses that define `__init__`

### DIFF
--- a/isaaclab_arena_environments/kitchen_pick_and_place_environment.py
+++ b/isaaclab_arena_environments/kitchen_pick_and_place_environment.py
@@ -46,7 +46,7 @@ class KitchenPickAndPlaceEnvironment(ExampleEnvironmentBase):
             objects = [self.asset_registry.get_asset_by_name(obj)() for obj in args_cli.object_set]
             pick_up_object = RigidObjectSet(name="object_set", objects=objects)
         else:
-            pick_up_object = self.asset_registry.get_asset_by_name(args_cli.object)(scale=args_cli.object_scale)
+            pick_up_object = self.asset_registry.get_asset_by_name(args_cli.object)()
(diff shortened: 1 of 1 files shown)


isaaclab_arena_environments/kitchen_pick_and_place_environment.py
```diff
@@ -88,6 +88,17 @@ def add_cli_args(parser: argparse.ArgumentParser) -> None:
...
+            "--object_scale",
+            nargs=3,
+            type=float,
+            default=None,
+            metavar=("X", "Y", "Z"),
+            help=(
+                "Scale (x, y, z) for the pick-up object(s). Example: --object_scale 1.0 1.0 1.0. If not set, uses the"
+                " object's default scale."
+            ),
+        )
```
alexmillane: Do we need this as an argument? Seems like something users wont generally use? Do you forsee this being used?

alexmillane (approved):
LGTM

isaaclab_arena_environments/kitchen_pick_and_place_environment.py
xyao-nv: Yeah I was a bit debating when adding it. I agree the chance of using it thru cli args is low. removed.

PR #400, opened 2026-02-09
Fix bounding box rotation in world frame

cvolkcvolk:
## Summary

Rotated background (`KitchenBackground` & the `kitchen_pick_and_place` environment) work with the ObjectPlacement's `OnRelation`. 

`Object.get_world_bounding_box()` only translated bounding boxes from local to world frame — rotations were not accounted for:
When a background asset with a non-identity rotation was used (e.g., the kitchen background with a -90° Z rotation), `ObjectReference` and `Object` bounding boxes were incorrect in world coordinates. 
This caused the relation solver to compute wrong valid placement ranges, leading to objects being placed at incorrect positions. For example, AtPosition(x=0.4, y=0.0) on the kitchen counter would place the object at the wrong location because the On constraint's valid Y range was computed from the unrotated (narrow) local Y extent instead of the rotated (wide) local X extent.

`Object.get_world_bounding_box()` now additionally rotates based on the initial pose. As we are dealing with axis-aligned bounding boxes, we only allow for rotations of 90-degree multiples (for an AABB to still remain valid).

### DIFF
--- a/isaaclab_arena/tests/test_relation_loss_strategies.py
+++ b/isaaclab_arena/tests/test_relation_loss_strategies.py
@@ -43,14 +43,12 @@ def test_on_loss_strategy_zero_loss_when_perfectly_placed():
     relation = On(table, clearance_m=0.01)
     strategy = OnLossStrategy(slope=10.0)
 
-    # Parent at origin
-    parent_pos = torch.tensor([0.0, 0.0, 0.0])
     # Child centered on table, at correct Z height
     # Table top = 0.1, clearance = 0.01, box bottom = 0.0 (relative)
     # So child_z should be 0.11 for box bottom to be at 0.11
     child_pos = torch.tensor([0.4, 0.4, 0.11])
 
-    loss = strategy.compute_loss(relation, child_pos, parent_pos, box.bounding_box, table.bounding_box)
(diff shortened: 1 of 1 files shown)


isaaclab_arena_environments/kitchen_pick_and_place_environment.py
```diff
@@ -47,7 +52,8 @@ def get_env(self, args_cli: argparse.Namespace):  # -> IsaacLabArenaEnvironment:
             pick_up_object = RigidObjectSet(name="object_set", objects=objects)
         else:
             pick_up_object = self.asset_registry.get_asset_by_name(args_cli.object)()
-        pick_up_object.set_initial_pose(pick_up_object_pose)
+        pick_up_object.add_relation(On(table_top_reference, clearance_m=0.02))
+        pick_up_object.add_relation(AtPosition(x=0.4, y=0.0))
```
alexmillane, line 56: Once we have embodiment support in the object placer, could we replace this with a contraint to be in front of the robot?

cvolkcvolk: Yes, that's exactly the plan :-)

PR #409, opened 2026-02-11
Add embodiment support to the relation solver via Placeable protocol

cvolkcvolk:
## Summary

Based on https://github.com/isaac-sim/IsaacLab-Arena/pull/414

Enables embodiments (robots) to participate in the relation solver, so a robot's position can be determined by spatial relations like NextTo(table) instead of requiring hardcoded poses.

Introduces a `Placeable` interface of what the solver expects: Any class implementing the required methods (add_relation, get_relations, get_bounding_box, get_world_bounding_box, set_initial_pose, get_initial_pose) satisfies the protocol.

Update the `kitchen_pick_and_place` to use relations to place the embodiment and the respective object.

### DIFF
--- a/isaaclab_arena/tests/test_robot_initial_position.py
+++ b/isaaclab_arena/tests/test_robot_initial_position.py
@@ -61,11 +61,12 @@ def _test_robot_initial_position(simulation_app):
         print(f"Robot position error: {robot_position_error}")
         assert robot_position_error < INITIAL_POSITION_EPS, "Robot ended up at the wrong position."
 
-        # Check the stand ended up at the correct position.
-        stand_position = env.scene["stand"].get_world_poses()[0].cpu().numpy()
-        stand_position_error = np.linalg.norm(stand_position - np.array(robot_init_position))
-        print(f"Stand position error: {stand_position_error}")
-        assert stand_position_error < INITIAL_POSITION_EPS, "Stand ended up at the wrong position."
+        # Check the stand ended up at the correct position (only if the embodiment has a separate stand entity).
+        if "stand" in env.scene.keys():
+            stand_position = env.scene["stand"].get_world_poses()[0].cpu().numpy()
+            stand_position_error = np.linalg.norm(stand_position - np.array(robot_init_position))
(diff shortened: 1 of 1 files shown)


isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -46,12 +52,104 @@ def __init__(
...
+        if self.scene_config is None:
+            return None
+        robot_cfg = getattr(self.scene_config, "robot", None)
+        if robot_cfg is None:
+            return None
+        spawn_cfg = getattr(robot_cfg, "spawn", None)
+        if spawn_cfg is None:
+            return None
+        usd_path = getattr(spawn_cfg, "usd_path", None)
+        return usd_path
```
alexmillane: The fallbacks here make this function kinda large. Do we have embodiments in Arena where these conditions are not fulfilled? Can we handle the situation where we request the robot to be placed via relations but it doesn't have a USD path? I.e. can we actually handle the returned `None` gracefully? I would just assert/make the access.

isaaclab_arena_environments/kitchen_pick_and_place_environment.py
```diff
@@ -53,7 +61,7 @@ def get_env(self, args_cli: argparse.Namespace):  # -> IsaacLabArenaEnvironment:
         else:
             pick_up_object = self.asset_registry.get_asset_by_name(args_cli.object)()
         pick_up_object.add_relation(On(table_top_reference, clearance_m=0.02))
-        pick_up_object.add_relation(AtPosition(x=0.4, y=0.0))
+        pick_up_object.add_relation(NextTo(embodiment, side=Side.POSITIVE_X, distance_m=0.2))
```
alexmillane, line 64: Really cool!

isaaclab_arena/embodiments/embodiment_base.py
cvolkcvolk: True, every embodiment has the` scene_config.robot.spawn.usd_path` set

PR #411, opened 2026-02-12
Distributed policy_runner on multi-gpu

xyao-nv:
## Summary
Product/PerfLab requires multi-gpu supports for benchmarking ILA envs. The ask is kinda urgent -- 'by early Feb'. It also introduces a great potential to support running large-scale eval, e.g. on cloud.

## Detailed description
- Following how [Lab handles multi-gpu ](https://isaac-sim.github.io/IsaacLab/main/source/features/multi_gpu.html), each gpu runs one process, each process has its own simulation instance and policy
- Process group management is handled by SimApp, policy_runner just needs to assign the right device_id.
- GR00T takes device_id to bring up its own copy on each gpu.
- Metrics are recorded & computed per gpu to avoid multi-process write.

## Results
- Tested on 2xRTX 6000 Ada
- Num_episodes in total = $\sum_{i=1}^{NumGpu}$  num_episodes_i
- Empirically I see the process from one gpu could occupy some VRAM (~200Mib) to the other gpu. Unclear why.

## TODO
- Expand to RL & eval_runner
- Add tests once mgpu CI runner is up

### Arch
- This setup may not be ideal, as policy may occupy a lot of RAM, and may be more efficient to run on one GPU. But the overhead from syncing between policy-gpu vs sim-gpu and tensor D2D copy are unknown.
- Lab does not have other ways to support distributed. So Im sticking to it for now and revisit once we have client-server merged.

### CI
- There is no runner with >1 gpu on our CI now. It will be really helpful to have >1 GPU CI runners and then I can push a test to catch any breaking changes. 

### Heterogeneous env
- Since each GPU spawns its own instance, it's feasible to load a different env_cfg on each rank. Like as a process launcher for multi-gpu.
- That means each gpu could run diff task/robot/obj etc.
-  I wonder what the cost is. Is it free lunch?


docs/pages/example_workflows/sequential_static_manipulation/step_5_evaluation.rst
```diff
@@ -108,18 +108,40 @@ Step 2: Run Parallel environments Evaluation
...
+         python -m torch.distributed.run --nnode=1 --nproc_per_node=2 isaaclab_arena/evaluation/policy_runner.py \
+           --policy_type isaaclab_arena_gr00t.policy.gr00t_closedloop_policy.Gr00tClosedloopPolicy \
+           --policy_config_yaml_path isaaclab_arena_gr00t/policy/config/gr1_manip_gr00t_closedloop_config.yaml \
+           --num_steps 2000 \
+           --num_envs 10 \
+           --enable_cameras \
+           --headless \
+           --distributed \
+           gr1_open_microwave \
+           --embodiment gr1_joint
```
alexmillane, line 143: Coool!

isaaclab_arena/cli/isaaclab_arena_cli.py
```diff
@@ -26,9 +26,7 @@ def add_isaac_lab_cli_args(parser: argparse.ArgumentParser) -> None:
     isaac_lab_group.add_argument(
         "--disable_fabric", action="store_true", default=False, help="Disable fabric and use USD I/O operations."
     )
-    isaac_lab_group.add_argument(
-        "--seed", type=int, default=None, help="Optional seed for the random number generator."
-    )
+    isaac_lab_group.add_argument("--seed", type=int, default=42, help="Optional seed for the random number generator.")
```
alexmillane, line 29: This seems to be a behaviour change, we've moved to always having a seed? Is the intention here that we get the same results (ish) every time we run?

I'm thinking that this is probably a good change. It will make our results more deterministic.

What was your thinking here?

isaaclab_arena/evaluation/policy_runner.py
```diff
@@ -109,35 +127,63 @@ def rollout_policy(env, policy: "PolicyBase", num_steps: int | None, num_episode
...
     args_cli, unknown = args_parser.parse_known_args()
 
-    # Start the simulation app
+    # local rank is the rank of the process on the current node, set by torchrun command
+    local_rank = int(os.environ.get("LOCAL_RANK", "0"))
+    # world size is the number of processes running the script, set by torchrun command
+    world_size = int(os.environ.get("WORLD_SIZE", "1"))
+    # Setting device to local rank before SimulationAppContext
+    if is_distributed(args_cli):
+        args_cli.distributed = True
```
alexmillane: Why set this `True`?  It's checked in `is_distributed`, I'm wondering if there's some specific reason to set it here.

isaaclab_arena/evaluation/policy_runner.py
```diff
@@ -109,35 +127,63 @@ def rollout_policy(env, policy: "PolicyBase", num_steps: int | None, num_episode
...
+
+        # Per-rank seed when distributed so each process has a different seed
+        seed = args_cli.seed
+        if seed is not None and is_distributed(args_cli):
+            seed = seed + local_rank
+        if seed is not None:
+            env.seed(seed)
+            torch.manual_seed(seed)
+            np.random.seed(seed)
+            random.seed(seed)
```
alexmillane: This is not from this MR, but shall we move this to `isaaclab_arena/utils/random.py` with a function called `set_seeds()`

isaaclab_arena/evaluation/policy_runner.py
```diff
@@ -109,35 +127,63 @@ def rollout_policy(env, policy: "PolicyBase", num_steps: int | None, num_episode
...
+    """
     args_parser = get_isaaclab_arena_cli_parser()
     # We do this as the parser is shared between the example environment and policy runner
     args_cli, unknown = args_parser.parse_known_args()
 
-    # Start the simulation app
+    # local rank is the rank of the process on the current node, set by torchrun command
+    local_rank = int(os.environ.get("LOCAL_RANK", "0"))
+    # world size is the number of processes running the script, set by torchrun command
+    world_size = int(os.environ.get("WORLD_SIZE", "1"))
```
alexmillane: Suggestion to wrap these in `isaaclab_arena/utils/multiprocess.py` with functions like `get_local_rank` and `get_world_size`

isaaclab_arena_gr00t/policy/gr00t_closedloop_policy.py
```diff
@@ -102,12 +103,17 @@ def __init__(self, config: Gr00tClosedloopPolicyArgs):
         """
         super().__init__(config)
         self.policy_config = create_config_from_yaml(config.policy_config_yaml_path, Gr00tClosedloopPolicyConfig)
+        self.num_envs = config.num_envs
+        self.device = config.policy_device
+        # Under torchrun (WORLD_SIZE > 1), pin policy to this process's GPU so each rank uses its own device.
+        world_size = int(os.environ.get("WORLD_SIZE", "1"))
+        if world_size > 1 and "cuda" in config.policy_device:
+            local_rank = int(os.environ.get("LOCAL_RANK", "0"))
```
alexmillane: Suggestion to use functions (suggested above) for these

isaaclab_arena/evaluation/policy_runner.py
```diff
@@ -109,35 +127,63 @@ def rollout_policy(env, policy: "PolicyBase", num_steps: int | None, num_episode
...
-        if args_cli.seed is not None:
-            env.seed(args_cli.seed)
-            torch.manual_seed(args_cli.seed)
-            np.random.seed(args_cli.seed)
-            random.seed(args_cli.seed)
+        name, cfg = arena_builder.build_registered()
+        # Avoid HDF5 file lock conflict when distributed: each rank uses a unique dataset filename
+        if is_distributed(args_cli) and hasattr(cfg, "recorders") and cfg.recorders is not None:
+            base = getattr(cfg.recorders, "dataset_filename", "dataset")
+            cfg.recorders.dataset_filename = f"{base}_rank{local_rank}"
```
alexmillane: What do you think about moving this into the `ArenaEnvBuilder`. The arena env builder could take an optional argument `distributed=False`. If `True`, we can update the file name where the recorder manager where it is composed.

Solving it in the `ArenaEnvBuilder`, rather than this script, means that the fix works where ever arena environments are built.

docs/pages/example_workflows/locomanipulation/step_4_evaluation.rst
```diff
@@ -94,21 +94,46 @@ Step 2: Run Parallel Environments Evaluation
...
+           --num_steps 1200 \
+           --num_envs 5 \
+           --enable_cameras \
+           --device cuda \
+           --policy_device cuda  \
+           --distributed \
+           --headless \
+           galileo_g1_locomanip_pick_and_place \
+           --object brown_box \
+           --embodiment g1_wbc_joint
```
alexmillane, line 135: Amazing!

alexmillane (approved):
Thanks for adding this capability. This is really awesome. I had a couple of small comments.

isaaclab_arena/cli/isaaclab_arena_cli.py
xyao-nv: Yes, I think it's important that locking the seed to make the run more deterministic from eval perspective.

isaaclab_arena/evaluation/policy_runner.py
xyao-nv: done

isaaclab_arena/evaluation/policy_runner.py
xyao-nv: great catch. forgot to remove, done.

isaaclab_arena/evaluation/policy_runner.py
xyao-nv: great suggestion. done

isaaclab_arena/evaluation/policy_runner.py
xyao-nv: agree. done

isaaclab_arena_gr00t/policy/gr00t_closedloop_policy.py
xyao-nv: done

PR #414, opened 2026-02-13
Franka robot and stand is a single usd

cvolkcvolk:
## Summary
Replace the separate Franka robot and stand USD assets with a single combined` franka_on_stand.usd` file. 
This is needed because the placement solver computes the robot's bounding box from `scene_config.robot.spawn.usd_path` — previously, the stand was a separate asset, so the computed bounding box only covered the robot arm and did not include the stand. This led to incorrect placement when using spatial relations.

### DIFF
--- a/isaaclab_arena/tests/test_robot_initial_position.py
+++ b/isaaclab_arena/tests/test_robot_initial_position.py
@@ -61,11 +61,12 @@ def _test_robot_initial_position(simulation_app):
         print(f"Robot position error: {robot_position_error}")
         assert robot_position_error < INITIAL_POSITION_EPS, "Robot ended up at the wrong position."
 
-        # Check the stand ended up at the correct position.
-        stand_position = env.scene["stand"].get_world_poses()[0].cpu().numpy()
-        stand_position_error = np.linalg.norm(stand_position - np.array(robot_init_position))
-        print(f"Stand position error: {stand_position_error}")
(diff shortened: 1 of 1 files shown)


isaaclab_arena/embodiments/franka/franka.py
```diff
@@ -41,6 +39,10 @@
 _DEFAULT_CAMERA_OFFSET = Pose(position_xyz=(0.11, -0.031, -0.074), rotation_wxyz=(-0.74896, 0.0, 0.0, -0.66262))
 
 
+_FRANKA_CFG = FRANKA_PANDA_HIGH_PD_CFG.copy()
+_FRANKA_CFG.spawn.usd_path = f"{ISAACLAB_STAGING_NUCLEUS_DIR}/Arena/assets/robot_library/franka_on_stand.usd"
```
xyao-nv: Can you rename the file s.t. it includes the name of the gripper? Sim enables the robot to [swap with variants](https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab_assets/isaaclab_assets/robots/franka.py#L90)(gripper in this case).
For example, in droid, it's a franka panda arm with robotiq gripper. Here I believe it's with the franka hand?

cvolkcvolk: Good point. Renamed to `franka_panda_hand_on_stand.usd` for now.

PR #415, opened 2026-02-13
Fix docker entrypoint failure when host user is 'ubuntu'

jo-fra:
## Problem

When the host username is `ubuntu`, `docker/run_docker.sh` fails at container startup with:

```
userdel: user 'ubuntu' does not exist
useradd: group 'ubuntu' does not exist
```

## Root Cause

In `entrypoint.sh`, `groupadd --force` runs **before** `userdel ubuntu`. Since the base image already has a `ubuntu` user/group, the sequence is:

1. `groupadd --force` — group `ubuntu` already exists, no-op
2. `userdel ubuntu` — deletes the user **and** its primary group
3. `useradd --gid ubuntu` — fails because the group was removed in step 2

## Fix

Move `userdel` calls **before** `groupadd`, so the group is always (re)created after any prior user/group removal.

### DIFF
--- a/docker/setup/entrypoint.sh
+++ b/docker/setup/entrypoint.sh
@@ -12,12 +12,15 @@ set -euo pipefail
 # at the end of the dockerfile
 ldconfig
 
-# Add the group of the user. User/group ID of the host user are set through env variables when calling docker run further down.
+# Remove any existing user with the same name (userdel may also remove its primary group)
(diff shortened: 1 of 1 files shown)


vaishkarni (approved):
This works when the host username is "ubuntu"

vaishkarni:
Please merge this, this works when the username is "ubuntu"

PR #416, opened 2026-02-14
Add DROID

xyao-nv:
## Summary
DROID is commonly used in Real exps. Add one referred from a public reference [here](https://github.com/arhanjain/sim-evals/blob/main/src/sim_evals/environments/droid_environment.py).

## Detailed description
- `FRANKA_ROBOTIQ_GRIPPER_CFG` from Lab has missing mimic joints at the gripper in USD. 
- The public sourced USD has verified to run in Lab env with joint position ctrl.

### DIFF
--- a/isaaclab_arena/embodiments/droid/actions.py
+++ b/isaaclab_arena/embodiments/droid/actions.py
@@ -16,7 +16,7 @@ class BinaryJointPositionZeroToOneAction(BinaryJointPositionAction):
         # compute the binary mask
         if actions.dtype == torch.bool:
             # true: close, false: open
-            binary_mask = actions == 0
+            binary_mask = actions == 1
         else:
             # true: close, false: open
             binary_mask = actions > 0.5
--- a/isaaclab_arena/embodiments/droid/droid.py
+++ b/isaaclab_arena/embodiments/droid/droid.py
@@ -94,7 +94,7 @@ class DroidSceneCfg:
         prim_path="{ENV_REGEX_NS}/Robot",
         spawn=sim_utils.UsdFileCfg(
             usd_path=(
-                f"{ISAACLAB_STAGING_NUCLEUS_DIR}Arena/assets/robot_library/droid/franka_robotiq_2f_85_flattened.usd"
+                f"{ISAACLAB_STAGING_NUCLEUS_DIR}/Arena/assets/robot_library/droid/franka_robotiq_2f_85_flattened.usd"
             ),
             activate_contact_sensors=True,
             rigid_props=sim_utils.RigidBodyPropertiesCfg(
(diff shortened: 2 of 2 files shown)


isaaclab_arena/embodiments/droid/actions.py
```diff
@@ -0,0 +1,30 @@
...
+    def process_actions(self, actions: torch.Tensor):
+        # store the raw actions
+        self._raw_actions[:] = actions
+        # compute the binary mask
+        if actions.dtype == torch.bool:
+            # true: close, false: open
+            binary_mask = actions == 0
+        else:
+            # true: close, false: open
+            binary_mask = actions > 0.5
```
alexmillane, line 22: This seems to reverse the polarity from the function in `BinaryJointPositionAction`. There they have have `binary_mask = actions < 0`. I guess that's intended? It's slight weird because:

`torch([1], dtype=torch.bool)` means open

but 

`torch([1.0], dtype=torch.float)` means close

Is that intended?

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -0,0 +1,345 @@
...
+@configclass
+class DroidSceneCfg:
+    """Additions to the scene configuration coming from the Franka embodiment."""
+
+    # The robot
+    robot: ArticulationCfg = ArticulationCfg(
+        prim_path="{ENV_REGEX_NS}/Robot",
+        spawn=sim_utils.UsdFileCfg(
+            usd_path=(
+                f"{ISAACLAB_STAGING_NUCLEUS_DIR}Arena/assets/robot_library/droid/franka_robotiq_2f_85_flattened.usd"
```
cvolkcvolk:
```suggestion
                f"{ISAACLAB_STAGING_NUCLEUS_DIR}/Arena/assets/robot_library/droid/franka_robotiq_2f_85_flattened.usd"
```

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -0,0 +1,345 @@
...
+        init_state=AssetBaseCfg.InitialStateCfg(pos=[-0.05, 0.0, 0.0], rot=[1.0, 0.0, 0.0, 0.0]),
+        spawn=UsdFileCfg(
+            usd_path="https://omniverse-content-production.s3-us-west-2.amazonaws.com/Assets/Isaac/4.5/Isaac/Props/Mounts/Stand/stand_instanceable.usd",
+            scale=(1.2, 1.2, 1.7),
+            activate_contact_sensors=False,
+        ),
+    )
+
+    # The end-effector frame marker
+    # TODO: Need to verify that pos is correct
```
cvolkcvolk: Note to add a name to the TODO(). Could we verify before merging?

isaaclab_arena/embodiments/droid/droid.py
xyao-nv: done, thx

isaaclab_arena/embodiments/droid/droid.py
xyao-nv: ;) I was thinking I don't need to be added. Alright, I will do it.

isaaclab_arena/embodiments/droid/actions.py
xyao-nv: Agree that does not make sense. Thx for catching, done

PR #420, opened 2026-02-18
Gr00t N1.6-Droid runs locally on a robolab scene

cvolkcvolk:
## Summary
Gr00t-N1.6-Droid runs locally, closed-loop on the Franka Droid setup on a robolab scene. 
Based on https://github.com/isaac-sim/IsaacLab-Arena/pull/421

In the Gr00t Docker:
`python isaaclab_arena/evaluation/policy_runner.py --num_steps 2000 --enable_cameras --policy_type isaaclab_arena_gr00t.policy.gr00t_closedloop_policy.Gr00tClosedloopPolicy --policy_config_yaml_path isaaclab_arena_gr00t/policy/config/droid_manip_gr00t_closedloop_config.yaml droid_pick_and_place_srl`

### DIFF
--- a/isaaclab_arena/embodiments/droid/droid.py
+++ b/isaaclab_arena/embodiments/droid/droid.py
@@ -5,6 +5,7 @@
 
 
 import torch
+from abc import ABC
 from dataclasses import MISSING
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -55,6 +56,7 @@ def __init__(
         initial_joint_pose: list[float] | None = None,
         concatenate_observation_terms: bool = False,
         arm_mode: ArmMode | None = None,
+        use_joint_position_actions: bool = False,
```
alexmillane: I don't think that this is used anywhere?

isaaclab_arena/embodiments/droid/droid.py
alexmillane: Is this an abstract base class? Do we need a child who implements an `action_cfg`?

If this is the case, suggestion to inherit from `abc`, and not to register it (only register the child that implements the class).

alexmillane (approved):
Thanks for adding this!!

isaaclab_arena/embodiments/droid/droid.py
cvolkcvolk: Yes, definetly an abstract class. Thanks.

PR #424, opened 2026-02-19
[GTC DLI] Add objects for zero-shot policy eval

xyao-nv:
## Summary
Running zero-shot eval of GN1.6 sequential task ckpt son similar cylindrical objects (ketch, bbq sauce, mayo)

## Detailed description
- 3 more sauce bottles are added from Robolab's Hope dataset, verified in zero-shot closedloop independently 
- Prim path for adding contact sensors: handling both cases
The composed prim at the reference site is that rigid body. Isaac Lab’s activate_contact_sensors(prim_path) checks that prim first; if it has RigidBodyAPI, it adds the contact sensor there and does not need to look at children.
1. Depth-0: default prim = rigid body (single root). 
2. Depth ≥ 1: default prim = parent so the referenced prim is a scope with rigid_body as a child. 
- Materials/bindings in cache: some references to texture files are using relative path. e.g. Hope packs all textures into one folder, and USDs are in the parent folder. It rewrites relationship targets and attribute connections from the old to the new rigid-body path on the root layer
- Add doc for heterogeneous eval and multi-task batch eval
- Remove OSMO-related internal-only steps in doc 

https://github.com/user-attachments/assets/ccdd27c3-c2f1-48bc-821c-2ddde48f90af

## TODO
Butter box & mustard bottle have some placement issues, sinking to the countertop. Syncing with @cvolkcvolk to debug.

image

## Note
- Beer bottle from LW cannot be spawned with Robolab's Hope objects in object_set. Their rigid body prim paths are at different levels.
- Zero-shot "works" because of trajectory overfitting to the similar object geometry, and we shall be honest about it.

### DIFF
--- a/docs/pages/example_workflows/sequential_static_manipulation/step_3_data_generation.rst
+++ b/docs/pages/example_workflows/sequential_static_manipulation/step_3_data_generation.rst
@@ -53,8 +53,8 @@ To start the annotation process run the following command:
 
    python isaaclab_arena/scripts/imitation_learning/annotate_demos.py \
      --device cpu \
-     --input_file $DATASET_DIR/ranch_bottle_into_fridge/ranch_bottle_into_fridge_recorded.hdf5 \
-     --output_file $DATASET_DIR/ranch_bottle_into_fridge/ranch_bottle_into_fridge_annotated.hdf5 \
+     --input_file $DATASET_DIR/ranch_bottle_into_fridge_recorded.hdf5 \
+     --output_file $DATASET_DIR/ranch_bottle_into_fridge_annotated.hdf5 \
      --enable_pinocchio \
      --mimic \
      put_item_in_fridge_and_close_door \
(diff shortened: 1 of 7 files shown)


isaaclab_arena/utils/usd/object_set_utils.py
```diff
@@ -33,18 +33,122 @@ def rescale_root(stage: Usd.Stage, asset: Asset) -> None:
...
+                    r = replace_path(c) if isinstance(c, Sdf.Path) else None
+                    if r is not None:
+                        new_conns.append(r)
+                        changed = True
+                    else:
+                        new_conns.append(c)
+                if changed and getattr(attr, "SetConnections", None) is not None:
+                    attr.SetConnections(new_conns)
+    finally:
+        stage.SetEditTarget(edit_target)
```
alexmillane, line 117: Thanks for writing this. Could you add a few comments? It's a bit hard to interpret otherwise.

isaaclab_arena_environments/gr1_put_and_close_door_environment.py
```diff
@@ -199,7 +197,7 @@ def add_cli_args(parser: argparse.ArgumentParser) -> None:
             "--object",
             type=str,
             default="ranch_dressing_bottle",
-            choices=["sweet_potato", "jug", "ranch_dressing_bottle"],
+            # choices=["beer_bottle", "ketchup_bottle", "ranch_dressing_bottle", "bbq_sauce_bottle"],
```
alexmillane: Should we remove this?

xyao-nv: done

isaaclab_arena/utils/usd/object_set_utils.py
xyao-nv: done

PR #428, opened 2026-02-23
Add hdr library and use robolab hdr

cvolkcvolk:
## Summary

Previously, applying an HDR texture to a dome light required manually constructing a DomeLightCfg with a hardcoded Nucleus path (eg in droid_pick_and_place_srl_environment.py). 

- Introduces an `HDRRegistry` for managing HDR/EXR environment map textures, following the same patterns used by `AssetRegistry` for objects and backgrounds. This replaces hardcoded texture file paths in environment definitions.

- Adds some of  the [RoboLab Backgrounds](https://gitlab-master.nvidia.com/xuningy/robolab/-/tree/main/assets/backgrounds?ref_type=heads)  to the HDRRegistry.

Now a texture can be directly added to a light instance. For example via:
```
# Create a dome light and attach an HDR by name
light = asset_registry.get_asset_by_name("light")()
light.add_hdr("billiard_hall_robolab")

# Or pass it directly to the constructor
light = asset_registry.get_asset_by_name("light")(hdr="billiard_hall_robolab")
```

### DIFF
--- a/isaaclab_arena/assets/asset_registry.py
+++ b/isaaclab_arena/assets/asset_registry.py
@@ -10,6 +10,7 @@ from isaaclab_arena.utils.singleton import SingletonMeta
 
 if TYPE_CHECKING:
     from isaaclab_arena.assets.asset import Asset
+    from isaaclab_arena.assets.hdr import HDR
     from isaaclab_arena.assets.teleop_device_base import TeleopDeviceBase
     from isaaclab_arena.policy.policy_base import PolicyBase
 
@@ -38,7 +39,7 @@ class Registry(metaclass=SingletonMeta):
             key (str): The name of the component.
         """
         # For AssetRegistry and DeviceRegistry, ensure assets are registered before checking
-        if isinstance(self, (AssetRegistry, DeviceRegistry, RetargeterRegistry, PolicyRegistry)):
+        if isinstance(self, (AssetRegistry, DeviceRegistry, RetargeterRegistry, PolicyRegistry, HDRRegistry)):
             ensure_assets_registered()
         return key in self._components
 
@@ -52,7 +53,7 @@ class Registry(metaclass=SingletonMeta):
             Any: The component.
         """
         # For AssetRegistry and DeviceRegistry, ensure assets are registered before accessing
-        if isinstance(self, (AssetRegistry, DeviceRegistry, RetargeterRegistry, PolicyRegistry)):
+        if isinstance(self, (AssetRegistry, DeviceRegistry, RetargeterRegistry, PolicyRegistry, HDRRegistry)):
             ensure_assets_registered()
         assert key in self._components, f"component {key} not found, please check if requested component is registered"
         return self._components[key]
@@ -64,7 +65,7 @@ class Registry(metaclass=SingletonMeta):
             list[str]: The list of keys.
         """
         # For AssetRegistry and DeviceRegistry, ensure assets are registered before accessing
-        if isinstance(self, (AssetRegistry, DeviceRegistry, RetargeterRegistry, PolicyRegistry)):
+        if isinstance(self, (AssetRegistry, DeviceRegistry, RetargeterRegistry, PolicyRegistry, HDRRegistry)):
             ensure_assets_registered()
         return list(self._components.keys())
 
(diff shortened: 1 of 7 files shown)


peterd-NV (approved):

isaaclab_arena/assets/asset_registry.py
```diff
@@ -10,6 +10,7 @@
 
 if TYPE_CHECKING:
     from isaaclab_arena.assets.asset import Asset
+    from isaaclab_arena.assets.hdr import HDR
```
alexmillane, line 13: This looks a bit strange. It looks like a constant because of the fact it's all caps. Also HDR stands for High Dynamic Range, as far as I'm aware. What do you think about renaming it to `HDRImage`. This has the advantage of being more descriptive of what it is (an image) and the class name looks better (because it doesn't look like a constant).

isaaclab_arena/assets/asset_registry.py
```diff
@@ -38,7 +39,7 @@ def is_registered(self, key: str) -> bool:
             key (str): The name of the component.
         """
         # For AssetRegistry and DeviceRegistry, ensure assets are registered before checking
-        if isinstance(self, (AssetRegistry, DeviceRegistry, RetargeterRegistry, PolicyRegistry)):
+        if isinstance(self, (AssetRegistry, DeviceRegistry, RetargeterRegistry, PolicyRegistry, HDRRegistry)):
```
alexmillane, line 42: if we take my suggestion above, this will be `HDRImageRegistry`

isaaclab_arena/assets/hdr.py
```diff
@@ -0,0 +1,36 @@
...
+        self,
+        name: str,
+        texture_file: str,
+        tags: list[str] | None = None,
+        description: str = "",
+        texture_format: TextureFormat = "latlong",
+    ):
+        self.name = name
+        self.texture_file = texture_file
+        self.tags = tags or []
```
alexmillane, line 31: Elsewhere we have inherited `self.name` and `self.tags` by inheriting from `Asset`. I feel we should do the same here? Wdyt?

In the end I feel that the class `Asset` is perhaps poorly named. But I would, for now, inherit from it.

isaaclab_arena/assets/object_library.py
```diff
@@ -417,9 +417,18 @@ def __init__(
...
+
+    The light can be used plain (uniform color) or combined with an HDR
+    texture via :meth:`add_hdr`::
+
+        dome_light = asset_registry.get_asset_by_name("light")()
+        dome_light.add_hdr("home_office_robolab")
+
+    You can also pass the HDR name or instance directly to the constructor::
+
+        dome_light = asset_registry.get_asset_by_name("light")(hdr="home_office_robolab")
```
alexmillane, line 431: Amazing!

isaaclab_arena/assets/object_library.py
```diff
@@ -435,9 +444,41 @@ def __init__(
...
+        """
+        from isaaclab_arena.assets.asset_registry import HDRRegistry
+        from isaaclab_arena.assets.hdr import HDR
+
+        if isinstance(hdr, str):
+            hdr = HDRRegistry().get_hdr_by_name(hdr)
+        # If it's a class (not yet instantiated), instantiate it.
+        if isinstance(hdr, type) and issubclass(hdr, HDR):
+            hdr = hdr()
+        assert isinstance(hdr, HDR), f"Expected HDR class, instance, or string name, got {type(hdr)}"
```
alexmillane, line 469: Just for simplicity and symmetry with the rest of the code I would rely on the user having called the registry outside, and just have this method accept an `HDR` object.

isaaclab_arena/assets/object_library.py
```diff
@@ -435,9 +444,41 @@ def __init__(
         prim_path: str | None = default_prim_path,
         initial_pose: Pose | None = None,
         spawner_cfg: sim_utils.DomeLightCfg = default_spawner_cfg,
+        hdr: "str | HDR | None" = None,  # noqa: F821
     ):
         self.spawner_cfg = spawner_cfg
         super().__init__(instance_name=instance_name, prim_path=prim_path, initial_pose=initial_pose)
```
alexmillane, line 450: This is not your code, but this design of the spawner object is poor. Basically functions in `Object`, in particular `_generate_spawner_cfg` rely on this variable `self.spawner_cfg`. It's pretty poor design to have a parent relying on variable that may only exist in a child.

I feel that we need to rework the lights. But for now, to minimize the changes, let's just proceed with what we have.

isaaclab_arena/assets/asset_registry.py
cvolkcvolk: Agree

isaaclab_arena/assets/object_library.py
cvolkcvolk: Noted

PR #429, opened 2026-02-23
Add robolab objects

cvolkcvolk:
## Summary

Adds robolab objects following the naming convention `{object_name}_{dataset}_robolab`

Removed 5 duplicate assets from object_library.py but keep the scales for the existing ones.

### DIFF
--- a/isaaclab_arena/assets/object_library.py
+++ b/isaaclab_arena/assets/object_library.py
@@ -4,9 +4,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import isaaclab.sim as sim_utils
+
+if TYPE_CHECKING:
+    from isaaclab_arena.assets.hdr_image import HDRImage
 from isaaclab.sim.spawners.from_files.from_files_cfg import GroundPlaneCfg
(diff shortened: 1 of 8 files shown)


isaaclab_arena/assets/object_library.py
```diff
@@ -949,3 +839,998 @@ def __init__(
...
+
+@register_asset
+class RanchDressingHopeRobolab(LibraryObject):
+    name = "ranch_dressing_hope_robolab"
+    tags = ["object", "graspable", "food", "robolab"]
+    usd_path = (
+        f"{ISAACLAB_STAGING_NUCLEUS_DIR}/Arena/assets/object_library/srl_robolab_assets/objects/hope/ranch_dressing.usd"
+    )
+    scale = (0.8, 0.8, 1.2)
+
```
xyao-nv, line 1297: Can you keep the init call? It's helpful to have diff scale of the same obj in diff env with changing the object_library.py.
[example](https://github.com/isaac-sim/IsaacLab-Arena/pull/429/changes#diff-7e6b39cffdf75e43512c31a891c83ae949746d0e3161880ca11808cd4efb0f96L727-L734)

cvolkcvolk: That __init__ just passes/ Inherits the same arguments through to LibraryObject.__init__ without any additions or modifications. Keeping the init in the object should not be necessary?

PR #439, opened 2026-02-25
ObjectPlacer: reject overlapping placements

cvolkcvolk:
## Summary

`_validate_placement` was a stub that always returned True, so the solver could converge to solutions where objects overlap

- Implement `_validate_placement` in ObjectPlacer to detect bounding-box overlaps between placed objects, causing the placer to retry again.

### DIFF
--- a/isaaclab_arena/relations/object_placer.py
+++ b/isaaclab_arena/relations/object_placer.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING
 from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
 from isaaclab_arena.relations.placement_result import PlacementResult
 from isaaclab_arena.relations.relation_solver import RelationSolver
-from isaaclab_arena.relations.relations import On, RandomAroundSolution, RotateAroundSolution, get_anchor_objects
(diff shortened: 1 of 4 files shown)


isaaclab_arena/relations/object_placer.py
```diff
@@ -187,20 +189,49 @@ def _validate_placement(
...
+
+                a_world = a.get_bounding_box().translated(positions[a])
+                b_world = b.get_bounding_box().translated(positions[b])
+
+                if a_world.overlaps_xy(b_world, margin=self.params.min_separation_xy_m):
+                    if self.params.verbose:
+                        print(f"  XY overlap between '{a.name}' and '{b.name}'")
+                    return False
+
         return True
```
alexmillane: I'm wondering why we're doing this in 2D (x,y). Why not go with full 3D. You can imagine situations where a user would want objects overlap in 2D but not in 3D. Those situations are probably rare, but wouldn't it be more parsimonious just to detect 3D overlap? Then we don't even have to exclude the on relations.

cvolkcvolk: I thought that 2d would be easier. But actually given the additional On() relation check required it adds some complexity as well there.
I added an z-axis overlap to check the full 3d range now

PR #445, opened 2026-02-26
Add video recording support to eval_runner.py

cvolkcvolk:
## Summary

Adds the ability to record videos of evaluation jobs via `--video` flag. Each job produces an mp4 in a per-job subdirectory under `--video_dir` 

```
python isaaclab_arena/evaluation/eval_runner.py \
  --eval_jobs_config isaaclab_arena_environments/eval_jobs_configs/droid_pnp_srl_gr00t_jobs_config.json \
  --video
```


isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -289,12 +289,17 @@ def build_registered(
...
 
     def make_registered_and_return_cfg(
-        self, env_cfg: None | IsaacLabArenaManagerBasedRLEnvCfg = None
+        self, env_cfg: None | IsaacLabArenaManagerBasedRLEnvCfg = None, render_mode: str | None = None
     ) -> tuple[ManagerBasedEnv, IsaacLabArenaManagerBasedRLEnvCfg]:
         name, cfg = self.build_registered(env_cfg)
-        return gym.make(name, cfg=cfg).unwrapped, cfg
+        env = gym.make(name, cfg=cfg, render_mode=render_mode)
+        if render_mode is None:
+            env = env.unwrapped
```
alexmillane: What's the reasoning here? If we're in a non-None render mode we don't wanna unwrap the env? Might be worth a comment.

isaaclab_arena/evaluation/policy_runner.py
```diff
@@ -107,16 +109,17 @@ def rollout_policy(
...
 
     else:
         # only compute metrics if env has metrics registered
-        if hasattr(env.cfg, "metrics"):
+        if hasattr(env.unwrapped.cfg, "metrics"):
             # NOTE(xinjieyao, 2025-10-07): lazy import to prevent app stalling caused by omni.kit
             from isaaclab_arena.metrics.metrics import compute_metrics
 
-            metrics = compute_metrics(env)
+            metrics = compute_metrics(env.unwrapped)
```
alexmillane: Is the robust to the env not being unwrapped? I think we had conditional unwrapping of env. Doesn't that mean if video is not requested, and the env is pre-unwrapped, that this will fail?

isaaclab_arena/environments/arena_env_builder.py
cvolkcvolk: VideoRecorder is a gym wrapper we don't want to strip away. If we don't require the render_mode, keep the default behaviour to return an unwrapped version

cvolkcvolk: Having looked deeper into this, we actually might want to return an wrapped gym environment as IsaacLab is doing it

PR #460, opened 2026-03-03
Collision Constraints for Objects

zhx06:
## Summary
Add NoCollision relation and loss strategy 

## Detailed description
Add NoCollision relation 
Add NoCollisionLossStrategy
Add Isaac Sim NoCollision example
Add GR1 table multi-object NoCollision environment

### DIFF
--- a/isaaclab_arena/relations/object_placer.py
+++ b/isaaclab_arena/relations/object_placer.py
@@ -191,8 +191,8 @@ class ObjectPlacer:
     ) -> bool:
         """Validate each On relation; logic matches OnLossStrategy (relation_loss_strategies.py).
 
-        1. X: child footprint entirely within parent X extent.
-        2. Y: child footprint entirely within parent Y extent.
+        1. X: child's footprint entirely within parent's X extent.
+        2. Y: child's footprint entirely within parent's Y extent.
         3. Z: child_bottom in (parent_top, parent_top+clearance_m], within on_relation_z_tolerance_m.
         """
         for obj in positions:
@@ -204,7 +204,7 @@ class ObjectPlacer:
                     continue
                 child_world = obj.get_bounding_box().translated(positions[obj])
                 parent_world = parent.get_bounding_box().translated(positions[parent])
-                # 1 & 2: Same as OnLossStrategy x_band_loss / y_band_loss (footprint within parent).
+                # 1 & 2: Same as OnLossStrategy X/Y band (child's footprint within parent).
                 if (
                     child_world.min_point[0] < parent_world.min_point[0]
                     or child_world.max_point[0] > parent_world.max_point[0]
--- a/isaaclab_arena/tests/test_no_collision_loss.py
+++ b/isaaclab_arena/tests/test_no_collision_loss.py
@@ -81,18 +81,33 @@ def test_no_collision_zero_loss_when_separated_on_one_axis_only():
 
 
 def test_no_collision_zero_loss_when_just_touching():
-    """Test that NoCollision loss is zero when intervals just touch (overlap length 0)."""
+    """Test that NoCollision loss is zero when intervals just touch (clearance_m=0)."""
     box_a = _create_box("box_a")
     box_b = _create_box("box_b")
-    relation = NoCollision(box_b)
+    relation = NoCollision(box_b, clearance_m=0.0)
     strategy = NoCollisionLossStrategy(slope=10.0)
 
-    # Child at (0,0,0) -> X [0, 0.2]. Parent at (0.2,0,0) -> X [0.2, 0.4]. Just touching at 0.2.
+    # Child X [0, 0.2], parent X [0.2, 0.4]. Just touching.
     child_pos = torch.tensor([0.0, 0.0, 0.0])
     parent_world_bbox = box_b.get_bounding_box().translated((0.2, 0.0, 0.0))
 
     loss = strategy.compute_loss(relation, child_pos, box_a.bounding_box, parent_world_bbox)
     assert torch.isclose(loss, torch.tensor(0.0), atol=1e-5)
+
+
+def test_no_collision_positive_loss_when_just_touching():
+    """Test that NoCollision loss is positive when just touching with default clearance."""
+    box_a = _create_box("box_a")
+    box_b = _create_box("box_b")
+    relation = NoCollision(box_b)
+    strategy = NoCollisionLossStrategy(slope=10.0)
+
+    # Child X [0, 0.2], parent X [0.2, 0.4]. Just touching; default clearance expands parent so overlap > 0.
+    child_pos = torch.tensor([0.0, 0.0, 0.0])
+    parent_world_bbox = box_b.get_bounding_box().translated((0.2, 0.0, 0.0))
+
+    loss = strategy.compute_loss(relation, child_pos, box_a.bounding_box, parent_world_bbox)
+    assert loss > 0.0
 
 
 def test_no_collision_positive_loss_when_3d_overlap():
@@ -143,18 +158,15 @@ def test_no_collision_loss_scales_with_relation_weight():
 
 
 def test_no_collision_loss_volume_formula():
-    """Test that NoCollision loss equals slope * (overlap_x * overlap_y * overlap_z) for known overlap."""
-    # Two 0.2 boxes: child [0.1, 0.3]^3, parent [0.15, 0.35]^3 -> overlap [0.15, 0.3]^3, side 0.15, volume 0.15^3.
+    """Test that NoCollision loss equals slope * overlap volume for known overlap (clearance_m=0)."""
     box_a = _create_box("box_a", size=0.2)
     box_b = _create_box("box_b", size=0.2)
-    relation = NoCollision(box_b, relation_loss_weight=1.0)
+    relation = NoCollision(box_b, relation_loss_weight=1.0, clearance_m=0.0)
     strategy = NoCollisionLossStrategy(slope=10.0)
 
     child_pos = torch.tensor([0.1, 0.1, 0.1])
     parent_world_bbox = box_b.get_bounding_box().translated((0.15, 0.15, 0.15))
-    # Child world: X [0.1, 0.3], Y [0.1, 0.3], Z [0.1, 0.3].
-    # Parent world: X [0.15, 0.35], etc. Overlap each axis: [0.15, 0.3], length 0.15.
-    # overlap_volume = 0.15^3 = 0.003375, total_loss = 10 * 0.003375 = 0.03375.
+    # Overlap [0.15, 0.3]^3, volume 0.15^3. Expected loss = 10 * 0.15^3.
     expected_loss = 10.0 * (0.15**3)
 
     loss = strategy.compute_loss(relation, child_pos, box_a.bounding_box, parent_world_bbox)
--- a/isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
+++ b/isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
@@ -10,14 +10,11 @@ No task — suitable for policy_runner with zero_action or any policy.
 
 Example:
   python isaaclab_arena/evaluation/policy_runner.py --policy_type zero_action --num_steps 500 \\
-    --num_envs 1 --enable_cameras \\
-    --environment isaaclab_arena_environments.gr1_table_multi_object_no_collision_environment:GR1TableMultiObjectNoCollisionEnvironment \\
-    gr1_table_multi_object_no_collision --embodiment gr1_joint
+    --num_envs 1 --enable_cameras gr1_table_multi_object_no_collision --embodiment gr1_joint
 """
 
 import argparse
 
-from isaaclab_arena.tasks.no_task import NoTask
 from isaaclab_arena_environments.example_environment_base import ExampleEnvironmentBase
 
 DEFAULT_TABLE_OBJECTS = [
@@ -44,6 +41,7 @@ class GR1TableMultiObjectNoCollisionEnvironment(ExampleEnvironmentBase):
         from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
         from isaaclab_arena.relations.relations import IsAnchor, On
         from isaaclab_arena.scene.scene import Scene
+        from isaaclab_arena.tasks.no_task import NoTask
         from isaaclab_arena.utils.pose import Pose
 
         enable_cameras = getattr(args_cli, "enable_cameras", False)
--- a/isaaclab_arena/relations/object_placer_params.py
+++ b/isaaclab_arena/relations/object_placer_params.py
@@ -40,5 +40,5 @@ class ObjectPlacerParams:
     adds a safety margin between objects."""
 
     on_relation_z_tolerance_m: float = 5e-3
-    """Optional tolerance (meters) for On-relation Z validation. Valid Z band is extended to
+    """Tolerance (meters) for On-relation Z validation. Valid Z band is extended to
     (parent_top - tolerance, parent_top + clearance_m + tolerance]. Default 5e-3 accommodates solver residual."""
--- a/isaaclab_arena/tests/test_validate_placement.py
+++ b/isaaclab_arena/tests/test_validate_placement.py
@@ -117,7 +117,7 @@ def test_validate_on_relations_child_z_above_clearance_fails():
 
 
 def test_validate_on_relations_child_z_within_tolerance_above_clearance_passes():
-    """Test that child bottom slightly above parent_top+clearance passes when on_relation_z_tolerance_m provides slack."""
+    """Test that child bottom slightly above parent_top+clearance passes when on_relation_z_tolerance_m is set."""
     # on_relation_z_tolerance_m=5e-3 → valid Z band (0.045, 0.065]; child_bottom 0.063 is inside band.
     placer = ObjectPlacer(params=ObjectPlacerParams(on_relation_z_tolerance_m=5e-3))
     desk = _make_desk()
--- a/isaaclab_arena_environments/cli.py
+++ b/isaaclab_arena_environments/cli.py
@@ -17,6 +17,9 @@ from isaaclab_arena_environments.galileo_g1_locomanip_pick_and_place_environment
 from isaaclab_arena_environments.galileo_pick_and_place_environment import GalileoPickAndPlaceEnvironment
 from isaaclab_arena_environments.gr1_open_microwave_environment import Gr1OpenMicrowaveEnvironment
 from isaaclab_arena_environments.gr1_put_and_close_door_environment import GR1PutAndCloseDoorEnvironment
+from isaaclab_arena_environments.gr1_table_multi_object_no_collision_environment import (
+    GR1TableMultiObjectNoCollisionEnvironment,
+)
 from isaaclab_arena_environments.gr1_turn_stand_mixer_knob_environment import Gr1TurnStandMixerKnobEnvironment
 from isaaclab_arena_environments.kitchen_pick_and_place_environment import KitchenPickAndPlaceEnvironment
 from isaaclab_arena_environments.lift_object_environment import LiftObjectEnvironment
@@ -44,6 +47,7 @@ ExampleEnvironments = {
     LiftObjectEnvironment.name: LiftObjectEnvironment,
     TableTopPlaceUprightEnvironment.name: TableTopPlaceUprightEnvironment,
     Gr1TurnStandMixerKnobEnvironment.name: Gr1TurnStandMixerKnobEnvironment,
+    GR1TableMultiObjectNoCollisionEnvironment.name: GR1TableMultiObjectNoCollisionEnvironment,
 }
 
 


isaaclab_arena_examples/relations/isaac_sim_no_collision_notebook.py
```diff
@@ -0,0 +1,172 @@
...
+    # Build without solving relations so objects start overlapping in the sim; we run the solver after reset.
+    args_cli = get_isaaclab_arena_cli_parser().parse_args(["--no-solve-relations"])
+    env_builder = ArenaEnvBuilder(isaaclab_arena_environment, args_cli)
+    env = env_builder.make_registered()
+
+    objects_with_relations = [tabletop_reference, cracker_box, mug, tomato_soup_can]
+    placeable_objects = (cracker_box, mug, tomato_soup_can)
+    num_envs = env.unwrapped.scene.num_envs
+    env_ids = torch.arange(num_envs, device=env.unwrapped.device)
+    identity_quat = (1.0, 0.0, 0.0, 0.0)
```
xyao-nv: I'd prefer if you explicitly state the quat order in its var name, i.e wxyz or xyzw.

isaaclab_arena_examples/relations/isaac_sim_no_collision_notebook.py
```diff
@@ -0,0 +1,172 @@
...
+                dtype=torch.float32,
+            )
+            .unsqueeze(0)
+            .expand(num_envs, 7)
+        )
+        root_pose = root_pose.clone()
+        root_pose[:, :3] += env.unwrapped.scene.env_origins[env_ids]
+        for obj in placeable_objects:
+            if obj.name in env.unwrapped.scene.rigid_objects:
+                env.unwrapped.scene.rigid_objects[obj.name].write_root_pose_to_sim(root_pose.clone(), env_ids=env_ids)
```
xyao-nv, line 123: Shall we do something like logging for those are not in the scene?

isaaclab_arena_examples/relations/isaac_sim_no_collision_notebook.py
zhx06: Will rename the variable to identity_quat_wxyz to make the order explicit

isaaclab_arena/relations/loss_primitives.py
```diff
@@ -122,3 +122,36 @@ def single_point_linear_loss(
         target = torch.tensor(target, dtype=value.dtype, device=value.device)
 
     return slope * torch.abs(value - target)
+
+
+def interval_overlap_axis_loss(
+    a_min: torch.Tensor,
+    a_max: torch.Tensor,
+    b_min: torch.Tensor | float,
+    b_max: torch.Tensor | float,
```
xyao-nv: Can you rename vars to child_min, child_max, parent_min, parent_max?

isaaclab_arena/relations/loss_primitives.py
```diff
@@ -122,3 +122,36 @@ def single_point_linear_loss(
...
+        a_min: Child interval min (tensor for gradient flow).
+        a_max: Child interval max (tensor).
+        b_min: Parent interval min (tensor or float).
+        b_max: Parent interval max (tensor or float).
+        slope: Gradient magnitude (default: 1.0).
+
+    Returns:
+        Zero when intervals are separated; otherwise slope * overlap length.
+    """
+    assert isinstance(a_min, torch.Tensor), f"a_min must be a torch.Tensor, got {type(a_min)}"
```
xyao-nv: any assert to ensure a_min is really less than a_max?

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
```diff
@@ -0,0 +1,208 @@
...
+            solver_params=solver_params,
+            verbose=False,
+        )
+        placer = ObjectPlacer(params=params)
+        print("[TableMultiObjectLayoutTask] Precomputing one table layout (On + NoCollision)...")
+        result = placer.place(objects=objects_with_relations)
+        layout = {obj.name: result.positions[obj] for obj in placeable_assets}
+        print(f"  Layout done (attempts={result.attempts}, success={result.success})")
+        self._layout = layout
+        self._events_cfg = TableLayoutEventCfg(layout=self._layout, object_names=self._object_names)
```
xyao-nv: Sorry Im confused here. Why is it a `Task`?  My understanding for yours is no matter what task user is introducing (e.g. OpenDoor, PnP), this no-collision solver will always work to pre-configure the scene.
Or if you just need a env to showcase it works, you can use `noTask` to by pass auto reset/event, success checking and manually trigger reset in your step call.
Is there any concerns of not adding pre-compute & set pose ops to `arena_env_builder.py`?

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
```diff
@@ -0,0 +1,208 @@
...
+            scene=scene,
+            task=task,
+            teleop_device=teleop_device,
+        )
+        return isaaclab_arena_environment
+
+    @staticmethod
+    def add_cli_args(parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--objects",
```
xyao-nv, line 110: I wonder if you want to make it more explicit in the naming, to differentiate 2 categories of objects, maybe even more. E.g. scene may have object of interest (task related, like object to be manipulated in PnP), and objects in the background (irrelevant to the task completion, like distractor object in PnP). Users may swap in and out object of interested with cli, but all objects are supposed to be processed by the solver. WDTY

xyao-nv (commented):
Thanks for doing it! Good work. I left a few comments regarding your example env. Please help me understand your dedicated thoughts about the design.

isaaclab_arena/relations/relations.py
```diff
@@ -117,6 +117,29 @@ def __init__(
         self.clearance_m = clearance_m
 
 
+class NoCollision(Relation):
+    """Represents a 'no collision' relationship between two objects.
+
+    This relation specifies that the child and parent bounding boxes must not
```
cvolkcvolk, line 123: Suggestion  to add a `clearance_m` to .__init__, similarly to what we have in the `On()` relation:                                                                                                                                                                                       

```
  class NoCollision(Relation):
      def __init__(self, parent, relation_loss_weight=1.0, clearance_m=0.01):                                                                                                                                                    
          super().__init__(parent, relation_loss_weight)
          self.clearance_m = clearance_m
```

  Then in NoCollisionLossStrategy.compute_loss, expand the parent bbox by clearance_m before computing overlap:
```
  c = relation.clearance_m
  parent_x_min -= c;  parent_x_max += c
  parent_y_min -= c;  parent_y_max += c
  parent_z_min -= c;  parent_z_max += c

```

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -305,6 +306,87 @@ def compute_loss(
...
+        # Parent world-space extents from the world bounding box
+        parent_x_min = parent_world_bbox.min_point[0]
+        parent_x_max = parent_world_bbox.max_point[0]
+        parent_y_min = parent_world_bbox.min_point[1]
+        parent_y_max = parent_world_bbox.max_point[1]
+        parent_z_min = parent_world_bbox.min_point[2]
+        parent_z_max = parent_world_bbox.max_point[2]
+
+        # Child world-space extents
+        child_x_min = child_pos[0] + child_bbox.min_point[0]
```
cvolkcvolk: I think we can write this                                                                                                                                                                                                                                                                                                                                                    
```
  child_x_min = child_pos[0] + child_bbox.min_point[0]                                                                                                                                                                                       
  child_x_max = child_pos[0] + child_bbox.max_point[0]                                                                                                                                                                                       
  child_y_min = child_pos[1] + child_bbox.min_point[1]                                                                                                                                                                                       
  child_y_max = child_pos[1] + child_bbox.max_point[1]
  child_z_min = child_pos[2] + child_bbox.min_point[2]
  child_z_max = child_pos[2] + child_bbox.max_point[2]
```

As
```
  child_world_min = child_pos + torch.tensor(child_bbox.min_point, dtype=child_pos.dtype, device=child_pos.device)
  child_world_max = child_pos + torch.tensor(child_bbox.max_point, dtype=child_pos.dtype, device=child_pos.device)
```

And  then access below viad
child_world_min[0], child_world_min[1], etc

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
cvolkcvolk: I second what Xinjie pointed out. This is more of a scene composition than a Task.
We could for example use the existing `kitchen_pick_and_place_environment.py` and upgrade it to handle `On()` and `NoCollision` constraints

cvolkcvolk: If the goal is to introduce a "reset" through a event_cfg, I would suggest to move that functionality to another MR. Its currently a known limitation that the solver only runs once

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
```diff
@@ -0,0 +1,208 @@
...
+        return None
+
+    def get_metrics(self) -> list:
+        return []
+
+    def get_viewer_cfg(self) -> ViewerCfg:
+        return ViewerCfg(eye=(-1.5, -1.5, 1.5), lookat=(0.0, 0.0, 0.5))
+
+
+class GR1TableMultiObjectNoCollisionEnvironment(ExampleEnvironmentBase):
```
cvolkcvolk, line 31: Suggestion to use an existing environemnt, for example `kitchen_pick_and_place`

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
```diff
@@ -0,0 +1,208 @@
...
+        tabletop_reference.add_relation(IsAnchor())
+
+        object_names = getattr(args_cli, "objects", None) or DEFAULT_TABLE_OBJECTS
+        placeable_assets = []
+        for name in object_names:
+            obj = self.asset_registry.get_asset_by_name(name)()
+            obj.add_relation(On(tabletop_reference))
+            placeable_assets.append(obj)
+        # Pairwise NoCollision so objects do not overlap
+        for i, obj_a in enumerate(placeable_assets):
```
cvolkcvolk: Is there a case where an object should _not_ get a NoCollision constraint? We could hide this from the user by letting the Solver attach a NoCollision constraint to every object before  solving.

isaaclab_arena/relations/object_placer.py
cvolkcvolk, line 233: Suggestion to move this to a function as well as you did with `self._validate_on_relations(positions)`

cvolkcvolk: Then this function could collapse to something like
```
def _validate_placement():
    return self._validate_on_relations(positions) and self._validate_no_overlap(object)
```

isaaclab_arena/relations/relations.py
```diff
@@ -117,6 +117,29 @@ def __init__(
...
+    overlap. Adding NoCollision on one side is enough; the solver counts each
```
cvolkcvolk, line 124: `RelationSolver._compute_total_loss` iterates every relation on every object with no deduplication. If both `box_a.add_relation(NoCollision(box_b))` and `box_b.add_relation(NoCollision(box_a))` are present, loss is computed twice no?

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
zhx06: yeah, here I just wanted to have a hacky solution for demo purpose, will use kitchen_pick_and_place_environment and have a one-time demo

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
zhx06: I think this is a good idea. Previously, I followed the setup of `gr1_put_and_close_door_environment` , where other objects were already placed on the kitchen top. Introducing background objects could make this approach more flexible. I’m now considering @cvolkcvolk 's idea and looking into `gr1_put_and_close_door_environment`. I’m still working out the implementation details, since this environment is not an empty table and there may be potential collision constraints for all objects in the scene.

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
zhx06: The idea of a clean table is that we can get rid of the impact of the other background objects. I have tested kitchen_pick_and_place, and it seems that the sink and tap will collide with the movable objects if we want to place them randomly on the counter top. This will be a future feature/fix as suggested by @xyao-nv that we have background objects and object of interests. Should we focus on a simple case and then we try to fix it in the future MR?

isaaclab_arena/relations/object_placer.py
```diff
@@ -185,6 +185,31 @@ def _generate_initial_positions(
...
+                if (
+                    child_world.min_point[0] < parent_world.min_point[0]
+                    or child_world.max_point[0] > parent_world.max_point[0]
+                    or child_world.min_point[1] < parent_world.min_point[1]
+                    or child_world.max_point[1] > parent_world.max_point[1]
+                ):
+                    if self.params.verbose:
+                        print(f"  On relation: '{obj.name}' XY outside parent (retrying)")
+                    return False
+        return True
```
alexmillane, line 226: We could also check that z is within some epsilon? 1cm?

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -305,6 +306,87 @@ def compute_loss(
...
+        child_z_max = child_pos[2] + child_bbox.max_point[2]
+
+        # 1. Per-axis overlap (0 when separated on that axis)
+        overlap_x = interval_overlap_axis_loss(child_x_min, child_x_max, parent_x_min, parent_x_max)
+        overlap_y = interval_overlap_axis_loss(child_y_min, child_y_max, parent_y_min, parent_y_max)
+        overlap_z = interval_overlap_axis_loss(child_z_min, child_z_max, parent_z_min, parent_z_max)
+
+        # 2. Volume loss: product of per-axis overlaps scaled by slope
+        overlap_volume = overlap_x * overlap_y * overlap_z
+        total_loss = self.slope * overlap_volume
```
alexmillane: The naming is a little weird here. The function `interval_overlap_axis_loss` seems to indicate it returns a loss, however, we then treat the returned value as a length (as we multiply to get a volume).

Should we rename `interval_overlap_axis_loss`, to `get_overlap_length`, or is the function not returning a length, in which case we should rename volume.

alexmillane: Ah I see, the default slope is 1.0, which gives you the overlap distance (if the distance is >0). Suggestion just to expand the comment to make this clear.

isaaclab_arena/tests/test_no_collision_loss.py
alexmillane, line 1: Thanks for adding great tests!

isaaclab_arena/relations/relations.py
zhx06: This relates to another question. If A and B add collision to each other, their relation graph with be cyclic and there will be an issue creating the environment. This will need to be addressed in a higher-level

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
```diff
@@ -0,0 +1,208 @@
...
+
+"""
+Table + multi-object NoCollision environment. Office table with objects placed via
+On(table) and pairwise NoCollision (relation solver). Includes a robot (e.g. GR1).
+No task — suitable for policy_runner with zero_action or any policy.
+
+Example:
+  python isaaclab_arena/evaluation/policy_runner.py --policy_type zero_action --num_steps 500 \\
+    --num_envs 1 --enable_cameras \\
+    --environment isaaclab_arena_environments.gr1_table_multi_object_no_collision_environment:GR1TableMultiObjectNoCollisionEnvironment \\
```
alexmillane: So you can put your environment in this list [here](https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena_environments/cli.py#L34) so you don't have to use the fully qualified name. (probably later we should add a recorator for environment registration).

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
```diff
@@ -0,0 +1,208 @@
...
+        def _apply_layout(env: ManagerBasedEnv, env_ids: torch.Tensor) -> None:
+            if env_ids is None or len(env_ids) == 0 or not layout or not object_names:
+                return
+            num_envs = env.scene.num_envs
+            for obj_name in object_names:
+                pose_list = [Pose(position_xyz=layout[obj_name])] * num_envs
+                set_object_pose_per_env(env, env_ids, SceneEntityCfg(obj_name), pose_list)
+            env.scene.write_data_to_sim()
+
+        self.apply_table_layout = EventTermCfg(func=_apply_layout, mode="reset")
```
alexmillane: I see this has been commented on below. typically objects have their pose set through 

`my_object.set_initial_pose()` which registers reset events.

We want to steer clear of adding environment specific reset events.

What was the purpose here?

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    Table-based scene with multiple objects (On(table) + NoCollision) and a robot.
+    Layout is precomputed once and applied at reset to all envs.
+    """
+
+    name: str = "gr1_table_multi_object_no_collision"
+
+    def get_env(self, args_cli: argparse.Namespace):  # -> IsaacLabArenaEnvironment:
+        # Task precomputes layout in TableMultiObjectLayoutTask; skip builder relation solving.
+        args_cli.solve_relations = False
```
alexmillane: Not sure I get the motivation here? Why not sure the defautl realation solving the `ArenaEnvBuilder`.

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
alexmillane: I would second this. Let's hide this in the `ArenaEnvBuilder` such that all pairs of objects get no collision constraints automatically added.I would second this. Let's hide this in the `ArenaEnvBuilder` such that all pairs of objects get no collision constraints automatically added.

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
```diff
@@ -0,0 +1,208 @@
...
+        if args_cli.teleop_device is not None:
+            teleop_device = self.device_registry.get_device_by_name(args_cli.teleop_device)()
+        else:
+            teleop_device = None
+
+        objects_with_relations = [tabletop_reference, *placeable_assets]
+        task = TableMultiObjectLayoutTask(
+            objects_with_relations=objects_with_relations,
+            placeable_assets=placeable_assets,
+        )
```
alexmillane: Hopefully we can remove this custom task.

isaaclab_arena_examples/relations/dummy_object_placer_notebook.py
```diff
@@ -161,12 +163,82 @@ def run_dummy_multi_anchor_demo():
...
+
+
 # %%
 if __name__ == "__main__":
     # Run single anchor demo
-    run_dummy_object_placer_demo()
+    # run_dummy_object_placer_demo()
+
+    # # Run multi-anchor demo
+    # run_dummy_multi_anchor_demo()
```
alexmillane: remember to uncomment these.

isaaclab_arena_examples/relations/isaac_sim_no_collision_notebook.py
```diff
@@ -0,0 +1,189 @@
...
+                        identity_quat_wxyz[1],
+                        identity_quat_wxyz[2],
+                        identity_quat_wxyz[3],
+                    ],
+                    device=env.unwrapped.device,
+                    dtype=torch.float32,
+                )
+                .unsqueeze(0)
+                .expand(num_envs, 7)
+            )
```
alexmillane, line 151: Formatting is a bit verbose. I guess this got caught by the formatter?

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
zhx06: Previously we setup a custom task and this is to address the issue of **setting the object poses twice**. After we remove the custom task, this line is removed.

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
cvolkc: Alright. That sounds reasonable.

isaaclab_arena/relations/relations.py
cvolkc: To not forget about this, could you add a NOTE or a TODO if this is still something that needs to be adressed in the future?

zhx06: sure, I will add a TODO in a coming MR

PR #477, opened 2026-03-09
[EXPERIMENTAL] cuRobo IK based A* for symbolic planning 

njawale42:
## Summary
Add IK feasibility heuristic to A* planner with multi-demo collection

## Detailed description
- **What was the reason for the change?**
  The A* symbolic planner in `droid_v3_tabletop_pick_and_place` only used Euclidean distance as its heuristic, with no awareness of whether the robot could actually reach grasp/place poses. This led to wasted sim time executing doomed pick-and-place sequences on kinematically infeasible layouts. There was also no support for collecting multiple successful demos in a loop.

- **What has been changed?**
  - Added `ik_utils.py` with standalone `check_ik_feasibility()` and `get_current_joint_config()` functions that wrap cuRobo's IK solver with user-controllable position/rotation thresholds and seed chaining. Kept outside the IsaacLab submodule so changes can be committed directly to Arena.
  - Modified the A* planner in `plan_pick_order()` to evaluate IK as a sequential kinematic chain (home → grasp(1) → place(1) → grasp(2) → ...), seeding each IK solve from the previous solution. Orderings with IK failures are penalized so A* prefers feasible sequences. Now returns `(order, total_ik_penalty)`.
  - Added a demo collection loop to the run script (`--num_demos`, `--max_attempts`). Layouts with any IK penalty are skipped immediately and re-randomized, avoiding ~30s of wasted sim time per infeasible layout.
  - Added `--ik_pos_threshold` and `--ik_rot_threshold` CLI args to control IK feasibility sensitivity.

- **What is the impact of this change?**
  The planner now filters out unreachable layouts before execution (~1s IK check vs 30s+ failed sim run), making multi-demo data collection practical. The A* produces pick orderings that are guaranteed IK-feasible when such an ordering exists. Users can tune feasibility sensitivity via CLI thresholds and collect N successful demos with a single command.

### DIFF
--- a/isaaclab_arena_environments/droid_v3_tabletop_pick_and_place_environment.py
+++ b/isaaclab_arena_environments/droid_v3_tabletop_pick_and_place_environment.py
@@ -286,18 +286,6 @@ class DroidV3TabletopPickAndPlaceEnvironment(ExampleEnvironmentBase):
             prev_jc = joint_configs.pop(node_id, None)
             prev_ik_total = ik_totals.pop(node_id, 0.0)
 
-            # Duplicate detection: order-based when IK is active (cost is
-            # sequence-dependent), set-based otherwise.
-            if ik_cost_fn is not None:
-                order_key = tuple(order)
-                if order_key in visited_order:
-                    continue
-                visited_order.add(order_key)
-            else:
-                if placed in visited_set:
-                    continue
-                visited_set.add(placed)
-
             if placed == goal:
                 _log(f"\n--- Result ---")
                 _log(f"  Nodes expanded: {nodes_expanded}")
@@ -309,6 +297,18 @@ class DroidV3TabletopPickAndPlaceEnvironment(ExampleEnvironmentBase):
                 joint_configs.clear()
                 ik_totals.clear()
                 return order, prev_ik_total
+
+            # Duplicate detection: order-based when IK is active (cost is
+            # sequence-dependent), set-based otherwise.
+            if ik_cost_fn is not None:
+                order_key = tuple(order)
+                if order_key in visited_order:
+                    continue
+                visited_order.add(order_key)
+            else:
+                if placed in visited_set:
+                    continue
+                visited_set.add(placed)
 
             nodes_expanded += 1
 


isaaclab_arena/scripts/curobo/run_droid_v2_tabletop_curobo_pick_place.py
```diff
@@ -182,200 +188,217 @@ def main() -> None:
...
-            )
-            results[object_name]["place"] = place_success
-            if not place_success:
-                print(f"[SKIP] Skipping retreat for '{object_name}' due to place failure")
-                execute_gripper_action(
-                    env, planner, gripper_binary_action=GRIPPER_OPEN_CMD, steps=args_cli.gripper_settle_steps, env_id=0
+            # ── Plan pick order ───────────────────────────────────────────
+            if is_v3:
+                target_positions = arena_builder.example_env.generate_target_positions()
+                planner.update_world()
```
YunchuZhang, line 265: did not have this line before? is this important?

isaaclab_arena_environments/droid_v3_tabletop_pick_and_place_environment.py
```diff
@@ -260,65 +265,92 @@ def heuristic(placed: frozenset[str]) -> float:
...
-                _log(f"    Total travel   : {g:.3f} m")
-                _log(f"    Pick order     : {' -> '.join(order)}")
-                _log(f"{'=' * 70}\n")
-                print(f"[SYMBOLIC A*] Pick order: {order} (total cost: {g:.3f}m)")
-                return order
-
-            if placed in visited:
-                continue
-            visited.add(placed)
+                _log(f"\n--- Result ---")
```
YunchuZhang, line 290: it will be cleaner if this goal checker be placed right after the heappop?

isaaclab_arena/scripts/curobo/run_droid_v2_tabletop_curobo_pick_place.py
njawale42: Yes, it's necessary. cuRobo's IK solver is collision-aware, it checks against the collision world. Without this call, the IK solver would check against stale collision geometry, especially after env.reset() in the multi demo loop.

isaaclab_arena_environments/droid_v3_tabletop_pick_and_place_environment.py
njawale42: I guess it would be cleaner this way, yes. But functionality wise, it wouldn't matter. Let me make the change anyway to make it cleaner.

PR #479, opened 2026-03-12
Sphere and initial object velocity

DavidRecasens:
## Summary
New sphere in object library and ObjectBase has now optional initial velocity

## Detailed description
Now it can be spawn spheres, and objects like the sphere can be initialized with linear velocity.

### DIFF
--- a/isaaclab_arena/tests/test_events.py
+++ b/isaaclab_arena/tests/test_events.py
@@ -155,16 +155,16 @@ def _test_object_moves_with_initial_velocity(simulation_app):
     env.reset()
 
     try:
-        initial_position = env.scene[sphere.name].data.root_pose_w[0, :3].clone()
-        initial_position[:3] -= env.scene.env_origins[0]
+        initial_position = env.unwrapped.scene[sphere.name].data.root_pose_w[0, :3].clone()
+        initial_position[:3] -= env.unwrapped.scene.env_origins[0]
 
         for _ in tqdm.tqdm(range(NUM_STEPS)):
             with torch.inference_mode():
                 actions = torch.zeros(env.action_space.shape, device=env.unwrapped.device)
                 env.step(actions)
 
-        final_position = env.scene[sphere.name].data.root_pose_w[0, :3].clone()
-        final_position[:3] -= env.scene.env_origins[0]
+        final_position = env.unwrapped.scene[sphere.name].data.root_pose_w[0, :3].clone()
+        final_position[:3] -= env.unwrapped.scene.env_origins[0]
         displacement = final_position - initial_position
         assert (
             displacement[0].item() < -OBJECT_VELOCITY_MIN_DISPLACEMENT  # - x because initial velocity is in -x


isaaclab_arena/assets/object_base.py
```diff
@@ -80,6 +81,21 @@ def set_initial_pose(self, pose: Pose | PoseRange) -> None:
             self.object_cfg.init_state.rot = initial_pose.rotation_wxyz
         self.event_cfg = self._init_event_cfg()
 
+    def set_initial_velocity(self, linear_velocity: tuple[float, float, float]) -> None:
```
viiik-inside: maybe `set_initial_linear_velocity` is better

isaaclab_arena/assets/object_base.py
```diff
@@ -80,6 +81,21 @@ def set_initial_pose(self, pose: Pose | PoseRange) -> None:
...
+
+        The velocity is applied as the ``init_state.lin_vel`` on the underlying
+        config (``RigidObjectCfg`` or ``ArticulationCfg``) and is also restored
+        on every environment reset via the reset event.
+
+        Args:
+            linear_velocity: Linear velocity ``(vx, vy, vz)`` in the world frame.
+        """
+        self.initial_velocity = linear_velocity
+        if self.object_cfg is not None and hasattr(self.object_cfg.init_state, "lin_vel"):
```
viiik-inside, line 95: are these two conditions possible to be false?

isaaclab_arena/assets/object_base.py
```diff
@@ -106,13 +122,16 @@ def _init_event_cfg(self) -> EventTermCfg | None:
                 },
             )
         else:
+            params: dict = {
+                "pose": initial_pose,
+                "asset_cfg": SceneEntityCfg(self.name),
+            }
+            if self.initial_velocity is not None:
```
viiik-inside: can we anyway set the velocity to none. So we wont need this check. I see None is being handled in the event function

isaaclab_arena/assets/object_library.py
```diff
@@ -419,6 +420,61 @@ def __init__(
         super().__init__(instance_name=instance_name, prim_path=prim_path, initial_pose=initial_pose)
 
 
+@register_asset
+class Sphere(LibraryObject):
```
viiik-inside, line 424: nice!!

isaaclab_arena/assets/object_base.py
DavidRecasens: Yes, because self.object_cfg can exist but not have "lin_vel" attribure (e.g. the walls). Only articulated and rigid objects have it

isaaclab_arena/assets/object_library.py
```diff
@@ -419,6 +420,61 @@ def __init__(
...
+        rigid_props=sim_utils.RigidBodyPropertiesCfg(
+            solver_position_iteration_count=16,
+            solver_velocity_iteration_count=1,
+            max_angular_velocity=1000.0,
+            max_linear_velocity=1000.0,
+            max_depenetration_velocity=5.0,
+            disable_gravity=False,
+        ),
+        mass_props=sim_utils.MassPropertiesCfg(mass=0.25),
+    )
```
alexmillane, line 446: Is there some reason we need to change these `RigidBodyPropertiesCfg`. Could we leave them at the default values?

isaaclab_arena/assets/object_library.py
```diff
@@ -419,6 +420,61 @@ def __init__(
...
+            prim_path=self.prim_path,
+            spawn=self.spawner_cfg,
+        )
+        object_cfg = self._add_initial_pose_to_cfg(object_cfg)
+        if self.initial_velocity is not None:
+            object_cfg.init_state.lin_vel = self.initial_velocity
+        return object_cfg
+
+    def _requires_reset_pose_event(self) -> bool:
+        return self.get_initial_pose() is not None and self.reset_pose
```
alexmillane, line 475: This is a bit messy. We don't want to define these functions per object in the library.

The problem you're trying to get around here is that the spawner object type only supports `AssetBaseCfg` and not `RigidObjectCfg`.

I think we need to solve this at a level higher in the heirachy though. Solving this per-specific library object is going to lead to a lot of duplication in the future.

I think we need to separate "does something provide a custom spawner?" from "what type of object is this?". Right now they are coupled together through `ObjectType`.

isaaclab_arena/tests/test_events.py
```diff
@@ -107,6 +110,76 @@ def _test_set_object_pose_per_env_event(simulation_app):
...
+            solver_position_iteration_count=16,
+            solver_velocity_iteration_count=1,
+            max_angular_velocity=1000.0,
+            max_linear_velocity=1000.0,
+            max_depenetration_velocity=5.0,
+            disable_gravity=True,
+        ),
+        mass_props=sim_utils.MassPropertiesCfg(mass=0.25),
+    )
+    sphere = asset_registry.get_asset_by_name("sphere")(spawner_cfg=no_gravity_cfg)
```
alexmillane, line 139: You shouldn't need to copy all these configs just to turn the gravity off. I think you should be able to access the particular property as an attribute.

`sphere.spawner_cfg.rigid_props.disable_gravity=True`

That should work (and if it doesn't we have a problem).

isaaclab_arena/tests/test_events.py
```diff
@@ -107,6 +110,76 @@ def _test_set_object_pose_per_env_event(simulation_app):
...
+            with torch.inference_mode():
+                actions = torch.zeros(env.action_space.shape, device=env.unwrapped.device)
+                env.step(actions)
+
+        final_position = env.unwrapped.scene[sphere.name].data.root_pose_w[0, :3].clone()
+        final_position[:3] -= env.unwrapped.scene.env_origins[0]
+        displacement = final_position - initial_position
+        assert (
+            displacement[0].item() < -OBJECT_VELOCITY_MIN_DISPLACEMENT  # - x because initial velocity is in -x
+        ), f"Object did not move with given velocity: displacement in x was {displacement[0].item()}"
```
alexmillane, line 171: This seems like a bit of a loose test. Shouldn't we be able to bound this amount more tightly? Shouldn't the amount moved be roughly equal to the `num_steps * time_per_step * velocity`?

PR #490, opened 2026-03-20
object Velocity class like Pose

DavidRecasens:
## Add `Velocity` dataclass

Introduce a `Velocity` dataclass (analogous to `Pose`) to replace raw tuples for specifying object velocities, with support for both linear and angular components.

### Changes

- Add `isaaclab_arena/utils/velocity.py` with `linear_xyz`, `angular_xyz` fields, validation, `zero()` factory, and `to_tensor()`.
- Rename `set_initial_linear_velocity(tuple)` to `set_initial_velocity(Velocity)` in `ObjectBase`.
- Update `set_object_pose` in `terms/events.py` to accept `Velocity`.
- Update usages in `dynamic_balls_environment.py` and `test_events.py`.

### Breaking changes

- `ObjectBase.set_initial_linear_velocity` renamed to `set_initial_velocity` and now takes a `Velocity` instead of a `tuple`.


alexmillane (approved):
Thanks for adding that! Looks great!

PR #495, opened 2026-03-23
Batch Support for AxisAlignedBoundingBox

zhx06:
## Summary
Add batch dimension support to AxisAlignedBoundingBox

## Detailed description
  - To support multi-environment batched placement, AxisAlignedBoundingBox needs to handle N bounding boxes simultaneously instead of only one.
- What has been changed?
  - Refactored AxisAlignedBoundingBox from a dataclass to a class with (N, 3) float32 tensor storage. Returns tuples/floats for N=1, tensors for N>1.
  - Fixed float32 rounding issue in `_validate_on_relations` Z boundary check.
  - Added `test_bounding_box.py` for single-env and multi-env coverage.
  - No behavior change for existing code: backward compatible for single env (N=1).
  - Enables future multi-env batch support for the coming relation solver MR.

### DIFF
--- a/isaaclab_arena/utils/bounding_box.py
+++ b/isaaclab_arena/utils/bounding_box.py
@@ -36,25 +36,25 @@ class AxisAlignedBoundingBox:
         assert self._min_point.shape[-1] == 3
 
     @staticmethod
-    def _to_batched_tensor(v: tuple[float, float, float] | torch.Tensor) -> torch.Tensor:
+    def _to_batched_tensor(value: tuple[float, float, float] | torch.Tensor) -> torch.Tensor:
         """Convert tuple, 1-D tensor, or (N, 3) tensor to (N, 3) float32 tensor."""
-        if isinstance(v, tuple):
-            return torch.tensor([v], dtype=torch.float32)
-        if v.dim() == 1:
-            return v.unsqueeze(0).float()
-        return v.float()
+        if isinstance(value, tuple):
+            return torch.tensor([value], dtype=torch.float32)
+        if value.dim() == 1:
+            return value.unsqueeze(0).float()
+        return value.float()
 
-    def _format_output(self, t: torch.Tensor):
+    def _format_output(self, tensor: torch.Tensor):
         """Return tuple when N=1, tensor when N>1."""
         if self._min_point.shape[0] == 1:
-            return tuple(t[0].tolist())
-        return t
+            return tuple(tensor[0].tolist())
+        return tensor
 
-    def _format_scalar(self, t: torch.Tensor):
+    def _format_scalar(self, tensor: torch.Tensor):
         """Return float when N=1, tensor when N>1."""
         if self._min_point.shape[0] == 1:
-            return t.item()
-        return t
+            return tensor.item()
+        return tensor
 
     @property
     def min_point(self) -> tuple[float, float, float] | torch.Tensor:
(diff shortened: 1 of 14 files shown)


isaaclab_arena/utils/bounding_box.py
```diff
@@ -9,57 +9,87 @@
...
 
-@dataclass
 class AxisAlignedBoundingBox:
-    """Axis-aligned bounding box storing local extents. Use get_corners_at(pos) for world-space corners."""
+    """Axis-aligned bounding box storing local extents. Use get_corners_at(pos) for world-space corners.
 
-    min_point: tuple[float, float, float]
-    """Local minimum extent (x, y, z) relative to object origin."""
+    Stores min/max extents as (N, 3) tensors where N is the number of environments.
+    Properties return tuples/floats when N=1 and tensors when N>1.
```
cvolkcvolk: I feel like the dual-mode return type is a reliability issue. Properties return tuple/float when N=1 and tensor when N>1. Code that works in single-env silently changes behavior when N increases. Recommend to always returning tensors

cvolkcvolk: Additionally we could simplify below some parts like `self._min_point = self._to_batched_tensor(min_point)`,as its always a tensor (the input as well).

cvolkcvolk: This would require some updates downstream, but it would be worth it in my opinion

isaaclab_arena/utils/bounding_box.py
```diff
@@ -9,57 +9,87 @@
...
+    @staticmethod
+    def _to_batched_tensor(value: tuple[float, float, float] | torch.Tensor) -> torch.Tensor:
+        """Convert tuple, 1-D tensor, or (N, 3) tensor to (N, 3) float32 tensor."""
+        if isinstance(value, tuple):
+            return torch.tensor([value], dtype=torch.float32)
+        if value.dim() == 1:
+            return value.unsqueeze(0).float()
+        return value.float()
+
+    def _format_output(self, tensor: torch.Tensor):
```
cvolkcvolk: Suggestion to rename this function to _what_ is formatted

cvolkcvolk: However can remove this if the class only handles tensors

isaaclab_arena/utils/bounding_box.py
cvolkcvolk: I think this would break silently for N>1?  When N>1, min_point returns a tensor and the subsequent torch.rand(3) arithmetic produces wrong-shaped results ((N,3) instead of (3,)). 
I would suggest to move to single input and output type (tensors)

alexmillane (approved):

kellyguo11 (commented):
## 🤖 Isaac Lab Review Bot

### Summary
This PR refactors `AxisAlignedBoundingBox` from a `@dataclass` with tuple storage to a class with `(N, 3)` float32 tensor storage, enabling future multi-env batched placement. All 14 changed files are consistent updates propagating the new tensor API. CI all green.

### Design Assessment
**Design is sound.** The always-tensor approach is clean — no dual code paths, no conditional returns based on N. The `_to_batched_tensor` helper keeps the constructor ergonomic for the N=1 case. Broadcasting in `overlaps()` is correct for N-vs-1 comparisons.

### Findings

🟡 **Warning: `_to_batched_tensor` doesn't handle list input** — `bounding_box.py:49-54`
While the type hints specify `tuple | Tensor`, passing a Python list would raise `AttributeError: 'list' object has no attribute 'dim'`. Consider accepting `Sequence[float]` or adding a list→tuple/tensor conversion for robustness:
```python
if isinstance(value, (tuple, list)):
    return torch.tensor([value], dtype=torch.float32)
```

🟡 **Warning: `_validate_on_relations` Z-fix changes semantics subtly** — `object_placer.py:280-286`
The old code used `translated()` world bboxes (float32 tensor math), the new code manually adds Python floats. This fixes float32 rounding ✅ but the X/Y checks still use the translated tensor path (`child_world.min_point[0, 0]`). Consider making X/Y consistent with Z for uniform precision, or add a comment explaining why Z needs special treatment.

🔵 **Suggestion: Constructor could validate `min ≤ max`** — `bounding_box.py:37-38`
The constructor asserts shape correctness but not value correctness. `min_point > max_point` per-axis creates an inverted bbox that `overlaps()`, `size`, etc. interpret incorrectly. Pre-existing issue, but with tensor storage a simple `assert (self._min_point <= self._max_point).all()` would catch bugs early.

🔵 **Suggestion: `get_random_pose_within_bounding_box` only uses env 0** — `bounding_box.py:256`
This function takes `bbox.min_point[0]` / `bbox.max_point[0]`, which is correct for N=1 but silently ignores other envs for N>1. Consider adding `assert bbox.num_envs == 1` or an `env_idx` parameter.

### Test Coverage
- **New tests:** `test_bounding_box.py` — 8 tests covering single-env and multi-env properties, transforms, overlaps, and corners. Good coverage.
- **Updated tests:** `test_no_collision_loss.py` and `test_object_placer_init.py` properly updated for `[0, axis]` indexing.
- **Minor gap:** No test for `scaled()` with batched (N>1) scale tensor, and no regression test for the float32 Z-rounding fix. Not blocking.
- **Verdict:** Coverage adequate. Test quality is good.

### CI Status
All checks passing ✅

### Verdict

**COMMENT**

Clean, well-executed refactor. The tensor-first approach is the right call for future batch support. No critical issues — the warnings are minor and the suggestions are nice-to-haves.

---
*Review by Isaac Lab Review Bot — [Expert Analysis] + [Error Handling Audit] + [Test Coverage Check]*

PR #512, opened 2026-03-30
Batch Support for Relation Solver

zhx06:
## Summary
Add batched multi-env placement to relation solver (`1 -> num_envs`)

### Solver
- `RelationSolverState` positions stored as `(num_envs, num_optimizable, 3)`; `get_position()` returns `(num_envs, 3)`
- `RelationSolver._compute_total_loss()` accumulates `(num_envs,)` loss per env, returns mean of losses; per-env loss exposed via `last_loss_per_env`
- All loss strategies (`On`, `NextTo`, `NoCollision`, `AtPosition`) updated to accept `(num_envs, 3)` with `(3,)` backward compat

### ObjectPlacer
- `place()` accepts `num_envs` and `result_per_env`; returns `MultiEnvPlacementResult` (one `PlacementResult` per env) or a single `PlacementResult`
- Pool-based placement: generates `max_placement_attempts * num_results` candidates in one batched solver call, sorts by (valid-first, lowest-loss), and selects the best `num_results`
- `result_per_env=False` solves a single layout applied to all environments
- `_validate_placement` skips `On`-related pairs in overlap check (fix for false-positive collisions)
- `mean_loss` calculation filters out non-finite values

### Unified Placement Path
- New `PosePerEnv` dataclass holds a list of `Pose` objects, one per environment
- `ObjectBase._init_event_cfg` handles `PosePerEnv` via `set_object_pose_per_env`, so each object registers its own reset event for both single-env and multi-env cases
- Removed `placement_events.py` — its functionality is now handled by object-level event registration
- `ArenaEnvBuilder._solve_relations()` simplified: no longer builds a separate placement event cfg or distinguishes single/multi-env code paths

### Example Update
- Update default objects in `gr1_table_multi_object_no_collision_environment`, add TODO for potential object collision issues after running the solver


isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -101,9 +103,35 @@ def _solve_relations(self) -> None:
...
-        result = placer.place(objects=objects_with_relations)
-
-        if result.success:
+        num_envs = self.args.num_envs
+        result = placer.place(objects_with_relations, num_envs=num_envs)
+        if isinstance(result, MultiEnvPlacementResult) and result.results:
+            positions_all_envs_by_name = [
+                {obj.name: result.results[e].positions[obj] for obj in result.results[0].positions}
+                for e in range(len(result.results))
+            ]
```
alexmillane: Wdyt about moving this into `MultiEnvPlacementResult`? It's a bit difficult to decipher. Perhaps would be nicers to hide inside a method?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -101,9 +103,35 @@ def _solve_relations(self) -> None:
...
+            object_names = [obj.name for obj in objects_with_relations]
+            anchor_names = [a.name for a in get_anchor_objects(objects_with_relations)]
+            self._placement_event_cfg = make_placement_event_cfg(
+                positions_all_envs_by_name,
+                object_names,
+                anchor_names,
+            )
+        else:
+            self._placement_event_cfg = None
```
alexmillane: I can see what the motivation is here, but we have two quite different code paths for the single environment and multi-environment case.

I think we should unify these code paths.

Both code paths (single and multi placement) should run the solver to produce a result (of type `PlacementResult` or `MultiEnvPlacementResult`).

This result should be passed to `ObjectPlacer._apply_positions`.

The placement should be passed to the object through it's `Object.set_initial_pose`, which will have to be expanded to take something like a `PosePerEnv` or something like that.

Right now, depending on whether your optimizing for a single or multiple environments results in two very different paths. In the single env case, the responsibility for applying the placement is in the `ObjectPlacer` and `Object` registers the event. In the multi-env case the placement is effectively applied in the `ArenaEnvBuilder`, which also registers the event. 

I notice you have a note somewhere that perhaps these options could be unified. I suggest we unify now.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -152,12 +180,15 @@ def compose_manager_cfg(self) -> IsaacLabArenaManagerBasedRLEnvCfg:
...
         )
-        events_cfg = combine_configclass_instances(
-            "EventsCfg",
+        events_sources = [
             embodiment.get_events_cfg(),
             self.arena_env.scene.get_events_cfg(),
             task.get_events_cfg(),
-        )
+        ]
+        placement_event = getattr(self, "_placement_event_cfg", None)
```
alexmillane: Prefer to set this in the constructor (`__init__`) as None unconditionally, such that the object is garunteed to have the member, rather than checking for the attribute every time

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -152,12 +180,15 @@ def compose_manager_cfg(self) -> IsaacLabArenaManagerBasedRLEnvCfg:
...
+        if placement_event is not None:
+            events_sources.append(placement_event)
+        events_cfg = combine_configclass_instances("EventsCfg", *events_sources)
```
alexmillane: See comment above. I think that the placement result should be registered through the objects, as it is the single env case.

isaaclab_arena/relations/object_placer.py
```diff
@@ -43,15 +45,18 @@ def __init__(self, params: ObjectPlacerParams | None = None):
     def place(
         self,
         objects: list[Object | ObjectReference],
-    ) -> PlacementResult:
+        num_envs: int = 1,
```
alexmillane, line 69: What do you think about the case that we have `num_envs` > 1, but we want a placement that is the same across all environments. I feel that that might be a capability we'd like to preserve. Wdyt @cvolkcvolk?

If we do want to preserve that functionality we should introduce another parameter `result_per_env: bool = True`

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +81,94 @@ def place(
...
+                if self._solver.last_loss_per_env is not None
+                else [float("inf")] * num_envs
+            )
 
-            # Solve
-            positions = self._solver.solve(objects, initial_positions)
-            loss = self._solver.last_loss_history[-1] if self._solver.last_loss_history else float("inf")
+            # Check if placement is valid (per env); update best valid and best-by-loss fallback
+            for e in range(num_envs):
+                loss_e = per_env_loss[e] if e < len(per_env_loss) else float("inf")
```
alexmillane: Try to avoid abbreviations

`e` -> `env_idx` (ironically we use "env" as an abbreviation for environment across the Isaac Lab stack so thats OK.)

`loss_e` -> `loss`

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +81,94 @@ def place(
...
-        best_positions: dict[Object | ObjectReference, tuple[float, float, float]] = {}
-        best_loss = float("inf")
-        success = False
+        # Placement loop with retries (per-env tracking)
+        best_valid_loss_per_env: list[float] = [float("inf")] * num_envs
+        best_valid_positions_per_env: list[dict | None] = [None] * num_envs
+        best_any_loss_per_env: list[float] = [float("inf")] * num_envs
+        best_any_positions_per_env: list[dict] = [dict() for _ in range(num_envs)]
 
         for attempt in range(self.params.max_placement_attempts):
```
alexmillane: Suggestion to add a longer comment about this loop as it's doing something non-obvious.

From my understanding: We run solver attempts until each environment reaches success at least onces OR we hit the max number of attempts.

Is that correct?

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +81,94 @@ def place(
...
 
             if self.params.verbose:
-                print(f"Attempt {attempt + 1}/{self.params.max_placement_attempts}: loss = {loss:.6f}")
-
-            # Check if placement is valid
-            if self._validate_placement(positions):
-                best_loss = loss
-                best_positions = positions
-                success = True
+                mean_loss = sum(per_env_loss) / num_envs
```
alexmillane: This sum will be a bit ruined by the presence of infinities no?

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +81,94 @@ def place(
...
-                best_positions = positions
+        # Per env: use best valid if any, else best-by-loss fallback
+        final_per_env: list[dict] = [
+            (
+                best_valid_positions_per_env[e]
+                if best_valid_positions_per_env[e] is not None
+                else best_any_positions_per_env[e]
+            )
+            for e in range(num_envs)
+        ]
```
alexmillane: The idea here is that we return some result, even if some of the environments do not result in a valid placement?

Do you think that that's a safe approach.

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +81,94 @@ def place(
...
-        if rng_state is not None:
-            torch.set_rng_state(rng_state)
-
-        return PlacementResult(
-            success=success,
-            positions=best_positions,
-            final_loss=best_loss,
-            attempts=attempt + 1,
-        )
+        # TODO(@zhx06): Consider applying via event for consistency with multi_env.
```
alexmillane: I think we should do this in this MR. See comments above.

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +81,94 @@ def place(
...
+            # Check if placement is valid (per env); update best valid and best-by-loss fallback
+            for e in range(num_envs):
+                loss_e = per_env_loss[e] if e < len(per_env_loss) else float("inf")
+                valid = self._validate_placement(positions_per_env[e])
+                if valid and loss_e < best_valid_loss_per_env[e]:
+                    best_valid_loss_per_env[e] = loss_e
+                    best_valid_positions_per_env[e] = positions_per_env[e]
+                if loss_e < best_any_loss_per_env[e]:
+                    best_any_loss_per_env[e] = loss_e
+                    best_any_positions_per_env[e] = positions_per_env[e]
```
alexmillane: One thing that is inefficient about this loop is that it if a single environment (say "A)) fails N times, by chance, the whole thing fails. Even if environment "B" has 10 successful solutions. Because the environments are homogenous, "A" could just use the solution from "B".

I wonder if we shouldn't just calculate `FACTOR * num_env` solutions, and take the best `num_envs`. There's no need to associate a particular solution to a particular environment (because they're homogeneous).

isaaclab_arena/relations/placement_events.py
```diff
@@ -0,0 +1,93 @@
...
+    anchor_set = set(anchor_names or [])
+    ordered_names = [n for n in object_names if n in anchor_set]
+    ordered_names += [n for n in object_names if n not in anchor_set]
+    identity_quat_wxyz = (1.0, 0.0, 0.0, 0.0)
+    for name in ordered_names:
+        if name not in env.scene.keys():
+            continue
+        asset = env.scene[name]
+        if not hasattr(asset, "write_root_pose_to_sim"):
+            continue
```
alexmillane: under which conditions would an object in out list not have this method? Is that something we want to handle?

isaaclab_arena/relations/placement_events.py
```diff
@@ -0,0 +1,93 @@
...
+    assert (
+        len(positions_all_envs_by_name) == env.num_envs
+    ), f"Expected {env.num_envs} layout dicts, got {len(positions_all_envs_by_name)}"
+    env_id_list = _env_ids_to_list(env, env_ids)
+    if not env_id_list:
+        return
+    env_ids_t = torch.tensor(env_id_list, device=env.device)
+    anchor_set = set(anchor_names or [])
+    ordered_names = [n for n in object_names if n in anchor_set]
+    ordered_names += [n for n in object_names if n not in anchor_set]
```
alexmillane: Why do the names have to be ordered. Consider writing a comment.

alexmillane (commented):
Good work.

I have some suggestions for how we might improve things.

isaaclab_arena/environments/arena_env_builder.py
zhx06: Done, currently it does not build layout dicts

isaaclab_arena/environments/arena_env_builder.py
zhx06: We unified these code paths in the current code. Each object's _init_event_cfg generates the appropriate reset event.

isaaclab_arena/environments/arena_env_builder.py
zhx06: This is removed in the new version of code, we no longer have `placement_event`

isaaclab_arena/environments/arena_env_builder.py
zhx06: Done. Both single-env and multi-env paths now register placement in object levels.

isaaclab_arena/relations/object_placer.py
zhx06: This one no longer exists when we have the pool-based selection.

isaaclab_arena/relations/object_placer.py
zhx06: This one no longer exists when we have the pool-based selection.

isaaclab_arena/relations/object_placer.py
zhx06: Done. Infinite losses are now filtered out via math.isfinite() before computing the mean

isaaclab_arena/relations/object_placer.py
zhx06: Done. We now generate `max_placement_attempts * num_envs` candidates in a single batched solver call

isaaclab_arena/relations/object_placer.py
zhx06: It is now implemented in this MR.

isaaclab_arena/relations/placement_events.py
zhx06: This part is outdated and is replaced by the pool-based approach

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -152,12 +166,12 @@ def compose_manager_cfg(self) -> IsaacLabArenaManagerBasedRLEnvCfg:
...
         )
-        events_cfg = combine_configclass_instances(
-            "EventsCfg",
+        events_sources = [
             embodiment.get_events_cfg(),
             self.arena_env.scene.get_events_cfg(),
             task.get_events_cfg(),
-        )
+        ]
+        events_cfg = combine_configclass_instances("EventsCfg", *events_sources)
```
alexmillane: Suggestion just to revert this change (it has no effect).

isaaclab_arena/relations/object_placer.py
alexmillane: Is see you've solved this! Thanks!

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +87,71 @@ def place(
...
+
+        initial_positions: list[dict] = []
+        for candidate_idx in range(num_candidates):
+            rng_state = None
+            if self.params.placement_seed is not None:
+                rng_state = torch.get_rng_state()
+                torch.manual_seed(self.params.placement_seed + candidate_idx)
+            initial_positions.append(self._generate_initial_positions(objects, anchor_objects_set, init_bounds))
+            if rng_state is not None:
+                torch.set_rng_state(rng_state)
```
alexmillane: Suggestion to abstract this logic into a function. Maybe `_generate_initial_positions` could be expanded to handle the single and multiple placement cases.

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +87,71 @@ def place(
...
+
+        all_positions: list[dict] = self._solver.solve(objects, initial_positions)  # type: ignore[assignment]  # overload returns list[dict] for list input
+        all_losses = (
+            self._solver.last_loss_per_env.cpu().tolist()
+            if self._solver.last_loss_per_env is not None
+            else [float("inf")] * num_candidates
+        )
```
alexmillane: Why is this needed? Seems like if we've run the solve, the last_loss_per_env should not be `None` or am I missing something?

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +87,71 @@ def place(
...
 
-        # Restore RNG state if we changed it
-        if rng_state is not None:
-            torch.set_rng_state(rng_state)
+        all_candidates: list[tuple[float, dict, bool]] = []
+        for idx in range(num_candidates):
+            loss = all_losses[idx] if idx < len(all_losses) else float("inf")
```
alexmillane: Do we need this check: `idx < len(all_losses)`. Shouldn't `all_losses` be defined to have the correct length?

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +87,71 @@ def place(
...
+            is_valid = self._validate_placement(all_positions[idx])
+            all_candidates.append((loss, all_positions[idx], is_valid))
+
+        # Sort: valid solutions first (by loss), then invalid (by loss)
+        all_candidates.sort(key=lambda c: (not c[2], c[0]))
+        selected = all_candidates[:num_results]
```
alexmillane: This is a bit difficult to read... I think because of the use the tuple. What do you think about creating a `@dataclass` which basically mimics this tuple but has named members. Then your sorting will be a bit more readable.

isaaclab_arena/relations/object_placer.py
```diff
@@ -76,61 +87,71 @@ def place(
...
+        final_per_env: list[dict] = [c[1] for c in selected]
+        results_per_env = [
+            PlacementResult(
+                success=c[2],
+                positions=c[1],
+                final_loss=c[0],
+                attempts=self.params.max_placement_attempts,
+            )
+            for c in selected
+        ]
```
alexmillane: See comment above. This accesses of `c` make things a bit difficult to read. Consider using a named composite.

isaaclab_arena/relations/object_placer.py
```diff
@@ -219,11 +240,25 @@ def _validate_no_overlap(
...
+                if isinstance(rel, On) and rel.parent in positions:
+                    on_pairs.add((id(obj), id(rel.parent)))
+                    on_pairs.add((id(rel.parent), id(obj)))
+
         objects = list(positions.keys())
         for i in range(len(objects)):
             for j in range(i + 1, len(objects)):
                 a, b = objects[i], objects[j]
+                if (id(a), id(b)) in on_pairs:
+                    continue
```
alexmillane, line 363: Add:
```
# Pairs related by an OnRelation are excluded from the check.
```

isaaclab_arena/relations/object_placer.py
```diff
@@ -250,29 +285,38 @@ def _validate_placement(
...
-        Rotation is taken from RotateAroundSolution marker if present, otherwise keep the identity rotation.
+        Handles both single-env and multi-env placement:
+        - Single-env: sets a fixed Pose or PoseRange (with RandomAroundSolution).
+        - Multi-env: sets a PosePerEnv with one Pose per environment.
+
+        Rotation is taken from RotateAroundSolution marker if present, otherwise identity.
         """
-        for obj, pos in positions.items():
+        num_envs = len(positions_per_env)
+        for obj in positions_per_env[0]:
```
alexmillane: This looks a bit weird because of the `[0]`. Could be clearer to split into two lines:

```
# Objects are the same for every environment. Extract them.
objects = [obj for obj in positions_per_env[0]]
# Apply pose for each object.
for obj in objects:
    ...
```

isaaclab_arena/relations/relation_solver.py
```diff
@@ -107,29 +108,38 @@ def _compute_total_loss(self, state: RelationSolverState, debug: bool = False) -
...
         objects: list[Object | ObjectReference],
-        initial_positions: dict[Object | ObjectReference, tuple[float, float, float]],
-    ) -> dict[Object | ObjectReference, tuple[float, float, float]]:
+        initial_positions: (
+            dict[Object | ObjectReference, tuple[float, float, float]]
+            | list[dict[Object | ObjectReference, tuple[float, float, float]]]
+        ),
+    ) -> (
+        dict[Object | ObjectReference, tuple[float, float, float]]
+        | list[dict[Object | ObjectReference, tuple[float, float, float]]]
```
alexmillane: These types are getting a bit awkward.

One way we can simply is: `Object | ObjectReference` -> `ObjectBase` (their parent class).

Do you see any opportunity to just use a list of length 1 for the single env case?

isaaclab_arena/relations/relation_solver_state.py
```diff
@@ -21,20 +21,30 @@ class RelationSolverState:
...
 
         Args:
             objects: List of all Object or ObjectReference instances to track. Must include at least one
                 object marked with IsAnchor() which serves as a fixed reference.
-            initial_positions: Starting positions for all objects (including anchors).
+            initial_positions: A single dict (backward compat, treated as single-env) or a list
+                of dicts (one per env). Length 1 = single-env, length > 1 = batched.
         """
+        if isinstance(initial_positions, dict):
+            initial_positions = [initial_positions]
```
alexmillane: Seems like maybe we should just use a list no matter what? What do you think?

alexmillane (commented):
Thank you for addressing the last comments. It's look a lot better (and simpler) now. Great!

I just have some smaller stylistic comments now.

isaaclab_arena/environments/arena_env_builder.py
zhx06: Done. This is reverted

isaaclab_arena/relations/object_placer.py
zhx06: Now the new `_generate_initial_positions` handles both cases.

isaaclab_arena/relations/object_placer.py
zhx06: This one is removed from the updated code.

isaaclab_arena/relations/object_placer.py
zhx06: Correct, This extra check is removed.

isaaclab_arena/relations/object_placer.py
zhx06: A new `@dataclass` is created to make it more readable.

isaaclab_arena/relations/object_placer.py
zhx06: The code has been updated following your suggestion.

isaaclab_arena/relations/relation_solver.py
zhx06: ll type annotations now use `ObjectBase`. 

`solve()` and `RelationSolverState` now always use `list[dict]`.

isaaclab_arena/relations/relation_solver_state.py
zhx06: This is removed and it takes `list[dict]` for all cases.

isaaclab_arena/relations/object_placer.py
zhx06: Done. This comment has been added.

isaaclab_arena/relations/relation_solver.py
```diff
@@ -68,9 +68,11 @@ def _compute_total_loss(self, state: RelationSolverState, debug: bool = False) -
             debug: If True, print detailed loss breakdown.
 
         Returns:
-            Total loss tensor.
+            Scalar loss tensor (mean over envs). Per-env loss stored in _last_loss_per_env.
         """
-        total_loss = torch.tensor(0.0)
+        N = state.num_envs
```
cvolkcvolk: Suggestion to not use abbreviations, eg to just name it `num_envs`

zhx06: Actually I've been thinking about the variable name here. The solver now takes a batch of `max_attempts * num_envs` candidates and we select `num_envs` ones with lowest losses.  Do you think we can rename it as `batch_size` in the solver/state layer?

isaaclab_arena/relations/object_placer.py
```diff
@@ -48,16 +65,24 @@ def __init__(self, params: ObjectPlacerParams | None = None):
...
         """Place objects according to their spatial relations.
 
         Args:
             objects: List of objects to place. Must include at least one object
                 marked with IsAnchor() which serves as a fixed reference.
+            num_envs: Number of environments. 1 for single-env; > 1 for batched
+                placement (one layout per env).
+            result_per_env: When True (default), each environment gets a distinct
+                layout. When False, a single best layout is solved and applied
+                identically to all environments (useful for deterministic evaluation).
```
cvolkcvolk, line 81: Suggestion to remove the statement about determinism. It implies that the solving for each env is not deterministic

isaaclab_arena/relations/relations.py
```diff
@@ -126,15 +125,16 @@ class NoCollision(Relation):
...
 
-    NOTE: If both A.add_relation(NoCollision(B)) and B.add_relation(NoCollision(A)) are present,
-    the solver will compute the loss twice and the relation graph becomes cyclic, which can cause
-    issues during environment creation. Deduplication or cycle detection should be addressed at a
-    higher level.
+    NOTE: RelationSolver._compute_total_loss iterates every relation on every object with no
+    deduplication. If both A.add_relation(NoCollision(B)) and B.add_relation(NoCollision(A))
+    are present, loss is computed twice. Bidirectional NoCollision can also make the relation
+    graph cyclic and cause issues when creating the environment. Deduplication and/or
+    higher-level handling of symmetric relations to be addressed in a future commit.
```
cvolkcvolk, line 132: Please make it a TODO, makes it easier to track later :-)

isaaclab_arena/relations/relations.py
```diff
@@ -126,15 +125,16 @@ class NoCollision(Relation):
...
     Note: Loss computation is handled by NoCollisionLossStrategy in relation_loss_strategies.py.
 
-    NOTE: If both A.add_relation(NoCollision(B)) and B.add_relation(NoCollision(A)) are present,
-    the solver will compute the loss twice and the relation graph becomes cyclic, which can cause
-    issues during environment creation. Deduplication or cycle detection should be addressed at a
-    higher level.
+    NOTE: RelationSolver._compute_total_loss iterates every relation on every object with no
+    deduplication. If both A.add_relation(NoCollision(B)) and B.add_relation(NoCollision(A))
+    are present, loss is computed twice. Bidirectional NoCollision can also make the relation
+    graph cyclic and cause issues when creating the environment. Deduplication and/or
```
cvolkcvolk, line 131: Could we assert that this does not happen or gives runtime warnings if that happens?

isaaclab_arena/assets/object_base.py
cvolkcvolk, line 207: Unrelated to this MR, but its (x, y, z, qx, qy, qz, qw) now on IsaacLab3

alexmillane (approved):

PR #517, opened 2026-03-31
Add per-job language_instruction to GR00T jobs config

cvolkcvolk:
## Summary

Policy YAML configs describe the model and embodiment and should not hold task-level _language instructions_. Task descriptions belong either in the task itself (auto-generated) or in the per-job `language_instruction` field of the jobs config.

Changes:
- Remove `language_instruction` from all GR00T policy YAML configs.
- Add `task_description` to `SequentialTaskBase` so sequential tasks can provide their own description via `get_task_description()`, matching the pattern already used by atomic tasks.
- Add `language_instruction` as an optional per-job field in the jobs config, letting individual evaluation jobs set the task description independently of the policy config.
- Add per-object language instructions to Droid GR00T evaluation jobs

### DIFF
--- a/isaaclab_arena_environments/gr1_put_and_close_door_environment.py
+++ b/isaaclab_arena_environments/gr1_put_and_close_door_environment.py
@@ -60,9 +60,13 @@ class GR1PutAndCloseDoorEnvironment(ExampleEnvironmentBase):
                 self,
                 subtasks: list[TaskBase],
                 episode_length_s: float | None = None,
+                task_description: str | None = None,
             ):
                 super().__init__(
-                    subtasks=subtasks, episode_length_s=episode_length_s, desired_subtask_success_state=[True, True]
+                    subtasks=subtasks,
+                    episode_length_s=episode_length_s,
+                    desired_subtask_success_state=[True, True],
+                    task_description=task_description,
                 )
 
             def get_viewer_cfg(self):
@@ -179,7 +183,11 @@ class GR1PutAndCloseDoorEnvironment(ExampleEnvironmentBase):
         )
 
         # Create sequential task
-        sequential_task = PutAndCloseDoorTask(subtasks=[pick_and_place_task, close_door_task], episode_length_s=10.0)
+        sequential_task = PutAndCloseDoorTask(
+            subtasks=[pick_and_place_task, close_door_task],
+            episode_length_s=10.0,
+            task_description=f"Place the {pickup_object.name} in the refrigerator and close the door.",
+        )
 
         # Create and return environment
         isaaclab_arena_environment = IsaacLabArenaEnvironment(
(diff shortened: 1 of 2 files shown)


isaaclab_arena/evaluation/policy_runner.py
```diff
@@ -67,9 +68,10 @@ def rollout_policy(
     try:
         obs, _ = env.reset()
         policy.reset()
-        # set task description (could be None) from the task being evaluated
-        # Use unwrapped to reach the base env through any gym wrappers (e.g. OrderEnforcing)
-        policy.set_task_description(env.unwrapped.cfg.isaaclab_arena_env.task.get_task_description())
+        # Determine language instruction: job-level override takes precedence over the task's own
+        # description. Use unwrapped to reach the base env through any gym wrappers (e.g. OrderEnforcing).
+        task_description = language_instruction or env.unwrapped.cfg.isaaclab_arena_env.task.get_task_description()
```
xyao-nv, line 73: Is there any chance both of them are None?

xyao-nv: If both of them are None, policy will [grab from the yaml](https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena_gr00t/policy/gr00t_closedloop_policy.py#L146).
And if you guarantee that it will be a language description at this point, you can also delete the [entry in the gr00t policy cfg class ](https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena_gr00t/policy/config/gr00t_closedloop_policy_config.py#L23 )

alexmillane: I feel that we should _require_ (and enforce) that our tasks provide a language instruction.

At the same time I think that this should be overridable by a user. So, if it's not suuper ugly, I would be a fan of leaving the override option in the config file.

Perhaps the order of priority could be (highest to lowest):
* job config file
* policy config file
* task default description

But maybe that results in complex error prone code. Wdyt @cvolkcvolk?

isaaclab_arena_environments/gr1_put_and_close_door_environment.py
```diff
@@ -179,7 +183,11 @@ def __post_init__(self):
         )
 
         # Create sequential task
-        sequential_task = PutAndCloseDoorTask(subtasks=[pick_and_place_task, close_door_task], episode_length_s=10.0)
+        sequential_task = PutAndCloseDoorTask(
+            subtasks=[pick_and_place_task, close_door_task],
+            episode_length_s=10.0,
+            task_description=f"Place the {pickup_object.name} in the refrigerator and close the door.",
```
alexmillane, line 189: It's an aside but I feel that the `SequentialCompositeTask` should be able to create a language instruction from it's sub-tasks. But that's something for v0.3

alexmillane (approved):
I think that this is a good move.

Could we add to `TaskBase` enforcement the inheriting class implement a language instruction?

xyao-nv (approved):
Remove changes request assuming we are deferring the language cleanups to later .

PR #518, opened 2026-03-31
docs: add dedicated page for using Arena from an external repository

cvolkcvolk:
## Summary

- Adds a new **Getting Started** page (`arena_in_your_repo.rst`) documenting the recommended pattern for consuming Arena as an unmodified git submodule from an external project, inspired by and based on the integration pattern used in [nvblox_next](https://github.com/nvidia-isaac/nvblox_next/tree/e3d4fec646004956ac24ed3446dbb41c531d5908/datagen)

## Notes
- **Not yet tested** end-to-end — the examples reflect the observed nvblox_next patterns but have not been validated in a fresh external repo

### DIFF
--- a/docs/pages/concepts/concept_policy_design.rst
+++ b/docs/pages/concepts/concept_policy_design.rst
@@ -77,7 +77,7 @@ Usage Examples
       .. code-block:: bash
 
          # Zero action policy for environment validation
-         python policy_runner.py --policy_type zero_action kitchen_pick_and_place --num_steps 1000
+         python isaaclab_arena/evaluation/policy_runner.py --policy_type zero_action --num_steps 1000 kitchen_pick_and_place --object cracker_box
 
    .. tab:: Distribute Multi-GPU
 
@@ -85,7 +85,7 @@ Usage Examples
 
          # Zero action policy for environment validation
          # --nproc_per_node represents the number of available GPUs
-         python -m torch.distributed.run --nnode=1 --nproc_per_node=2 policy_runner.py --policy_type zero_action kitchen_pick_and_place --num_steps 1000 --distributed --headless
+         python -m torch.distributed.run --nnode=1 --nproc_per_node=2 isaaclab_arena/evaluation/policy_runner.py --distributed --headless --policy_type zero_action --num_steps 1000 kitchen_pick_and_place --object cracker_box
 
 **Demonstration Replay**
 
@@ -96,7 +96,7 @@ Usage Examples
       .. code-block:: bash
 
          # Replay recorded demonstrations
-         python policy_runner.py --policy_type replay --replay_file_path demos.h5 kitchen_pick_and_place
+         python isaaclab_arena/evaluation/policy_runner.py --policy_type replay --replay_file_path demos.h5 kitchen_pick_and_place --object cracker_box
 
    .. tab:: Distribute Multi-GPU
 
@@ -104,7 +104,7 @@ Usage Examples
 
          # Replay recorded demonstrations
          # --nproc_per_node represents the number of available GPUs
-         python -m torch.distributed.run --nnode=1 --nproc_per_node=2 policy_runner.py --policy_type replay --replay_file_path demos.h5 kitchen_pick_and_place --distributed --headless
+         python -m torch.distributed.run --nnode=1 --nproc_per_node=2 isaaclab_arena/evaluation/policy_runner.py --distributed --headless --policy_type replay --replay_file_path demos.h5 kitchen_pick_and_place --object cracker_box
 
 **Neural Policy Execution**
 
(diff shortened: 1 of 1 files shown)


docs/pages/quickstart/arena_in_your_repo.rst
```diff
@@ -0,0 +1,128 @@
...
+
+   from isaaclab_arena_environments.cli import (
+       ExampleEnvironments,
+       get_arena_builder_from_cli,
+       get_isaaclab_arena_environments_cli_parser,
+   )
+   from my_package.isaaclab_arena_environments.my_environment import MyEnvironment
+
+   # Register the custom environment
+   ExampleEnvironments[MyEnvironment.name] = MyEnvironment
```
cvolkcvolk: @alexmillane Is this required?

docs/pages/quickstart/arena_in_your_repo.rst
```diff
@@ -0,0 +1,128 @@
+Using IsaacLab-Arena in Your Own Repository
+============================================
+
+The recommended way to consume IsaacLab-Arena from an external project is to include it as an
+**unmodified git submodule** and extend it purely through its registration API — without editing
+any file inside the Arena source tree.
```
alexmillane, line 6: Agreed!

docs/pages/quickstart/arena_in_your_repo.rst
```diff
@@ -0,0 +1,128 @@
...
+   my_project/
+   ├── submodules/
+   │   └── IsaacLab-Arena/          ← unmodified Arena submodule
+   ├── my_package/
+   │   ├── pyproject.toml
+   │   ├── isaaclab_arena_environments/
+   │   │   ├── __init__.py
+   │   │   └── my_environment.py    ← custom environment class
+   │   └── scripts/
+   │       └── run_datagen.py       ← entry script
```
alexmillane: oh gosh. This is non ideal. The requirement to copy the scripts is not ideal. I think/hope that this isn't required. Let me do some testing.

docs/pages/quickstart/arena_in_your_repo.rst
```diff
@@ -0,0 +1,128 @@
...
+   ├── docker/
+   │   ├── Dockerfile
+   │   └── install_isaaclab_arena.sh
```
alexmillane: Hmm. I feel we should maybe remove this script and add a section about how arena is installed. Such a section will state the assumption that the environment being installed to already has lab installed. Then the installation (I believe?) is as simple as `/isaac-sim/python.sh -m pip install -e submodules/IsaacLab-Arena`

alexmillane: Having looked at the nvblox_next version of this script it's not as easy as pip installing. But that is actually due to an issue with how we're handling dependencies. I'll fix that.

docs/pages/quickstart/arena_in_your_repo.rst
alexmillane: I think this shouldn't be required as we can run externally defined environments. Let me try it out first.

docs/pages/quickstart/arena_in_your_repo.rst
cvolkcvolk: [run_datagen.py](https://github.com/nvidia-isaac/nvblox_next/blob/e3d4fec646004956ac24ed3446dbb41c531d5908/datagen/run_isaaclab_arena_datagen.py) is is actually not the isaaclab copy. It does lots of nvlox specific data saving. Maybe the whole wording here is not correct calling it "entry script"

alexmillane: I feel that we should just remove this script and show an example of using an environment in our of our existing workflows that doesn't require a custom script. For example using our zero-action runner.

cvolkcvolk: Agree

xyao-nv (approved):

docs/pages/quickstart/arena_in_your_repo.rst
```diff
@@ -0,0 +1,177 @@
...
+               scene=scene,
+               task=NoTask(),
+           )
+           return isaaclab_arena_environment
+
+       @staticmethod
+       def add_cli_args(parser: argparse.ArgumentParser) -> None:
+           parser.add_argument("--object", type=str, default="cracker_box")
+
+External environments can be used in Isaac Lab Arena workflows by using a particular
```
cvolkcvolk, line 135:
```suggestion
External environments can be used in Isaac Lab-Arena workflows by using a particular
```

alexmillane: Will make this change in the next MR to avoid re-running CI.

alexmillane: We've made this "mistake" all over the place. Also the spacing with the hyphen is still not clear to me. We should make a call on the hyphen and then add a test to test for this in the docs.

PR #557, opened 2026-04-08
[CI]  Fix CI subprocess test hangs

xyao-nv:
## Summary
Subprocess-spawning tests hang indefinitely on CI.

### Problems

From Lab:

1. Lab reports "AppLauncher doesnt quit properly after app.close(), app.quit() doesn't help either."
2. Cold startup times for tests using IS can be upwards of 10 min on Lab CI machines.

Above issues apply to us, because tests hang during sub-process tests section, between the end of last test and the beginning of the next test. See detailed logs and analysis from reproducing locally [here](https://github.com/isaac-sim/IsaacLab-Arena/pull/568)

### Fixes

1. `SimulationApp` Force Exit: Skips `app.close()` (which can hang indefinitely in Kit's shutdown path) when the env var `ISAACLAB_ARENA_FORCE_EXIT_ON_COMPLETE=1` is set.
Calls a new `_kill_child_processes()` helper that walks `/proc` to `SIGKILL` all direct children before doing `os._exit(0)`, preventing orphaned Kit processes from holding GPU resources.

2. `run_subprocess` has a configuarable wall-clock timeouts and process isolation, such that when needed, it could trigger the force exit path above.

3. Add wall-clock timing and logging inside the SimulationApp start method. Keep track of how much startup time is taking on CI.

## Minor fixes

1. Add timing stats into pytest cmds such that it reports the slowests test func at the end of each section.

2. Parametrize multi-config tests: Convert nested for-loops in `test_zero_action_policy_kitchen_pick_and_place` (6 configs) and `test_zero_action_policy_gr1_open_microwave` (3 configs) into `@pytest.mark.parametrize.` Each config gets its own timeout, pass/fail, and timing.

3. Reduce num_envs in gr00t eval_runner test to speed up.

### Local validation
With the repro script https://github.com/isaac-sim/IsaacLab-Arena/pull/568, I do not have local stalling.
Log for more details.
[repro_20260410_041313.log](https://github.com/user-attachments/files/26620524/repro_20260410_041313.log)

### CI Before -- timeout
image

### CI After
image
image

## TODOs
- test_camera_observations takes 10mins to start the app due to Kit cold start. Experimenting with a warm start before tests process here https://github.com/isaac-sim/IsaacLab-Arena/pull/565
- Kit itself intermittently deadlocks during startup — not because of orphans, but because Kit's internal thread synchronization fails on low-CPU runners. Experimenting with retry here https://github.com/isaac-sim/IsaacLab-Arena/pull/570

### DIFF
--- a/isaaclab_arena/tests/test_policy_runner.py
+++ b/isaaclab_arena/tests/test_policy_runner.py
@@ -65,21 +65,18 @@ def test_zero_action_policy_press_button():
 
 
 @pytest.mark.with_subprocess
-def test_zero_action_policy_kitchen_pick_and_place():
+@pytest.mark.parametrize("embodiment", ["franka_ik", "gr1_pink", "gr1_joint"])
+@pytest.mark.parametrize("object_name", ["cracker_box", "tomato_soup_can"])
+def test_zero_action_policy_kitchen_pick_and_place(embodiment, object_name):
     # TODO(alexmillane, 2025.07.29): Get an exhaustive list of all scenes and embodiments
     # from a registry when we have one.
-    example_environment = "kitchen_pick_and_place"
-    embodiments = ["franka_ik", "gr1_pink", "gr1_joint"]
-    object_names = ["cracker_box", "tomato_soup_can"]
-    for embodiment in embodiments:
-        for object_name in object_names:
-            run_policy_runner(
-                policy_type="zero_action",
-                example_environment=example_environment,
-                embodiment=embodiment,
-                object_name=object_name,
-                num_steps=NUM_STEPS,
-            )
+    run_policy_runner(
+        policy_type="zero_action",
+        example_environment="kitchen_pick_and_place",
+        embodiment=embodiment,
+        object_name=object_name,
+        num_steps=NUM_STEPS,
+    )
 
 
 @pytest.mark.with_subprocess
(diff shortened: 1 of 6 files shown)


xyao-nv:
@greptile-apps

isaaclab_arena/tests/utils/subprocess.py
```diff
@@ -31,26 +31,70 @@
...
+    timeout_sec: int | None = None,
+    capture_output: bool = False,
+) -> subprocess.CompletedProcess | None:
+    """Run a command in a subprocess with timeout.
+
+    ``start_new_session=True`` isolates the child into its own process group.
+    The child-side ``SimulationAppContext`` uses this to SIGTERM its entire
+    group before ``os._exit()``, preventing orphaned Kit children (shader
+    compiler, GPU workers, …) from holding GPU resources and blocking the
+    next subprocess.
```
alexmillane: Suggestion to start this comment block with:

"This function runs the child subprocess with ``start_new_session=True`` such that the child is isolated..."

Right now the comment looks like it's describing an argument to _this_ function, which is a bit confusing because there is no such argument. Whereas it's actually describing an argument used inside the function.

Another option is to move this comment down below the subprocess call.

isaaclab_arena/tests/test_policy_runner.py
```diff
@@ -65,21 +65,18 @@ def test_zero_action_policy_press_button():
...
-                object_name=object_name,
-                num_steps=NUM_STEPS,
-            )
+    run_policy_runner(
+        policy_type="zero_action",
+        example_environment="kitchen_pick_and_place",
+        embodiment=embodiment,
+        object_name=object_name,
+        num_steps=NUM_STEPS,
+    )
```
alexmillane, line 79: Nice!

isaaclab_arena_gr00t/tests/test_gr00t_closedloop_policy.py
```diff
@@ -220,7 +220,7 @@ def test_g1_locomanip_gr00t_closedloop_policy_runner_multi_envs(gr00t_finetuned_
     assert result, "Test test_g1_locomanip_gr00t_closedloop_policy_runner_multi_envs failed"
 
 
-@pytest.mark.with_subprocess
+@pytest.mark.skip(reason="Skipping because of CI stalling")
```
alexmillane: Is this still required? This MR solves this issue right?

alexmillane (commented):
Thank you addressing this!

I'll spin the job a bunch of times to check that this solves the issue!

isaaclab_arena_gr00t/tests/test_gr00t_closedloop_policy.py
xyao-nv: yeah I forgot. revert

isaaclab_arena/tests/utils/subprocess.py
xyao-nv: done

alexmillane (approved):

PR #559, opened 2026-04-08
[CI] Revert eval_runner tests

xyao-nv:
## Summary
As CI seems to run smoothly agin, bring back previously disabled tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,7 @@ jobs:
           /isaac-sim/python.sh -m pytest -sv --durations=0 -m "with_cameras and not with_subprocess and not with_newton" \
             isaaclab_arena/tests/
 
-      - name: Run subprocess-spawning  PhysX tests
+      - name: Run subprocess-spawning PhysX tests
         run: |
           ISAACLAB_ARENA_SUBPROCESS_TIMEOUT=900 \
           /isaac-sim/python.sh -m pytest -sv --durations=0 -m with_subprocess \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,6 +157,7 @@ jobs:
 
       - name: Run subprocess-spawning  PhysX tests
         run: |
+          ISAACLAB_ARENA_SUBPROCESS_TIMEOUT=900 \
           /isaac-sim/python.sh -m pytest -sv --durations=0 -m with_subprocess \
             isaaclab_arena/tests/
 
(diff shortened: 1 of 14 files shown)


kellyguo11 (approved):
## Review: PR #559 — revert test

**Summary:** This PR re-enables 5 eval runner tests that were previously skipped with `@pytest.mark.skip(reason="Skipping because of CI stalling")` by replacing those skip markers with `@pytest.mark.with_subprocess`.

**Context understood:** The base branch `xyao/exp/ci_stall` added infrastructure to separate subprocess-spawning tests from persistent SimulationApp tests (via the `with_subprocess` marker and CI workflow changes). This PR is the natural follow-up that actually reverts the skips and uses the new marker.

### ✅ What looks good
- The change is straightforward: 5 mechanical replacements of `skip` → `with_subprocess`.
- The `with_subprocess` marker is properly registered in `pytest.ini` and the CI workflow already has a dedicated step that runs `-m with_subprocess` tests with `ISAACLAB_ARENA_SUBPROCESS_TIMEOUT=900`.
- The marker semantics are correct — these tests all use `run_eval_runner_and_check_no_failures()`, which spawns `eval_runner.py` via `subprocess.run()`.

### ⚠️ Potential concern: missing timeout / process group isolation in `run_eval_runner_and_check_no_failures`

The utility `run_subprocess()` in `isaaclab_arena/tests/utils/subprocess.py` was carefully written with:
- `start_new_session=True` to isolate GPU child processes
- `timeout=_SUBPROCESS_TIMEOUT_SEC` to prevent hangs

However, `run_eval_runner_and_check_no_failures()` in this test file uses raw `subprocess.run(args, capture_output=True, text=True, check=True)` — **no timeout, no `start_new_session=True`**. If the original CI stalling was caused by subprocess hangs or orphaned GPU processes, these tests could still stall even with the marker separation, since the subprocess itself has no timeout.

**Suggestion:** Consider either:
1. Refactoring `run_eval_runner_and_check_no_failures()` to use `run_subprocess()` from the utils module, or
2. At minimum, adding `timeout=_SUBPROCESS_TIMEOUT_SEC` and `start_new_session=True` to the `subprocess.run()` call.

This isn't a blocker if the CI workflow-level `timeout-minutes: 60` is sufficient, but it would be more robust to have per-subprocess timeouts too.

### PR description
The PR description still has the default template text. A brief note about reverting the skips now that the CI stalling fix is in place would help future readers.

**Overall:** The change itself is correct and minimal. Approving since the marker infrastructure is in place and CI will validate.

peterd-NV (approved):

PR #571, opened 2026-04-10
Refactor locomanipulation task

alexmillane:
## Summary
Generalize and simplify our loco-manipulation task.

## Detailed description
- Remove G1, galileo, and object specific details, to change `g1_locomanip_pick_and_place_task`, which was specific to the galileo environment, into `locomanip_pick_and_place_task`, which is general to any environment.
- Required expanding our contact sensor creation to support contact _against_ objects with rigid bodies not at their root.
- `LocomanipPickAndPlaceTask` is now a shallow extention of `PickAndPlaceTask`. It inherits all its functionality from the pick-and-place task, except for the mimic config, which is special for 2 hands + navigation.


docker/run_docker.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 DOCKER_IMAGE_NAME='isaaclab_arena'
-DOCKER_VERSION_TAG='latest'
+DOCKER_VERSION_TAG='latest1'
```
kellyguo11: **Unintentional change?** `latest` → `latest1` looks like a local dev override that shouldn't be committed. Please revert or explain the intent.

isaaclab_arena/tasks/g1_locomanip_pick_and_place_task.py
```diff
@@ -32,6 +33,7 @@ def __init__(
         task_description: str | None = None,
     ):
         super().__init__(episode_length_s=episode_length_s)
+        print("HERE")
```
kellyguo11: **Debug print — remove before merge.**
```suggestion
```

isaaclab_arena/tasks/terminations.py
```diff
@@ -35,13 +35,18 @@ def object_on_destination(
     # NOTE(alexmillane, 2025-08-04): We expect the binary flags to have shape (N, )
     # where N is the number of envs.
     force_matrix_norm = torch.norm(wp.to_torch(sensor.data.force_matrix_w), dim=-1).reshape(-1)
+    print(f"force_matrix_norm: {force_matrix_norm}")
```
kellyguo11: **Debug prints — remove before merge.** These five `print()` calls in a hot-path termination function will spam stdout every step across every env. Either remove them or gate behind a `logger.debug()` call.

Same applies to the prints on lines 40, 43, 45, and 47.

isaaclab_arena/assets/object_base.py
```diff
@@ -216,13 +218,15 @@ def set_object_pose(self, env: ManagerBasedEnv, pose: Pose, env_ids: torch.Tenso
             torch.zeros(env.unwrapped.num_envs, 6, device=env.unwrapped.device), env_ids=env_ids
         )
 
-    def get_contact_sensor_cfg(self, contact_against_prim_paths: list[str] | None = None) -> ContactSensorCfg:
+    def get_contact_sensor_cfg(self, contact_against_object: ObjectBase | None = None) -> ContactSensorCfg:
         assert self.object_type == ObjectType.RIGID, "Contact sensor is only supported for rigid objects"
-        if contact_against_prim_paths is None:
-            contact_against_prim_paths = []
+        assert (
```
kellyguo11: **Bug: crashes when `contact_against_object` is `None`.** The assertion `contact_against_object.object_type == ObjectType.RIGID` runs unconditionally, but `contact_against_object` can be `None` (the default). This will raise `AttributeError: 'NoneType' object has no attribute 'object_type'`.

Wrap it:
```python
if contact_against_object is not None:
    assert (
        contact_against_object.object_type == ObjectType.RIGID
    ), "Contact sensor is only supported for rigid objects"
filter_prim_paths = [contact_against_object.get_prim_path()] if contact_against_object else []
```

isaaclab_arena/assets/object_set.py
```diff
@@ -90,16 +90,18 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
         # We override this function from the parent class because in some assets, the rigid body
         # is not at the root of the USD file. To be robust to this, we find the shallowest rigid body
         # and add the contact sensor to it.
         # TODO(alexmillane, 2026.01.29): This capability to search for the correct place
         # to add the contact sensor is not yet supported for ObjectReferences and RigidObjectSet.
         # For these objects we just (try to) add the contact sensor to the root prim.
         assert self.object_type == ObjectType.RIGID, "Contact sensor is only supported for rigid objects"
-        if contact_against_prim_paths is None:
-            contact_against_prim_paths = []
+        assert (
```
kellyguo11: **Same `None`-safety bug as `object_base.py`.** `contact_against_object` defaults to `None`, but `contact_against_object.object_type` is accessed unconditionally on the next line. Guard the assertion:
```python
if contact_against_object is not None:
    assert (
        contact_against_object.object_type == ObjectType.RIGID
    ), "Contact sensor is only supported for rigid objects"
```

isaaclab_arena/tasks/g1_locomanip_pick_and_place_task.py
```diff
@@ -42,18 +44,33 @@ def __init__(
...
+            pick_up_object_contact_sensor=self.pick_up_object.get_contact_sensor_cfg(
+                contact_against_object=self.destination_bin,
+            ),
+        )
+
     def get_scene_cfg(self):
-        pass
+        return self.scene_config
 
     def get_termination_cfg(self):
```
kellyguo11: **Dead code — commented-out block.** The old `objects_in_proximity` termination is left as a 10-line comment. If this is intentionally kept for reference, add a note explaining when/why it might be re-enabled. Otherwise, remove it to keep the codebase clean.

isaaclab_arena/tests/test_g1_locomanip_object_on_termination.py
```diff
@@ -0,0 +1,112 @@
...
+
+import torch
+import traceback
+
+import warp as wp
+
+from isaaclab_arena.tests.utils.subprocess import run_simulation_app_function
+
+NUM_STEPS = 500
+HEADLESS = False
```
kellyguo11: **`HEADLESS = False` — should be `True` for CI.** This will try to open a display window, which will fail in headless CI environments.

isaaclab_arena/tests/test_object_on_termination.py
```diff
@@ -12,8 +12,8 @@
 
 from isaaclab_arena.tests.utils.subprocess import run_simulation_app_function
 
-NUM_STEPS = 50
-HEADLESS = True
+NUM_STEPS = 500
+HEADLESS = False
 PLOT = False
```
kellyguo11, line 17: **Existing test changed to `HEADLESS = False` and `NUM_STEPS = 50 → 500`.** Same CI concern — this will break headless test runs. If these values are needed for local debugging, please revert them so CI stays green.

isaaclab_arena/tasks/g1_locomanip_pick_and_place_task.py
```diff
@@ -42,18 +44,33 @@ def __init__(
         )
         self.events_cfg = None
```
kellyguo11: Nice — having the task own its scene config and return it from `get_scene_cfg()` is a clean pattern. The previous `pass` was clearly a stub.

isaaclab_arena/assets/object.py
```diff
@@ -94,24 +94,43 @@ def enable_reset_pose(self) -> None:
         self.reset_pose = True
         self.event_cfg = self._init_event_cfg()
 
-    def get_contact_sensor_cfg(self, contact_against_prim_paths: list[str] | None = None) -> ContactSensorCfg:
+    def get_contact_sensor_cfg(self, contact_against_object: ObjectBase | None = None) -> ContactSensorCfg:
```
kellyguo11: Good API change — accepting `ObjectBase` instead of raw prim paths pushes the rigid-body-path discovery into the right layer. The `isinstance` dispatch is reasonable given the different capabilities of `Object` vs other `ObjectBase` subclasses.

kellyguo11 (changes requested):
Good direction — generalizing the locomanip task and handling non-root rigid bodies in contact sensors is solid work. A few issues to address before merging, mostly around debug leftovers and a potential crash in the base class.

cvolkcvolk (approved):
Thanks for cleaning this up!

PR #572, opened 2026-04-10
Update newton example doc to clearly state failure using physx eval

qianl-nv:
## Summary
This is a doc change requested by QA in https://nvbugspro.nvidia.com/bug/6063011
It clarifies that evaluated newton trained model using physx is expected to completely fail the dexsuite task.


alexmillane (approved):
👍 Thanks for this.

PR #573, opened 2026-04-10
[OUTDATED] Make no-overlap a built-in solver behavior

cvolkcvolk:
## Summary

- No-overlap is now a built-in behavior of `RelationSolver`, controlled by a single `clearance_m` parameter and exposed through `RelationSolverParams` (default 1cm).
- `ArenaEnvBuilder._add_pairwise_no_collision()`  removed from ArenaEnvBuilder -> This is solver specific and now handled within the solver (Eg just using the RelationSolver without the ArenaEnvBuilder would have not gotten NoCollisionConstraints)
 - `NoCollision` class removed from the public API : Users should not need to manage collision avoidance manually. This is now a default behaviour.

## Test plan
- [x] All 9 no-collision tests pass (6 strategy-level, 3 solver-level)
- [x] All 52 relation/placer tests pass
- [x] Pre-commit checks pass

### DIFF
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -29,7 +29,6 @@ from isaaclab_arena.environments.isaaclab_arena_manager_based_env import (
 from isaaclab_arena.metrics.recorder_manager_utils import metrics_to_recorder_manager_cfg
 from isaaclab_arena.relations.object_placer import ObjectPlacer
 from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
-from isaaclab_arena.relations.relations import IsAnchor, NoCollision
 from isaaclab_arena.tasks.no_task import NoTask
 from isaaclab_arena.utils.configclass import combine_configclass_instances
 from isaaclab_arena.utils.isaaclab_utils.simulation_app import reapply_viewer_cfg
@@ -72,25 +71,13 @@ class ArenaEnvBuilder:
                 objects_with_relations.append(asset)
         return objects_with_relations
 
-    def _add_pairwise_no_collision(self, objects_with_relations: list[Object | ObjectReference]) -> None:
-        """Add NoCollision between every pair of non-anchor objects (if not already present)."""
-        non_anchors = [
-            obj for obj in objects_with_relations if not any(isinstance(r, IsAnchor) for r in obj.get_relations())
-        ]
-        for i, obj_a in enumerate(non_anchors):
-            for obj_b in non_anchors[i + 1 :]:
-                has_no_collision = any(isinstance(r, NoCollision) and r.parent is obj_b for r in obj_a.get_relations())
-                if not has_no_collision:
-                    obj_a.add_relation(NoCollision(obj_b))
-
     def _solve_relations(self) -> None:
         """Solve spatial relations for objects in the scene.
 
         This method:
         1. Collects all objects from the scene that have relations
-        2. Adds NoCollision between every pair of non-anchor objects (if not already present)
-        3. Runs the ObjectPlacer to solve spatial constraints
-        4. Applies solved positions to objects
+        2. Runs the ObjectPlacer to solve spatial constraints
+        3. Applies solved positions to objects
         """
         # All objects with relations are subjects of the relation solving.
         objects_with_relations = self._get_objects_with_relations()
@@ -98,8 +85,6 @@ class ArenaEnvBuilder:
         if not objects_with_relations:
             print("No objects with relations found in scene. Skipping relation solving.")
             return
-
-        self._add_pairwise_no_collision(objects_with_relations)
 
         # Run the ObjectPlacer (default on_relation_z_tolerance_m accommodates solver residual).
         placement_seed = getattr(self.args, "placement_seed", None)
--- a/isaaclab_arena/relations/object_placer.py
+++ b/isaaclab_arena/relations/object_placer.py
@@ -298,28 +298,53 @@ class ObjectPlacer:
         self,
         positions: dict[Object | ObjectReference, tuple[float, float, float]],
     ) -> bool:
-        """Check that no two non-anchor objects overlap in 3D (axis-aligned bbox with clearance margin).
+        """Check that no two objects overlap in 3D (axis-aligned bbox check).
 
-        Only checks non-anchor pairs. Anchor-to-non-anchor overlap is expected when the
-        non-anchor sits On(anchor), so those pairs are skipped here; the solver's built-in
-        no-overlap loss handles them during optimization.
+        Skips a pair when one object is an anchor and the other has an On relation
+        targeting that anchor (overlap is expected because the child sits on the parent).
+        For non-anchor pairs, applies ``clearance_m`` as a safety margin. For non-anchor
+        vs anchor pairs without an On relation between them, checks actual overlap only
+        (margin=0.0) since clearance is meant for sibling objects.
         """
         anchor_objects = get_anchor_objects(list(positions.keys()))
         anchor_set = set(anchor_objects)
-        non_anchors = [obj for obj in positions if obj not in anchor_set]
+        all_objects = list(positions.keys())
 
-        for i in range(len(non_anchors)):
-            for j in range(i + 1, len(non_anchors)):
-                a, b = non_anchors[i], non_anchors[j]
+        for i in range(len(all_objects)):
+            for j in range(i + 1, len(all_objects)):
+                a, b = all_objects[i], all_objects[j]
+                a_is_anchor = a in anchor_set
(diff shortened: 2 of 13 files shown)


isaaclab_arena/relations/relation_solver.py
```diff
@@ -60,6 +65,70 @@ def _get_strategy(self, relation: RelationBase) -> RelationLossStrategy | UnaryR
             )
         return strategy
 
+    def _compute_no_overlap_loss(self, state: RelationSolverState, debug: bool = False) -> torch.Tensor:
+        """Compute pairwise no-overlap loss for all object pairs.
+
+        Covers (non-anchor, non-anchor) pairs and (non-anchor, anchor) pairs.
```
kellyguo11: Nit: The slope `10000.0` is now hardcoded here instead of being configurable via the strategy dict. Consider exposing a `no_overlap_slope` parameter on `RelationSolverParams` for power users who need to tune the balance between no-overlap and other relation losses.

isaaclab_arena/relations/relation_solver.py
```diff
@@ -60,6 +65,70 @@ def _get_strategy(self, relation: RelationBase) -> RelationLossStrategy | UnaryR
...
+        for i in range(len(opt)):
+            for j in range(i + 1, len(opt)):
+                obj_i, obj_j = opt[i], opt[j]
+                pos_i = state.get_position(obj_i)
+                pos_j = state.get_position(obj_j)
+
+                # i as child, j as parent (gradient flows through pos_i)
+                world_bbox_j = obj_j.get_bounding_box().translated((pos_j[0].item(), pos_j[1].item(), pos_j[2].item()))
+                loss_ij = self._no_collision_strategy.compute_loss(
+                    clearance_m=clearance,
```
kellyguo11: The `.item()` calls to build the translation tuple detach from the computation graph, which is correct here — `parent_world_bbox` is used as a fixed target for the loss, and gradients flow only through `child_pos`. Good.

One thing to keep in mind: if you ever wanted gradients to flow through both positions simultaneously (e.g., for a different optimizer), you'd need to restructure this. Current approach (compute in both directions with one detached) is fine for the Adam optimizer loop.

isaaclab_arena/relations/object_placer.py
```diff
@@ -298,21 +298,54 @@ def _validate_no_overlap(
...
                 a_world = a.get_bounding_box().translated(positions[a])
                 b_world = b.get_bounding_box().translated(positions[b])
 
-                if a_world.overlaps(b_world, margin=self.params.min_separation_m):
+                # Non-anchor vs non-anchor: use clearance_m margin
+                # Non-anchor vs anchor (no On relation): check actual overlap only
+                if not a_is_anchor and not b_is_anchor:
+                    margin = self.params.solver_params.clearance_m
+                else:
+                    margin = 0.0
```
kellyguo11: Nice handling of the On-parent skip. One edge case to consider: if object A has `On(table)` and object B is placed *next to* the table (no On relation) — B could overlap the table's bbox. The `margin=0.0` check catches actual overlap but not clearance. Seems like the right tradeoff since B's position is constrained by other relations, but worth documenting.

kellyguo11 (commented):
## Review: Make no-overlap a built-in solver behavior

Nice refactor. Moving no-overlap from a user-managed `NoCollision` relation to a built-in solver behavior is the right call — reduces API surface and eliminates a common footgun (forgetting to add pairwise constraints). The commit stack is clean and logical.

### What works well

- **Single parameter (`clearance_m`) on `RelationSolverParams`** — clean, discoverable, with a sensible 1cm default and `__post_init__` validation.
- **Bidirectional loss for non-anchor pairs** — computing loss in both directions (`i→j` and `j→i`) ensures both objects receive gradient. Good detail.
- **Smart On-parent skipping in validation** — `_has_on_relation_with()` correctly lets child-on-parent overlap pass while still checking non-On anchor pairs. The asymmetric margin (clearance for sibling pairs, 0.0 for non-On anchor pairs) is a reasonable choice.
- **Test coverage** — 9 tests cover the strategy API change, solver separation, clearance_m respect, negative validation, On-parent overlap acceptance, non-anchor overlap rejection, and reproducibility. Solid.
- **Clean removal** — `NoCollision` class, `_add_pairwise_no_collision()`, `min_separation_m`, and all example/env usages are fully removed. No stale references found in docs or other source files.

### Issues / Suggestions

**1. Solver applies clearance to non-anchor↔anchor pairs, but validation doesn't (intentional asymmetry?)**

In `_compute_no_overlap_loss`, non-anchor vs anchor pairs use `clearance_m`. But in `_validate_no_overlap`, those same pairs use `margin=0.0`. The docstring says "clearance is meant for sibling objects" — but the solver loss pushes non-anchors away from anchors by `clearance_m` during optimization, then validation only checks hard overlap.

This is probably fine in practice (the solver already enforced the gap), but it means a scenario that barely passes optimization could technically fail validation if the solver residual ate into the margin. Consider whether the solver should also use `margin=0.0` for non-anchor↔anchor pairs to keep them consistent, or document the rationale more explicitly.

**2. O(n²) complexity for large object counts**

The pairwise loop is `O(n²)` per iteration for n non-anchor objects. With the current use cases (handful of objects on a table) this is fine, but worth a brief comment noting that scaling to 50+ objects may want spatial hashing or a grid-based approach.

**3. `NoCollisionLossStrategy` slope is hardcoded at 10000.0**

```python
self._no_collision_strategy = NoCollisionLossStrategy(slope=10000.0)
```

Previously this was configurable via the strategy dict. Now it's hardcoded. For most users this won't matter, but power users may want to tune it (e.g., if clearance loss dominates their On/NextTo losses). Consider exposing `no_overlap_slope` on `RelationSolverParams` for completeness.

**4. `get_world_bounding_box()` assumption for anchors**

In `_compute_no_overlap_loss`, anchor bboxes are retrieved via `anchor.get_world_bounding_box()`, while in `_compute_total_loss` for regular relations, the same is done with `parent.get_world_bounding_box()` for anchor parents. This is consistent, but it means the solver assumes anchors always have their world bbox precomputed. If that's guaranteed by `RelationSolverState` initialization, great — just worth noting.

**5. Spec doc not included**

The PR body references `docs/superpowers/specs/2026-04-10-no-collision-solver-builtin-design.md` but this file is not in the branch. Not blocking, but consider including it or removing the reference.

**6. Minor: `relation_loss_weight` no longer applies to no-overlap**

Previously users could set `relation_loss_weight` on `NoCollision` to increase/decrease its priority relative to other relations. The removed test `test_no_collision_loss_scales_with_relation_weight` confirms this is intentionally dropped. Worth a migration note if any downstream code relied on per-pair weight tuning.

### Verdict

Clean, well-tested, good API simplification. The items above are mostly minor / future-proofing. 👍

isaaclab_arena/relations/object_placer.py
```diff
@@ -298,16 +298,29 @@ def _validate_no_overlap(
         self,
         positions: dict[Object | ObjectReference, tuple[float, float, float]],
     ) -> bool:
-        """Check that no two objects overlap in 3D (axis-aligned bbox with margin)."""
+        """Check that no two objects overlap in 3D (axis-aligned bbox with clearance margin).
+
+        All pairs must be separated by at least ``clearance_m``.
```
zhx06, line 303: A minor question: Do we plan to have the same ``clearance_m`` for all objects?  Will we expect that different object sizes sometimes benefit from different clearance values in the future?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -72,25 +71,13 @@ def _get_objects_with_relations(self) -> list[Object | ObjectReference]:
...
-    def _add_pairwise_no_collision(self, objects_with_relations: list[Object | ObjectReference]) -> None:
-        """Add NoCollision between every pair of non-anchor objects (if not already present)."""
-        non_anchors = [
-            obj for obj in objects_with_relations if not any(isinstance(r, IsAnchor) for r in obj.get_relations())
-        ]
-        for i, obj_a in enumerate(non_anchors):
-            for obj_b in non_anchors[i + 1 :]:
-                has_no_collision = any(isinstance(r, NoCollision) and r.parent is obj_b for r in obj_a.get_relations())
-                if not has_no_collision:
-                    obj_a.add_relation(NoCollision(obj_b))
```
alexmillane, line 84: Yea good move to remove this from the EnvBuilder.

isaaclab_arena/relations/object_placer.py
cvolkcvolk: Yes that's a good point, we might want to add this in the future. For now I think we should go with simplicity.

alexmillane (approved):
Looks good!

cvolkcvolk:
Superseded by #589 (re-implemented on develop branch)

PR #574, opened 2026-04-10
Add PositionLimits relation for constraining object placement

cvolkcvolk:
## Motivation

When building tasks, users often need to restrict object placement to a sub-region of a surface -- for example, only within the robot's reachable workspace. `On(table)` allows placement anywhere on the table, and `AtPosition` pins to a single point. There was no way to constrain to a region or set bounds on individual axes.

## Summary
- New `PositionLimits` unary relation that constrains object position in world coordinates. Supports full ranges (box), single bounds (half-plane), or a mix per axis.
- New `UnaryRelation` base class so `get_spatial_relations()` automatically includes any new unary relation without updating isinstance checks.
- `PositionLimitsLossStrategy` using `linear_band_loss` (both bounds) and `single_boundary_linear_loss` (single bound).
- Registered in solver strategies with slope=100.0 (matching `AtPosition`).
- Fixed `_print_unary_relation_debug` to work with any unary relation type.

## Usage

```python
# Full box constraint (reachable region)
apple.add_relation(On(table))
apple.add_relation(PositionLimits(x_min=-0.3, x_max=0.3, y_min=-0.2, y_max=0.2))

# Single bound (half-plane)
apple.add_relation(PositionLimits(x_min=0.5))

# Mix
apple.add_relation(PositionLimits(x_min=-0.3, x_max=0.3, y_min=-0.2))
```

## Test plan
- [x] 12 PositionLimits tests pass (11 strategy-level, 1 solver integration)
- [x] All relation/placer tests pass
- [x] Pre-commit checks pass


isaaclab_arena/relations/relation_solver.py
```diff
@@ -78,8 +78,8 @@ def _compute_total_loss(self, state: RelationSolverState, debug: bool = False) -
                 child_pos = state.get_position(obj)
                 strategy = self._get_strategy(relation)
 
-                # Handle unary relations (no parent) like AtPosition
-                if isinstance(relation, AtPosition):
+                # Handle unary relations (no parent) like AtPosition and WithinBox
+                if isinstance(relation, (AtPosition, WithinBox)):
                     loss = strategy.compute_loss(
                         relation=relation,
```
kellyguo11, line 84: Bug: `_print_unary_relation_debug` (line 259) uses `getattr(relation, axis)` for `axis in ('x', 'y', 'z')`, which will raise `AttributeError` when called with a `WithinBox` relation (which has `x_min`/`x_max`, not `x`/`y`/`z`).

This path is exercised when `debug=True` in `_compute_total_loss`. Consider either:
1. Making the debug function dispatch on relation type, or
2. Adding a `__repr__`/`debug_str()` method to each relation class

kellyguo11 (commented):
Nice, clean addition — follows the existing `AtPosition` pattern well and the test coverage is solid. A few observations:

## Scalability concern: `isinstance` type lists

Every new unary relation requires updating `isinstance` checks in **four** places:
1. `object_base.py:get_spatial_relations` → `isinstance(r, (Relation, AtPosition, WithinBox))`
2. `dummy_object.py:get_spatial_relations` → same
3. `relation_solver.py:_compute_total_loss` → `isinstance(relation, (AtPosition, WithinBox))`
4. The debug function type hint

This is an O(n) maintenance burden per new relation type. Consider a marker base class (e.g., `SpatialRelation`) or a class attribute (`is_spatial = True`) so `get_spatial_relations` doesn't need to enumerate types. Not blocking — the current approach matches the existing pattern — but worth thinking about before the next relation lands.

## Bug: `_print_unary_relation_debug` will crash for `WithinBox`

The debug function at `relation_solver.py:259` assumes `AtPosition`-style attributes (`x`, `y`, `z`) via `getattr(relation, axis)`. `WithinBox` has `x_min`/`x_max` etc., so this will raise `AttributeError` when `debug=True` is passed. This isn't exercised by tests since the solver integration test uses `verbose=False`, but it's a latent bug.

## Loss strategy slope mismatch

`WithinBoxLossStrategy.__init__` defaults to `slope=10.0`, but `_default_strategies()` in `relation_solver_params.py` registers it with `slope=100.0`. The default in `__init__` is effectively dead code since the solver always uses the registered strategy, but the mismatch could confuse someone instantiating `WithinBoxLossStrategy()` directly (as the tests do with `slope=10.0`). Minor — just note it or align them.

## Test quality

12 tests covering construction validation, all loss branches (both-bounds, min-only, max-only, unconstrained axes), weight scaling, and solver integration. Good coverage. The solver integration test at the end is a nice touch.

Overall: clean implementation, correct math, good tests. The `_print_unary_relation_debug` crash is the only functional issue.

isaaclab_arena/assets/object_base.py
```diff
@@ -15,7 +15,7 @@
 from isaaclab_tasks.manager_based.manipulation.stack.mdp.franka_stack_events import randomize_object_pose
 
 from isaaclab_arena.assets.asset import Asset
-from isaaclab_arena.relations.relations import AtPosition, Relation, RelationBase
+from isaaclab_arena.relations.relations import AtPosition, Relation, RelationBase, WithinBox
 from isaaclab_arena.terms.events import set_object_pose
 from isaaclab_arena.utils.pose import Pose, PoseRange
```
kellyguo11, line 20: Nit (not blocking): Every new unary relation requires updating this `isinstance` tuple in both `object_base.py` and `dummy_object.py`. A shared base class like `SpatialRelation` (or even a simple class attribute `is_spatial = True` on `RelationBase`) would make this filter future-proof:
```python
return [r for r in self.relations if getattr(r, 'is_spatial', False)]
```

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -433,3 +433,78 @@ def compute_loss(
...
+        child_bbox: AxisAlignedBoundingBox,
+    ) -> torch.Tensor:
+        """Compute loss for WithinBox relation.
+
+        Args:
+            relation: WithinBox relation with optional x/y/z min/max bounds.
+            child_pos: Child object position tensor (x, y, z) in world coords.
+            child_bbox: Child object local bounding box (unused, for signature consistency).
+
+        Returns:
```
kellyguo11, line 466: Minor: default `slope=10.0` here but registered with `slope=100.0` in `_default_strategies()`. The registration wins at runtime, but the mismatch could mislead someone instantiating `WithinBoxLossStrategy()` directly. Consider aligning to `100.0` to match the registered default, or documenting the intentional difference.

isaaclab_arena/assets/object_base.py
cvolkcvolk: Good Idea

kellyguo11: 👍 Nice — feel free to tackle it in this PR or a follow-up, either works.

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -433,3 +433,63 @@ def compute_loss(
...
+            elif lower_bound is not None:
+                # Only lower bound: penalize positions below it
+                total_loss = total_loss + single_boundary_linear_loss(
+                    child_pos[axis_index], lower_bound, slope=self.slope, penalty_side="less"
+                )
+            elif upper_bound is not None:
+                # Only upper bound: penalize positions above it
+                total_loss = total_loss + single_boundary_linear_loss(
+                    child_pos[axis_index], upper_bound, slope=self.slope, penalty_side="greater"
+                )
```
alexmillane, line 492: clean!

isaaclab_arena/relations/relations.py
```diff
@@ -340,6 +350,49 @@ def __init__(
...
+        relation_loss_weight: float = 1.0,
+    ):
+        assert (
+            x_min is not None
+            or x_max is not None
+            or y_min is not None
+            or y_max is not None
+            or z_min is not None
+            or z_max is not None
+        ), "At least one bound (x_min, x_max, y_min, y_max, z_min, or z_max) must be specified for WithinBox"
```
alexmillane: Is it required that is `x_min` is not None then `x_max` is not none. In other words you need non-none pairs. What you have here is a looser condition.

Is it valid to go `WithinBox(x_min=1.0)`?

alexmillane: *edit: looking at the test, this is valid. Ok fair enough. Not really what you'd expect but it does represent something valid

alexmillane (approved):
Really clean!

xyao-nv (approved):

isaaclab_arena/relations/relations.py
cvolkcvolk: Mh good point. The naming does not reflect that really. I think I'll keep the loose constraint (one sided only possible). And rename it to something generic like `PositionLimits` .

PR #575, opened 2026-04-10
Clean up type annotation in the environment files.

alexmillane:
## Summary
Clean up type annotations in the environment files

## Detailed description
- Type annotation were not properly done at the start of the project, and that propagated over time to all environment files.
- This cleans that up.


cvolkcvolk (approved):
Nice! Those notes have been with us for so long I will probably start to miss them.

PR #576, opened 2026-04-10
Deactivate all tests except for the subprocess tests.

alexmillane:
## Summary
Tests if we still get stalls when only running the subprocess-spawning tests (without the in-process tests).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,20 +140,20 @@ jobs:
       # - All subprocess-spawning tests (each spawns their own SimulationApp)
       #
       # To restore sanity here is a goal for Isaac Lab Arena v0.3.
-      - name: Run in-process Newton tests
-        run: |
-          /isaac-sim/python.sh -m pytest -sv -m with_newton \
-            isaaclab_arena/tests/
+      # - name: Run in-process Newton tests
+      #   run: |
+      #     /isaac-sim/python.sh -m pytest -sv -m with_newton \
+      #       isaaclab_arena/tests/
 
-      - name: Run in-process PhysX tests without cameras
-        run: |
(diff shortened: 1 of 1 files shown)


kellyguo11 (commented):
## Review — Diagnostic CI change

**What this does:** Comments out all three in-process test steps (Newton, PhysX without cameras, PhysX with cameras) from the CI workflow, leaving only the subprocess-spawning PhysX tests running. The goal is to isolate whether CI stalls originate from the in-process tests or from mixing them with subprocess-spawning tests.

### Assessment

The change is mechanically correct — clean commenting with no accidental syntax breakage in the YAML. The subprocess test step and all other jobs (`pre_commit`, `test_policy`, `build_docs_pre_merge`, `build_and_push_image_post_merge`) are untouched.

### Concerns

1. **Not safe to merge into `main` as-is.** If this lands on `main`, every CI run will silently skip the bulk of the test suite. Regressions in Newton support, camera-related codepaths, and non-subprocess PhysX tests would go undetected. The branch name (`alex/test/ci_stalling_only_subprocess`) signals this is experimental, but the PR targets `main` — so either:
   - Retarget to a non-default branch so the experiment runs without affecting `main` CI, **or**
   - Add a clear `TODO`/`FIXME` comment in the YAML (e.g., `# TEMP: re-enable after CI stall investigation — see #576`) so it doesn't get forgotten, **and** create a follow-up tracking issue.

2. **Minor nit — double space in remaining step name.** Line ~162 has `Run subprocess-spawning  PhysX tests` (two spaces before "PhysX"). Pre-existing but worth fixing while you're in this file.

### Verdict

Reasonable diagnostic approach. 👍 for running the experiment on a feature branch or behind a clear revert plan — just wouldn't merge to `main` without safeguards to restore coverage.

PR #578, opened 2026-04-10
Install missing arena package into NGC docker.

alexmillane:
## Summary
Install missing arena package into NGC docker.

## Detailed description
- We forgot to install our new package `isaaclab_arena_examples` into the docker image.
- This was masked in CI due to mounting a branch and correctly installing there.


kellyguo11 (approved):
LGTM — straightforward one-liner.

**Verification:**
- `setup.py` already includes `isaaclab_arena_examples*` in `find_packages()`, so the package was expected but never copied into the image.
- The new `COPY` is correctly placed after the other `isaaclab_arena_*` copies and before the `pip install -e` step, so the editable install will pick it up.
- Consistent with how `isaaclab_arena`, `isaaclab_arena_g1`, `isaaclab_arena_gr00t`, and `isaaclab_arena_environments` are handled.
- Good catch that CI masked this — the branch mount + install there meant the package was available in CI but not in a standalone Docker build.

No issues found. Safe to merge.

xyao-nv (approved):

cvolkcvolk (approved):
Thanks

PR #581, opened 2026-04-10
Fix CI pre-commit file-mode failure 

xyao-nv:
## Summary
Fix pre-commit CI job failing with "pre-commit hook(s) made changes". E.g. https://github.com/isaac-sim/IsaacLab-Arena/actions/runs/24260814760/job/70843875557

## Detailed description

- Every file showing a 644→755 mode-only diff in pre-commit failed CI jobs
- It is caused by the runner's persistent workspace retaining executable permissions across runs, which hooks then propagate when rewriting files.
- Set to ignore it form git as a fix.


kellyguo11 (approved):
LGTM — clean, minimal fix.

**What it does:** On self-hosted runners, the persistent workspace can leave files with executable (755) permissions from previous runs. Pre-commit hooks that rewrite files preserve those modes, producing spurious 644→755 diffs that fail the `check-executables-have-shebangs` and similar checks. Setting `core.fileMode false` tells git to ignore these permission-only changes.

**Step ordering is correct:** The `git config` runs after checkout (so `.git/` exists) and right before `pre-commit/action`, which is the ideal placement. Uses local config (not `--global`), scoped to this repo only — good.

**CI confirms:** Pre-commit job passes on the latest commit (`a18564ca`). The `Run policy-related tests` failure is unrelated to this change.

**Note:** Greptile's comment about step ordering was based on the first commit; the second commit (`reorder`) fixed that, so it's a stale concern.

Nit (non-blocking): Commit messages could be more descriptive (`"try if it fix"` → something like `"ci: ignore file-mode changes in pre-commit job"`), but that's cosmetic.

alexmillane (dismissed):
Looks reasonable if it works!

Issue #592, opened 2026-04-13
[Improvement] Separate spawner and object type concepts.


alexmillane:
## Issue Summary

Suggestion to move SPAWNER out to ObjectType.

## Issue Details
- We have mixing of concerns in ObjectType (in file object_base.py) between BASE/RIGID/ARTICULATION, that denote the type of object, and SPAWNER which denotes that this object, which can be any of the types above, should use a user-specified spawn function.
- These two concepts should be independent. The type of the object, and the optional presence of a user-specified spawner.
- Right now, all SPAWNER type objects, i.e. objects with custom spawn functions, have to generate a object config of type a AssetBaseCfg.
- This is starting to cause issues where we have objects that require custom spawn functions, but want to be rigid objects, for example. See the object Sphere, for example, in the object_library.py, which is forced to hack around this by overriding _generate_spawner_cfg.

## Fix suggestion.
- Suggested to fix by removing SPAWNER from the ObjectType, and treating the concept of having a custom spawn function separately to the issue of what type the object is.

PR #593, opened 2026-04-13
Separate spawners and object types.

alexmillane:
## Summary
Refactors how we handle spawner objects, to separate their specification from object type

## Detailed description
- Addresses #592 
- Separate `ObjectType` and `SPAWNER`


docker/run_docker.sh
```diff
@@ -132,6 +132,7 @@ else
                     $(add_volume_if_it_exists $EVAL_HOST_MOUNT_DIRECTORY /eval)
                     "-v" "$HOME/.bash_history:/home/$(id -un)/.bash_history"
                     "-v" "$HOME/.config/osmo:/home/$(id -un)/.config/osmo"
+                    "-v" "$HOME/.config/gh:/home/$(id -un)/.config/gh"
```
alexmillane, line 135: Separate to another MR

isaaclab_arena/assets/object_library.py
```diff
@@ -453,28 +454,14 @@ def __init__(
...
-            spawn=self.spawner_cfg,
-        )
-        object_cfg = self._add_initial_pose_to_cfg(object_cfg)
-        if self.initial_velocity is not None:
-            object_cfg.init_state.lin_vel = self.initial_velocity.linear_xyz
-            object_cfg.init_state.ang_vel = self.initial_velocity.angular_xyz
-        return object_cfg
-
-    def _requires_reset_pose_event(self) -> bool:
-        return self.get_initial_pose() is not None and self.reset_pose
```
alexmillane, line 476: Remove hacky overriding of the `_generate_spawner_cfg` to get around the coupling of concerns between `object_type` and `spawner_cfg`.

Now we can cleanly call a sphere a rigid object with a custom spawner.

isaaclab_arena/assets/object.py
```diff
@@ -32,13 +32,15 @@ def __init__(
...
         spawn_cfg_addon: dict[str, Any] = kwargs.pop("spawn_cfg_addon", {}) or {}
         asset_cfg_addon: dict[str, Any] = kwargs.pop("asset_cfg_addon", {}) or {}
-        if object_type is not ObjectType.SPAWNER:
-            assert usd_path is not None
+        self.spawner_cfg = spawner_cfg
+        if usd_path is None and self.spawner_cfg is None:
+            raise ValueError(f"Object '{name}' requires either usd_path or a spawner_cfg")
         # Detect object type if not provided
         if object_type is None:
             object_type = detect_object_type(usd_path=usd_path)
```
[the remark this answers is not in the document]
alexmillane: Fixes

isaaclab_arena/assets/object.py
```diff
@@ -136,63 +138,50 @@ def get_contact_sensor_cfg(
...
-    def _generate_spawner_cfg(self) -> AssetBaseCfg:
-        assert self.object_type == ObjectType.SPAWNER
-        object_cfg = AssetBaseCfg(
-            prim_path=self.prim_path,
-            spawn=self.spawner_cfg,
-            **self.asset_cfg_addon,
-        )
-        object_cfg = self._add_initial_pose_to_cfg(object_cfg)
-        return object_cfg
+        return self._add_initial_pose_to_cfg(object_cfg)
```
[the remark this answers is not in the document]
alexmillane: fixed

xyao-nv (approved):

isaaclab_arena/assets/object.py
```diff
@@ -32,15 +33,21 @@ def __init__(
...
         # Pull out addons (and remove them from kwargs before passing to super)
         spawn_cfg_addon: dict[str, Any] = kwargs.pop("spawn_cfg_addon", {}) or {}
         asset_cfg_addon: dict[str, Any] = kwargs.pop("asset_cfg_addon", {}) or {}
-        if object_type is not ObjectType.SPAWNER:
-            assert usd_path is not None
+        self.spawner_cfg = spawner_cfg
+        if usd_path is None and self.spawner_cfg is None:
+            raise ValueError(f"Object '{name}' requires either usd_path or a spawner_cfg")
         # Detect object type if not provided
         if object_type is None:
```
viiik-inside, line 47: Does a spawner_cfg always have an object type? It seems to me like spawner cfg and usd path are independent with these checks. Is that true? If thats the case then we should probably assert that spawner cfg has a object type set maybe?

isaaclab_arena/assets/object.py
```diff
@@ -136,63 +143,50 @@ def get_contact_sensor_cfg(
...
         assert self.object_type == ObjectType.RIGID
         object_cfg = RigidObjectCfg(
             prim_path=self.prim_path,
-            spawn=UsdFileCfg(
-                usd_path=self.usd_path,
-                scale=self.scale,
-                activate_contact_sensors=True,
-                **self.spawn_cfg_addon,
-            ),
+            spawn=self._get_spawn_cfg(activate_contact_sensors=True),
```
viiik-inside, line 163: Consider adding a comment on why the contact sensors are swtiched on

isaaclab_arena/assets/object.py
```diff
@@ -136,63 +143,50 @@ def get_contact_sensor_cfg(
...
         assert self.object_type == ObjectType.ARTICULATION
         object_cfg = ArticulationCfg(
             prim_path=self.prim_path,
-            spawn=UsdFileCfg(
-                usd_path=self.usd_path,
-                scale=self.scale,
-                activate_contact_sensors=True,
-                **self.spawn_cfg_addon,
-            ),
+            spawn=self._get_spawn_cfg(activate_contact_sensors=True),
```
viiik-inside, line 172: see comment above

isaaclab_arena/assets/object.py
```diff
@@ -136,63 +143,50 @@ def get_contact_sensor_cfg(
...
+                    )
         object_cfg = AssetBaseCfg(
             prim_path=self.prim_path,
-            spawn=self.spawner_cfg,
+            spawn=self._get_spawn_cfg(),
             **self.asset_cfg_addon,
         )
-        object_cfg = self._add_initial_pose_to_cfg(object_cfg)
-        return object_cfg
+        return self._add_initial_pose_to_cfg(object_cfg)
```
viiik-inside, line 191: The bot seems to have a valid comment

viiik-inside (approved):
Looks great

isaaclab_arena/assets/object.py
alexmillane: I've reworked the check. You're right, for a `spawner_cfg` we require an `object_type`. For a usd path we don't because we can use the USD to check the type.

isaaclab_arena/assets/object.py
alexmillane: I think we just set contact sensors to true every where because we use contact checking all over the framework. In a perfect world we'd turn it on as needed, but to avoid the complication we just set it on everywhere. I think I'll skip adding that long winded and poor explanation 😅

isaaclab_arena/assets/object.py
alexmillane: yea I agree. I fixed it though.

PR #594, opened 2026-04-13
Homogeneous Object Placement on Environment Reset

zhx06:
## Summary
Pool-based object placement with fresh layouts on environment reset

## Detailed description
What changed:
- Added `PlacementPool` that pre-solves valid layouts at build time and serves them on reset via `placement_events.py`.
- Rewritten `ArenaEnvBuilder._solve_relations()` to use the pool, with `--resolve_on_reset` / `--no-resolve_on_reset` controlling behavior.
- Added `--episode_length_s` to `gr1_table_multi_object_no_collision` for periodic resets.

## Impact
- Fresh collision-free layouts on every reset, different across envs and episodes.
- Solver runs once at build time (batched), not per-reset.

### DIFF
--- a/isaaclab_arena/cli/isaaclab_arena_cli.py
+++ b/isaaclab_arena/cli/isaaclab_arena_cli.py
@@ -44,6 +44,19 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
         "Isaac Lab Arena Arguments", "Arguments specific to Isaac Lab Arena framework"
     )
     arena_group.add_argument(
+        "--no-solve-relations",
+        action="store_false",
+        dest="solve_relations",
+        default=True,
+        help="Disable solving spatial relations in the environment.",
+    )
+    arena_group.add_argument(
+        "--placement_seed",
+        type=int,
+        default=None,
+        help="Seed for object placement. If set, objects are placed at the same positions across runs.",
+    )
+    arena_group.add_argument(
         "--presets",
         type=str,
         default=None,
@@ -53,23 +66,7 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
             "When not set, each environment uses its own default."
         ),
     )
-
-    placement_group = parser.add_argument_group(
-        "Placement Arguments", "Arguments controlling object placement and relation solving"
-    )
-    placement_group.add_argument(
-        "--solve_relations",
-        action=argparse.BooleanOptionalAction,
-        default=True,
-        help="Solve spatial relations in the environment. Use --no-solve_relations to disable.",
-    )
-    placement_group.add_argument(
-        "--placement_seed",
-        type=int,
-        default=None,
-        help="Seed for object placement. If set, objects are placed at the same positions across runs.",
-    )
-    placement_group.add_argument(
+    arena_group.add_argument(
         "--resolve_on_reset",
         action=argparse.BooleanOptionalAction,
         default=None,
--- a/isaaclab_arena/cli/isaaclab_arena_cli.py
+++ b/isaaclab_arena/cli/isaaclab_arena_cli.py
@@ -44,19 +44,6 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
         "Isaac Lab Arena Arguments", "Arguments specific to Isaac Lab Arena framework"
     )
     arena_group.add_argument(
-        "--no-solve-relations",
-        action="store_false",
-        dest="solve_relations",
-        default=True,
-        help="Disable solving spatial relations in the environment.",
-    )
-    arena_group.add_argument(
-        "--placement_seed",
-        type=int,
-        default=None,
-        help="Seed for object placement. If set, objects are placed at the same positions across runs.",
-    )
-    arena_group.add_argument(
         "--presets",
         type=str,
         default=None,
@@ -64,6 +51,31 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
             "Physics backend preset: 'physx' or 'newton'. "
             "Mirrors Isaac Lab's ``presets=newton`` Hydra syntax. "
             "When not set, each environment uses its own default."
+        ),
+    )
+
+    placement_group = parser.add_argument_group(
+        "Placement Arguments", "Arguments controlling object placement and relation solving"
+    )
+    placement_group.add_argument(
+        "--solve_relations",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help="Solve spatial relations in the environment. Use --no-solve_relations to disable.",
+    )
+    placement_group.add_argument(
+        "--placement_seed",
+        type=int,
+        default=None,
+        help="Seed for object placement. If set, objects are placed at the same positions across runs.",
+    )
+    placement_group.add_argument(
+        "--resolve_on_reset",
+        action=argparse.BooleanOptionalAction,
+        default=None,
+        help=(
+            "Re-place objects from the pool on each reset (default: True). Use --no-resolve_on_reset to keep the same"
+            " layout."
         ),
     )
 
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -33,7 +33,7 @@ from isaaclab_arena.relations.placement_events import solve_and_place_objects
 from isaaclab_arena.relations.placement_pool import PlacementPool
 from isaaclab_arena.relations.relation_solver_params import RelationSolverParams
 from isaaclab_arena.tasks.no_task import NoTask
-from isaaclab_arena.utils.configclass import combine_configclass_instances
+from isaaclab_arena.utils.configclass import combine_configclass_instances, make_configclass
 from isaaclab_arena.utils.isaaclab_utils.simulation_app import reapply_viewer_cfg
 from isaaclab_arena.utils.multiprocess import get_local_rank
 
@@ -155,8 +155,6 @@ class ArenaEnvBuilder:
         )
         placement_event_cfg = None
         if self._placement_event_cfg is not None:
-            from isaaclab_arena.utils.configclass import make_configclass
-
             PlacementEventCfg = make_configclass(
                 "PlacementEventCfg",
                 [("placement_reset", EventTermCfg, self._placement_event_cfg)],
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -118,9 +118,9 @@ class ArenaEnvBuilder:
         )
 
         if placer_params.resolve_on_reset:
-            anchor_objects = set(get_anchor_objects(objects_with_relations))
+            anchor_objects_set = set(get_anchor_objects(objects_with_relations))
             for obj in objects_with_relations:
-                if obj not in anchor_objects:
+                if obj not in anchor_objects_set:
                     assert obj.event_cfg is None, (
                         f"Non-anchor object '{obj.name}' has an explicit pose-reset event. "
                         "Relational solving should not be combined with explicit setting of "
@@ -128,7 +128,7 @@ class ArenaEnvBuilder:
                     )
             # Set init_state so objects spawn at valid positions (not origin).
             # The placement event will override these on every reset.
-            self._set_init_state_from_pool(objects_with_relations, placement_pool, anchor_objects)
+            self._set_init_state_from_pool(objects_with_relations, placement_pool, anchor_objects_set)
             self._placement_event_cfg = EventTermCfg(
                 func=solve_and_place_objects,
                 mode="reset",
@@ -144,7 +144,7 @@ class ArenaEnvBuilder:
         self,
         objects: list[Object | ObjectReference],
         pool: PlacementPool,
-        anchor_objects: set,
+        anchor_objects_set: set,
     ) -> None:
         """Set ``object_cfg.init_state`` from a pool layout so objects spawn at valid positions.
 
@@ -153,7 +153,7 @@ class ArenaEnvBuilder:
         """
         layout = pool.sample(1)[0]
         for obj in objects:
-            if obj in anchor_objects or obj.object_cfg is None:
+            if obj in anchor_objects_set or obj.object_cfg is None:
                 continue
             pos = layout.positions.get(obj)
             if pos is None:
@@ -176,10 +176,10 @@ class ArenaEnvBuilder:
         from isaaclab_arena.utils.pose import Pose, PosePerEnv
 
         layouts = pool.sample(num_envs)
-        anchor_objects = set(get_anchor_objects(objects))
+        anchor_objects_set = set(get_anchor_objects(objects))
 
         for obj in objects:
-            if obj in anchor_objects:
+            if obj in anchor_objects_set:
                 continue
             rotation_xyzw = get_rotation_xyzw(obj)
             poses = [
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -164,11 +164,14 @@ class ArenaEnvBuilder:
             if obj in anchor_objects_set:
                 continue
             rotation_xyzw = get_rotation_xyzw(obj)
-            poses = [
-                Pose(position_xyz=layouts[env_idx].positions[obj], rotation_xyzw=rotation_xyzw)
-                for env_idx in range(num_envs)
-            ]
-            obj.set_initial_pose(PosePerEnv(poses=poses))
+            poses = []
+            for env_idx in range(num_envs):
+                pos = layouts[env_idx].positions.get(obj)
+                if pos is None:
+                    break
+                poses.append(Pose(position_xyz=pos, rotation_xyzw=rotation_xyzw))
+            else:
+                obj.set_initial_pose(PosePerEnv(poses=poses))
 
     def _modify_recorder_cfg_dataset_filename(self, recorder_cfg: RecorderManagerBaseCfg) -> RecorderManagerBaseCfg:
         """Modify the recorder dataset filename to include the timestamp and rank."""
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -104,8 +104,8 @@ class ArenaEnvBuilder:
         if placer_params.resolve_on_reset:
             anchor_objects_set = set(get_anchor_objects(objects_with_relations))
             for obj in objects_with_relations:
-                if obj not in anchor_objects_set:
-                    assert obj.event_cfg is None, (
+                if obj not in anchor_objects_set and obj.event_cfg is not None:
+                    raise RuntimeError(
                         f"Non-anchor object '{obj.name}' has an explicit pose-reset event. "
                         "Relational solving should not be combined with explicit setting of "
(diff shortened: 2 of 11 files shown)


isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -102,28 +102,26 @@ def _solve_relations(self) -> None:
...
-        else:
-            print(f"Warning: Relation solving not completed after {result.attempts} attempt(s)")
+        # Clear per-object reset events on solver-managed non-anchor objects.
+        # Without this, both the per-object event (set_object_pose) and the
+        # solver event (solve_and_place_objects) would fire on reset, with
+        # unpredictable ordering.
+        anchor_objects_set = set(get_anchor_objects(objects_with_relations))
+        for obj in objects_with_relations:
+            if obj not in anchor_objects_set and obj.event_cfg is not None:
+                obj.event_cfg = None
```
alexmillane: I'm wondering if it's an error that non-anchor objects have pose-reset events in the case of relational solving. Do you know if what cases this is actually doing something?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -102,28 +102,26 @@ def _solve_relations(self) -> None:
...
+
+        placer_params = ObjectPlacerParams(
+            placement_seed=getattr(self.args, "placement_seed", None),
+        )
```
alexmillane, line 102: I think we should make an unguarded access of `self.args.placement_seed`. Something has gone wrong if we end up here without `placement_seed`.

isaaclab_arena/relations/placement_events.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Solver needs autograd, so temporarily exit any active inference_mode context.
+    reset_solver_params = replace(placer_params.solver_params, save_position_history=False)
+    reset_params = replace(
+        placer_params,
+        solver_params=reset_solver_params,
+        apply_positions_to_objects=False,
+        verbose=False,
+        placement_seed=None,
+    )
```
alexmillane: Why do we have to force the following params:
* save_position_history=False,
* apply_positions_to_objects=False
* verbose=False

I can see the justification for the `placement_seed` (although perhaps we should set that to false one level above and then warn if it's not `None` at this level).

isaaclab_arena/relations/placement_events.py
```diff
@@ -0,0 +1,87 @@
...
+    placer = ObjectPlacer(params=reset_params)
+    with torch.inference_mode(False):
+        result = placer.place(objects, num_envs=num_reset_envs, result_per_env=True)
+
+    if isinstance(result, MultiEnvPlacementResult):
+        results_per_env = result.results
+    else:
+        results_per_env = [result]
```
alexmillane: This strikes me as a little strange. If we immediately convert the single environment solution to a list on return, should we just always return a `MultiEnvPlacementResult`?

isaaclab_arena/relations/placement_events.py
```diff
@@ -0,0 +1,87 @@
...
+
+    anchor_objects_set = set(get_anchor_objects(objects))
+    rotations: dict[ObjectBase, tuple[float, float, float, float]] = {}
+    for obj in objects:
+        if obj not in anchor_objects_set:
+            rotate_marker = next((r for r in obj.get_relations() if isinstance(r, RotateAroundSolution)), None)
+            rotations[obj] = rotate_marker.get_rotation_xyzw() if rotate_marker else (0.0, 0.0, 0.0, 1.0)
+
+    # Always write positions, even when validation failed — the solver's best-effort
+    # is still better than the USD defaults (objects at origin).
```
alexmillane: I guess we warn if the solving failed? Seems like we should warn before writing.

isaaclab_arena/relations/placement_events.py
```diff
@@ -0,0 +1,87 @@
...
+        positions = results_per_env[local_idx].positions
+        for obj, pos in positions.items():
+            if obj in anchor_objects_set:
+                continue
+            asset = env.scene[obj.name]
+            pose = Pose(position_xyz=pos, rotation_xyzw=rotations[obj])
+            pose_t_xyz_q_xyzw = pose.to_tensor(device=env.device).unsqueeze(0)
+            pose_t_xyz_q_xyzw[0, :3] += env.scene.env_origins[cur_env, :]
+            asset.write_root_pose_to_sim(pose_t_xyz_q_xyzw, env_ids=env_id_tensor)
+            asset.write_root_velocity_to_sim(zero_velocity, env_ids=env_id_tensor)
```
alexmillane, line 67: Should we put this in a function somewhere if it's not already?

isaaclab_arena/relations/placement_events.py
```diff
@@ -0,0 +1,87 @@
...
+    pose events for solver-managed objects. Each call solves for the subset of
+    environments being reset, producing a new random layout per env.
+
+    Args:
+        env: The Isaac Lab environment.
+        env_ids: 1-D tensor of environment indices being reset.
+        objects: All objects (including anchors) participating in relation solving.
+        placer_params: Parameters forwarded to ``ObjectPlacer``. Seed is forced to
+            ``None`` so each reset gets fresh randomness.
+    """
```
alexmillane, line 45: One big question I have about this approach is that it's effectively gotten rid of our batching behaviour. One can imagine that over time, the (potentially 1000s) of environments in an evaluation will all eventually reset at different indices, which will mean that this event fires for 1 `env` frequently.

With the current design that will mean solving a lot of small problems frequently, rather than solving one big problem infrequently. This is likely to mean poor performance on the GPU.

How does the performance look. What do you think about the approach of solving a bigger batched problem and saving to a solution pool?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -102,28 +102,26 @@ def _solve_relations(self) -> None:
 
         self._add_pairwise_no_collision(objects_with_relations)
 
-        # Run the ObjectPlacer (default on_relation_z_tolerance_m accommodates solver residual).
-        # Positions are applied to objects via set_initial_pose (single-env: Pose/PoseRange,
-        # multi-env: PosePerEnv), so each object's event_cfg handles its own reset.
-        placement_seed = getattr(self.args, "placement_seed", None)
-        placer = ObjectPlacer(params=ObjectPlacerParams(placement_seed=placement_seed))
-        num_envs = self.args.num_envs
-        result = placer.place(objects_with_relations, num_envs=num_envs)
```
alexmillane, line 96: So here, we've removed the upfront solve. So at the moment we're _enforcing_ that we have a new solution per reset.

I could imagine that some users may still want that behavior. It would be great if we had a parameter that controlled `resolve_on_reset`.

This actually relates to another comment further down about a solution pool. If we change the design slightly to draw solutions from a pool, implementing `reset_on_solve = False` is easy: we just don't remove solutions from the pool when they're used.

alexmillane (commented):
This is looking pretty good!

I have some small comments, but the big one is about speed and the potential use of a solution pool, in order to run bigger problem sizes, less frequently.

Let's discuss this. It's a good first stab at the problem.

isaaclab_arena/environments/arena_env_builder.py
zhx06: Right now non-anchor objects don't have set_initial_pose() called, so it's not an error and these object never gets a pose-reset event

isaaclab_arena/environments/arena_env_builder.py
zhx06: Sure, we will use `self.args.placement_seed`.

isaaclab_arena/relations/placement_events.py
zhx06: We have now moved them to construction into `arena_env_builder.py`. `save_position_history` is turned off for object reset because it is for verification purpose. We don't use `apply_positions_to_objects` because we write them directly to sim. `verbose` is turned off so we removed this forcing here.

isaaclab_arena/relations/placement_events.py
zhx06: This comes from the ObjectPlacer where we have `MultiEnvPlacementResult` for n_envs>1 and `PlacementResult` for n_envs=1. Do we want to edit the design in that level?

isaaclab_arena/relations/placement_events.py
zhx06: Warning added

isaaclab_arena/relations/placement_events.py
zhx06: These 4 lines are similar in some other code for setting object poses, but has slight differences. I would suggest we keep it for now and do the cleaning later since this touches lots of things in relation solver.

isaaclab_arena/relations/placement_events.py
zhx06: I tested with `num_envs=200` and `num_envs=10` (both 10 max_attempts). They both take like 10 seconds. So the solver batches well, but the per-call overhead is significant for repeated resets. We could have a pool, but let me think about it and have a better design

isaaclab_arena/environments/arena_env_builder.py
alexmillane: Rather than setting the their `event_cfg` to `None` should we just `assert obj.event_cfg = None, "Relational object solving should not be combined with explicit setting of poses on non-anchor objects`.

Right now, if a user tries to combine these things, the code will silently ignore the explicit poses. IMO it's better to fail and tell the user that they're doing something incorrect.

isaaclab_arena/relations/placement_events.py
alexmillane: Yea, maybe we should leave that for another MR. I think we should consider it in general.

isaaclab_arena/relations/placement_events.py
alexmillane: :+1:

isaaclab_arena/relations/placement_events.py
alexmillane: Yea that's matches expectation. The solve time is basically independent of batch size (at the small sizes we're dealing with). I think that makes having a pool look like a good option (because we get additional solutions for free at solve-time).

Let me know when you'd like me to re-review :)

isaaclab_arena/cli/isaaclab_arena_cli.py
```diff
@@ -66,6 +66,15 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
             "When not set, each environment uses its own default."
         ),
     )
+    arena_group.add_argument(
+        "--resolve_on_reset",
+        action=argparse.BooleanOptionalAction,
```
alexmillane, line 71: I llike this `BooleanOptionalAction`, I didn't know about that!

isaaclab_arena/environments/arena_env_builder.py
alexmillane: Suggestion to move this to `Scene`? Wdyt?

alexmillane: What are we protecting against here? I'm pretty sure that the scene only contain objects (derived from `ObjectBase`.

Looking a bit deeper I think that the issue is that the type hinting is wrong in `Scene` which is leading Claude to be overly protective here.

Unless there's a legit situation were protecting against, I would remove this whole if clause and just assume everything in this list has the function `get_relations`.

alexmillane: *Edit: I am making another MR to correct the type annotations in `Scene`.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -78,37 +82,113 @@ def _solve_relations(self) -> None:
...
-            if n_succeeded == num_envs:
-                print(f"Relation solving succeeded for all {num_envs} env(s) after {result.attempts} attempt(s)")
-            else:
-                print(
-                    f"Relation solving: {n_succeeded}/{num_envs} env(s) passed validation after"
-                    f" {result.attempts} attempt(s)."
-                )
-        elif result.success:
-            print(f"Relation solving succeeded after {result.attempts} attempt(s)")
+        cli_resolve = getattr(self.args, "resolve_on_reset", None)
```
alexmillane: just access directly.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -78,37 +82,113 @@ def _solve_relations(self) -> None:
...
+        anchor_objects_set: set,
+    ) -> None:
+        """Set ``object_cfg.init_state`` from a pool layout so objects spawn at valid positions.
+
+        Only touches ``init_state.pos`` / ``init_state.rot`` — does NOT create
+        per-object reset events (the placement event handles resets).
+        """
+        layout = pool.sample(1)[0]
+        for obj in objects:
+            if obj in anchor_objects_set or obj.object_cfg is None:
```
alexmillane: What's with this second condition `obj.object_cfg is None`?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -78,37 +82,113 @@ def _solve_relations(self) -> None:
...
+        objects: list[Object | ObjectReference],
+        pool: PlacementPool,
+        num_envs: int,
+    ) -> None:
+        """Draw layouts from the pool and apply them to objects via ``set_initial_pose``.
+
+        Each non-anchor object gets a :class:`~isaaclab_arena.utils.pose.PosePerEnv`
+        so that per-object reset events restore these positions.
+        """
+        from isaaclab_arena.utils.pose import Pose, PosePerEnv
```
alexmillane: move this to the top of the file with the other imports.

isaaclab_arena/relations/placement_events.py
```diff
@@ -0,0 +1,67 @@
...
+        env: The Isaac Lab environment.
+        env_ids: 1-D tensor of environment indices being reset.
+        objects: All objects (including anchors) participating in relation solving.
+        placement_pool: Pre-solved pool of layouts to draw from.
+    """
+    if env_ids is None or len(env_ids) == 0:
+        return
+
+    num_reset_envs = len(env_ids)
+    results_per_env = placement_pool.acquire(num_reset_envs)
```
alexmillane: I've seen `placement_pool.acquire(N)` and `placement_pool.sample(N)`.

What's the difference?

isaaclab_arena/relations/placement_pool.py
```diff
@@ -0,0 +1,126 @@
...
+from isaaclab_arena.relations.object_placer import ObjectPlacer
+from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
+from isaaclab_arena.relations.placement_result import MultiEnvPlacementResult, PlacementResult
+
+if TYPE_CHECKING:
+    from isaaclab_arena.assets.object_base import ObjectBase
+
+
+class PlacementPool:
+    """Pre-solved pool of valid placement layouts.
```
alexmillane: "pre-solved" is incorrect here. This object actually does the solve regularly.

isaaclab_arena/relations/placement_pool.py
```diff
@@ -0,0 +1,126 @@
...
+
+from isaaclab_arena.relations.object_placer import ObjectPlacer
+from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
+from isaaclab_arena.relations.placement_result import MultiEnvPlacementResult, PlacementResult
+
+if TYPE_CHECKING:
+    from isaaclab_arena.assets.object_base import ObjectBase
+
+
+class PlacementPool:
```
alexmillane: This object is more than a pool, because it does the solve.

`PooledObjectPlacer`?

isaaclab_arena/relations/placement_pool.py
```diff
@@ -0,0 +1,126 @@
...
+    """Pre-solved pool of valid placement layouts.
+
+    Solves ``pool_size`` layouts at build time and keeps only those that
+    pass validation.  Valid layouts are shuffled so successive draws
+    provide variety across resets.
```
alexmillane: I'm not sure I understand the need to shuffle? If each solution is unique and only used once, why bother shuffling?

isaaclab_arena/relations/placement_pool.py
```diff
@@ -0,0 +1,126 @@
...
+
+    * :meth:`acquire` — returns the next *count* layouts sequentially
+      (without replacement).  Auto-refills when exhausted.
+    * :meth:`sample` — picks *count* layouts at random with replacement
+      (non-consuming).  Used for static initial positions.
```
alexmillane: It's hard to know which of these means which option.

What do you think about making this more explicit:
- `sample_with_replacement`
- `sample_without_replacement`

isaaclab_arena/relations/placement_pool.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Internally the pool is a single list with a read index, following the
+    same pattern as ``stable-baselines3.ReplayBuffer`` (single storage +
+    position index).
```
alexmillane: Suggestion to remove.

isaaclab_arena/relations/placement_pool.py
```diff
@@ -0,0 +1,126 @@
...
+    def __init__(
+        self,
+        objects: list[ObjectBase],
+        placer_params: ObjectPlacerParams,
+        pool_size: int = 100,
+    ) -> None:
+        if pool_size < 1:
+            raise ValueError(f"pool_size must be >= 1, got {pool_size}")
+
+        self._objects = list(objects)
```
alexmillane: This seems unnecessary given that the input type annotation say that this guy should be a list? Are we expecting non-lists to be passed in (if so we should probably change the type annotation above).

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
```diff
@@ -97,12 +98,28 @@ def get_env(self, args_cli: argparse.Namespace):  # -> IsaacLabArenaEnvironment:
                 light,
             ]
         )
+
+        episode_length_s = getattr(args_cli, "episode_length_s", None)
```
alexmillane: args_cli.episode_length_s

alexmillane (approved):
Looks pretty good.

I have a few nits, but I think it's basically ready to go. I'm approving. Feel free to merge after addressing the issues.

isaaclab_arena/relations/placement_pool.py
zhx06: Yeah, we don't need to shuffle. This is removed.

isaaclab_arena/environments/arena_env_builder.py
zhx06: Sure, this is moved to `Scene`.

isaaclab_arena/relations/placement_pool.py
zhx06: The objects should always be a list. Now `self._objects = objects `.

isaaclab_arena_environments/gr1_table_multi_object_no_collision_environment.py
zhx06: Now it's `episode_length_s = args_cli.episode_length_s`

isaaclab_arena/relations/placement_pool.py
zhx06: Sure, this will make the function names easier to read. Changes adopted.

isaaclab_arena/environments/arena_env_builder.py
zhx06: `object_cfg` can  never be `None`. I have removed the check.

PR #596, opened 2026-04-14
Skip slow camera test and use run_subprocess() for eval_runner tests

xyao-nv:
## Summary
CI subprocess tests are slow and faced with timeout without stalling

## Detailed description
- Skipped `test_eval_runner_enable_cameras` as cold-start camera rendering takes ~1165s, making CI exceeding the timeout.
- Replaced raw `subprocess.run()` with the shared `run_subprocess()` helper, which enforces `ISAACLAB_ARENA_SUBPROCESS_TIMEOUT` (900s in CI).
- Removed redundant stdout-regex failure check; the eval_runner already exits non-zero on job failure (no` --continue_on_error`).

### DIFF
--- a/isaaclab_arena/tests/test_eval_runner.py
+++ b/isaaclab_arena/tests/test_eval_runner.py
@@ -5,12 +5,11 @@
 
 import json
 import os
-import re
-import subprocess
 
 import pytest
 
 from isaaclab_arena.tests.utils.constants import TestConstants
+from isaaclab_arena.tests.utils.subprocess import run_subprocess
 
 HEADLESS = True
 NUM_STEPS = 2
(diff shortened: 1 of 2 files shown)


isaaclab_arena/tests/test_eval_runner.py
```diff
@@ -23,41 +22,24 @@ def write_jobs_config_to_file(jobs: list[dict], tmp_file_path: str):
...
-    # Filter out the header row
-    job_statuses = [(name.strip(), status.strip()) for name, status in matches if name.strip() != "Job Name"]
-
-    # Check for failed jobs
-    failed_jobs = [name for name, status in job_statuses if status.lower() == "failed"]
-
-    if failed_jobs:
-        print("\n" + output)  # Print full output for debugging
-        raise AssertionError(f"The following jobs failed: {', '.join(failed_jobs)}\nAll job statuses: {job_statuses}")
+    run_subprocess(args)
```
alexmillane, line 42: good cleanup!

alexmillane (approved):
Seems reasonable. Of course we'd like to re-enable this, but this seems reasonable while we work out a path to doing so.

PR #598, opened 2026-04-14
Refactor environment registration to use a registry.

alexmillane:
## Summary
Move's our environment registration from a static map to the registry pattern that we use for assets etc.

## Detailed description
- Using a static map meant that a user registering a new environment would have to add their environment by-hand to a static map.
- Now registration is automatic when a class subclassing `ExampleEnvironment` is added to the folder in `issaclab_arena_environments`


isaaclab_arena_environments/cli.py
```diff
@@ -5,56 +5,40 @@
...
+                continue
+            module = importlib.import_module(f"isaaclab_arena_environments.{modname}")
+            for _attr_name, obj in inspect.getmembers(module, inspect.isclass):
+                if (
+                    issubclass(obj, ExampleEnvironmentBase)
+                    and obj is not ExampleEnvironmentBase
+                    and getattr(obj, "name", None) is not None
+                    and obj.name not in registry._components
+                ):
+                    registry.register(obj, obj.name)
```
alexmillane: Suggestion to move this to another file. `cli.py` doesn't quite fit.

isaaclab_arena_environments/cli.py
```diff
@@ -5,56 +5,40 @@
...
-    CubeGoalPoseEnvironment.name: CubeGoalPoseEnvironment,
-    DexsuiteLiftEnvironment.name: DexsuiteLiftEnvironment,
-    LiftObjectEnvironment.name: LiftObjectEnvironment,
-    TableTopPlaceUprightEnvironment.name: TableTopPlaceUprightEnvironment,
-    Gr1TurnStandMixerKnobEnvironment.name: Gr1TurnStandMixerKnobEnvironment,
-    GR1TableMultiObjectNoCollisionEnvironment.name: GR1TableMultiObjectNoCollisionEnvironment,
-}
+from isaaclab_arena_environments.example_environment_base import ExampleEnvironmentBase
+
+_environments_registered = False
```
alexmillane: Can we get away without the global? Just re-register on each call to this function (and be OK if an environment is already registered).

isaaclab_arena_environments/cli.py
```diff
@@ -90,19 +74,23 @@ def parse_and_return_external_environment_from_string(environment_path: str) ->
...
+    env_registry = EnvironmentRegistry()
+
     args, unknown = args_parser.parse_known_args()
     environment = getattr(args, "external_environment_class_path", None)
     if environment is not None:
-        # Update the ExampleEnvironments dictionary with the new external environment
         print(f"Adding external environment: {environment}")
-        ExampleEnvironments.update(parse_and_return_external_environment_from_string(environment))
+        for name, cls in parse_and_return_external_environment_from_string(environment).items():
+            env_registry.register(cls, name)
```
alexmillane: This seems wrong. Why are we looping over something. The parsing of the string should return a single environment.

alexmillane (commented):
Self review

isaaclab_arena/assets/register.py
```diff
@@ -57,3 +58,13 @@ def register_hdr(cls):
     else:
         HDRImageRegistry().register(cls, cls.name)
     return cls
+
+
+# Decorator to register an environment with the EnvironmentRegistry.
+def register_environment(cls):
+    registry = EnvironmentRegistry()
+    if cls.name in registry._components:
```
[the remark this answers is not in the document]
xyao-nv: I like what bot suggests better

isaaclab_arena/assets/asset_registry.py
```diff
@@ -210,6 +210,13 @@ def get_random_hdr_by_tag(self, tag: str) -> type["HDRImage"]:
         return random.choice(hdrs)
 
 
+class EnvironmentRegistry(Registry):
```
xyao-nv, line 213: Shall this file be renamed as `registry.py`?

isaaclab_arena/assets/asset_registry.py
xyao-nv, line 235: Are those assets? Or shall we rename func to reflect what are included exactly?

xyao-nv, line 159: Minor: Policy is not asset, and it stays in this asset_registry.py.

isaaclab_arena_environments/cli.py
```diff
@@ -5,56 +5,40 @@
...
+        _environments_registered = True
+        registry = EnvironmentRegistry()
+        for _importer, modname, ispkg in pkgutil.iter_modules(isaaclab_arena_environments.__path__):
+            if ispkg:
+                continue
+            module = importlib.import_module(f"isaaclab_arena_environments.{modname}")
+            for _attr_name, obj in inspect.getmembers(module, inspect.isclass):
+                if (
+                    issubclass(obj, ExampleEnvironmentBase)
+                    and obj is not ExampleEnvironmentBase
```
xyao-nv: Is that because those envs could not be within the repo (e.g. collaborators use Arena), that it has to implement the registration ops similar to the decreator of `def register_environment`? Can it share the same func?

isaaclab_arena_environments/cli.py
```diff
@@ -5,52 +5,40 @@
...
+                    and getattr(obj, "name", None) is not None
+                    and obj.name not in registry._components
+                ):
+                    registry.register(obj, obj.name)
```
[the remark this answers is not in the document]
alexmillane: I see your point. This is indeed redundant. What do you think about removing the loop, and using a top level import in the __init__.py file which imports all environments with `import isaaclab_arena_environments`

isaaclab_arena_environments/cli.py
alexmillane: fixed

isaaclab_arena/assets/register.py
```diff
@@ -57,3 +58,13 @@ def register_hdr(cls):
...
+        print(f"WARNING: Environment {cls.name} is already registered. Doing nothing.")
+    else:
+        registry.register(cls, cls.name)
+    return cls
```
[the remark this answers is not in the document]
alexmillane: fixed

isaaclab_arena/assets/register.py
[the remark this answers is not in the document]
alexmillane: Agreed. Fixed

isaaclab_arena/assets/asset_registry.py
alexmillane: I will do this in a follow up as it's likely to touch many files beyond what this MR is about. But it's a good call 😅

isaaclab_arena/assets/asset_registry.py
alexmillane: Agreed. Will do that in a follow up. See comment above.

alexmillane: See comment above.

isaaclab_arena_environments/cli.py
alexmillane: I rewrote this. Thanks for pointing out that this is weird.

xyao-nv (approved):

PR #605, opened 2026-04-15
Correct the CloudXR config for Quest handtracking

qianl-nv:
## Summary
The config in the doc is mistakenly set for AVP instead of for Quest/Pico handtracking. Correcting the doc.
This fixes issue reported from https://nvbugspro.nvidia.com/bug/6076546


alexmillane (approved):
Thanks for doing this!

PR #464, opened 2026-03-04
Lab 3.0 (Newton)

alexmillane:
## Summary
This is the working branch for upgrading Isaac Lab Arena to Lab 3.0 with Newton.

## Detailed description
- Upgrade to latest lab.

PR #611, opened 2026-04-16
Add G1 loco-manipulation apple-to-plate environment

bokalnv:
## Summary

Generalize the existing `galileo_g1_locomanip_pick_and_place` environment so the same entry point drives both the original brown-box-to-blue-bin workflow (unchanged) and a new apple-to-plate variant.

### Environment (`isaaclab_arena_environments/galileo_g1_locomanip_pick_and_place_environment.py`)

- Add `--destination` (default `blue_sorting_bin`) and `--task_description` CLI args. Destination is no longer hardcoded.
- Auto-template the natural-language task description from `--object` / `--destination`, but **preserve the exact legacy string verbatim** when both are at their defaults. This keeps the v0.2 pretrained gr00t workflow bit-identical — any wording drift is gated behind an explicit CLI override or a non-default destination.
- Pass tighter 10 cm proximity thresholds to the task when running the apple-on-plate combo (`--object apple_01_objaverse_robolab --destination clay_plates_hot3d_robolab`). The box+bin defaults of 0.26 / 0.13 / 0.15 m are too loose for a ~30 cm plate.
- Replace `hasattr(args_cli, ...)` guards on the mimic patch with `getattr(args_cli, ..., False)`. The previous `not hasattr(args_cli, "auto")` was also subtly wrong — it would skip the mimic patch whenever `--auto` was registered as an argparse flag, even when left at `False`.

### Asset registry (`isaaclab_arena/assets/object_library.py`)

- Add `Apple01ObjaverseRobolab` and `Apple02ObjaverseRobolab` with baked-in `scale=(0.01, 0.01, 0.01)`. Objaverse meshes ship at ~100× real-world scale, so embedding the downscale in the registry class keeps callers free of per-site scale factors.

### Task (`isaaclab_arena/tasks/g1_locomanip_pick_and_place_task.py`)

- Add `pick_up_object_name` to `G1LocomanipPickPlaceMimicEnvCfg` (default `"brown_box"`). `G1LocomanipPickAndPlaceTask.get_mimic_env_cfg` passes `self.pick_up_object.name` so the Mimic pipeline works with arbitrary objects.
- Include `pick_up_object_name` in `datagen_config.name` so apple and box runs no longer collide on a shared `g1_locomanip_pick_and_place_D0` dataset name.
- Parameterize `max_x_separation`, `max_y_separation`, `max_z_separation` on `__init__`, defaulting to the existing box+bin values (0.260 / 0.130 / 0.150 m). Callers that need tighter thresholds (e.g. apple-on-plate) pass them in at construction time.

### Tests (`isaaclab_arena/tests/test_g1_locomanip_apple_to_plate.py`)

- `test_initial_state_not_terminated`: apple spawned well off the plate, confirmed the task does not terminate.
- `test_apple_on_plate_succeeds`: apple re-teleported to plate center each step so short physics drift cannot push it past the proximity threshold before termination fires. Exercises the production 10 cm thresholds.
- `test_mimic_cfg_uses_object_name`: non-sim unit test verifying `pick_up_object_name`, `datagen_config.name`, and every subtask's `object_ref` propagate correctly.

## Backward compatibility
- The brown-box-to-blue-bin task description is preserved verbatim when `--object` / `--destination` are at their defaults, so the v0.2 pretrained gr00t model keeps bit-identical behavior. Follow-up SQA run on the apple-to-plate wording can happen independently of this merge.
- `G1LocomanipPickAndPlaceTask` proximity thresholds default to the pre-PR values, so any other caller of this task class is unaffected.


isaaclab_arena/tests/test_g1_locomanip_apple_to_plate.py
```diff
@@ -0,0 +1,233 @@
...
+def _replace_apple_at_initial_pose(env, apple) -> None:
+    """Teleport the apple back to its initial position with zero velocity."""
+    from isaaclab.assets import RigidObject
+
+    with torch.inference_mode():
+        apple_object: RigidObject = env.unwrapped.scene[apple.name]
+        init_pos = torch.tensor([[0.15, 0.15, 0.05]], device=env.unwrapped.device)
+        init_quat = torch.tensor([[1.0, 0.0, 0.0, 0.0]], device=env.unwrapped.device)
+        apple_object.write_root_pose_to_sim(root_pose=torch.cat([init_pos, init_quat], dim=-1))
+        apple_object.write_root_velocity_to_sim(root_velocity=torch.zeros((1, 6), device=env.unwrapped.device))
```
alexmillane: We have a function that allows for skipping this duplicated code:

```
apple.set_object_pose(
   env,
   Pose(
      position_xyz=(0.15, 0.15, 0.05),
      rotation_xyzw=(0.0, 0.0, 0.0, 1.0),
   )
)
```

Btw Isaac Lab uses xyzw on Isaac Lab 3.0. Not sure if you intend this to be the identity quaternion?

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
```diff
@@ -0,0 +1,115 @@
...
+    def get_env(self, args_cli: argparse.Namespace) -> IsaacLabArenaEnvironment:
+        from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
+        from isaaclab_arena.scene.scene import Scene
+        from isaaclab_arena.tasks.g1_locomanip_pick_and_place_task import G1LocomanipPickAndPlaceTask
+        from isaaclab_arena.utils.pose import Pose, PoseRange
+
+        background = self.asset_registry.get_asset_by_name("galileo_locomanip")()
+        # Objaverse assets are ~100x real-world scale; downscale to realistic size.
+        pick_up_object = self.asset_registry.get_asset_by_name(args_cli.object)(scale=(0.01, 0.01, 0.01))
+        destination = self.asset_registry.get_asset_by_name(args_cli.destination)(scale=(0.5, 0.5, 0.5))
```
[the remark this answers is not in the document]
alexmillane: I kinda agree here.

Given that we're scaling by some object-specific factor, let's just omit the CLI arguments and load a specified, fixed object from the registry.

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
```diff
@@ -0,0 +1,115 @@
...
+            Pose(
+                position_xyz=(-0.2450, -1.6272, -0.2641),
+                rotation_xyzw=(0.0, 0.0, 1.0, 0.0),
+            )
+        )
+        embodiment.set_initial_pose(Pose(position_xyz=(0.0, 0.18, 0.0), rotation_xyzw=(0.0, 0.0, 0.0, 1.0)))
+
+        if (
+            args_cli.embodiment == "g1_wbc_pink"
+            and hasattr(args_cli, "mimic")
```
alexmillane: suggestion to remove this `hasattr(args_cli, "mimic")`.

I feel that claude is overly protective on these `hasattr` checks.

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
```diff
@@ -0,0 +1,115 @@
...
+            and args_cli.mimic
+            and not hasattr(args_cli, "auto")
```
alexmillane: I can't see where we have a CLI argument `auto`. Do you know where this comes from? Or is this something inherited from the other g1_locomanip env?

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
alexmillane: Is there a reason we need to scale the plate?

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
alexmillane, line 1: This file appears (to the eye) to be an exact duplicate of `galileo_g1_locomanip_pick_and_place_environment` with:
- the default object arguments changed
- Some scales changed.

Could we change the default scale of the apple in the asset library and then delete this whole file and use `galileo_g1_locomanip_pick_and_place_environment` with the appropriate CLI arguments?

Or is there something I'm missing?

alexmillane (commented):
Thanks for this!

Looks pretty good. My only qualm is that the environment appears to be a near duplicate of an existing environment with the asset names changed? Can we get rid of this duplication?

alexmillane:
@cvolkcvolk Another question, and I suspect I know the answer, but have you tried the object placer on this scene? It would be great if we could use it to place the objects to elimate these hardcoded positions. I suspect the issue would be getting a bounding box for the particular shelf (i.e. is it seperable from the other shelves above).

cvolkcvolk:
> @cvolkcvolk Another question, and I suspect I know the answer, but have you tried the object placer on this scene? It would be great if we could use it to place the objects to elimate these hardcoded positions. I suspect the issue would be getting a bounding box for the particular shelf (i.e. is it seperable from the other shelves above).

Exactly. The shelve here is only one single prim. We would need each level to be its own mesh/ subprim

isaaclab_arena/tests/test_g1_locomanip_apple_to_plate.py
```diff
@@ -0,0 +1,234 @@
...
+    from isaaclab_arena.environments.arena_env_builder import ArenaEnvBuilder
+    from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
+    from isaaclab_arena.scene.scene import Scene
+    from isaaclab_arena.tasks.g1_locomanip_pick_and_place_task import G1LocomanipPickAndPlaceTask
+    from isaaclab_arena.utils.pose import Pose
+
+    asset_registry = AssetRegistry()
+    background = asset_registry.get_asset_by_name("table")()
+    apple = asset_registry.get_asset_by_name("apple_01_objaverse_robolab")()
+    plate = asset_registry.get_asset_by_name("clay_plates_hot3d_robolab")()
```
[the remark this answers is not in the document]
bokalnv: Fixed by baking the scale into the registry instead of applying it
per-caller: `Apple01ObjaverseRobolab` (and `Apple02ObjaverseRobolab`) now
default to `scale=(0.01, 0.01, 0.01)`. Also removed the plate scale —
empirical bbox measurement in-sim showed `clay_plates_hot3d_robolab` ships
at a realistic ~30 cm diameter, so the 0.5x factor was wrong. The test and
env now both load the assets through the registry without any external
scale factors. (22cb8d54, e7b07a4f)

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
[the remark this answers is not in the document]
bokalnv: Resolved by deleting the dedicated apple-to-plate env and moving canonical
scale into the asset registry. `galileo_g1_locomanip_pick_and_place` no
longer applies any external scale — each asset is responsible for its own
canonical size. (22cb8d54)

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
bokalnv: Dropped both `hasattr` guards in e7b07a4f:

```python
if (
    args_cli.embodiment == "g1_wbc_pink"
    and getattr(args_cli, "mimic", False)
    and not getattr(args_cli, "auto", False)
):
```

Slight bonus: the previous `not hasattr(args_cli, "auto")` was subtly
wrong — it would skip the mimic patch whenever `--auto` was registered as
an argparse flag at all (even when left at `False`). The
`getattr(..., False)` form means "mimic is truthy AND auto is not truthy",
which is the behaviour we actually want.

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
bokalnv: Inherited from the existing box pick-and-place env. `--auto` is registered
by the automatic-evaluation entry point (the headless eval wrapper adds it
before calling `get_env`); interactive teleop doesn't register it. The
original `not hasattr(args_cli, "auto")` was trying to say "we're not in
the auto-eval flow", which is now `not getattr(args_cli, "auto", False)` —
same intent, and also correct when `--auto` is registered but left
`False`. See the reply on the `hasattr("mimic")` thread for the updated
snippet.

isaaclab_arena/tests/test_g1_locomanip_apple_to_plate.py
bokalnv: Thanks — switched the test to
`apple.set_object_pose(env, Pose(position_xyz=..., rotation_xyzw=...))`.
The quaternion is `(0.0, 0.0, 0.0, 1.0)` under the Isaac Lab xyzw
convention, which is the identity — intentional for keeping the apple
upright. (22cb8d54)

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
bokalnv: No — the scale was wrong. I spawned the plate unscaled and measured the
bounding box in-sim: `clay_plates_hot3d_robolab` is ~30 cm in diameter at
`scale=(1,1,1)`, which is what we want. The 0.5x factor was a copy-paste
residue. The plate is now loaded with the default scale, and
`clay_plates_hot3d_robolab` has no `scale` override in the registry.
(22cb8d54)

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
bokalnv: Done. `galileo_g1_locomanip_apple_to_plate_environment.py` and its CLI
registration are deleted. `galileo_g1_locomanip_pick_and_place_environment.py`
grew a `--destination` CLI arg (default `blue_sorting_bin`) plus a
templated `--task_description` derived from `--object`/`--destination` if
not explicitly overridden. Apple-to-plate is now:

```
--env galileo_g1_locomanip_pick_and_place \
--object apple_01_objaverse_robolab \
--destination clay_plates_hot3d_robolab
```

Apple scale baked into `Apple01ObjaverseRobolab` / `Apple02ObjaverseRobolab`
in the asset registry (e7b07a4f), plate scale dropped (verified visually —
see separate thread on `--destination`). No external scale factors remain
anywhere in the env code. (22cb8d54)

isaaclab_arena_environments/galileo_g1_locomanip_pick_and_place_environment.py
```diff
@@ -75,20 +74,27 @@ def get_env(self, args_cli: argparse.Namespace) -> IsaacLabArenaEnvironment:
...
-        scene = Scene(assets=[background, pick_up_object, blue_sorting_bin])
+        if args_cli.task_description is not None:
+            task_description = args_cli.task_description
+        else:
+            object_label = args_cli.object.replace("_", " ")
+            destination_label = args_cli.destination.replace("_", " ")
+            task_description = (
+                f"Pick up the {object_label} from the shelf, and place it on the {destination_label} on the table"
+                " located at the right of the shelf."
+            )
```
alexmillane, line 135: @bokalnv This would be a good change... but unfortunately we have a slight issue here...

The problem is that this changes the language instruction before and after this change, because "blue bin" -> "blue sorting bin"

This file is used in a workflow for v0.2, which comes with a pretrained gr00t model. This workflow+model was just (last week) was SQA'ed for release with v0.2. This minor change may change gr00t behaviour, which would mean we can't merge it into `main` (which is going to be released).

To confirm what we should do, we need to test [this command](https://isaac-sim.github.io/IsaacLab-Arena/main/pages/example_workflows/sequential_static_manipulation/step_5_evaluation.html#step-1-run-single-environment-evaluation) on this branch, with the slightly changed language instruction, and confirm that the performance of the model is still acceptable.

@xyao-nv for viz.

alexmillane (commented):
Thank you for addressing the comments. This looks clean!

However, we need to run one test to ensure this change is safe to merge. See comment above.

isaaclab_arena_environments/galileo_g1_locomanip_pick_and_place_environment.py
bokalnv: Thanks for flagging this, I addressed it by adding explicit backwards compat handling so the old/existing instructions works without change.

isaaclab_arena/tests/test_g1_locomanip_apple_to_plate.py
```diff
@@ -0,0 +1,247 @@
...
+
+    from isaaclab.assets import RigidObject
+
+    env, apple, plate = get_test_environment(num_envs=1)
+
+    try:
+        _step_with_standing_actions(env, WARMUP_STEPS)
+
+        plate_object: RigidObject = env.unwrapped.scene[plate.name]
+        plate_pos_world = wp.to_torch(plate_object.data.root_pos_w)[0]
```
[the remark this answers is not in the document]
bokalnv: Thanks for the flag, but I believe this is a false positive. Arena is on Lab 3.0 (Newton) (#464), where RigidObjectData.root_pos_w returns a wp.array, not a torch.Tensor (see isaaclab_newton/.../rigid_object_data.py: def root_pos_w(self) -> wp.array). The wp.to_torch(object.data.root_pos_w) idiom is used throughout Arena — tasks/terminations.py, tasks/rewards/rewards.py, tasks/observations/observations.py, terms/transforms.py, and the existing tests/test_sorting_task.py — so if this were raising a TypeError the whole task framework (including the objects_in_proximity termination this test exercises) would be broken.

Also: the try/except in the test returns False on any exception, and the outer pytest wrapper asserts result, so a TypeError would fail the test loudly rather than silently pass. Happy to add a pytest.fail(f"unexpected: {e}") in the except if you'd prefer an even louder failure mode.

isaaclab_arena/tasks/g1_locomanip_pick_and_place_task.py
```diff
@@ -95,17 +104,29 @@ class TerminationsCfg:
...
+
     def __post_init__(self):
         # post init of parents
         super().__post_init__()
 
-        self.datagen_config.name = "g1_locomanip_pick_and_place_D0"
+        if self.pick_up_object_name == "brown_box":
+            self.datagen_config.name = _LEGACY_BROWN_BOX_DATAGEN_NAME
+        else:
+            self.datagen_config.name = f"g1_locomanip_pick_and_place_{self.pick_up_object_name}_D0"
```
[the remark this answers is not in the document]
bokalnv: Fixed as well

isaaclab_arena/assets/object_library.py
```diff
@@ -1646,6 +1648,7 @@ class Apple02ObjaverseRobolab(LibraryObject):
     usd_path = (
         f"{ISAACLAB_STAGING_NUCLEUS_DIR}/Arena/assets/object_library/srl_robolab_assets/objects/objaverse/apple_02.usd"
     )
+    scale = (0.01, 0.01, 0.01)
```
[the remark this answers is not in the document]
bokalnv: fixed

isaaclab_arena_environments/galileo_g1_locomanip_pick_and_place_environment.py
```diff
@@ -75,20 +74,27 @@ def get_env(self, args_cli: argparse.Namespace) -> IsaacLabArenaEnvironment:
...
+
+        scene = Scene(assets=[background, pick_up_object, destination])
         isaaclab_arena_environment = IsaacLabArenaEnvironment(
```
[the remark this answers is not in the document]
bokalnv: Updated

isaaclab_arena_environments/galileo_g1_locomanip_apple_to_plate_environment.py
```diff
@@ -0,0 +1,115 @@
...
+        # NOTE: Shelf and destination poses are inherited from the brown-box/bin
+        # locomanip variant. The z-heights (0.0707 on the shelf, -0.2641 on the
+        # table) may need tuning for the smaller apple and flatter plate once
+        # validated in the simulator.
+        XY_RANGE_M = 0.025
+        pick_up_object.set_initial_pose(
+            PoseRange(
+                position_xyz_min=(0.5785 - XY_RANGE_M, 0.18 - XY_RANGE_M, 0.0707),
+                position_xyz_max=(0.5785 + XY_RANGE_M, 0.18 + XY_RANGE_M, 0.0707),
+                rpy_min=(math.pi, 0.0, math.pi),
```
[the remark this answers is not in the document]
bokalnv: updated

isaaclab_arena/tests/test_g1_locomanip_apple_to_plate.py
```diff
@@ -0,0 +1,234 @@
...
+            apple_target_pos = plate_pos.clone().unsqueeze(0)
+            apple_target_pos[0, 2] += 0.05
+
+            apple_object.write_root_pose_to_sim(root_pose=torch.cat([apple_target_pos, target_quat], dim=-1))
+            apple_object.write_root_velocity_to_sim(root_velocity=torch.zeros((1, 6), device=env.unwrapped.device))
+
+            terminated_ever = False
+            for _ in range(NUM_STEPS * 10):
+                actions = torch.zeros(env.action_space.shape, device=env.unwrapped.device)
+                actions[:, -4] = 0.75
```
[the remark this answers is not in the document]
bokalnv: updated

alexmillane (approved):

PR #629, opened 2026-04-22
Force isaaclab newton specified mujoco version

qianl-nv:
## Summary
This fixes the isaaclab newton training crashing from arena docker bug https://nvbugspro.nvidia.com/bug/6100149

## Detailed description
isaaclab source/isaaclab_newton/setup.py specified EXTRAS_REQUIRE to ping the mujoco/newton dep versions to "mujoco==3.5.0" / "newton @ git+https://github.com/newton-physics/newton.git@2684d75bfa4bb8b058a93b81c458a74b7701c997"

however, arena docker installs the isaaclab dependencies through isaaclab.sh which only specifies "mujoco>=3.5" in isaaclab source/isaaclab/setup.py

Result: the arena container ends up with mujoco-warp==3.7.0.1 + newton==1.1.0, from recent updates, which crashes with the arena submodule version of isaaclab.

- What has been changed?
Update the arena dockerfile to force submodules/IsaacLab/source/isaaclab_newton[all] specifically.

Fixes the newton training example crash

### DIFF
--- a/docker/Dockerfile.isaaclab_arena
+++ b/docker/Dockerfile.isaaclab_arena
@@ -52,6 +52,15 @@ RUN ${ISAACLAB_PATH}/isaaclab.sh -i
 # Install Isaac Teleop Python APIs (retargeters, device I/O, OpenXR bindings)
 RUN /isaac-sim/python.sh -m pip install isaacteleop~=1.1.0 --extra-index-url https://pypi.nvidia.com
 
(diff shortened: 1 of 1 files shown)


xyao-nv (approved):
I think Newton related deps shall be an optional module, not bake into Arena core by default. But that can come after Arena core pip installable.

PR #632, opened 2026-04-23
Add teleop reset known issue and workaround to doc (#631)

qianl-nv:
## Summary
Current isaacteleop version has buggy reset (squat height is wrong). Won't be able to fix it in 0.2 release (requires isaaclab / isaacteleop version update which is too risky). Add this known issue ans work around instruction to doc.
This is for bug https://nvbugspro.nvidia.com/bug/6103425

## Summary

## Detailed description
- What has been changed?


alexmillane (approved):
Thanks for doing this!

PR #636, opened 2026-04-24
Upgrade CI to 4-core machines

xyao-nv:
## Summary
CI tests related workflows are on Arena-only 4-core machines. Deploy doc is still on 2-core runners.


qianl-nv (approved):

alexmillane (approved):
I'm thinking that we should move the non-GPU-test-running jobs to non-GPU machines. I'll ask where that should be. For now let's merge this as is.

PR #637, opened 2026-04-24
Proposed v3 timeline

alexmillane:
## Summary
Add the proposed v0.3 timeline to the repo

## Detailed description
- This is just a draft.

2026_04_24_v3_propsed_timeline


submodules/IsaacLab
qianl-nv, line 1: is the isaaclab update intentional?

alexmillane: No it's not. Good catch 😅

PR #644, opened 2026-04-28
Test all environments.

alexmillane:
## Summary
Adds a test that starts up all environments to check that there are no errors on start-up.

## Detailed description
- Test (mostly) tries to start environments with the default arguments. (Some have to have their embodiments changed to allow the zero-action policy to run)

### DIFF
--- a/isaaclab_arena/tests/test_all_environments.py
+++ b/isaaclab_arena/tests/test_all_environments.py
@@ -1,0 +1,96 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Smoke-test every registered environment via eval_runner.
+
+Auto-discovers environments from the EnvironmentRegistry and runs each one for
+a few steps with the zero_action policy in a single eval_runner subprocess.
+The test passes if no environment errors out during startup or stepping.
+"""
+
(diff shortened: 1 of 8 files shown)


isaaclab_arena/tests/test_all_environments.py
```diff
@@ -0,0 +1,97 @@
...
+
+    config_path = str(tmp_path / "test_eval_runner_all_environments.json")
+    write_jobs_config_to_file(jobs, config_path)
+    run_eval_runner(config_path, headless=HEADLESS)
+
+
+def test_all_environments_have_default_args():
+    """Every registered env's ``add_cli_args`` parser must accept zero extra args.
+
+    Enforces that envs declare defaults for all their CLI options so that they
```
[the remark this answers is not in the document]
alexmillane: I think we're fine with the code possibly (loudly) breaking in the future.

isaaclab_arena/tests/test_eval_runner.py
```diff
@@ -38,6 +39,9 @@ def run_eval_runner(jobs_config_path: str, headless: bool = HEADLESS):
     args.append(jobs_config_path)
     if headless:
         args.append("--headless")
+    else:
+        args.append("--visualizer")
```
xyao-nv: --viz?

--visualizer is deprecated, viz is the new standard.

xyao-nv (approved):

isaaclab_arena/tests/test_eval_runner.py
alexmillane: good catch.

PR #651, opened 2026-05-01
Add G1 static (no-nav) apple-to-plate env for tabletop manipulation

bokalnv:
## Summary

Adds `galileo_g1_static_pick_and_place`: a same-shelf-only sibling of `galileo_g1_locomanip_pick_and_place` for tabletop manipulation without locomotion. Reuses the `g1_wbc_pink` embodiment and OpenXR retargeter unchanged — WBC balances the standing pose; only the upper body is teleoperated.

- **Env** (`isaaclab_arena_environments/galileo_g1_static_pick_and_place_environment.py`) — apple and destination plate placed on the same shelf in the `galileo_locomanip` background. Object Z values are tuned empirically against the visible shelf top (`z = -0.030`) and account for per-asset USD-origin offsets so apple and plate land flush.
- **Task** (`isaaclab_arena/tasks/static_pick_and_place_task.py`) — `StaticPickAndPlaceTask` subclasses `LocomanipPickAndPlaceTask` and overrides only `get_mimic_env_cfg` to return `StaticPickAndPlaceMimicEnvCfg`. The Mimic cfg collapses the locomanip's 4-phase nav body subtask sequence into a single no-op subtask (no nav term signals fire in the static env, so the locomanip body sequence would deadlock data generation). Datagen name uses a `static_pick_and_place_*` prefix to avoid collisions with locomanip datasets in the converter.
- **Tests** (`isaaclab_arena/tests/test_g1_static_pick_and_place.py`) — mirrors `test_g1_locomanip_apple_to_plate.py` (same `g1_wbc_pink` standing-hold action pattern, same `table` background) plus Mimic-cfg regression guards for the datagen-name prefix and the single-no-op body subtask group.

## Test plan

- [x] `pytest isaaclab_arena/tests/test_g1_static_pick_and_place.py` — 4/4 pass (2 sim, 2 Mimic-cfg)
- [x] Smoke test: production env loads end-to-end with `g1_wbc_pink`, action space is 23-D, 20 standing-hold steps run cleanly
- [x] Visual inspection: apple and plate sit flush on the shelf, robot stands on the floor
- [x] No regression in `test_g1_locomanip_apple_to_plate.py` Mimic-cfg tests

PR #656, opened 2026-05-06
Delete LocomanipPickAndPlaceTask, add mimic_env_cfg_factory

cvolkcvolk:
## Summary

Deletes `LocomanipPickAndPlaceTask`. Adds a `mimic_env_cfg_factory` kwarg to `PickAndPlaceTask` so the locomanip env can inject `G1PickAndPlaceMimicEnvCfg` directly without a task subclass.

## Motivation

`LocomanipPickAndPlaceTask` was a configuration-disguised-as-subclass: its only behavior was to override `get_mimic_env_cfg` so the locomanip env got a different cfg class than the parent's `PickPlaceMimicEnvCfg`. Same `__init__`, same termination, same metrics, same scene events, just a different cfg class wired in.

That pattern was about to grow another floor:
PR #651 (`bokal/g1-static-pick-and-place`) is introducing `StaticPickAndPlaceTask(LocomanipPickAndPlaceTask)`. Under this refactor, that class becomes unnecessary: the static env can construct `PickAndPlaceTask` directly with its own `mimic_env_cfg_factory` closure that builds its preferred cfg. 

## How the new construction works

The locomanip env now uses a small factory closure:

```python
def _build_g1_pick_and_place_mimic_cfg(arm_mode):
    return G1PickAndPlaceMimicEnvCfg(
        pick_up_object_name=pick_up_object.name,
        destination_name=destination.name,
        arm_mode=arm_mode,
    )

task = PickAndPlaceTask(
    pick_up_object, destination, background,
    ...,
    mimic_env_cfg_factory=_build_g1_pick_and_place_mimic_cfg,
)
```

The factory receives `arm_mode` from the env builder and returns a constructed cfg. When set, the factory short-circuits the default `PickPlaceMimicEnvCfg` construction in `PickAndPlaceTask.get_mimic_env_cfg`. When not set (every other existing call site), behavior is unchanged.

`mimic_env_cfg_factory` is a first-class customization point, not a stopgap. It exists for tasks whose Mimic cfg shape genuinely diverges from `PickPlaceMimicEnvCfg` (different arm subtask sequences, different per-subtask numerical knobs, bespoke fields). The G1 cfg is one example; other tasks with bespoke cfg shapes can use the same mechanism.

## Naming

The cfg class is renamed from `LocomanipPickAndPlaceMimicEnvCfg` to `G1PickAndPlaceMimicEnvCfg`. The old name implied "this is for locomanipulation tasks" but the cfg is really "this is the G1 shape for pick-and-place" (3 subtasks per arm with idle settle, 4-phase nav body channel, G1-specific numerical tunings). The new name names what it is.

The cfg now lives in `isaaclab_arena/tasks/pick_and_place_task.py` next to the parent `PickPlaceMimicEnvCfg`, matching the repo convention that every file in `isaaclab_arena/tasks/` holds a task plus its related cfgs.

## Files changed

- `isaaclab_arena/tasks/pick_and_place_task.py`: new `mimic_env_cfg_factory` kwarg on `PickAndPlaceTask` with class docstring; `G1PickAndPlaceMimicEnvCfg` moved here from its standalone file.
- `isaaclab_arena/tasks/locomanip_pick_and_place_task.py`: deleted. The `LocomanipPickAndPlaceTask` class is gone; the cfg moved to `pick_and_place_task.py` (renamed to `G1PickAndPlaceMimicEnvCfg`) and gained an `arm_mode` field whose `__post_init__` re-asserts `DUAL_ARM`, preserving the guard that lived on the deleted task class.
- `isaaclab_arena_environments/galileo_g1_locomanip_pick_and_place_environment.py`: migrates to `PickAndPlaceTask(mimic_env_cfg_factory=...)`.
- Tests in `test_g1_locomanip_apple_to_plate.py` and `test_locomanip_object_on_termination.py` migrated to the new pattern with no assertion changes.
- `docs/pages/example_workflows/example_environments.rst` and `docs/pages/example_workflows/locomanipulation/step_1_environment_setup.rst` updated to reflect the new construction and naming.

## Test plan

- [x] `pytest -m 'not with_cameras and not with_subprocess' isaaclab_arena/tests/test_g1_locomanip_apple_to_plate.py isaaclab_arena/tests/test_locomanip_object_on_termination.py` passes (4 tests).
- [x] Smoke: `policy_runner.py --policy_type zero_action --num_steps 5 galileo_g1_locomanip_pick_and_place --embodiment g1_wbc_joint` runs clean.
- [x] Pre-commit hooks (black, flake8, isort, pyupgrade, codespell) pass on every commit.
- [ ] CI's `test_all_environments` smoke (please verify in CI).

Note: `test_apple_on_plate_succeeds` is a pre-existing flake on `main` from the older "re-teleport every step" pattern. Manifests on pristine `main` too; unrelated to this branch.


isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -36,6 +37,8 @@ def __init__(
         task_description: str | None = None,
         force_threshold: float = 1.0,
         velocity_threshold: float = 0.1,
+        mimic_subtask_boundaries: dict[str, tuple[str, ...]] | None = None,
```
peterd-NV: Question about this interface: Is the plan to use `mimic_subtask_boundaries` and `mimic_env_cfg_factory` as a common interface across all tasks but for now only the PickAndPlace task needs to have them?

cvolkcvolk: Yes, I kept it on PickAndPlaceTask only for this PR because lifting it to TaskBase requires updating every task's MimicEnvCfg to consume it. 
Similar for `mimic_env_cfg_factory`. PickAndPlaceTask's current dual-cfg-shape problem (PickPlaceMimicEnvCfg vs LocomanipPickAndPlaceMimicEnvCfg) requires it. Other tasks each have one MimicEnvCfg shape and don't have the problem _yet_. If OpenDoorTask or another task grows a similar dual shape, I'd add the factory there too. 

I didn't go for the full update as a) we need the refactor for the upcoming RA workflow and b) Mimic is refactored out fully anyway in the near future. 
Open to lifting either now if you'd prefer though.

cvolkcvolk: Okay I just refactored it again: The only interface is now the `mimic_env_cfg_factory` to pass the config. Looks much cleaner now.

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -229,3 +255,291 @@ def __post_init__(self):
...
+
+    Encodes G1-specific cfg shape: 3 subtasks per arm
+    (idle / grasp_and_idle / final), a 4-phase nav body subtask sequence
+    (navigate_to_table / navigate_turn_inplace / navigate_to_bin / final), and
+    G1-specific per-subtask numerical tunings. Used by the locomanip env via
+    ``PickAndPlaceTask(mimic_env_cfg_factory=...)``.
+    """
+
+    pick_up_object_name: str = MISSING
+    destination_name: str = MISSING
```
peterd-NV: Very minor nit: Since we are moving this config into the pickplace task file, could consider unifying the name with the `PickPlaceMimicEnvCfg` and set this to `destination_location_name`

peterd-NV (approved):
Thanks a lot for this refactor. It's a great improvement from before. Regarding upstreaming the API change to TaskBase, I don't think that is needed since the upcoming DataGen-next arch will eliminate the need for Mimic configs in Arena. In the meantime, this is clean refactor for our practical needs right now.

isaaclab_arena/tasks/pick_and_place_task.py
cvolkcvolk: Done

PR #658, opened 2026-05-07
Pluggable action schedulers for the GR00T remote policy

xyao-nv:
## Summary
Pluggable action schedulers for the GR00T remote policy

## Detailed description
- Why: Let the remote GR00T policy choose between per-env chunk fetching and a synchronized full-batch mode for higher GR00T-server utilization, and lift the fetch/play logic out of the policy class.
- What: New `isaaclab_arena/policy/action_scheduling/` package with `ActionScheduler` (abstract), `ActionChunkScheduler` (per-env), and `SyncedBatchActionScheduler` (waits for all envs, holds waiting envs at their current state); pure-torch tests; `Gr00tRemoteClosedloopPolicy` rewired to take a pluggable scheduler with a new `--scheduler chunk|synced_batch` CLI flag and a _extract_hold_action helper func.
- Impact: Default behavior unchanged (chunked); opt-in --scheduler synced_batch for batched inference; no new deps.

### DIFF
--- a/isaaclab_arena/policy/action_chunking.py
+++ b/isaaclab_arena/policy/action_chunking.py
@@ -11,7 +11,7 @@ import torch
 from collections.abc import Callable
 
 
-# TODO(xinjieyao, 2026-05-06): to be recycled
+# TODO(xinjieyao, 2026-05-06): Delete once all callers have migrated to action_scheduling.ActionChunkScheduler
 class ActionChunkingState:
     """Holds chunk buffer, per-env index, and refill flag; provides get_action(fetch_chunk_fn).
 
--- a/isaaclab_arena/policy/action_chunking.py
+++ b/isaaclab_arena/policy/action_chunking.py
@@ -11,6 +11,7 @@ import torch
 from collections.abc import Callable
 
 
+# TODO(xinjieyao, 2026-05-06): to be recycled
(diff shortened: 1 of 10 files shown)


isaaclab_arena/policy/action_chunking.py
```diff
@@ -11,6 +11,7 @@
 from collections.abc import Callable
 
 
+# TODO(xinjieyao, 2026-05-06): to be recycled
```
cvolkcvolk: Nit: Maybe a more declarative comment like "TODO(xinjieyao): Delete once all callers have migrated to action_scheduling.ActionChunkScheduler"

isaaclab_arena/policy/action_chunking_client.py
```diff
@@ -21,6 +21,7 @@
 from isaaclab_arena.remote_policy.remote_policy_config import RemotePolicyConfig
 
 
+# TODO(xinjieyao, 2026-05-06): to be recycled
```
cvolkcvolk: Same

cvolkcvolk: If someone takes over your work that person knows what to "recycle" ;)

cvolkcvolk (approved):
lgtm

PR #663, opened 2026-05-08
Add isaaclab_arena_openpi package for pi0 droid eval

cvolkcvolk:
## Summary

Adds `isaaclab_arena_openpi/`, a new top-level package that lets arena evaluate
the upstream openpi pi0/pi05 droid policies. Ships the current
openpi server as a dockerized container fixed to a pinned commit.

## What's in the package

- `policy/pi0_remote_policy.py` — `Pi0RemotePolicy` (`PolicyBase` subclass).
  Embodiment-agnostic openpi WebSocket client. Single-env only.
- `policy/droid_adapter.py` — `Pi0DroidAdapter` (subclass of
  `Pi0EmbodimentAdapter`) and the `DroidObservation` dataclass. Holds
  DROID-specific wire format: 8-dim action, image / proprio keys, variant→horizon table.
- `policy/pi0_remote_config.py` — `Pi0RemotePolicyArgs` dataclass +
  `MAX_RECONNECT_ATTEMPTS`.
- `docker/build_openpi_server.sh` — clones upstream
  `Physical-Intelligence/openpi` at the pinned commit (from `docker/OPENPI_COMMIT`), builds the server image with upstream's own `serve_policy.Dockerfile`. Supports
  `--src-dir=<path>` to skip the clone and build from a local checkout.

## Docs

New Sphinx page under **Running a Real Policy → OpenPi** at
`docs/pages/quickstart/first_experiments/running_a_real_policy/openpi.rst`. The
former single-page `running_a_real_policy.rst` is now a dropdown with `gr00t.rst`
and `openpi.rst` as children.

## Usage

```bash
./isaaclab_arena_openpi/docker/build_openpi_server.sh    # one-time, ~3 min, ~19 GB

# Terminal 1: start server (pi05 default)
docker run --rm -it --gpus all --network=host \
  -e XLA_PYTHON_CLIENT_MEM_FRACTION=0.5 \
  isaaclab_arena_openpi-server:latest \
  uv run scripts/serve_policy.py policy:checkpoint \
    --policy.config=pi05_droid_jointpos_polaris \
    --policy.dir=gs://openpi-assets-simeval/pi05_droid_jointpos

# Terminal 2: arena client (inside the arena docker container)
python isaaclab_arena/evaluation/policy_runner.py \
  --viz kit \
  --policy_type isaaclab_arena_openpi.policy.pi0_remote_policy.Pi0RemotePolicy \
  --num_steps 500 \
  --enable_cameras --num_envs 1 \
  --language_instruction "Pick up the Rubik's cube and place it in the bowl." \
  pick_and_place_maple_table \
    --embodiment droid_abs_joint_pos \
    --pick_up_object rubiks_cube_hot3d_robolab \
    --destination_location bowl_ycb_robolab \
    --hdr home_office_robolab
```


docker/Dockerfile.isaaclab_arena
```diff
@@ -120,6 +120,11 @@ RUN /isaac-sim/python.sh -m pip install msgpack==1.1.0 msgpack-numpy==0.4.8 pyzm
     /isaac-sim/python.sh -m pip install --no-deps --ignore-requires-python -e ${WORKDIR}/submodules/Isaac-GR00T/
 RUN /isaac-sim/python.sh -m pip install av
 
+# openpi remote-client. Pin OPENPI_COMMIT to the commit the server runs at.
+ARG OPENPI_COMMIT=c23745b5ad24e98f66967ea795a07b2588ed6c79
```
cvolkcvolk: For now I put the openpi dependency on the client into the arena base docker. We could in the future layer this instead ontop as a new container

isaaclab_arena_openpi/docker/build_openpi_server.sh
```diff
@@ -0,0 +1,63 @@
...
+#   ./build_openpi_server.sh <commit>               # build at a specific commit
+#   ./build_openpi_server.sh --push                 # build and push to NGC
+#   ./build_openpi_server.sh <commit> --push
+
+set -euo pipefail
+
+IMAGE_NAME="${IMAGE_NAME:-isaaclab_arena_openpi-server}"
+NGC_PATH="${NGC_PATH:-nvcr.io/nvstaging/isaac-amr/${IMAGE_NAME}}"
+OPENPI_REPO="https://github.com/Physical-Intelligence/openpi"
+DEFAULT_COMMIT="c23745b5ad24e98f66967ea795a07b2588ed6c79"
```
xyao-nv: Can you unify this spec defining the src of openpi with Dockerfile? I wonder if having a shared .env could work.

isaaclab_arena_openpi/README.md
```diff
@@ -0,0 +1,99 @@
+# isaaclab_arena_openpi
```
xyao-nv: Can it be doc rst instead of readme?

isaaclab_arena_openpi/docker/build_openpi_server.sh
```diff
@@ -0,0 +1,63 @@
...
+# Usage:
+#   ./build_openpi_server.sh                        # build at DEFAULT_COMMIT
+#   ./build_openpi_server.sh <commit>               # build at a specific commit
+#   ./build_openpi_server.sh --push                 # build and push to NGC
+#   ./build_openpi_server.sh <commit> --push
+
+set -euo pipefail
+
+IMAGE_NAME="${IMAGE_NAME:-isaaclab_arena_openpi-server}"
+NGC_PATH="${NGC_PATH:-nvcr.io/nvstaging/isaac-amr/${IMAGE_NAME}}"
```
xyao-nv: External users may not have write access to this path

isaaclab_arena_openpi/docker/build_openpi_server.sh
```diff
@@ -0,0 +1,63 @@
...
+    esac
+done
+
+COMMIT="${COMMIT:-$DEFAULT_COMMIT}"
+
+TMPDIR=$(mktemp -d)
+trap 'rm -rf "$TMPDIR"' EXIT
+
+echo "Cloning openpi at ${COMMIT} ..."
+GIT_LFS_SKIP_SMUDGE=1 git clone --quiet "$OPENPI_REPO" "$TMPDIR/openpi"
```
xyao-nv: I wonder if the clone could be optional as users may already have a PI src repo and just want to build server docker (if Pi did not ship it).

isaaclab_arena_openpi/docker/build_openpi_server.sh
```diff
@@ -0,0 +1,63 @@
...
+    .
+
+echo "Built ${IMAGE_NAME}:${SHORT_HASH} (also tagged :latest)"
+
+if [ "$PUSH" = true ]; then
+    echo "Pushing to ${NGC_PATH}:${SHORT_HASH}"
+    docker tag "${IMAGE_NAME}:${SHORT_HASH}" "${NGC_PATH}:${SHORT_HASH}"
+    docker push "${NGC_PATH}:${SHORT_HASH}"
+    echo "Pushed ${NGC_PATH}:${SHORT_HASH}"
+fi
```
xyao-nv: If we imagine every policy builds their own docker, then can we make the docker build & push more universal? 
Not bind to Pi server docker.

I personally think it's a valid assumption.

isaaclab_arena_openpi/policy/pi0_remote_policy.py
```diff
@@ -0,0 +1,182 @@
...
+        chunk_exhausted = self._cached_action_chunk is None or self._next_chunk_step >= self.open_loop_horizon
+        if chunk_exhausted:
+            self._cached_action_chunk = self._fetch_action_chunk(observation)
+            self._next_chunk_step = 0
+
+        next_action_np = self._cached_action_chunk[self._next_chunk_step]
+        self._next_chunk_step += 1
+        return torch.from_numpy(next_action_np).to(dtype=torch.float32, device=self.device).unsqueeze(0)
+
+    def reset(self, env_ids: torch.Tensor | None = None) -> None:
```
xyao-nv, line 166: Can you add a TODO for env_id reset when parallel is supported?

isaaclab_arena_openpi/policy/pi0_remote_policy.py
```diff
@@ -0,0 +1,182 @@
...
+        return Pi0DroidRemotePolicy(
+            Pi0DroidRemotePolicyArgs(
+                policy_variant=args.policy_variant,
+                policy_device=args.policy_device,
+                remote_host=args.remote_host,
+                remote_port=args.remote_port,
+            )
+        )
+
+    def get_action(self, env: gym.Env, observation: dict[str, Any]) -> torch.Tensor:
```
xyao-nv, line 145: Can you add a TODO for extending to parallel envs?

isaaclab_arena_openpi/policy/pi0_droid_remote_policy.py
```diff
@@ -0,0 +1,182 @@
...
+        next_action_np = self._cached_action_chunk[self._next_chunk_step]
+        self._next_chunk_step += 1
+        return torch.from_numpy(next_action_np).to(dtype=torch.float32, device=self.device).unsqueeze(0)
+
+    def reset(self, env_ids: torch.Tensor | None = None) -> None:
+        self._cached_action_chunk = None
+        self._next_chunk_step = 0
+
+    def _fetch_action_chunk(self, observation: dict[str, Any]) -> np.ndarray:
+        droid_obs = self._extract_droid_observation(observation)
```
xyao-nv: Does it have to droid dependent? Can you extend to other embodiments?
Can the dict key be a dataclass/cfg?

isaaclab_arena_openpi/policy/pi0_droid_remote_policy.py
```diff
@@ -0,0 +1,182 @@
...
+            action_chunk.ndim == 2 and action_chunk.shape[1] == ACTION_DIM
+        ), f"Expected actions of shape (H, {ACTION_DIM}); got {action_chunk.shape}"
+        assert (
+            action_chunk.shape[0] >= self.open_loop_horizon
+        ), f"Server returned horizon {action_chunk.shape[0]} < configured open_loop_horizon {self.open_loop_horizon}"
+        return action_chunk[: self.open_loop_horizon].astype(np.float32, copy=True)
+
+    def _extract_droid_observation(self, observation: dict[str, Any]) -> dict[str, np.ndarray]:
+        cam = observation["camera_obs"]
+        proprio = observation["policy"]
```
xyao-nv: We kinda enforce "camera_obs" as the key when inserting camera to embodiment. But how about "policy"? Is it fixed?

isaaclab_arena_openpi/policy/pi0_droid_remote_policy.py
```diff
@@ -0,0 +1,182 @@
...
+    ARENA_WRIST_CAMERA_KEY,
+    DEFAULT_VARIANT,
+    MAX_RECONNECT_ATTEMPTS,
+    OPEN_LOOP_HORIZON_BY_VARIANT,
+    TARGET_IMAGE_SIZE,
+    Pi0DroidRemotePolicyArgs,
+)
+
+
+class Pi0DroidRemotePolicy(PolicyBase):
```
xyao-nv: I don't like the idea of limiting the class to only robot specific closedloop policy.
It's okay that you have only tested with Droid, but when building the policy class, it shall be embodiment agnostic, i.e. `Pi0RemotePolicy`.
You can provide embodiment specific cfg as the data member.

xyao-nv (changes requested):
Sorry I have to request changes -- can policy class be embodiment agostic?

isaaclab_arena_openpi/policy/pi0_remote_policy.py
```diff
@@ -0,0 +1,182 @@
...
+    def from_args(args: argparse.Namespace) -> Pi0DroidRemotePolicy:
+        return Pi0DroidRemotePolicy(
+            Pi0DroidRemotePolicyArgs(
+                policy_variant=args.policy_variant,
+                policy_device=args.policy_device,
+                remote_host=args.remote_host,
+                remote_port=args.remote_port,
+            )
+        )
+
```
xyao-nv, line 131: When closing the connection (esp it's used in eval_runner), what's the expected components needed? [Gr00t side](https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena_gr00t/policy/gr00t_remote_closedloop_policy.py#L258) I have added closing socket and cleaning context to prevent stalling memory.

isaaclab_arena_openpi/policy/pi0_remote_policy.py
```diff
@@ -0,0 +1,182 @@
...
+            ),
+            "observation/wrist_image_left": image_tools.resize_with_pad(
+                droid_obs["wrist_image"], target_height, target_width
+            ),
+            "observation/joint_position": droid_obs["joint_position"],
+            "observation/gripper_position": droid_obs["gripper_position"],
+            "prompt": language_instruction,
+        }
+
+    def _call_server_with_retry(self, server_request: dict[str, Any]) -> dict[str, Any]:
```
xyao-nv, line 195: I wonder if this func could be in the parent class `RemotePolicy`. Retry with max attempts will always be needed.

isaaclab_arena_openpi/policy/pi0_droid_remote_policy.py
```diff
@@ -0,0 +1,182 @@
...
+    """openpi remote closed-loop policy for DROID, single-env only."""
+
+    name = "pi0_droid_remote"
+    config_class = Pi0DroidRemotePolicyArgs
+
+    def __init__(self, config: Pi0DroidRemotePolicyArgs) -> None:
```
xyao-nv: Do you need `action_scheduler_cls`? This MR seems to only contain obs -> action_chunk. Can you add a TODO for `action_chunk` to `action`

isaaclab_arena_openpi/policy/pi0_droid_config.py
```diff
@@ -0,0 +1,33 @@
...
+# How many actions to replay before refetching a new chunk from the server.
+OPEN_LOOP_HORIZON_BY_VARIANT: dict[str, int] = {
+    "pi05": 15,
+    "pi0": 10,
+    "pi0_fast": 10,
+}
+DEFAULT_VARIANT = "pi05"
+
+ARENA_EXTERNAL_CAMERA_KEY = "external_camera_rgb"
+ARENA_WRIST_CAMERA_KEY = "wrist_camera_rgb"
```
xyao-nv: Shall we do a closedloop-embodiment dataclass? Hydra? 
So we can unify how cfg are supposed to set, you have py, and g00t has yaml.

isaaclab_arena_openpi/policy/pi0_remote_policy.py
cvolkcvolk: We don't have a RemotePolicy class _yet_. But I think it would make sense though to maybe introduce another layer `PolicyBase` -> `RemotePolicy` -> `PiRemote` / `Gr00tRemote` / ...

isaaclab_arena_openpi/policy/pi0_droid_remote_policy.py
cvolkcvolk: Good feedback, that was actually planned for a follow up. I added an `EmbodimentAdapter` now that encapsulates the Embodiment specific dimensions.

isaaclab_arena_openpi/policy/pi0_remote_policy.py
cvolkcvolk: Done

isaaclab_arena_openpi/policy/pi0_remote_policy.py
cvolkcvolk: Done

isaaclab_arena_openpi/policy/pi0_droid_remote_policy.py
cvolkcvolk: `Pi0RemotePolicy` is now embodiment-agnostic. Moved the droid-specific logic lives in `Pi0DroidAdapter`

isaaclab_arena_openpi/policy/pi0_remote_policy.py
cvolkcvolk: Done

isaaclab_arena_openpi/policy/pi0_remote_policy.py
cvolkcvolk: Thanks. Added a `close()` teardown function. 
We could align this with the `Gr00tRemotePolicy` by introducing another abstaction `RemotePolicy` that sits ontop of `PolicyBase`. That's a follow up.

isaaclab_arena_openpi/policy/pi0_droid_remote_policy.py
cvolkcvolk: Yes fixed by us, in the respective ObservationCfgs

cvolkcvolk: I made them variables and provided a comment to make it a bit clearer

isaaclab_arena_openpi/docker/build_openpi_server.sh
cvolkcvolk: Added a `.OPENPI_COMMIT` file to share among both places

isaaclab_arena_openpi/docker/build_openpi_server.sh
```diff
@@ -0,0 +1,63 @@
...
+
+COMMIT="${COMMIT:-$DEFAULT_COMMIT}"
+
+TMPDIR=$(mktemp -d)
+trap 'rm -rf "$TMPDIR"' EXIT
+
+echo "Cloning openpi at ${COMMIT} ..."
+GIT_LFS_SKIP_SMUDGE=1 git clone --quiet "$OPENPI_REPO" "$TMPDIR/openpi"
+cd "$TMPDIR/openpi"
+git checkout "$COMMIT"
```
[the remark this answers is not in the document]
cvolkcvolk: Done

isaaclab_arena_openpi/docker/build_openpi_server.sh
cvolkcvolk: Added an optional `--src-dir`

isaaclab_arena_openpi/docker/build_openpi_server.sh
cvolkcvolk: NGC_PATH documented and able to be overwritten now. Changing the default registry feels like an arena-wide thing (would also touch `push_to_ngc.sh`) so I'd leave it consistent with the rest of arena here. Happy to revisit in a separate PR if you'd like a different default across the project.

isaaclab_arena_openpi/policy/pi0_droid_remote_policy.py
cvolkcvolk: Done

isaaclab_arena_openpi/docker/build_openpi_server.sh
cvolkcvolk: Agree. I would prefer to this in a follow up to keep it seperated.
For now, I delete the openpi push functionality from this PR as its not used yet.

isaaclab_arena_openpi/README.md
cvolkcvolk: Written as `.rst` it should then move to `/docs/pages/<>..`

cvolkcvolk: Moved it to our docs pages next to Gr00t

isaaclab_arena_openpi/policy/pi0_droid_config.py
cvolkcvolk: Good idea. Agreed there's a inconsistency that gr00t uses YAML configs and openpi uses a Python dataclass. 
I feel like unifying is a broader change, so leaving this out of scope for this PR. Left a TODO though.

docker/Dockerfile.isaaclab_arena
```diff
@@ -120,6 +120,15 @@ RUN /isaac-sim/python.sh -m pip install msgpack==1.1.0 msgpack-numpy==0.4.8 pyzm
     /isaac-sim/python.sh -m pip install --no-deps --ignore-requires-python -e ${WORKDIR}/submodules/Isaac-GR00T/
 RUN /isaac-sim/python.sh -m pip install av
 
+# openpi remote-client. The pinned commit lives in
+# isaaclab_arena_openpi/docker/OPENPI_COMMIT and is shared with
+# isaaclab_arena_openpi/docker/build_openpi_server.sh
```
alexmillane: Do you mean that you use this file, `OPENPI_COMMIT` to ensure that the Pi docker and Arena docker get the same hash of the server/client combo?

docs/pages/quickstart/first_experiments/running_a_real_policy/index.rst
```diff
@@ -0,0 +1,12 @@
...
+
+The zero-action experiments keep the robot still and success rates at zero. In this
+section we will see actual pre-trained models in action. Arena ships clients for two
+foundation-model policy servers — pick whichever you want to try.
+
+.. toctree::
+   :maxdepth: 1
+
+   gr00t
+   openpi
```
alexmillane: Am I correct in assuming that there is differential success between these two? Should we maybe note the approximate success rates on bracket?

Am I correct in assuming the gr00t gets basically zero while pi gets something non-zero? I feel that a user experience might be better if they end up running the policy that gets more than zero...

docs/pages/quickstart/first_experiments/running_a_real_policy/openpi.rst
```diff
@@ -0,0 +1,116 @@
...
+~11 GB checkpoint into the container layer.
+
+.. code-block:: bash
+
+   docker run --rm -it --gpus all --network=host \
+     -e XLA_PYTHON_CLIENT_MEM_FRACTION=0.5 \
+     isaaclab_arena_openpi-server:latest \
+     uv run scripts/serve_policy.py policy:checkpoint \
+       --policy.config=pi05_droid_jointpos_polaris \
+       --policy.dir=gs://openpi-assets-simeval/pi05_droid_jointpos
```
alexmillane: What do you think about running this as we do elsewhere: having a single script that builds and runs the docker. Such a script will be much similar to, but much simpler than our `run_docker.sh` because it needn't worry about the mounts etc.

alexmillane (commented):
Partial review. Looking good so far!

xyao-nv (approved):

isaaclab_arena_openpi/policy/pi0_remote_policy.py
```diff
@@ -0,0 +1,255 @@
...
+        """Pull env tensors out of the arena gym observation dict.
+
+        Concrete return type is adapter-defined (typically a frozen dataclass);
+        the policy treats it as an opaque value to round-trip through
+        :meth:`pack_request`.
+        """
+
+    @abstractmethod
+    def pack_request(self, extracted: Any, language_instruction: str) -> dict[str, Any]:
+        """Build the wire-format request payload openpi server expects."""
```
alexmillane, line 43: Is this something that we want across all policies?

How is this conversion done in the gr00t model?

I'm wondering if this interface should be shared between models.

docker/Dockerfile.isaaclab_arena
cvolkcvolk: Exactly, extended the comment

docs/pages/quickstart/first_experiments/running_a_real_policy/index.rst
cvolkcvolk: Yes quite a bit. Reverted the order to first run into the openpi workflow. And added a heads up about the success rate.

docs/pages/quickstart/first_experiments/running_a_real_policy/openpi.rst
cvolkcvolk: Good Idea! Merged into one step.

docs/pages/quickstart/first_experiments/running_a_real_policy/index.rst
cvolkcvolk: Actually running for num_steps doesn't make much sense here regarding the success rate... I will provide an additional json workflow file to run with the eval_runner as we do in the gr00t section

isaaclab_arena_openpi/policy/pi0_remote_policy.py
```diff
@@ -0,0 +1,255 @@
...
+    Action handling today is straight chunk replay: the policy fetches one
+    ``(open_loop_horizon, action_dim)`` chunk from the server and yields
+    rows in order. A future action scheduler (interpolation, smoothing,
+    chunk-overlap blending) belongs as a pluggable component.
+    """
+
+    # TODO(cvolk, 2026-05-18): add an action_scheduler_cls so action_chunk
+    # -> action is configurable (today it is a row-by-row replay).
+
+    # TODO(cvolk, 2026-05-18): Add a RemotePolicy base class.
```
alexmillane, line 58: Yea I'm wondering if there's any structure we should be sharing between policies. (Such an improvement can, naturally, be done in a follow up MR).

isaaclab_arena_openpi/policy/pi0_remote_policy.py
```diff
@@ -0,0 +1,255 @@
...
+
+    name = "pi0_remote"
+    config_class = Pi0RemotePolicyArgs
+
+    def __init__(self, config: Pi0RemotePolicyArgs, openpi_embodiment_adapter: Pi0EmbodimentAdapter) -> None:
+        super().__init__(config)
```
alexmillane, line 64: Given the base class doesn't know about this config type, I guess it just stores it?

isaaclab_arena_openpi/policy/pi0_remote_policy.py
```diff
@@ -0,0 +1,255 @@
...
+        assert config.policy_variant in openpi_embodiment_adapter.open_loop_horizon_by_variant, (
+            f"Unknown policy_variant {config.policy_variant!r} for adapter"
+            f" {type(openpi_embodiment_adapter).__name__};"
+            f" known: {sorted(openpi_embodiment_adapter.open_loop_horizon_by_variant)}"
+        )
+        self._openpi_embodiment_adapter = openpi_embodiment_adapter
+        self.open_loop_horizon = openpi_embodiment_adapter.open_loop_horizon_by_variant[config.policy_variant]
```
alexmillane: I feel that across the project we have not been consistent with leading `_`.

I'm a bit unsure about what to do at this point.

For example `self._openpi_embodiment_adapter` indicates that an user shouldn't touch this variable, however, I feel like we also don't expect `self.open_loop_horizon` to be touched I assume. So IMO the use of leading underscores is inconsistent, even in this class.

I guess we should make a policy on this and try to work towards adhering to it. I think we'd need to look through the code to see what such a policy should actually be.

isaaclab_arena_openpi/policy/pi0_remote_policy.py
```diff
@@ -0,0 +1,255 @@
...
+        # TODO(cvolk, 2026-05-18): extend to parallel envs once the openpi
+        # server supports batched observations; today it accepts one obs
+        # per request.
+        assert (
+            env.unwrapped.num_envs == 1
+        ), f"Pi0RemotePolicy only supports num_envs=1 (openpi server limitation), got num_envs={env.unwrapped.num_envs}"
+        assert self.task_description, (
+            "Pi0RemotePolicy requires a non-empty language instruction"
+            " (set via --language_instruction or on the task definition)."
+        )
```
alexmillane, line 155: Uff. This is a bit of a limitation.

What do we plan to do about this?

At a minimum, I feel we should loop over the envs. We totally expect to almost-always be evaluating on parallel envs.

Perhaps we should just add the loop on a follow up MR. A longer term solution might be adding batching to the Pi server... But we could wait until there's a clear performance reason for doing so?

isaaclab_arena_openpi/policy/pi0_remote_policy.py
```diff
@@ -0,0 +1,255 @@
...
+                    f" (attempt {attempt_index + 1}/{MAX_RECONNECT_ATTEMPTS - 1}) ..."
+                )
+                # Best-effort cleanup of the old client before replacing it.
+                try:
+                    _old = self._websocket_client
+                    ws = getattr(_old, "_ws", None)
+                    if ws is not None:
+                        ws.close()
+                except (websockets.exceptions.ConnectionClosed, OSError):
+                    pass
```
alexmillane: nit: It's a bit hard to read this function because of the nested `try`s. Suggestion to just move this to a (pure) function

alexmillane (approved):
Looks great! I left a few nits, which can be defered.

isaaclab_arena_openpi/policy/pi0_remote_policy.py
cvolkcvolk: Yeah this class was inconsistent, Im applying underscores in this class. 

For the rest of the repo I would say we align to the `self._underscore` convention, [as pep8 points out here] (https://peps.python.org/pep-0008/#naming-conventions:~:text=_single_leading_underscore%3A%20weak%20%E2%80%9Cinternal%20use%E2%80%9D%20indicator.%20E.g.%20from%20M%20import%20*%20does%20not%20import%20objects%20whose%20names%20start%20with%20an%20underscore.)

isaaclab_arena_openpi/policy/pi0_remote_policy.py
cvolkcvolk: Yes, I have a TODO to extend this to a batched call from our side, via simply looping. Down the road Ill update the OpenPi repo to accept batched observations.

isaaclab_arena_openpi/policy/pi0_remote_policy.py
cvolkcvolk: yep basically just sets `self.config = config`

isaaclab_arena_openpi/policy/pi0_remote_policy.py
cvolkcvolk: Agree. Moved out to a function

isaaclab_arena_openpi/policy/pi0_remote_policy.py
xyao-nv: Yeah I think its the way to go -- on my plate

PR #664, opened 2026-05-11
Update IL GR00T training docs

xyao-nv:
## Summary
Update IL GR00T training docs

## Detailed description
- Reason: Clarify GR00T finetuning should use native GR00T uv deps outside Arena Base Docker.
- Changed: 
1. Policy training docs now link to the GR00T installation guide and show only finetuning launch commands.
2. Evaluation docs now consistently use Arena Base Docker while GR00T servers run from the GR00T uv environment.
- Impact: Users get a cleaner split between Arena simulation/evaluation and GR00T finetuning/server workflows.


cvolkcvolk (approved):
Thanks!

PR #676, opened 2026-05-13
Heterogenous Object Placement

zhx06:
## Summary
Add heterogeneous object placement support

## Detailed description
- Add per-environment variant assignment and bounding boxes for `RigidObjectSet`.
- Update `ObjectPlacer`, `RelationSolver`, and `PooledObjectPlacer` to solve, validate, pool, and reset with env-specific bounding boxes.
- Add `--mode heterogeneous` to the GR1 table example with stable Robolab object sets.
- Make the office table a kinematic background asset to prevent object/table bouncing.
- Add tests for heterogeneous placement, pooled reset behavior, variant assignment, and reset atomicity.

### DIFF
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -37,6 +37,8 @@ __all__ = [
 class ObjectBase(Asset, ABC):
     """Parent class for (spawnable) Object and ObjectReference."""
 
+    has_env_specific_bboxes: bool = False
+
     def __init__(
         self,
         name: str,
@@ -54,8 +56,6 @@ class ObjectBase(Asset, ABC):
         self.object_cfg = None
         self.event_cfg = None
         self.relations: list[RelationBase] = []
-        if not hasattr(self, "has_env_specific_bboxes"):
-            self.has_env_specific_bboxes: bool = False
 
     def get_initial_pose(self) -> Pose | PoseRange | PosePerEnv | None:
         """Return the current initial pose of this object.
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -54,7 +54,8 @@ class ObjectBase(Asset, ABC):
         self.object_cfg = None
         self.event_cfg = None
         self.relations: list[RelationBase] = []
-        self.has_env_specific_bboxes: bool = False
+        if not hasattr(self, "has_env_specific_bboxes"):
+            self.has_env_specific_bboxes: bool = False
 
     def get_initial_pose(self) -> Pose | PoseRange | PosePerEnv | None:
         """Return the current initial pose of this object.
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -37,8 +37,6 @@ __all__ = [
 class ObjectBase(Asset, ABC):
     """Parent class for (spawnable) Object and ObjectReference."""
 
-    has_env_specific_bboxes: bool = False
-
     def __init__(
         self,
         name: str,
@@ -74,27 +72,6 @@ class ObjectBase(Asset, ABC):
     def get_world_bounding_box(self) -> AxisAlignedBoundingBox:
         """Get bounding box in world coordinates (local bbox rotated and translated)."""
         ...
-
-    def get_bounding_box_per_env(self, num_envs: int) -> AxisAlignedBoundingBox:
-        """Get local bounding boxes for each environment.
-
-        This default implementation is for objects with the same geometry in
-        every environment: it expands the single local bbox to ``(num_envs, 3)``.
-        ``RigidObjectSet`` overrides this to return the bbox for each env's
-        assigned variant.
-
-        Args:
-            num_envs: Number of environments.
-
-        Returns:
-            ``AxisAlignedBoundingBox`` with ``min_point`` / ``max_point`` of shape
-            ``(num_envs, 3)``.
-        """
-        bbox = self.get_bounding_box()
-        return AxisAlignedBoundingBox(
-            min_point=bbox.min_point.expand(num_envs, 3),
-            max_point=bbox.max_point.expand(num_envs, 3),
-        )
 
     def _get_initial_pose_as_pose(self) -> Pose | None:
         """Return a single ``Pose`` suitable for *init_state* and bounding-box calculations.
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -74,11 +74,12 @@ class ObjectBase(Asset, ABC):
         ...
 
     def get_bounding_box_per_env(self, num_envs: int) -> AxisAlignedBoundingBox:
-        """Get per-environment local bounding boxes.
+        """Get local bounding boxes for each environment.
 
-        For homogeneous objects the single local bbox is expanded to ``(num_envs, 3)``.
-        ``RigidObjectSet`` overrides this to return the actual bbox of each env's
-        variant, enabling heterogeneous placement.
+        This default implementation is for objects with the same geometry in
+        every environment: it expands the single local bbox to ``(num_envs, 3)``.
+        ``RigidObjectSet`` overrides this to return the bbox for each env's
+        assigned variant.
 
         Args:
             num_envs: Number of environments.
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -73,6 +73,26 @@ class ObjectBase(Asset, ABC):
         """Get bounding box in world coordinates (local bbox rotated and translated)."""
         ...
 
+    def get_bounding_box_per_env(self, num_envs: int) -> AxisAlignedBoundingBox:
+        """Get per-environment local bounding boxes.
+
+        For homogeneous objects the single local bbox is expanded to ``(num_envs, 3)``.
+        ``RigidObjectSet`` overrides this to return the actual bbox of each env's
+        variant, enabling heterogeneous placement.
+
+        Args:
+            num_envs: Number of environments.
+
+        Returns:
+            ``AxisAlignedBoundingBox`` with ``min_point`` / ``max_point`` of shape
+            ``(num_envs, 3)``.
+        """
+        bbox = self.get_bounding_box()
+        return AxisAlignedBoundingBox(
+            min_point=bbox.min_point.expand(num_envs, 3),
+            max_point=bbox.max_point.expand(num_envs, 3),
+        )
+
     def _get_initial_pose_as_pose(self) -> Pose | None:
         """Return a single ``Pose`` suitable for *init_state* and bounding-box calculations.
 
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -54,6 +54,7 @@ class ObjectBase(Asset, ABC):
         self.object_cfg = None
         self.event_cfg = None
         self.relations: list[RelationBase] = []
+        self.has_env_specific_bboxes: bool = False
 
     def get_initial_pose(self) -> Pose | PoseRange | PosePerEnv | None:
         """Return the current initial pose of this object.
--- a/isaaclab_arena/assets/object_set.py
+++ b/isaaclab_arena/assets/object_set.py
@@ -71,6 +71,12 @@ class RigidObjectSet(Object):
         self.random_choice = random_choice
         self.heterogeneous_bbox: bool = len(objects) > 1
 
+        assert not (self.heterogeneous_bbox and self.random_choice), (
+            f"RigidObjectSet '{name}': random_choice=True is not supported with heterogeneous "
+            "placement (len(objects) > 1). The placement pool assumes round-robin variant "
+            "assignment (env_idx % num_variants) which conflicts with random spawning order."
+        )
+
         # Set default prim_path if not provided
         if prim_path is None:
             prim_path = f"{{ENV_REGEX_NS}}/{name}"
--- a/isaaclab_arena/assets/object_set.py
+++ b/isaaclab_arena/assets/object_set.py
@@ -93,7 +93,7 @@ class RigidObjectSet(Object):
         """
         return max(self.objects, key=lambda obj: obj.get_bounding_box().size[0, 2].item()).get_bounding_box()
 
-    def get_variant_indices(self, num_envs: int) -> list[int]:
+    def get_variant_indices(self, num_envs: int, seed: int | None = None) -> list[int]:
         """Return which member object index is assigned to each environment.
 
         When ``random_choice`` is False the mapping is round-robin
@@ -103,6 +103,9 @@ class RigidObjectSet(Object):
 
         Args:
             num_envs: Number of environments.
+            seed: Optional RNG seed for reproducible variant assignment
+                when ``random_choice`` is True. If None, uses the global
+                torch RNG.
 
         Returns:
             List of length ``num_envs`` with indices into ``self.objects``.
@@ -112,7 +115,11 @@ class RigidObjectSet(Object):
             return [i % n for i in range(num_envs)]
 
         if not hasattr(self, "_cached_variant_indices") or len(self._cached_variant_indices) != num_envs:
-            self._cached_variant_indices = [int(i) for i in torch.randint(n, (num_envs,)).tolist()]
+            generator = None
+            if seed is not None:
+                generator = torch.Generator()
+                generator.manual_seed(seed)
+            self._cached_variant_indices = [int(i) for i in torch.randint(n, (num_envs,), generator=generator).tolist()]
         return self._cached_variant_indices
 
     def get_bounding_box_per_env(self, num_envs: int) -> AxisAlignedBoundingBox:
--- a/isaaclab_arena/assets/object_set.py
+++ b/isaaclab_arena/assets/object_set.py
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import torch
+
 import isaaclab.sim as sim_utils
 from isaaclab.assets import RigidObjectCfg
 from isaaclab.sensors.contact_sensor.contact_sensor_cfg import ContactSensorCfg
@@ -67,6 +69,7 @@ class RigidObjectSet(Object):
 
         self.objects: list[Object] = objects
         self.random_choice = random_choice
+        self.heterogeneous_bbox: bool = len(objects) > 1
 
         # Set default prim_path if not provided
         if prim_path is None:
@@ -89,6 +92,49 @@ class RigidObjectSet(Object):
         This is a heuristic to avoid objects spawning inside their support surfaces.
         """
         return max(self.objects, key=lambda obj: obj.get_bounding_box().size[0, 2].item()).get_bounding_box()
+
+    def get_variant_indices(self, num_envs: int) -> list[int]:
+        """Return which member object index is assigned to each environment.
+
+        When ``random_choice`` is False the mapping is round-robin
+        (``env_idx % len(objects)``).  When True, a random permutation is
+        generated (and cached so repeated calls with the same ``num_envs``
+        are deterministic within a session).
+
+        Args:
+            num_envs: Number of environments.
+
+        Returns:
+            List of length ``num_envs`` with indices into ``self.objects``.
+        """
+        n = len(self.objects)
+        if not self.random_choice:
+            return [i % n for i in range(num_envs)]
+
+        if not hasattr(self, "_cached_variant_indices") or len(self._cached_variant_indices) != num_envs:
+            self._cached_variant_indices = [int(i) for i in torch.randint(n, (num_envs,)).tolist()]
+        return self._cached_variant_indices
+
+    def get_bounding_box_per_env(self, num_envs: int) -> AxisAlignedBoundingBox:
+        """Get the actual bounding box for each env's variant.
+
+        Unlike ``get_bounding_box()`` (which uses a max-z heuristic), this
+        returns the real local bbox of the variant assigned to each env,
+        enabling correct collision-free placement for heterogeneous scenes.
+
+        Args:
+            num_envs: Number of environments.
+
+        Returns:
+            ``AxisAlignedBoundingBox`` with ``min_point`` / ``max_point`` of
+            shape ``(num_envs, 3)``.
+        """
+        variant_indices = self.get_variant_indices(num_envs)
+        member_bboxes = [obj.get_bounding_box() for obj in self.objects]
+
+        min_pts = torch.stack([member_bboxes[idx].min_point[0] for idx in variant_indices], dim=0)
+        max_pts = torch.stack([member_bboxes[idx].max_point[0] for idx in variant_indices], dim=0)
+        return AxisAlignedBoundingBox(min_point=min_pts, max_point=max_pts)
 
     def get_contact_sensor_cfg(self, contact_against_object: ObjectBase | None = None) -> ContactSensorCfg:
         # We assume that by here, our USDs have been modified to be compatible with each other
--- a/isaaclab_arena/assets/object_set.py
+++ b/isaaclab_arena/assets/object_set.py
@@ -151,8 +151,8 @@ class RigidObjectSet(Object):
 
     def get_contact_sensor_cfg(self, contact_against_object: ObjectBase | None = None) -> ContactSensorCfg:
         # We assume that by here, our USDs have been modified to be compatible with each other
-        # and we can use the first USD path to find the shallowest rigid body.
-        return super().get_contact_sensor_cfg(contact_against_object, usd_path=self.object_usd_paths[0])
+        # and we can use the canonical first member USD to find the shallowest rigid body.
+        return super().get_contact_sensor_cfg(contact_against_object, usd_path=self._member_object_usd_paths[0])
 
     def _generate_variant_indices(self, num_envs: int) -> list[int]:
         n = len(self.objects)
@@ -171,6 +171,7 @@ class RigidObjectSet(Object):
 
         self.variant_indices_by_env = list(variant_indices_by_env)
         if len(self.objects) > 1:
+            # Keep Isaac Lab's MultiUsdFileCfg aligned with the fixed per-env variant assignment.
             self.object_usd_paths = [self._member_object_usd_paths[idx] for idx in self.variant_indices_by_env]
             spawn_cfg = self.object_cfg.spawn if getattr(self, "object_cfg", None) is not None else None
             if isinstance(spawn_cfg, sim_utils.MultiUsdFileCfg):
--- a/isaaclab_arena/assets/object_set.py
+++ b/isaaclab_arena/assets/object_set.py
@@ -29,7 +29,8 @@ class RigidObjectSet(Object):
         objects: list[Object],
         prim_path: str | None = None,
         scale: tuple[float, float, float] = (1.0, 1.0, 1.0),
-        random_choice: bool = False,
+        random_choice: bool = True,
+        variant_indices_by_env: list[int] | None = None,
         initial_pose: Pose | None = None,
         **kwargs,
     ):
@@ -42,7 +43,9 @@ class RigidObjectSet(Object):
             scale: The scale of the object set. Note all objects can only have the same scale, if
                 different scales are needed, considering scaling the object USD file.
             random_choice: Whether to randomly choose an object from the object set to spawn in
-                each environment. If False, object is spawned based on the order of objects in the list.
+                each environment. The assignment is sampled once when ``num_envs`` is known and
+                then reused across resets.
+            variant_indices_by_env: Optional fixed variant index for each environment.
             initial_pose: The initial pose of the object from this object set.
         """
         if not self._are_all_objects_type_rigid(objects):
@@ -65,18 +68,19 @@ class RigidObjectSet(Object):
             self.object_usd_paths = self._modify_assets(objects)
             print(f"Modified object USD paths: {self.object_usd_paths}")
         else:
-            self.object_usd_paths = [object.usd_path for object in objects]
+            self.object_usd_paths = []
+            for obj in objects:
+                assert obj.usd_path is not None
+                self.object_usd_paths.append(obj.usd_path)
 
         self.objects: list[Object] = objects
+        self._member_object_usd_paths: list[str] = list(self.object_usd_paths)
         self.random_choice = random_choice
         self.has_env_specific_bboxes: bool = len(objects) > 1
+        self.variant_indices_by_env: list[int] | None = None
 
-        if self.has_env_specific_bboxes and self.random_choice:
-            raise ValueError(
-                f"RigidObjectSet '{name}': random_choice=True is not supported with heterogeneous "
-                "placement (len(objects) > 1). The placement pool assumes round-robin variant "
-                "assignment (env_index % num_variants) which conflicts with random spawning order."
-            )
+        if variant_indices_by_env is not None:
+            self._set_variant_indices_by_env(variant_indices_by_env)
 
         # Set default prim_path if not provided
         if prim_path is None:
@@ -103,10 +107,9 @@ class RigidObjectSet(Object):
     def get_variant_indices(self, num_envs: int) -> list[int]:
         """Return which member object index is assigned to each environment.
 
-        Multi-variant sets use round-robin assignment
-        (``env_index % len(objects)``). ``random_choice=True`` with multiple
-        variants is rejected in ``__init__`` because placement needs to know
-        the assigned variant for each env.
+        Multi-variant sets use one fixed assignment for the lifetime of the
+        object set. By default, each env independently samples one variant
+        once, then keeps it across resets.
 
         Args:
             num_envs: Number of environments.
@@ -114,8 +117,15 @@ class RigidObjectSet(Object):
         Returns:
             List of length ``num_envs`` with indices into ``self.objects``.
         """
-        n = len(self.objects)
-        return [i % n for i in range(num_envs)]
+        if self.variant_indices_by_env is None:
+            self._set_variant_indices_by_env(self._generate_variant_indices(num_envs))
+        elif len(self.variant_indices_by_env) != num_envs:
+            raise ValueError(
+                f"RigidObjectSet '{self.name}' has variant assignments for "
+                f"{len(self.variant_indices_by_env)} envs, got request for {num_envs}."
+            )
+        assert self.variant_indices_by_env is not None
+        return self.variant_indices_by_env
 
     def get_bounding_box_per_env(self, num_envs: int) -> AxisAlignedBoundingBox:
         """Get the actual bounding box for each env's variant.
@@ -143,10 +153,41 @@ class RigidObjectSet(Object):
         # and we can use the first USD path to find the shallowest rigid body.
         return super().get_contact_sensor_cfg(contact_against_object, usd_path=self.object_usd_paths[0])
 
-    def _are_all_objects_type_rigid(self, objects: list[ObjectBase]) -> bool:
+    def _generate_variant_indices(self, num_envs: int) -> list[int]:
+        n = len(self.objects)
+        if n == 1:
+            return [0 for _ in range(num_envs)]
+        if not self.random_choice:
+            raise ValueError(
+                f"RigidObjectSet '{self.name}' has {n} variants and random_choice=False, "
+                "but no variant_indices_by_env were provided."
+            )
+        return torch.randint(low=0, high=n, size=(num_envs,)).tolist()
+
+    def _set_variant_indices_by_env(self, variant_indices_by_env: list[int]) -> None:
+        n = len(self.objects)
+        if any(idx < 0 or idx >= n for idx in variant_indices_by_env):
+            raise ValueError(
+                f"RigidObjectSet '{self.name}' variant indices must be in [0, {n}); "
+                f"got {variant_indices_by_env}."
+            )
+
+        self.variant_indices_by_env = list(variant_indices_by_env)
+        if self.has_env_specific_bboxes:
+            self.object_usd_paths = [self._member_object_usd_paths[idx] for idx in self.variant_indices_by_env]
+            spawn_cfg = self.object_cfg.spawn if getattr(self, "object_cfg", None) is not None else None
+            if isinstance(spawn_cfg, sim_utils.MultiUsdFileCfg):
+                spawn_cfg.usd_path = self.object_usd_paths
+                spawn_cfg.random_choice = False
+
+    def _are_all_objects_type_rigid(self, objects: list[Object]) -> bool:
         if objects is None or len(objects) == 0:
             raise ValueError(f"Object set {self.name} must contain at least 1 object.")
-        return all(detect_object_type(usd_path=object.usd_path) == ObjectType.RIGID for object in objects)
+        for obj in objects:
+            assert obj.usd_path is not None
+            if detect_object_type(usd_path=obj.usd_path) != ObjectType.RIGID:
+                return False
+        return True
 
     def _generate_rigid_cfg(self) -> RigidObjectCfg:
         assert self.object_type == ObjectType.RIGID
@@ -154,11 +195,12 @@ class RigidObjectSet(Object):
             prim_path=self.prim_path,
             spawn=sim_utils.MultiUsdFileCfg(
                 usd_path=self.object_usd_paths,
-                random_choice=self.random_choice,
+                random_choice=self.random_choice if self.variant_indices_by_env is None else False,
                 activate_contact_sensors=True,
             ),
         )
         object_cfg = self._add_initial_pose_to_cfg(object_cfg)
+        assert isinstance(object_cfg, RigidObjectCfg)
         return object_cfg
 
     def _generate_articulation_cfg(self):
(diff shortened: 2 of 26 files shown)


isaaclab_arena/environments/isaaclab_arena_environment.py
```diff
@@ -26,7 +26,9 @@ def __init__(
         embodiment: EmbodimentBase | None = None,
         task: TaskBase | None = None,
         teleop_device: TeleopDeviceBase | None = None,
-        env_cfg_callback: Callable[IsaacLabArenaManagerBasedRLEnvCfg] | None = None,
+        env_cfg_callback: (
+            Callable[[IsaacLabArenaManagerBasedRLEnvCfg], IsaacLabArenaManagerBasedRLEnvCfg] | None
```
cvolkcvolk: Is this change needed?

isaaclab_arena/relations/pooled_object_placer.py
```diff
@@ -20,64 +21,122 @@
 class PooledObjectPlacer:
     """Object placer that maintains a pool of valid placement layouts.
 
-    Wraps :class:`ObjectPlacer` and solves layouts in batches of ``pool_size``,
-    keeping only those that pass validation.  The pool is refilled automatically
-    when consumed layouts run out.
+    Storage: ``num_envs`` independent layout pools, each with its own read
+    cursor (this replaces the single ``_layouts`` list + ``_next_idx`` cursor
```
cvolkcvolk:
```suggestion

```

cvolkcvolk: Suggestion to remove: "(this replaces the single ``_layouts`` list + ``_next_idx`` cursor used before heterogeneous placement)" Ai leftover

isaaclab_arena/assets/object_base.py
```diff
@@ -69,6 +71,27 @@ def get_world_bounding_box(self) -> AxisAlignedBoundingBox:
         """Get bounding box in world coordinates (local bbox rotated and translated)."""
         ...
 
+    def get_bounding_box_per_env(self, num_envs: int) -> AxisAlignedBoundingBox:
```
cvolkcvolk: I feel like this should not live in `object_base.py`. It's a runtime detail of the placement/simulation &`num_envs` is a placer/scene concept, not an object concept. 
Same for the introduced variable `has_env_specific_bboxes`: It only matters from the placer's perspective.

Could we move this to a relation side helper?

isaaclab_arena/relations/relation_solver.py
```diff
@@ -66,12 +67,32 @@ def _get_strategy(self, relation: RelationBase) -> RelationLossStrategy | UnaryR
...
+        """Return the per-env or default local bbox for *obj*, moved to *device*."""
+        if env_bboxes is not None and obj in env_bboxes:
+            return env_bboxes[obj].to(device)
+        return obj.get_bounding_box().to(device)
+
+    def _compute_total_loss(
+        self,
+        state: RelationSolverState,
+        debug: bool = False,
+        env_bboxes: dict[ObjectBase, AxisAlignedBoundingBox] | None = None,
```
cvolkcvolk: Is there another place where this could be stored and accessed within this function to not need to thread it trough here? 
Maybe store env_bboxes on RelationSolverState alongside positions as state.get_position(obj) and state.get_bbox(obj)?

isaaclab_arena/environments/isaaclab_arena_environment.py
zhx06: No it is not, will revert it

isaaclab_arena/relations/pooled_object_placer.py
zhx06: I reviewed the comments and docstring and removed the leftover or confusing statement.

isaaclab_arena/assets/object_base.py
zhx06: Thanks! We now have a new file` bbox_helpers.py ` where we introduced environment-specific variable to bboxes if necessary, if not just keep the same as in the original bbox.

isaaclab_arena/relations/relation_solver.py
zhx06: Done. Now `RelationSolverState` stores bbox, will return env_bboxs if it is used, so users can get the env-specific variables from the solver states.

isaaclab_arena/assets/object_set.py
```diff
@@ -63,10 +68,19 @@ def __init__(
...
             print(f"Modified object USD paths: {self.object_usd_paths}")
         else:
-            self.object_usd_paths = [object.usd_path for object in objects]
+            self.object_usd_paths = []
+            for obj in objects:
+                assert obj.usd_path is not None
+                self.object_usd_paths.append(obj.usd_path)
 
         self.objects: list[Object] = objects
+        self._member_object_usd_paths: list[str] = list(self.object_usd_paths)
```
alexmillane: @cvolkcvolk . This relates to our discussion on your MR. Basically, agents are starting to insert leading underscores into our code. Honestly, probably correctly. The problem is that we haven't done this consistently.

We need to make a decision as a team about our coding style here and try to adhere to it (and put in the agent.md our decision.

@zhx06 For now I would avoid the leading underscore in this class. Basically all these members are intended to be private (even though they don't have a leading underscore). Let's aim for class-level consistency until we make a call.

isaaclab_arena/assets/object_set.py
```diff
@@ -90,27 +104,101 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
-    def _are_all_objects_type_rigid(self, objects: list[ObjectBase]) -> bool:
+        # and we can use the canonical first member USD to find the shallowest rigid body.
+        return super().get_contact_sensor_cfg(contact_against_object, usd_path=self._member_object_usd_paths[0])
+
+    def _generate_variant_indices(self, num_envs: int) -> list[int]:
+        n = len(self.objects)
+        if n == 1:
+            return [0 for _ in range(num_envs)]
+        if not self.random_choice:
+            return [env_idx % n for env_idx in range(num_envs)]
```
alexmillane: I think we don't need the extra branch here for `n == 1`?

The following two options generate a list of zeros.

isaaclab_arena/assets/object_set.py
```diff
@@ -90,27 +104,101 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
             prim_path=self.prim_path,
             spawn=sim_utils.MultiUsdFileCfg(
                 usd_path=self.object_usd_paths,
-                random_choice=self.random_choice,
+                random_choice=self.random_choice if self.variant_indices_by_env is None else False,
                 activate_contact_sensors=True,
             ),
         )
         object_cfg = self._add_initial_pose_to_cfg(object_cfg)
+        assert isinstance(object_cfg, RigidObjectCfg)
```
alexmillane: What's the meaning of this check? Above we create the `RigidObjectCfg` and here we check it? Are we making sure we dont get weird behaviour in `_add_initial_pose_to_cfg`?

isaaclab_arena/assets/object_set.py
```diff
@@ -90,27 +104,101 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+        if not self.random_choice:
+            return [env_idx % n for env_idx in range(num_envs)]
+        return torch.randint(low=0, high=n, size=(num_envs,)).tolist()
+
+    def _set_variant_indices_by_env(self, variant_indices_by_env: list[int]) -> None:
+        n = len(self.objects)
+        if any(idx < 0 or idx >= n for idx in variant_indices_by_env):
+            raise ValueError(
+                f"RigidObjectSet '{self.name}' variant indices must be in [0, {n}); got {variant_indices_by_env}."
+            )
```
alexmillane: `assert`

alexmillane: We should probably add this to the `AGENTS.md` that we prefer `assert` rather than `if...raise`

isaaclab_arena/assets/object_set.py
```diff
@@ -90,27 +104,101 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+
+        self.variant_indices_by_env = list(variant_indices_by_env)
```
alexmillane: why `list`. The type hint say that this thing should already be a list?

isaaclab_arena/assets/object_set.py
```diff
@@ -90,27 +104,101 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+
+        Args:
+            num_envs: Number of environments.
+
+        Returns:
+            ``AxisAlignedBoundingBox`` with ``min_point`` / ``max_point`` of
+            shape ``(num_envs, 3)``.
+        """
+        variant_indices = self.get_variant_indices(num_envs)
+        member_bboxes = [obj.get_bounding_box() for obj in self.objects]
```
alexmillane: `member_bounding_boxes` or just `bounding_boxes`.

Try to avoid abbreviations.

isaaclab_arena/assets/object_set.py
```diff
@@ -90,27 +104,101 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+
+    def _generate_variant_indices(self, num_envs: int) -> list[int]:
+        n = len(self.objects)
+        if n == 1:
+            return [0 for _ in range(num_envs)]
+        if not self.random_choice:
+            return [env_idx % n for env_idx in range(num_envs)]
+        return torch.randint(low=0, high=n, size=(num_envs,)).tolist()
+
+    def _set_variant_indices_by_env(self, variant_indices_by_env: list[int]) -> None:
```
alexmillane, line 182: Suggestion to add a short docstring. Hard to know what this function is doing.

isaaclab_arena/assets/object_set.py
alexmillane: What's the reason for having `self.object_usd_paths` AND `self._member_object_usd_paths`

They are at least initialized to being the same thing? Perhaps one is shuffled with respect to the other?

Can we get away with one of these members? It's making things very difficult to interpret what's going on.

isaaclab_arena/assets/object_set.py
```diff
@@ -63,10 +68,19 @@ def __init__(
...
+            self.object_usd_paths = []
+            for obj in objects:
+                assert obj.usd_path is not None
+                self.object_usd_paths.append(obj.usd_path)
 
         self.objects: list[Object] = objects
+        self._member_object_usd_paths: list[str] = list(self.object_usd_paths)
         self.random_choice = random_choice
+        self.variant_indices_by_env: list[int] | None = None
+        self.has_env_specific_bboxes = len(objects) > 1
```
alexmillane: Why do we have this? I would assume that a user who is using the `RigidObjectSet`, has more than a single object in the set. I think that that's a safe assumption. I would prefer that we just `assert len(objects) > 1` and then use that assumption to simplify everything.

isaaclab_arena/relations/bbox_helpers.py
```diff
@@ -0,0 +1,44 @@
...
+
+from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox
+
+if TYPE_CHECKING:
+    from isaaclab_arena.assets.object_base import ObjectBase
+
+
+def is_heterogeneous(obj: ObjectBase) -> bool:
+    """True if obj has different bounding boxes per environment (e.g. multi-variant RigidObjectSet)."""
+    return getattr(obj, "has_env_specific_bboxes", False)
```
alexmillane: Suggestion to just substitute with `isinstance(obj, RigidObjectSet)`

isaaclab_arena/assets/object_set.py
```diff
@@ -90,27 +104,101 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+        order across environments.
+
+        Args:
+            num_envs: Number of environments.
+
+        Returns:
+            List of length ``num_envs`` with indices into ``self.objects``.
+        """
+        if self.variant_indices_by_env is None:
+            self._set_variant_indices_by_env(self._generate_variant_indices(num_envs))
```
alexmillane: Here we appear to be setting `variant_indices_by_env`, in a function called `get_variant_indices`. That is quite confusing behaviour, and is making it quite hard to tell what's going on in this class.

Let's have a larger discussion about the function of this class.

isaaclab_arena/assets/object_set.py
```diff
@@ -90,27 +104,101 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+        return True
 
     def _generate_rigid_cfg(self) -> RigidObjectCfg:
         assert self.object_type == ObjectType.RIGID
         object_cfg = RigidObjectCfg(
             prim_path=self.prim_path,
             spawn=sim_utils.MultiUsdFileCfg(
                 usd_path=self.object_usd_paths,
-                random_choice=self.random_choice,
+                random_choice=self.random_choice if self.variant_indices_by_env is None else False,
```
alexmillane: It's quite confusing what's going on in this class.

It appears to me that if you request bounding boxes, the random choice being handled by Isaac Lab is overridden and the randomness is handled by us? Is that correct?

That makes sense to me - we need to handle handle the obj -> env mapping, such that we have access to which bounding box ends up where.

The dataflow in this class is very confusing. I think because we're trying to maintain the option that randomness is handled by Isaac Lab.

Am I correct that we could simplify this class substantially by just _always_ handling randomization in the Arena layer?

alexmillane (commented):
Hi @zhx06.

Thank you for implementing this.

I have given it a (very) partial review that focused on the `RigidObjectSet`. It's becoming quite complicated to follow what's going on in that class.

I think that if we make some additional assumptions and simplifications we'll get something more easily understood.

isaaclab_arena/assets/object_set.py
```diff
@@ -77,11 +90,22 @@ def __init__(
             object_type=ObjectType.RIGID,
             usd_path="",
             prim_path=prim_path,
-            scale=(1.0, 1.0, 1.0),  # We rewrite the USDs to handle scaling
+            scale=(1.0, 1.0, 1.0),
```
alexmillane: Suggestion to add the comment back in.

isaaclab_arena/assets/object_set.py
```diff
@@ -90,23 +114,97 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+                    f"RigidObjectSet '{self.name}' has variant assignments for "
+                    f"{len(self.variant_indices_by_env)} envs, got request for {num_envs}."
+                )
+            return
+        self._set_variant_indices_by_env(self._generate_variant_indices(num_envs))
+
+    def get_bounding_box_per_env(self, num_envs: int) -> AxisAlignedBoundingBox:
+        """Get the actual bounding box for each env's variant.
+
+        Unlike ``get_bounding_box()`` (which uses a max-z heuristic), this
```
alexmillane: Do we still use the max-z heuristic somewhere? Is there a reason to still have that around?

isaaclab_arena/relations/bounding_box_helpers.py
alexmillane, line 1: -> `bounding_box_helpers` (avoid abbreviations)

isaaclab_arena/relations/object_placer.py
```diff
@@ -85,6 +106,10 @@ def place(
             result_per_env: When True (default), each environment gets a distinct
                 layout. When False, a single best layout is solved and applied
                 identically to all environments.
+            env_bboxes: Pre-computed per-env bounding boxes (shape (num_envs, 3)
+                per object). When provided, _place_heterogeneous uses these
+                instead of calling get_bounding_box_per_env(obj, num_envs). This
+                allows the caller to tile real-env bboxes for pooled solving.
```
alexmillane: This seems a little strange to me. Is there a reason to have the bounding boxes computed outside?

Is this some requirement coming from the pooling?

isaaclab_arena/relations/object_placer.py
```diff
@@ -119,67 +144,192 @@ def place(
...
-        num_candidates = self.params.max_placement_attempts * num_results
+        max_attempts = self.params.max_placement_attempts
+        num_candidates = max_attempts * num_results
+
+        # Two solve paths produce a list[PlacementResult] of length num_results, sorted by
+        # (is_valid, loss):
+        #   - homogeneous: any solved layout serves any env; pick the top num_results.
+        #   - heterogeneous: some objects vary per env, so each env owns a fixed slice
+        #     of candidates and we pick the best within that slice.
+        uses_env_specific_bboxes = result_per_env and has_heterogeneous_objects(objects)
```
alexmillane: Is it true that in the hetrogeneous case we ALWAYS want a result per env?

In the past, we had heterogeneous environments which got the same placement per env (using an approximation of the bounding box). However, this was always a stepping stone to the true solution, which is the subject of this MR.

I would suggest removing the old behaviour. That is, if you heterogeneous objects, always compute a solution per env. (unless you have a good reason for maintaining this behaviour?)

isaaclab_arena/relations/object_placer.py
```diff
@@ -119,67 +144,192 @@ def place(
...
+
+        if uses_env_specific_bboxes:
+            results_per_env = self._place_heterogeneous(
+                objects,
+                anchor_objects_set,
+                num_envs,
+                max_attempts,
+                num_candidates,
+                generator,
+                env_bboxes=env_bboxes,
```
alexmillane: Can we request these inside this function, to make the interface to `place` a bit cleaner?

isaaclab_arena/relations/object_placer.py
```diff
@@ -119,67 +144,192 @@ def place(
...
+            )
+        else:
+            results_per_env = self._place_homogeneous(
+                objects, anchor_objects_set, num_results, max_attempts, num_candidates, generator
+            )
+
+        final_per_env = [r.positions for r in results_per_env]
```
alexmillane: `positions_per_env`

isaaclab_arena/relations/object_placer.py
```diff
@@ -119,67 +144,192 @@ def place(
...
+        if self.params.apply_positions_to_objects:
+            self._apply_positions(final_per_env, anchor_objects_set)
+
+        if num_results == 1:
+            return results_per_env[0]
+        return MultiEnvPlacementResult(results=results_per_env)
```
alexmillane: Could we simplify things by wrapping these two into a single placement result object, that is either multi-env or single env?

isaaclab_arena/relations/object_placer.py
```diff
@@ -119,67 +144,192 @@ def place(
...
 
-        all_candidates: list[PlacementCandidate] = []
-        for idx in range(num_candidates):
-            loss = all_losses[idx]
-            is_valid = self._validate_placement(all_positions[idx])
-            all_candidates.append(PlacementCandidate(loss, all_positions[idx], is_valid))
-
-        # Sort: valid solutions first (by loss), then invalid (by loss)
+        all_candidates = [
+            PlacementCandidate(all_losses[idx], all_positions[idx], self._validate_placement(all_positions[idx]))
```
alexmillane: nit: move the validation call above. something like

```
all_validatations = [self._validate_placement(positions) for positions in all_positions]
```

isaaclab_arena/relations/object_placer.py
```diff
@@ -119,67 +144,192 @@ def place(
...
+        num_candidates: int,
+        generator: torch.Generator | None,
+        env_bboxes: dict[ObjectBase, AxisAlignedBoundingBox] | None = None,
+    ) -> list[PlacementResult]:
+        """Per-env placement: each candidate is tied to its env's object variants.
+
+        Candidates [cur_env * max_attempts : (cur_env + 1) * max_attempts]
+        belong to cur_env and use that env's variant geometry. We solve all
+        num_envs * max_attempts candidates in one batched solver.solve()
+        call and pick the best candidate within each env's slice.
```
alexmillane: Suggestion to describe the shape of the return value.

alexmillane: So there's a bit of a difference between homogeneous and heterogeneous.

In homogeneous, we return ranked results, in heterogenous we return a single result per env? Is that correct?

Do we then loose the ability to do pooling? A clearer equivalent to me would be to have ranked results per env.

What's your thinking there?

alexmillane (commented):
Another partial review. My main suggestion is that we try to do some unification between the heterogeneous and homogeneous paths in the placer and solver.

Is there any reason that we can't treat a homogeneous solve as a homogeneous solve, except that the bounding boxes happen to be the same over each environment? Is there a reason we can't do that?

One potential reason I can see is that right now (I think) we have batching behaviour on the homogeneous code path, and we don't on the heterogeneous code path. Is my reading correct? Could we also do some unification there? I.e. both code paths have batching behaviour?

My central fear is that with all this branching, the solver is going to become quite difficult to maintain over time, because each path has to be separately tested, maintained, and fixed.

isaaclab_arena/assets/object_set.py
zhx06: Heterogeneous placement uses `get_bounding_box_per_env() `and does not use this max-z heuristic.

isaaclab_arena/assets/object_set.py
zhx06: This is reverted now.

isaaclab_arena/relations/bounding_box_helpers.py
zhx06: File renamed.

isaaclab_arena/assets/object_set.py
zhx06: Just to confirm, we don't want to add new underscored members, but underscored helper methods are still fine?

isaaclab_arena/relations/object_placer.py
zhx06: I changed the design so  both homogeneous and heterogeneous placement share the same ranked placement designed. The differences lie in the ranking mode, whether it is shared geometry ranks from one common pool, or env-specific geometry ranks for each single environment.

isaaclab_arena/assets/object_set.py
```diff
@@ -77,36 +88,134 @@ def __init__(
...
+        Args:
+            num_envs: Number of environments to assign variants for.
+            variant_seed: Optional seed used when random_choice=True.
+        """
+        if self.variant_indices_by_env is not None:
+            assert len(self.variant_indices_by_env) == num_envs, (
+                f"RigidObjectSet '{self.name}' has variant assignments for "
+                f"{len(self.variant_indices_by_env)} envs, got request for {num_envs}."
+            )
+            return
```
alexmillane: suggestion to do if..else, rather than if...return.

isaaclab_arena/assets/object_set.py
```diff
@@ -77,36 +88,134 @@ def __init__(
...
+
+        Args:
+            num_envs: Number of environments to assign variants for.
+            variant_seed: Optional seed used when random_choice=True.
+        """
+        if self.variant_indices_by_env is not None:
+            assert len(self.variant_indices_by_env) == num_envs, (
+                f"RigidObjectSet '{self.name}' has variant assignments for "
+                f"{len(self.variant_indices_by_env)} envs, got request for {num_envs}."
+            )
```
alexmillane: I feel that this function handle being called twice. If it's called twice regenerate the indices. On the second call it should maybe print a warning saying that it's regenerating the indices.

isaaclab_arena/assets/object_set.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         prim_path: str | None = None,
         scale: tuple[float, float, float] = (1.0, 1.0, 1.0),
         random_choice: bool = False,
+        variant_indices_by_env: list[int] | None = None,
```
alexmillane: I think that this constructor argument is only used in a test? Suggestion to remove it and rely on the member function.

isaaclab_arena/assets/object_set.py
```diff
@@ -77,36 +88,134 @@ def __init__(
...
+        return True
 
     def _generate_rigid_cfg(self) -> RigidObjectCfg:
         assert self.object_type == ObjectType.RIGID
         object_cfg = RigidObjectCfg(
             prim_path=self.prim_path,
             spawn=sim_utils.MultiUsdFileCfg(
                 usd_path=self.object_usd_paths,
-                random_choice=self.random_choice,
+                random_choice=False,
```
alexmillane, line 210: Maybe add a small inline comment here that this is hardcoded false because the randomization is handled at the Arena level in this class.

isaaclab_arena/relations/bounding_box_helpers.py
```diff
@@ -0,0 +1,61 @@
...
+
+def assign_variants_for_envs(objects: list[ObjectBase], num_envs: int, placement_seed: int | None = None) -> None:
+    """Assign per-env variants on every RigidObjectSet in the list.
+
+    Placers call this once they know the real environment count, before
+    requesting per-env bounding boxes. Non-RigidObjectSet objects are skipped.
+    Seeded assignments offset each set by its index so multiple sets do not
+    reuse the same random sequence.
+    """
+    from isaaclab_arena.assets.object_set import RigidObjectSet
```
alexmillane, line 37: Do we need to make these local imports? Do we need to import this file in a context that Isaac Sim has not started in?

isaaclab_arena/relations/object_placer.py
```diff
@@ -61,14 +66,19 @@ class ObjectPlacer:
...
-        """Initialize the ObjectPlacer.
-
-        Args:
-            params: Configuration parameters. If None, uses defaults.
-        """
         self.params = params or ObjectPlacerParams()
         self._solver = RelationSolver(params=self.params.solver_params)
 
+    @staticmethod
+    def _resolve_bbox(
```
alexmillane: `_resolve_bounding_box`

isaaclab_arena/relations/object_placer.py
```diff
@@ -61,14 +66,19 @@ class ObjectPlacer:
...
+        obj: ObjectBase,
+        overrides: dict[ObjectBase, AxisAlignedBoundingBox] | None,
+    ) -> AxisAlignedBoundingBox:
+        """Return overrides[obj] if present, otherwise the object's default bbox."""
+        if overrides is not None and obj in overrides:
+            return overrides[obj]
+        return obj.get_bounding_box()
```
alexmillane: I'm struggling to see why we need this function.

From what I can tell, `overrides` is always set?

Is there a code path where we're not using the "overrides"? (I mean in production code, not in the tests. I suggest that we _always_ use computed bounding boxes and we delete this function.

isaaclab_arena/relations/object_placer.py
```diff
@@ -83,103 +93,266 @@ def place(
...
+        if ranking_mode == "shared_geometry":
+            assert not has_heterogeneous_objects(objects), (
+                "Shared-geometry ranking requires homogeneous objects; candidates are not interchangeable across"
+                " variants."
+            )
+
+        # Variant assignment fixes the env-to-USD mapping before bbox expansion.
+        assign_variants_for_envs(objects, num_envs, placement_seed=self.params.placement_seed)
+        env_bboxes = {obj: get_bounding_box_per_env(obj, num_envs) for obj in objects}
+        num_candidates = num_envs * candidates_per_env
```
alexmillane, line 185: move below

isaaclab_arena/relations/object_placer.py
```diff
@@ -83,103 +93,266 @@ def place(
...
+        """
+        if ranking_mode == "shared_geometry":
+            assert not has_heterogeneous_objects(objects), (
+                "Shared-geometry ranking requires homogeneous objects; candidates are not interchangeable across"
+                " variants."
+            )
+
+        # Variant assignment fixes the env-to-USD mapping before bbox expansion.
+        assign_variants_for_envs(objects, num_envs, placement_seed=self.params.placement_seed)
+        env_bboxes = {obj: get_bounding_box_per_env(obj, num_envs) for obj in objects}
```
alexmillane: suggestion to move this into `_build_candidate_bboxes`

isaaclab_arena/relations/object_placer.py
```diff
@@ -83,103 +93,266 @@ def place(
...
+        num_candidates = num_envs * candidates_per_env
+        candidate_bboxes, per_env_bbox_overrides = self._build_candidate_bboxes(
+            env_bboxes, num_envs, candidates_per_env
+        )
```
alexmillane: Can we rename `per_env_bbox_overrides`.

From what I can tell, we now always use the bounding boxes in `per_env_bbox_overrides`. So it's not really an override any more right? It's the default behaviour?

isaaclab_arena/relations/object_placer.py
```diff
@@ -83,103 +93,266 @@ def place(
...
+        per_env_bbox_overrides: list[dict[ObjectBase, AxisAlignedBoundingBox]] = [
+            {
+                obj: AxisAlignedBoundingBox(
+                    min_point=bbox.min_point[cur_env : cur_env + 1],
+                    max_point=bbox.max_point[cur_env : cur_env + 1],
+                )
+                for obj, bbox in env_bboxes.items()
+            }
+            for cur_env in range(num_envs)
+        ]
```
alexmillane: Can we make this not a list comprehension, it's quite difficult to interpret

isaaclab_arena/relations/object_placer.py
```diff
@@ -83,103 +93,266 @@ def place(
...
+
+        candidate_bboxes: dict[ObjectBase, AxisAlignedBoundingBox] = {}
+        for obj, bbox in env_bboxes.items():
+            candidate_bboxes[obj] = AxisAlignedBoundingBox(
+                min_point=bbox.min_point.repeat_interleave(candidates_per_env, dim=0),
+                max_point=bbox.max_point.repeat_interleave(candidates_per_env, dim=0),
             )
```
alexmillane: Am I correct that at a fundamental level, this function takes a specification of the per-env bounding boxes and reformats that information in two different ways?

It's quite hard to tell from the names what the different formats are. There's three formats floating around the code, the input to this function, and it's two outputs. 

There's a solution to this problem: encapsulate the per-bounding boxes in a new class, allow retrieval of the information, formatted in the three different ways, through member functions.

Then we can have just a single object that represents per-env bounding boxes and pass it around the code, and retrieve the correctly formatted option at the correct location.

isaaclab_arena/relations/object_placer.py
```diff
@@ -190,45 +363,74 @@ def _generate_initial_positions(
...
         positions: dict[ObjectBase, tuple[float, float, float]] = {}
         for obj in objects:
             if obj in anchor_objects:
-                positions[obj] = obj.get_initial_pose().position_xyz
+                initial_pose = obj.get_initial_pose()
+                if not isinstance(initial_pose, Pose):
+                    raise TypeError(
+                        f"Anchor object '{obj.name}' must have a fixed Pose before placement, got"
+                        f" {type(initial_pose).__name__}."
+                    )
```
alexmillane: `assert`

isaaclab_arena/relations/object_placer.py
```diff
@@ -190,45 +363,74 @@ def _generate_initial_positions(
...
+                    obj, anchor_objects, anchor_bbox, generator, child_bbox=bbox_override, bbox_overrides=child_bboxes
+                )
             else:
                 positions[obj] = (cx, cy, cz)
         return positions
 
+    @staticmethod
+    def _get_world_bbox_for_init(
+        obj: ObjectBase,
+        bbox_overrides: dict[ObjectBase, AxisAlignedBoundingBox] | None,
```
alexmillane: We're using the bounding boxes unconditionally now right?

Should we remove "overrides" in all the variable names and remove the option that the parameter is `None`?

This shoulds simplify some stuff.

isaaclab_arena/relations/object_placer.py
```diff
@@ -190,45 +363,74 @@ def _generate_initial_positions(
...
+    ) -> AxisAlignedBoundingBox:
+        bbox_override = bbox_overrides.get(obj) if bbox_overrides else None
+        if bbox_override is None:
+            return obj.get_world_bounding_box()
```
alexmillane: Is this still a valid option?

isaaclab_arena/relations/object_placer.py
```diff
@@ -83,103 +93,266 @@ def place(
...
+            f"Solved {num_candidates} candidates in one batch (env-specific geometry): "
+            f"{n_valid}/{num_envs} env(s) valid"
+        )
 
     def _generate_initial_positions(
         self,
         objects: list[ObjectBase],
         anchor_objects: set[ObjectBase],
         generator: torch.Generator | None = None,
+        child_bboxes: dict[ObjectBase, AxisAlignedBoundingBox] | None = None,
```
alexmillane: I think that we never use the `child_bboxes=None` branch.

This is making the code a lot more complex because we're maintaining a bunch of code paths that are no longer used.

isaaclab_arena/relations/object_placer.py
```diff
@@ -190,45 +363,74 @@ def _generate_initial_positions(
...
+                initial_pose = obj.get_initial_pose()
+                if not isinstance(initial_pose, Pose):
+                    raise TypeError(
+                        f"Anchor object '{obj.name}' must have a fixed Pose before placement, got"
+                        f" {type(initial_pose).__name__}."
+                    )
+                positions[obj] = initial_pose.position_xyz
             elif any(isinstance(r, On) for r in obj.get_relations()):
-                positions[obj] = self._compute_on_guided_position(obj, anchor_objects, anchor_bbox, generator)
+                bbox_override = child_bboxes.get(obj) if child_bboxes else None
```
alexmillane: see comment above, I think that the branches related to Non specified bounding boxes are no longer used.

isaaclab_arena/relations/object_placer.py
alexmillane: The code branches related to `bbox_overrides=None` are no longer used.

This is making the code more complex because we're always checking for conditions that are no longer valid.

alexmillane (commented):
Partial review.

It's looking better. Thanks for aligning the code paths between homogeneous and hetero geneous placement.

However, the clean up isn't complete. From what I can tell, the code has a bunch of branching for the case where we don't have per-env bounding boxes. This is left over from the homogeneous placement code-branch, which we have now removed. The problem with this is that it's making the code quite difficult to follow because there's lot of branching, even though that branching is not used in production (from what I can tell).

I would suggest that we complete the transition. We _always_ extract per-env bounding boxes, even in the homogeneous case where such bounding boxes are the same across all environments, and we remove all code paths that switch on the non-presence of per-env bounding boxes.

I would also recommend that we abstract the handling of per-env bounding boxes into a class, and move it out of the object placer. Right now we have the same informatoin formatted 3 ways, with similar names, being passed around the code (also incorrectly called "override"). If we group these in a class with different ways of accessing the information, this will simplify matters.

isaaclab_arena/assets/object_set.py
zhx06: Addressed. Now it uses `if else`

isaaclab_arena/assets/object_set.py
zhx06: Yeah, we remove this from the constructor argument

isaaclab_arena/assets/object_set.py
zhx06: Doctring added

isaaclab_arena/relations/object_placer.py
zhx06: Addressed. We removed the optional `bbox_overrides=None`

isaaclab_arena/relations/object_placer.py
zhx06: Addressed. We removed the unused `child_bboxes=None` branch

isaaclab_arena/relations/object_placer.py
zhx06: Addressed as above

isaaclab_arena/relations/object_placer.py
zhx06: Addressed. We removed it.

isaaclab_arena/relations/object_placer.py
zhx06: We have removed the `overrides`.

isaaclab_arena/relations/object_placer.py
zhx06: This helper has been removed, so the function is no longer needed.

isaaclab_arena/relations/object_placer.py
zhx06: Addressed.  `_resolve_bbox()` is removed entirely and ObjectPlacer  always uses computed env bboxes

isaaclab_arena/relations/object_placer.py
zhx06: Addressed by moving bbox construction into `_build_candidate_bboxes()`.

isaaclab_arena/relations/object_placer.py
zhx06: Function renamed and the bboxes are now the default placement input.

isaaclab_arena/relations/object_placer.py
zhx06: Addressed. We have rewritten it as explicit for loops.

isaaclab_arena/relations/object_placer.py
zhx06: Addressed

isaaclab_arena/assets/object_set.py
zhx06: For refills, we only want new placements, not new object variants. The warning at this time might be noisy and not informative

isaaclab_arena/assets/object_set.py
zhx06: Addressed. inline comment added.

isaaclab_arena/relations/object_placer.py
zhx06: This was discussed in the MR for homogeneous object-placement. Lots of code outside use single placement result. I would suggest having a separate MR for this

isaaclab_arena/relations/object_placer.py
zhx06: Addressed. We now have a new dataclass `PerEnvBoundingBoxes` for this.

isaaclab_arena/relations/object_placer.py
zhx06: Addressed.

isaaclab_arena/relations/bounding_box_helpers.py
zhx06: Addressed. We do not need this to stay as a local import.

isaaclab_arena/relations/bounding_box_helpers.py
```diff
@@ -0,0 +1,110 @@
...
+        for obj, bbox in self.object_bboxes.items():
+            assert (
+                bbox.min_point.shape[0] == self.num_envs
+            ), f"Object '{obj.name}' bbox min_point has {bbox.min_point.shape[0]} envs, expected {self.num_envs}."
+            assert (
+                bbox.max_point.shape[0] == self.num_envs
+            ), f"Object '{obj.name}' bbox max_point has {bbox.max_point.shape[0]} envs, expected {self.num_envs}."
+
+    def env_bboxes(self, env_id: int) -> dict[ObjectBase, AxisAlignedBoundingBox]:
+        """Return one-env bboxes for initialization and validation."""
```
alexmillane: Could we add to these functions an explanation of the shape of the returns? (I know that the shape is actually hidden inside `AxisAlignedBoundingBox` but I feel that that would be useful anyway).

isaaclab_arena/relations/bounding_box_helpers.py
```diff
@@ -0,0 +1,110 @@
...
+        assert self.num_envs >= 1, f"num_envs must be >= 1, got {self.num_envs}"
+        for obj, bbox in self.object_bboxes.items():
+            assert (
+                bbox.min_point.shape[0] == self.num_envs
+            ), f"Object '{obj.name}' bbox min_point has {bbox.min_point.shape[0]} envs, expected {self.num_envs}."
+            assert (
+                bbox.max_point.shape[0] == self.num_envs
+            ), f"Object '{obj.name}' bbox max_point has {bbox.max_point.shape[0]} envs, expected {self.num_envs}."
+
+    def env_bboxes(self, env_id: int) -> dict[ObjectBase, AxisAlignedBoundingBox]:
```
alexmillane: Suggestion to make method names more verbose is general. Don't worry too much about horizontal space

-> `get_bounding_boxes_for_env_id`

isaaclab_arena/relations/bounding_box_helpers.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def solver_candidate_bboxes(self, candidates_per_env: int) -> dict[ObjectBase, AxisAlignedBoundingBox]:
+        """Return bboxes expanded to match candidate rows passed to the solver."""
+        return {
+            obj: AxisAlignedBoundingBox(
+                min_point=bbox.min_point.repeat_interleave(candidates_per_env, dim=0),
+                max_point=bbox.max_point.repeat_interleave(candidates_per_env, dim=0),
+            )
+            for obj, bbox in self.object_bboxes.items()
+        }
```
alexmillane: See comment above.

Thank you for separating these functions into a class that's a great change that cleans up the placer.

One further comment is to add a longer explanation in these three functions about the format of the output. The problem is that the returned data is all very similar, so it's a bit tough to know what's going on.

alexmillane (approved):
Thanks for addressing all the comments. This is looking really good.

I have made a couple of comments about the `PerEnvBoundingBoxes`. I'm still a bit confused about the format of the various outputs. So I'd suggest expanding the comments there (the implementation is good now. thanks for putting it into a separate class).

I've approved so merge when you've addressed the above :)

PR #677, opened 2026-05-13
Add order-does-not-matter composite tasks

peterd-NV:
## Summary
This PR refactors composite tasks to create two classes:

1. A new `CompositeTaskBase` class that enables "order-does-not-matter" subtask completion. Mostly copied over from existing SequentialTaskBase.
2. A refactored `SequentialTaskBase` that is dervied from CompositeTaskBase for "order-does-matter" tasks.

Tests are updated as follows:

1. Test cases for both TaskBase classes are updated accordingly.
2. A new runtime test for CompositeTaskBase is added. 

Other updates include:

1. Added support for "don't cares" when specifying the `desired_subtask_success_state` by setting subtask value to `None`.

API usage of composite tasks remain the same as before.


isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -170,11 +60,17 @@ def sequential_task_success_func(
                 if current_subtask_idx < len(subtasks) - 1:
                     env._current_subtask_idx[env_idx] += 1
```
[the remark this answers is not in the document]
peterd-NV: fixed

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -0,0 +1,343 @@
...
+
+        Args:
+            recorded_metric_data: List of arrays, each shape (num_subtasks,) with bool values.
+
+        Returns:
+            List of success rates for each subtask.
+        """
+        num_demos = len(recorded_metric_data)
+        if num_demos == 0:
+            return [0.0]
```
alexmillane, line 86: thanks 😅 . I notice some of our other rates return Nan. (my fault).

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -0,0 +1,343 @@
...
+    """
+    A base class for composite tasks composed of multiple subtasks.
+    Completion ordering of subtasks does not matter.
+    """
+
+    def __init__(
+        self,
+        subtasks: list[TaskBase],
+        episode_length_s: float | None = None,
+        desired_subtask_success_state: list[bool | None] | None = None,
```
alexmillane, line 116: The meaning of this argument is not immediately obvious to me. Are there cases where we'd like to build a composite task, but that we don't want some of the task to be successful?

Suggestion to add a short docstring.

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -0,0 +1,343 @@
...
+
+        # Check success of subtask for each env
+        for subtask_idx in range(len(subtasks)):
+            subtask_success_func = subtasks[subtask_idx].get_termination_cfg().success.func
+            subtask_success_params = subtasks[subtask_idx].get_termination_cfg().success.params
+            results = subtask_success_func(env, **subtask_success_params)
+            for env_idx in range(env.num_envs):
+                if results[env_idx]:
+                    current_subtask_success_state[env_idx][subtask_idx] = True
+                    env._subtask_success_state[env_idx][subtask_idx] = True
```
alexmillane: What's the difference between these two variables? I guess they must represent different things, but it's not clear from the naming. Suggestion to a) rename to make things clearer, and b) add a comment if "a)" is not possible.

alexmillane: Hmm I think I see. One is if a sub-task has ever been successful and the other is if it is successful right now. Suggestion to rename :)

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -0,0 +1,343 @@
...
+        # at some point and (b) currently match the desired value.
+        if desired_subtask_success_state is not None:
+            per_env_success = [
+                all(
+                    env._subtask_success_state[env_idx][i] and current_subtask_success_state[env_idx][i] == desired
+                    for i, desired in enumerate(desired_subtask_success_state)
+                    if desired is not None
+                )
+                for env_idx in range(env.num_envs)
+            ]
```
alexmillane: Personally, and maybe I'm not Pythonic enough, I would break this nested list composition into for loops. It's quite hard to interpret what's happening here. I can't quite tell. I suspect it would be easier if this was formatted as a raw loop.

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -0,0 +1,343 @@
...
+        else:
+            for env_id in env_ids:
+                env._subtask_success_state[env_id] = [False for _ in subtasks]
+
+    def get_scene_cfg(self) -> Any:
+        "Make combined scene cfg from all subtasks."
+        # Check for duplicate fields across subtask scene configs and warn if found
+        duplicates = check_configclass_field_duplicates(*(subtask.get_scene_cfg() for subtask in self.subtasks))
+        if duplicates:
+            import warnings
```
alexmillane: Any reason to import this locally? Suggestion to move this to the top if not

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -0,0 +1,343 @@
...
+        "Make combined events cfg from all subtasks."
+        # Collect events_cfgs from subtasks with renamed fields to avoid collisions
+        renamed_events_cfgs = []
+        for i, subtask in enumerate(self.subtasks):
+            subtask_events_cfg = subtask.get_events_cfg()
+            renamed_cfg = transform_configclass_instance(
+                subtask_events_cfg, partial(self.add_suffix_configclass_transform, suffix=f"_subtask_{i}")
+            )
+            if renamed_cfg is not None:
+                renamed_events_cfgs.append(renamed_cfg)
```
alexmillane: Am I correct that `renamed_cfg` is `None` if and only if `subtask_events_cfg`?

If that's true could we move the condition to test `subtask_events_cfg` for `None`ness. Right now it reads as if something valid could come out of `get_events_cfg` and then fail the transform and not make it into the resulting combine config. That would be problematic I think. The resulting composite task would be missing some events due to a renaming issue.

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -0,0 +1,343 @@
...
+        "Make combined termination cfg from all subtasks."
+        # Collect termination cfgs from subtasks with 'success' field removed
+        subtask_termination_cfgs = []
+        for subtask in self.subtasks:
+            termination_cfg = subtask.get_termination_cfg()
+            cleaned_cfg = transform_configclass_instance(
+                termination_cfg, partial(self.remove_configclass_transform, exclude_fields={"success"})
+            )
+            if cleaned_cfg is not None:
+                subtask_termination_cfgs.append(cleaned_cfg)
```
alexmillane, line 314: See comment above. I think that applies here too.

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -0,0 +1,343 @@
...
+        return combined_termination_cfg
+
+    def combine_subtask_metrics(self, subtask_idxs: list[int]) -> list[MetricBase]:
+        "Combine metrics from subtasks with the given ids."
+        combined_metrics = []
+
+        for subtask_idx in subtask_idxs:
+            subtask_metrics = self.subtasks[subtask_idx].get_metrics()
+            for metric in subtask_metrics:
+                if metric.name != "success_rate":
```
alexmillane, line 335: Is it true that although we strip out the subtask success rates, we have some other mechanism for reporting subtask success rate? I guess that's true. I would suggest we expand the docstring of the function or add a comment indicating that (i.e. that a maintainer shouldn't worry about losing the success rate terms because we're tracking them another way).

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -0,0 +1,343 @@
...
+        # Check that all subtasks have the same Mimic eef_names
+        mimic_eef_names = set(self.subtasks[0].get_mimic_env_cfg(arm_mode).subtask_configs.keys())
+
+        for subtask in self.subtasks[1:]:
+            subtask_eef_names_set = set(subtask.get_mimic_env_cfg(arm_mode).subtask_configs.keys())
+            if subtask_eef_names_set != mimic_eef_names:
+                raise ValueError(
+                    f"All subtasks must have the same Mimic eef_names.\nSubtask 0 has eef_names: {mimic_eef_names}, but"
+                    f" subtask {self.subtasks.index(subtask)} has eef_names: {subtask_eef_names_set}."
+                )
```
alexmillane: Suggestion to move out to another function.

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -0,0 +1,343 @@
...
+    def get_mimic_env_cfg(self, arm_mode: ArmMode) -> MimicEnvCfg:
+        "Get the Mimic environment configuration for the sequential task."
+        mimic_env_cfg = MimicEnvCfg()
+
+        # Assign all default config values to mimic_env_cfg.datagen_config
+        for key, value in MIMIC_DATAGEN_CONFIG_DEFAULTS.items():
+            setattr(mimic_env_cfg.datagen_config, key, value)
+
+        mimic_env_cfg.subtask_configs = self.combine_mimic_subtask_configs(arm_mode)
+        return mimic_env_cfg
```
alexmillane, line 400: One thing about this class and (unfortunately, the framework), is that we have been inconsistent with the naming of functions with leading underscores. I feel that some of these functions should have leading underscores to indicate that they're internal functions... But admittedly we've not done that consistently, so up to you.

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -106,37 +22,11 @@ class SequentialTaskBase(TaskBase):
...
-        "Config transformation to remove all fields in an exclude set."
-        return [(name, ftype, value) for name, ftype, value in fields if name not in exclude_fields]
-
     @staticmethod
-    def sequential_task_success_func(
+    def composite_task_success_func(
         env,
         subtasks: list[TaskBase],
-        desired_subtask_success_state: list[bool] | None,
+        desired_subtask_success_state: list[bool | None] | None,
```
alexmillane, line 29: Suggestion to add an expanded docstring to indicate the meaning of this variable.

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -148,33 +38,40 @@ def sequential_task_success_func(
...
+            subtasks_to_evaluate = sorted(set(env._current_subtask_idx))
+
+        # Compute the success state for the subtasks
+        for subtask_idx in subtasks_to_evaluate:
+            subtask_success_func = subtasks[subtask_idx].get_termination_cfg().success.func
+            subtask_success_params = subtasks[subtask_idx].get_termination_cfg().success.params
+            results = subtask_success_func(env, **subtask_success_params)
+            for env_idx in range(env.num_envs):
+                if results[env_idx]:
+                    current_subtask_success_state[env_idx][subtask_idx] = True
```
alexmillane: This is a (near) duplicate of a loop used in the base class. Could we move this to a shared function?

isaaclab_arena/tests/test_composite_open_door.py
```diff
@@ -0,0 +1,379 @@
...
+            super().__init__(subtasks=subtasks, episode_length_s=episode_length_s)
+
+        def get_metrics(self):
+            return []
+
+        def get_prompt(self):
+            return ""
+
+        def get_mimic_env_cfg(self, embodiment_name: str):
+            return None
```
alexmillane: Do we need to subclass `CompositeTaskBase` here?

We don't really add anything to the base class. This structure makes constructing a composite task quite verbose.

Can we not do `CompositeTaskBase([subtask_1, subtask_2])`?

Also with the current approach we can do

```
pnp_1  = PickAndPlaceTask(apple, bowl)
pnp_2 = PickAndPlaceTask(banana, bowl)

open_door_composite = OpenDoorCompositeTask([pnp_1, pnp_2])
```

alexmillane (approved):
Looks good. A few minor comments.

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: Added a docstring to explain. It's an optional argument that can be used to specify the precise subtask success value at the end of the episode. When not provided, once a subtask flips True, it is deemed "completed" for the rest of the episode regardless of it's current success state is True or False at the end of the episode. This is desirable for cases when multiple subtask successes should not be true at the same time (e.g. bottle grasped, bottle in fridge)

This argument gives the user the flexibility to enforce when subtask successes "should" be true (e.g. bottle in fridge, fridge door closed).

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: Updated the naming to "subtask_ever_succeeded" and "subtask_currently_succeeding" to be more clear.

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: moved to the top

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: yes, renamed_cfg == None iff subtask_events_cfg == None. I've updated the structure to do check none-ness check on subtask_events_cfg instead and assert that renamed_cfg is not none.

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: This case is slightly different as cleaned_cfg can be None if the termination_cfg was not empty but only has "success". So here we do want to keep the guard as is. 

I've added a comment to explain.

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: Yes, that is correct. Added a doc string to explain that this is intentional and we shouldn't worry about the other success rate terms.

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: done

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: Did a cleanup and renamed relevant functions with underscores

isaaclab_arena/tasks/sequential_task_base.py
peterd-NV: done

isaaclab_arena/tasks/sequential_task_base.py
peterd-NV: done

isaaclab_arena/tests/test_composite_open_door.py
peterd-NV: Yea, it's not necessary and was originally added like that to be more clear. I've gone ahead and removed the subclassing from both composite/sequential tests

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: Separated it out to nested for loops for clarity

PR #678, opened 2026-05-13
Update main with fixes on release 0.2.1 branch:  G1 static apple-to-plate workflow 

pulkitg01:
## Summary

This cherry-picks the G1 static apple-to-plate workflow fixes from release 0.2.1 branch into main.

Changes:
- Adds explicit apple and plate test scales so the static pick-and-place test uses realistic object geometry.
- Adds an invisible `procedural_table` support patch under the Galileo static pick-and-place shelf area.

## Motivation

The imported Galileo locomanip shelf mesh appears to have uneven or perforated collision in part of the task region. Small objects like the apple or plate can fall through the visible shelf surface depending on where they spawn.

The support patch gives the task objects a clean collision surface without changing the visible scene geometry.

### DIFF
--- a/isaaclab_arena_environments/galileo_g1_static_pick_and_place_environment.py
+++ b/isaaclab_arena_environments/galileo_g1_static_pick_and_place_environment.py
@@ -44,6 +44,7 @@ if TYPE_CHECKING:
 #   shelf on the first sim tick (which would otherwise launch them upward).
 SHELF_SURFACE_Z = -0.030
 SHELF_AIRGAP = 0.005
+SHELF_SUPPORT_PATCH_CENTER = (0.62, 0.0, SHELF_SURFACE_Z - 0.02)
 
 # Object XY spawn pose (env-local frame, shelf-relative). X mirrors the locomanip env
 # (the only on-shelf X we have ground-truth data for via the brown_box flow). The pickup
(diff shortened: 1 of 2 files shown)


isaaclab_arena_environments/galileo_g1_static_pick_and_place_environment.py
```diff
@@ -135,6 +136,13 @@ def get_env(self, args_cli: argparse.Namespace) -> IsaacLabArenaEnvironment:
         # Reuse the locomanip background USD: it bakes in lighting and provides the same
         # shelf-in-front-of-robot geometry the locomanip env was tuned against.
         background = self.asset_registry.get_asset_by_name("galileo_locomanip")()
+        # The imported shelf mesh has uneven/perforated collision in the task region:
```
xyao-nv, line 140: Can you comment here running on CPU does not show the problem?

xyao-nv (approved):

isaaclab_arena_environments/galileo_g1_static_pick_and_place_environment.py
```diff
@@ -44,6 +44,7 @@
 #   shelf on the first sim tick (which would otherwise launch them upward).
 SHELF_SURFACE_Z = -0.030
 SHELF_AIRGAP = 0.005
+SHELF_SUPPORT_PATCH_CENTER = (0.62, 0.0, SHELF_SURFACE_Z - 0.02)
```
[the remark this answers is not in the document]
pulkitg01: Done.

isaaclab_arena_environments/galileo_g1_static_pick_and_place_environment.py
pulkitg01: Done.

steple (approved):

PR #679, opened 2026-05-13
Deterministic Object Placement

zhx06:
## Summary
Deterministic Object Placement

## Detailed description
- Add seed and random number generator to make optimized layouts deterministic for both homogeneous and heterogeneous object placement
- Add several related tests to guarantee the correctness

### DIFF
--- a/isaaclab_arena/relations/pooled_object_placer.py
+++ b/isaaclab_arena/relations/pooled_object_placer.py
@@ -67,8 +67,8 @@ class PooledObjectPlacer:
     * sample_for_envs(env_ids) consumes one layout for each requested
       absolute env id (used for partial resets).
     * sample_with_replacement(count) is non-consuming. Env-specific layouts
-      are sampled from matching env slots; reusable layouts are sampled IID from
-      all stored layouts.
+      are sampled from matching env slots; reusable layouts stay marginally
+      uniform over all stored layouts.
 
     Args:
         objects: All objects (including anchors) participating in relation solving.
@@ -102,7 +102,7 @@ class PooledObjectPlacer:
         self._base_placement_seed = placer_params.placement_seed
         self._next_seed_offset = 0
         # Per-env sampling RNG keyed by (placement_seed, env_id): env i's draws are reproducible
-        # and independent of other envs. placement_seed=None falls back to system entropy.
+        # and independent of other envs.
         self._env_rngs = get_rngs(self._num_envs, placer_params.placement_seed)
         self._env_pools: list[EnvLayoutPool] = [EnvLayoutPool([]) for _ in range(self._num_envs)]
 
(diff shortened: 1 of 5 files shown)


isaaclab_arena/relations/pooled_object_placer.py
```diff
@@ -115,6 +118,13 @@ def __init__(
     # Pool storage internals
     # ------------------------------------------------------------------
 
+    def _build_env_rngs(self, base_seed: int | None) -> list[random.Random]:
+        """Build one sampling RNG per env, keyed by (base_seed, env_id); None seeds from system entropy."""
+        if base_seed is None:
+            return [random.Random() for _ in range(self._num_envs)]
+        seeder = random.Random(base_seed)
+        return [random.Random(seeder.getrandbits(64)) for _ in range(self._num_envs)]
```
alexmillane: Can we use if..else rather than this early return construction?

isaaclab_arena/relations/pooled_object_placer.py
```diff
@@ -401,17 +411,20 @@ def sample_with_replacement(self, count: int) -> list[PlacementResult]:
...
                 pool = self._env_pools[cur_env].layouts
                 assert pool, f"Env {cur_env} has no valid layouts to sample from."
-                results.append(random.choice(pool))
+                results.append(self._env_rngs[cur_env].choice(pool))
             return results
+        # Reusable layouts are interchangeable, so each slot stays marginally uniform over the full
+        # pool; per-env RNGs only fix which stream a slot draws from, for parity with the env-specific branch.
         all_layouts = [layout for pool in self._env_pools for layout in pool.layouts]
-        return random.choices(all_layouts, k=count)
+        return [self._env_rngs[i % self._num_envs].choice(all_layouts) for i in range(count)]
```
alexmillane: nit: Could we split these list comprehensions over multiple lines. They're getting really hard to interpret quickly.

The more normal equivalent is:
```
# Serialize all layouts.
all_layouts = []
for pool in self._env_pools:
    for layout in pool.layouts:
         all_layouts.append(layout)
# Select random layouts from the serialized list
for layout_idx in range(count):
    rng = self._env_rngs[layout_idx % self._num_envs]
    rng.choose(all_layouts)
```
which is much easier to interpret.

isaaclab_arena/relations/pooled_object_placer.py
alexmillane: Could we move this to `isaaclab_arena/utils/random.py` and rename

```
get_rngs(num: int, base_seed: int | None) -> List[random.Random]:
    ...
```

alexmillane (approved):
A couple of nits. Feel free to merge once addressed.

isaaclab_arena/relations/pooled_object_placer.py
zhx06: Done — the RNG builder is now `get_rngs` in `isaaclab_arena/utils/random.py` and uses an explicit `if/else` instead of the early return.

isaaclab_arena/relations/pooled_object_placer.py
zhx06: Done — the reusable branch now builds `all_layouts` and `results` with explicit `for` loops instead of the nested comprehensions.

isaaclab_arena/relations/pooled_object_placer.py
zhx06: Done — moved to `isaaclab_arena/utils/random.py` with the exact signature `get_rngs(num: int, base_seed: int | None)`. Also added direct unit tests for it in `tests/test_random.py`.

zhx06: Done. It is now in `utils/random.py` and uses an explicit if...else

zhx06: Done. It is moved to `isaaclab_arena/utils/random.py`

isaaclab_arena/relations/pooled_object_placer.py
zhx06: Done. It is now in the style you suggested.

PR #688, opened 2026-05-20
Support parallel envs in Pi0RemotePolicy

cvolkcvolk:
## Summary

Arena rollouts can now request ``num_envs > 1`` instead of being restricted to single-env.

- Loop over envs on the Arena client side, with a per-env action-chunk cache and step counter (**openpi's format is still one obs per request**). 
- ``reset(env_ids)`` clears only the listed envs' caches, so per-env termination leaves untouched envs replaying uninterrupted.


alexmillane (approved):
Clean! Thanks!

PR #691, opened 2026-05-21
Add Agent Skills library and refocus AGENTS.md

cvolkcvolk:
## Summary

Starts an Agent Skills library for Arena under `.agents/skills/` with a `.claude/skills/` symlink for Claude Code compatibility, and refocuses `AGENTS.md`.

Three skills introduced:

- **`dev-container`** — Docker container lifecycle (build, start, exec, the `python` aliasing).
- **`run-tests`** — three-phase pytest suite with a `smoke` argument for fast install verification.
- **`commit-and-pr`** — DCO sign-off, branch naming, no-AI-attribution, `main` as base, new-commits-on-feedback rule. `disable-model-invocation: true` so it stays user-initiated.

`AGENTS.md` procedures move into skills, and new sections (Project / Repository layout / Boundaries with paired Don't/Do) cover the standing context that recent AGENTS.md best practices recommend.

## Test plan
On a fresh clone and newly spawned agent:
- [x] Prompt via `"Lets set up arena"` installs the container, checks mounted repo and installs pre-commit hook
- [x] skills are listed under `/skills`
- [x] `/dev-container` brings the container up; `docker exec ... import isaaclab_arena` resolves.
- [x] `/run-tests` (no argument) runs all three phases in order.
- [ ] Codex discovers all three skills from `.agents/skills/` without the symlink.
- [x] Claude Code discovers all three skills via the `.claude/skills/` symlink.

## References

- [A good AGENTS.md is a model upgrade — Augment Code, Apr 22 2026](https://www.augmentcode.com/blog/how-to-write-good-agents-dot-md-files)
- [How to Build Your AGENTS.md (2026) — Augment Code](https://www.augmentcode.com/guides/how-to-build-agents-md)

### DIFF
--- a/.agents/skills/commit-and-pr/SKILL.md
+++ b/.agents/skills/commit-and-pr/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: commit-and-pr
-description: Creates Arena-conformant commits and pull requests. Enforces DCO sign-off (git commit -s), no AI attribution lines (no Co-Authored-By, no Generated-with-Claude footers), <username>/<feature-desc> branch naming, separate new commits rather than --amend when iterating on PR feedback, and main as the default base branch. Use when the user asks to commit, stage changes, push the current branch, or open a pull request.
+description: Creates Arena-conformant commits and pull requests. Enforces DCO sign-off (git commit -s), no AI attribution lines (no Co-Authored-By, no Generated-with-Claude footers), <username>/<feature-desc> branch naming, separate new commits rather than --amend when iterating on PR feedback, and main as the default base branch. Use when the user asks to commit, stage changes, push the current branch, open a pull request, make a PR, submit changes, or mark work as ready to merge.
 disable-model-invocation: true
 allowed-tools: Bash(git *) Bash(pre-commit *) Bash(gh pr create *) Bash(gh pr view *)
 ---
--- a/.agents/skills/commit-and-pr/SKILL.md
+++ b/.agents/skills/commit-and-pr/SKILL.md
@@ -40,7 +40,7 @@ EOF
 
 ## Branch
 
-Branch naming: `<username>/<feature-desc>` — e.g. `cvolk/feature-video-recording`, `cvolk/refactor-no-unwrap`, `mimil/create-arena-external-workflow-skill`.
+Branch naming: `<username>/<feature-desc>` with a kebab-case feature description — e.g. `<username>/feature-video-recording`, `<username>/refactor-no-unwrap`.
 
 Do not use `feature/...`, `fix/...`, or `docs/...` prefixes.
 
--- a/.agents/skills/commit-and-pr/SKILL.md
+++ b/.agents/skills/commit-and-pr/SKILL.md
@@ -2,7 +2,7 @@
 name: commit-and-pr
 description: Creates Arena-conformant commits and pull requests. Enforces DCO sign-off (git commit -s), no AI attribution lines (no Co-Authored-By, no Generated-with-Claude footers), <username>/<feature-desc> branch naming, separate new commits rather than --amend when iterating on PR feedback, and main as the default base branch. Use when the user asks to commit, stage changes, push the current branch, or open a pull request.
 disable-model-invocation: true
-allowed-tools: Bash(git *) Bash(gh pr create *) Bash(gh pr view *)
+allowed-tools: Bash(git *) Bash(pre-commit *) Bash(gh pr create *) Bash(gh pr view *)
 ---
 
 # Commit and Pull Request
--- a/.agents/skills/commit-and-pr/SKILL.md
+++ b/.agents/skills/commit-and-pr/SKILL.md
@@ -1,0 +1,92 @@
+---
+name: commit-and-pr
+description: Creates Arena-conformant commits and pull requests. Enforces DCO sign-off (git commit -s), no AI attribution lines (no Co-Authored-By, no Generated-with-Claude footers), <username>/<feature-desc> branch naming, separate new commits rather than --amend when iterating on PR feedback, and develop as the default base branch. Use when the user asks to commit, stage changes, push the current branch, or open a pull request.
+disable-model-invocation: true
+allowed-tools: Bash(git *) Bash(gh pr create *) Bash(gh pr view *)
+---
+
+# Commit and Pull Request
+
+Creates commits and PRs that follow Arena's contributor guidelines.
+
+## Commit
+
+1. Run the `pre-commit-check` skill (or `pre-commit run --all-files`) until the working tree is clean. Do not commit through hook failures.
+2. Stage **specific files** rather than using `git add -A` — this avoids accidentally committing dotfiles, credentials, or large untracked artifacts.
+3. Sign off the commit (DCO is required by Arena):
+
+   ```bash
+   git commit -s -m "<subject>"
+   ```
+
+4. **Subject line:** imperative mood, ~50 characters, no trailing period.
+   - Good: `Fix attribute access on wrapped env`
+   - Bad: `Fixed the attribute access on the wrapped env.` (past tense, trailing period)
+5. **Body** (when needed): blank line after the subject, wrap at 72 chars, explain *what* and *why* — not *how* (the diff shows that).
+6. **No AI attribution lines.** Do not include `Co-Authored-By: Claude...`, `Generated with Claude Code`, or similar footers. The DCO sign-off is the only required trailer.
+
+Use a heredoc for multi-paragraph messages so quoting stays clean:
+
+```bash
+git commit -s -m "$(cat <<'EOF'
+<subject>
+
+<body paragraph 1>
+
+<body paragraph 2>
+EOF
+)"
+```
+
+## Branch
+
+Branch naming: `<username>/<feature-desc>` — e.g. `cvolk/feature-video-recording`, `cvolk/refactor-no-unwrap`, `mimil/create-arena-external-workflow-skill`.
+
+Do not use `feature/...`, `fix/...`, or `docs/...` prefixes.
+
+## Pull request
+
+1. Push the branch (set upstream on first push):
+
+   ```bash
+   git push -u origin <branch-name>
+   ```
+
+2. Open the PR against `develop` (the active branch — `main` is under SQA and only receives fixes):
+
+   ```bash
+   gh pr create --base develop \
+     --title "<short imperative subject>" \
+     --body "<summary + test plan>"
+   ```
+
+3. PR title follows the same rules as a commit subject: imperative, ~70 chars max, no trailing period.
+
+## Iterating on review feedback
+
+When addressing review comments, **add new commits**. Do not `--amend` and force-push, because reviewers need to see each round of changes as a discrete commit.
+
+```bash
+# good
+git commit -s -m "Address review: rename foo to bar"
+git push
+
+# avoid
+git commit --amend
+git push --force
+```
+
+The exception: if a single commit is *purely* a typo fix in your own as-yet-unmerged work and no one has reviewed it, `--amend` is fine.
+
+## Verify
+
+Before declaring the work done:
+
+```bash
+gh pr view --json title,baseRefName,statusCheckRollup
+```
+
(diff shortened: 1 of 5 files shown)


.agents/skills/run-tests/SKILL.md
```diff
@@ -0,0 +1,66 @@
...
+name: run-tests
+description: Runs the Isaac Lab-Arena pytest suite across its three required phases (no-cameras, with-cameras, with-subprocess) inside the running isaaclab_arena Docker container. Use when the user asks to run the tests, test their changes, verify the test suite passes, check whether a change broke anything (e.g. "did I break anything", "do the tests still pass", "is everything still working"), smoke-test the install, run smoke tests, or run coverage for a specific module.
+argument-hint: "[smoke]"
+allowed-tools: Bash(docker exec *)
+---
+
+# Run Tests
+
+**Default (no argument):** run all three phases in order.
+**`smoke`:** Phase 1 only (no-cameras, no-subprocess) — the fast install / sanity check used in the first-touch journey.
```
cvolkcvolk: This is actually not a smoke test but runs for quite a while. We currently state this test to verify the installation of Arena in our [Installation Docs](https://isaac-sim.github.io/IsaacLab-Arena/main/pages/quickstart/installation.html). 
Shall we add a simple _import succeeds_ test to verify the installation/ or something similar? @alexmillane @xyao-nv

cvolkcvolk: Im now just not calling this smoke test anymore here for now

.agents/skills/commit-and-pr/SKILL.md
```diff
@@ -0,0 +1,91 @@
...
+
+# Commit and Pull Request
+
+Creates commits and PRs that follow Arena's contributor guidelines.
+
+## Commit
+
+1. Run the `pre-commit-check` skill (or `pre-commit run --all-files`) until the working tree is clean. Do not commit through hook failures.
+2. Stage **specific files** rather than using `git add -A` — this avoids accidentally committing dotfiles, credentials, or large untracked artifacts.
+3. Sign off the commit (DCO is required by Arena):
```
alexmillane: I'm at a bit of a loss what to do here? I'm not sure our other (human) commits are being signed off 😅

.agents/skills/commit-and-pr/SKILL.md
```diff
@@ -0,0 +1,91 @@
...
+
+   ```bash
+   git commit -s -m "<subject>"
+   ```
+
+4. **Subject line:** imperative mood, ~50 characters, no trailing period.
+   - Good: `Fix attribute access on wrapped env`
+   - Bad: `Fixed the attribute access on the wrapped env.` (past tense, trailing period)
+5. **Body** (when needed): blank line after the subject, wrap at 72 chars, explain *what* and *why* — not *how* (the diff shows that).
```
alexmillane, line 25: Shall we update to include our template? I.e. one summary bullet and 2-5 details bullets? We can actually point the agent to the template file [here](https://github.com/isaac-sim/IsaacLab-Arena/blob/main/.github/pull_request_template.md)

Personally I find the (default) agent PR messages way to verbose, so I feel that these guidelines will actually be helpful. Perhaps we should also put these guidelines in the `AGENTS.md` - a lot (probably most) PRs will be created with coding agents not using the skills.

.agents/skills/commit-and-pr/SKILL.md
```diff
@@ -0,0 +1,91 @@
...
+<body paragraph 2>
+EOF
+)"
+```
+
+## Branch
+
+Branch naming: `<username>/<feature-desc>` with a kebab-case feature description — e.g. `<username>/feature-video-recording`, `<username>/refactor-no-unwrap`.
+
+Do not use `feature/...`, `fix/...`, or `docs/...` prefixes.
```
alexmillane: I think most of us are using `<username>/<type>/<description>

where type is one of (feature, fix, docs, etc.)

It's funny, telling an agent this stuff, forces you to come to an agreement.

.agents/skills/commit-and-pr/SKILL.md
```diff
@@ -0,0 +1,91 @@
...
+When addressing review comments, **add new commits**. Do not `--amend` and force-push, because reviewers need to see each round of changes as a discrete commit.
+
+```bash
+# good
+git commit -s -m "Address review: rename foo to bar"
+git push
+
+# avoid
+git commit --amend
+git push --force
```
alexmillane, line 92: great! agreed!

.agents/skills/dev-container/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+./docker/run_docker.sh
+```
+
+Idempotent: builds the image if it does not exist, starts the container if it is not running, then attaches.
+
+## Common flag combinations
+
+| Flag | Purpose |
+|---|---|
+| `-r` | Force image rebuild |
```
alexmillane, line 23: I believe there's another flag `-R` that rebuilds without cache?

.agents/skills/dev-container/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+| `-g` | Include GR00T N1.6 dependencies |
```
alexmillane: I guess this will updating soon 😅. Which raises another question, how do we keep these skills in line with the code (i.e. the truth). I wonder if occasionally we can introduce another skill that checks the skills for validity. i.e. "check that dev-container skill describes the current set of flags in `run_docker.sh`).

I wonder if we could actually run such a skill as a nightly and let the agent decide if it passes or fails. This might make a start on our agentic VDR goal.

.agents/skills/run-tests/SKILL.md
```diff
@@ -0,0 +1,68 @@
+---
+name: run-tests
+description: Runs the Isaac Lab-Arena pytest suite across its three required phases (no-cameras, with-cameras, with-subprocess) inside the running isaaclab_arena Docker container. Use when the user asks to run the tests, test their changes, verify the test suite passes, check whether a change broke anything (e.g. "did I break anything", "do the tests still pass", "is everything still working"), run only one specific phase (e.g. "run Phase 1", "run only the no-camera tests", "skip the camera and subprocess tests"), or run coverage for a specific module.
```
alexmillane, line 3: Dumb question, this description indicates that an agent has access to the skill without the user explicitly calling it. Perhaps the desriptions are loaded into the context for all agents running in the repo?

.agents/skills/run-tests/SKILL.md
```diff
@@ -0,0 +1,68 @@
...
+---
+
+# Run Tests
+
+**Default (no argument):** run all three phases in order.
+**`phase1`:** no-cameras, no-subprocess only.
+**`phase2`:** with-cameras only.
+**`phase3`:** with-subprocess only.
+
+The three phases use mutually exclusive pytest markers and must be invoked separately. For a full pre-PR run, every phase must pass before moving on to the next.
```
alexmillane: is it true that the _markers_ are mutually exclusive? I think not, otherwise we wouldn't have to use multiple tag based filtering. The sets of tests we end up running are mutually exclusive though.

AGENTS.md
```diff
@@ -2,91 +2,40 @@
 
 This file provides guidance to AI coding agents (Claude Code, OpenAI Codex, etc.) when working with code in this repository.
 
-## Docker Environment
+## Project
 
-All commands (tests, linting, training scripts, etc.) must be run inside the Docker container. The default container is `isaaclab_arena-latest`, started via:
+Isaac Lab-Arena is a composable environment-creation and policy-evaluation library for robotics simulation, built on Isaac Sim 5.1 and Isaac Lab 2.3. Status: alpha (`v0.2.x`); APIs are unstable. `main` is the active development branch.
```
alexmillane: Isaac Sim 6.0 and Isaac Lab 3.0 Beta

AGENTS.md
```diff
@@ -2,91 +2,40 @@
...
-/isaac-sim/python.sh -m pytest -sv -m "not with_cameras and not with_subprocess" isaaclab_arena/tests
-/isaac-sim/python.sh -m pytest  -sv -m "with_cameras and not with_subprocess" isaaclab_arena/tests
-/isaac-sim/python.sh -m pytest  -sv -m "with_subprocess" isaaclab_arena/tests
-
-# Run a single test file
-/isaac-sim/python.sh -m pytest isaaclab_arena/tests/test_asset_registry.py
-
-# Run a specific test function
-/isaac-sim/python.sh -m pytest isaaclab_arena/tests/test_asset_registry.py::test_default_assets_registered
+All commands (tests, linting, training scripts) must run inside the `isaaclab_arena-latest` Docker container. The repo root is mounted at `/workspaces/isaaclab_arena`. Inside the container, `python` is aliased to `/isaac-sim/python.sh` — prefer the explicit path in `docker exec` invocations from outside the container, where the alias is not active.
```
alexmillane: Do we do linting inside the docker? I have to admit, I've been running it outside. However, I think running it inside is a better idea. 

The line above this though, recommends installing `pre-commit` outside.

What shall we do here?

alexmillane (approved):
Great!! Thanks for adding this!

xyao-nv (approved):

.agents/skills/dev-container/SKILL.md
xyao-nv: Let's drop this. It's on my todo list to drop gr00t-deps in our docker.

AGENTS.md
cvolkcvolk: Thanks

AGENTS.md
cvolkcvolk: Good point. We are running pre-commit _outside_ the docker. I think pre-commit is designed to run wherever git runs. And since we ran git from the host-side, this evolved naturally. 

I'll update the description above. Let's stick with host side invocation for now.

.agents/skills/run-tests/SKILL.md
cvolkcvolk: Yea you are right, the markers themselves aren't mutually exclusive. Updated

.agents/skills/run-tests/SKILL.md
cvolkcvolk: Yep: At session start, every session running in this repo loads each skill's name and description from this directory _.agents/skills/_ into the models context. When a phrasing matches a description, the model can auto-invoke the skill. Then, the body below si loaded

cvolkcvolk: or by manually invoking `/run-tests` during a session

.agents/skills/dev-container/SKILL.md
cvolkcvolk: On Xinjies todo list to remove the dependencies. Dropped  the -g flag.

.agents/skills/dev-container/SKILL.md
cvolkcvolk: Updated. Thanks.

.agents/skills/dev-container/SKILL.md
cvolkcvolk: Dropped

.agents/skills/commit-and-pr/SKILL.md
cvolkcvolk: Agree. Let's go with  `<username>/<type>/<short-description>` instead. That's what most of us use already

.agents/skills/commit-and-pr/SKILL.md
cvolkcvolk: Fully agree. It happily writes you a full novel into the PR body

.agents/skills/commit-and-pr/SKILL.md
cvolkcvolk: Mh probably most, if not all of the human commits were never signed off. And arguably we are licenced under Apache 2.0, which doesn't require DCO.

It's currently stated in our `CONTRIBUTING.md` but I feel like we should either strictly enforece this or loosen it to not be required at all.

cvolkcvolk: Ill keep it in for this PR

PR #696, opened 2026-05-21
Honor --enable_cameras in XR teleop

rwiltz:
## Summary
- `teleop.py` unconditionally called `remove_camera_configs(env_cfg)` whenever `args_cli.xr` was true, even with `--enable_cameras` set. `args_cli.xr` is auto-set by openxr teleop devices, so any user combining `--teleop_device openxr` with `--enable_cameras` ended up with camera scene entities stripped but the matching `camera_obs.<cam>_rgb` observation terms still in place. The env failed to construct with `The scene entity 'robot_head_cam' does not exist`.
- Gate the call on `not args_cli.enable_cameras`, mirroring the existing logic in `record_demos.py`.


xyao-nv (approved):

xyao-nv:
Close it as the fix has been merged in #697

PR #699, opened 2026-05-22
Add script to merge teleop HDF5 demo recordings

rwiltz:
## Summary
Merge multi-session teleop HDF5 recordings

## Detailed description
- Operators collect 100+ demos across multiple sittings, producing one HDF5 per session; downstream training and conversion expect a single dataset.
- Adds `merge_demos.py` (pure `h5py`, no Isaac Sim) that validates `format_version`/schema/`env_args` across inputs, renumbers demos sequentially, recomputes `data.attrs["total"]`, and uses a recursive `h5py.Group.copy` so new recorder terms remain forward-compatible.
- Adds 20 Phase-1 unit tests (3s runtime) covering schema validation, dry-run, forward-compat keys, and a property test via Isaac Lab's `HDF5DatasetFileHandler`.
- Documents the workflow in `static_apple/step_2_teleoperation.rst` (new Step 4b) with cross-references from the other three teleop pages.

### DIFF
--- a/docs/pages/example_workflows/static_apple/step_2_teleoperation.rst
+++ b/docs/pages/example_workflows/static_apple/step_2_teleoperation.rst
@@ -252,8 +252,8 @@ Step 4: Record with the headset device
 Step 4b: Merge Multiple Recording Sessions (Optional)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Collecting 100+ clean demonstrations in a single sitting is impractical because of XR latency,
-operator fatigue, and the realities of stopping and starting the Arena app for breaks. The
+Collecting 100+ clean demonstrations in a single sitting is impractical because of operator fatigue,
+and the realities of stopping and starting the Arena app for breaks. The
 recommended workflow is to record one HDF5 per session by pointing ``--dataset_file`` at a fresh
 path each time:
 
@@ -303,9 +303,8 @@ geometry, and prints a per-file summary with the demo and step counts:
 
 If a session was recorded against a slightly different environment (e.g. a different physics
 timestep) the merge will warn but still proceed. Schema-level differences (different action
-dimensions, missing observation keys, different camera resolutions) are hard errors by default —
-pass ``--allow_schema_mismatch`` to downgrade them to warnings if you understand the consequences
-for downstream training.
+dimensions, missing observation keys, different camera resolutions) are hard errors: re-record
+the offending session against the canonical environment instead.
 
 
 Step 5: Replay Recorded Demos (Optional)
--- a/docs/pages/example_workflows/static_apple/step_2_teleoperation.rst
+++ b/docs/pages/example_workflows/static_apple/step_2_teleoperation.rst
@@ -247,6 +247,67 @@ Step 4: Record with the headset device
       Example static apple-to-plate demonstration trajectory.
 
 
+.. _static_apple_merge_demos:
+
+Step 4b: Merge Multiple Recording Sessions (Optional)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Collecting 100+ clean demonstrations in a single sitting is impractical because of XR latency,
+operator fatigue, and the realities of stopping and starting the Arena app for breaks. The
+recommended workflow is to record one HDF5 per session by pointing ``--dataset_file`` at a fresh
+path each time:
+
+.. code-block:: bash
+
+   # Session 1 (e.g. morning)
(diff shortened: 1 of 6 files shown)


rwiltz:
@greptile

isaaclab_arena/scripts/imitation_learning/merge_demos.py
```diff
@@ -0,0 +1,549 @@
...
+        "--overwrite",
+        action="store_true",
+        help="Overwrite the output file if it already exists.",
+    )
+    parser.add_argument(
+        "--dry_run",
+        action="store_true",
+        help="Validate inputs and print the merge report without writing the output file.",
+    )
+    parser.add_argument(
```
peterd-NV: Is there a good reason to ever have this even be an option? HDF5s in Lab are used in a very specific way and I feel like giving the users to arbitrarily combine data from different envs/schemas might actually cause more confusion. 

For example, if a data set had different action shapes, then it will throw an error eventually when using replay_demos script. Same if they come from different envs.

Wonder if it would it be better to eliminate this path completely?

isaaclab_arena/scripts/imitation_learning/merge_demos.py
```diff
@@ -0,0 +1,549 @@
...
+        "--allow_schema_mismatch",
+        action="store_true",
+        help=(
+            "Downgrade per-demo dataset schema mismatches (different shapes, missing keys) from"
+            " hard errors to warnings. The format_version check is always hard."
+        ),
+    )
+    parser.add_argument(
```
peterd-NV: See comment above on whether such a use case would ever actually be beneficial. If not then I would be in favour of removing it altogether.

peterd-NV (approved):
Overall lgtm, I have some comments regarding whether some of the functionality being exposed with this tool is actually necessary and if not, maybe they should be removed to reduce user confusion. Approved since we are short on time and this does not necessarily prevent merging.

.dockerignore
```diff
@@ -1,3 +1,9 @@
 **/.mypy_cache
 **/__pycache__
 /.git
+/.agents
+/.claude
```
xyao-nv: Could you help me understand why letting docker ignore those fix the CI parsing issue from your local repo?

rwiltz: https://github.com/isaac-sim/IsaacLab-Arena/commit/3f2c5c954a42b606430dd28ce3d32ca4571dd7b6#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

This fixes the CI parsing issue (hopefully)

This was keeping my local docker build from working.

rwiltz: Context:

The COPY *.* ${WORKDIR}/ in Dockerfile.isaaclab_arena:146 matches dot-prefixed directories too, and your agent-skills directories collide.

Here's what's happening:

.cursor/, .agents/, and .claude/ all match the *.* glob pattern (since Docker's filepath.Match allows the leading * to be empty).
For each matched directory, Docker COPY copies the contents of that directory into the destination (not the directory itself).
.cursor/ contains skills/inspect-usd-assets/ (a real dir) → tries to create ${WORKDIR}/skills/...
.agents/ contains skills/{run-tests,dev-container,commit-and-pr}/ → tries to merge into ${WORKDIR}/skills/...
.claude/ contains skills (a symlink to ../.agents/skills) → tries to write a symlink at ${WORKDIR}/skills, but Docker already created it as a directory from the previous step.
Hence: cannot replace to directory ... /skills with file.

This was introduced by commit a221208f1 "Add Agent Skills library and refocus AGENTS.md" (#691). The COPY *.* line has been a latent footgun all along — these new dot-prefixed dirs just tripped it.

Even ignoring the immediate symlink failure, the silent copies are wrong too: ${WORKDIR}/skills/, ${WORKDIR}/CODEOWNERS, ${WORKDIR}/launch.json, etc. would all end up at the wrong places from .cursor/, .github/, .vscode/, .pytest_cache/.

docs/pages/example_workflows/static_apple/step_2_teleoperation.rst
```diff
@@ -247,6 +247,66 @@ Step 4: Record with the headset device
       Example static apple-to-plate demonstration trajectory.
 
 
+.. _static_apple_merge_demos:
+
+Step 4b: Merge Multiple Recording Sessions (Optional)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Collecting 100+ clean demonstrations in a single sitting is impractical because of XR latency,
```
tifchen-nvda: Remove this?

rwiltz: Good catch, will remove

pulkitg01 (approved):

PR #707, opened 2026-05-22
 Refactor relation placement solver API

xyao-nv:
## Summary

## Detailed description
 - Introduces `ArenaRelationSolver.solve()` as the high-level API: collect related objects, build `PlacementProblem`, solve object-only placement, select from the pool, then apply the result.
  - Keeps `ArenaEnvBuilder` as the only owner of relation placement setup and removes the standalone preparation helper.
  - Renames pooled placement to `PlacementCandidatePool` and leaves joint object/embodiment optimization as future TODO.


isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+@dataclass
+class PlacementProblem:
+    """Placement problem assembled from the Arena Env components."""
+
+    objects: list[ObjectBase]
+    """Objects with spatial predicates that should be relation-solved."""
+
+    object_placer_params: ObjectPlacerParams
```
zhx06: We have object_placer_params for both `PlacementProblem` and `ObjectRelationSolveResult`. Is there a way to organize the dataclass more efficiently?

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,262 @@
...
+    def _build_problem(self, objects: list[ObjectBase]) -> PlacementProblem:
+        """Build placement inputs from scene objects and solver params."""
+        return PlacementProblem(
+            objects=objects,
+            object_placer_params=deepcopy(self.object_placer_params),
+            num_envs=self.num_envs,
+        )
+
+    def _solve_object_only_placement(self, problem: PlacementProblem) -> ObjectRelationSolveResult:
+        """Reuse the existing v0.2 object relation placer."""
```
zhx06: May consider remove information like `v0.2`

xyao-nv: removed this for now.

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: good point, the param shall be only visible to the problem

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,262 @@
...
+    ) -> ArenaRelationSolveResult:
+        """Apply selected candidates to object spawn state and build reset event config."""
+        anchor_objects_set = set(get_anchor_objects(problem.objects))
+        # Prevent external pose-reset events from conflicting with relation-solved objects.
+        self._validate_no_conflicting_pose_reset_events(problem.objects, anchor_objects_set)
+
+        placement_candidate_pool = placement_candidates.placement_candidate_pool
+        placement_event_cfg = None
+        # Anchor objects do not move, so no need to apply reset event.
+        if anchor_objects_set == set(problem.objects):
```
zhx06: Nit: could we wrap the object-placement code into a helper? The if/elif/else chain is getting hard to read when we have more complicated logics

xyao-nv: sure

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -67,104 +62,14 @@ def _solve_relations(self) -> None:
...
-            print("No objects with relations found in scene. Skipping relation solving.")
-            return
-
-        num_envs = self.args.num_envs
-        cli_resolve = self.args.resolve_on_reset
-
-        # The pool applies positions itself, so disable ObjectPlacer's built-in apply.
-        # Position history and verbose logging are unnecessary for batch-solving a pool.
-        placer_params = ObjectPlacerParams(
+        relation_solver = ArenaRelationSolver(
```
alexmillane: Suggestion to drop the `Arena` from the name. Everything in this repo could have `Arena` in front of the name, so it's not necessary in my opinion.

(One exception to this is the stuff related to environments because they're contrasted to their Isaac Lab equivalents).

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+    """Number of environments that the solver must prepare."""
+
+    @property
+    def pool_size(self) -> int:
+        """Number of object-layout candidates to pre-solve."""
+        return self.num_envs * self.object_placer_params.min_unique_layouts_per_env
+
+
+@dataclass
+class ArenaRelationSolveResult:
```
alexmillane: nit: drop `Arena`

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+
+    placement_candidate_pool: PlacementCandidatePool | None = None
+    """Placement candidate pool used by reset-time placement, when one exists."""
+
+    placement_event_cfg: EventTermCfg | None = None
+    """Reset event config to attach to the Arena environment, if needed."""
+
+
+class ObjectRelationSolver:
+    """Solve and validate object-object spatial relations using the v0.2 object placer."""
```
alexmillane: Nit: Suggestion to drop `v0.2`. This description outdates very quickly - arguably it's already outdated because we're working on v0.3.

isaaclab_arena/relations/placement_candidate_pool.py
```diff
@@ -0,0 +1,152 @@
...
+
+from isaaclab_arena.relations.object_placer import ObjectPlacer
+from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
+from isaaclab_arena.relations.placement_result import MultiEnvPlacementResult, PlacementResult
+
+if TYPE_CHECKING:
+    from isaaclab_arena.assets.object_base import ObjectBase
+
+
+class PlacementCandidatePool:
```
alexmillane: I see we've changed the name from `PooledObjectPlacer` to `PlacementCandidatePool`.

In my opinion, this is a confusing change. To me, a pool stores solutions while a placer computes them. This class computes and stores solutions, so `PooledObjectPlacer` actually seems like quite a good name to me.

What was the reason for the change?

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+            candidate_validator=self.validate_layout,
+        )
+        self.validate_placement_candidate_pool(placement_candidate_pool)
+        return placement_candidate_pool
+
+    def validate_placement_candidate_pool(self, placement_candidate_pool: PlacementCandidatePool) -> None:
+        """Hook for subclasses that need to validate the solved placement candidate pool."""
+
+    def validate_layout(self, layout: PlacementResult) -> None:
+        """Hook for subclasses that need to validate one sampled layout."""
```
alexmillane: What's the purpose of this class?

It appears to just wrap the solve in `PlacementCandidatePool`? This makes our already deep hierarchy (of `ArenaRelationSolver`, `ObjectRelationSolver`, `PlacementCandidatePool`, `ObjectPlacer`. `RelationSolver`), one layer deeper.

Given that this class doesn't appear to add anything, could we eliminate it?

isaaclab_arena/relations/placement_candidate_pool.py
alexmillane: _Edit: Reading on I think I see the reason. This thing is a solution pool, in that it is attached to a placement result and then solutions are drawn from it_.

The naming in that context makes sense.

I'm thinking it still makes sense to retain the original name `PooledObjectPlacer`, given that this thing actually owns the solver.

isaaclab_arena/tests/test_relation_placement.py
```diff
@@ -0,0 +1,80 @@
...
+"""Tests for the Arena relation placement orchestration API."""
+
+from isaaclab_arena.assets.dummy_object import DummyObject
+from isaaclab_arena.relations.arena_env_relation_solver import ArenaRelationSolver
+from isaaclab_arena.relations.relations import IsAnchor, On
+from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox
+from isaaclab_arena.utils.pose import Pose, PosePerEnv
+
+
+class FastArenaRelationSolver(ArenaRelationSolver):
```
alexmillane: `TestArenaRelationSolver`

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+    """Object placer params used for solving, or None when no solve was needed."""
+
+    placement_candidate_pool: PlacementCandidatePool | None = None
+    """Placement candidate pool used by reset-time placement, when one exists."""
+
+    placement_event_cfg: EventTermCfg | None = None
+    """Reset event config to attach to the Arena environment, if needed."""
+
+
+class ObjectRelationSolver:
```
cvolkcvolk: Suggestion to inlie this into `ArenaRelationSolver.solve()` (eg construct `PlacementCandidatePool` directly), and removing this abstaction again as `ObjectRelationSolver` looks like premature abstraction:
- stateless class without init(). with `validate_` that are not _yet_ uoverwritten/ used anywher if I didn't miss anything.

Right now it looks like we don't gain anything except a slight duplication. Could we defer and we introduce this once we need the additional abstaction?

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+    building a placement problem, selecting layouts, and applying the selected
+    result to Arena reset/static placement state.
+    """
+
+    def __init__(
+        self,
+        *,
+        num_envs: int,
+        scene: Scene | None = None,
+        objects: list[ObjectBase] | None = None,
```
alexmillane: Wdyt about moving the objects to an argument to `solve`. This way, each `ArenaRelationSolver` is only valid for a single set of objects. To me, it would make more sense that an `ArenaRelationSolver` can solve multiple problems with different sets of objects.

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+        placement_candidate_pool = self.object_solver.solve(problem)
+        return self._apply_result(problem, placement_candidate_pool)
+
+    def _collect_related_objects(self) -> list[ObjectBase]:
+        """Collect objects with spatial predicates from explicit input or the scene."""
+        if self.objects is not None:
+            return list(self.objects)
+        if self.scene is None:
+            return []
+        return self.scene.get_objects_with_relations()
```
alexmillane: Is there a reason that we need to handle accepting the scene either as a `Scene` or as `list[ObjectBase]`.

Can we simplify by accepting only one of these forms of specification of the scene?

We can then drop this function

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def _build_problem(self, objects: list[ObjectBase]) -> PlacementProblem:
+        """Build placement inputs from scene objects and solver params."""
+        return PlacementProblem(
+            objects=objects,
+            object_placer_params=deepcopy(self.object_placer_params),
+            num_envs=self.num_envs,
+        )
+
+    def _apply_result(
```
alexmillane: I notice that `ObjectPlacer` also has functionality to apply the result to the scene `ObjectPlace._apply_positions`

Is there a reason we need this functionality at both levels of the hierarchy? Can we eliminate applying the result of a solve at one of these levels?

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+            self.object_placer_params.resolve_on_reset = resolve_on_reset
+
+    def solve(self) -> ArenaRelationSolveResult:
+        """Solve Arena relation placement and apply the selected result."""
+        objects = self._collect_related_objects()
+        if not objects:
+            print("No objects with relations found in scene. Skipping relation solving.")
+            return ArenaRelationSolveResult(objects=[])
+
+        problem = self._build_problem(objects)
```
cvolkcvolk: Could we remove this? Everything that the "problem" holds is already an individual member variable

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+
+if TYPE_CHECKING:
+    from isaaclab.managers import EventTermCfg
+
+    from isaaclab_arena.assets.object_base import ObjectBase
+    from isaaclab_arena.scene.scene import Scene
+
+
+@dataclass
+class PlacementProblem:
```
cvolkcvolk: Its currently used in `ArenaRelationSolver`, which already holds `objects`, and the `object_placer_params` as member variables.  I feel like adding this adds a layer which is not needed and might open the backdoor to bugs.

alexmillane (commented):
Hi @xyao-nv.

Thank you for abstracting all the object placement-related code out of the Env Builder. That simplifies a class that was becoming quite complicated. Amazing!

I have a few minor comments.

My main concern is that we're starting to get quite a deep hierarchy in the object placement-related code. We have: `ArenaRelationSolver`, `ObjectRelationSolver`, `PlacementCandidatePool`, `ObjectPlacer`. `RelationSolver`. 

I think we can just eliminate `ObjectRelationSolver`, which is added in this MR - it appears to be an empty class that just adds a layer to the hierarchy.

At the top level, the `ArenaRelationSolver` seems to mainly(/entirely) add functions for converting the results of a placement into an effect on the arena environment (either a reset term or directly setting the initial pose). Could we eliminate this layer of the hierarchy by just moving this functionality to a few freestanding functions in a separate file?

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+
+        # Static initial poses for each environment.
+        self._apply_static_initial_poses(
+            problem.objects,
+            placement_candidate_pool,
+            anchor_objects_set,
+        )
+        return None
+
+    def _apply_dynamic_spawn_pose(
```
cvolkcvolk: Could be moved out of this function as it does not need access to any private (self) variables

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+            rotation_xyzw = get_rotation_xyzw(obj)
+            poses = []
+            for env_idx in range(self.num_envs):
+                pos = layouts[env_idx].positions.get(obj)
+                if pos is None:
+                    raise RuntimeError(f"Placement candidate is missing a solved position for '{obj.name}'.")
+                poses.append(Pose(position_xyz=pos, rotation_xyzw=rotation_xyzw))
+            obj.set_initial_pose(PosePerEnv(poses=poses))
+
+    def _validate_no_conflicting_pose_reset_events(
```
cvolkcvolk: Same, could be moved out into a free function

isaaclab_arena/relations/arena_env_relation_solver.py
```diff
@@ -0,0 +1,257 @@
...
+        return EventTermCfg(
+            func=solve_and_place_objects,
+            mode="reset",
+            params={
+                "objects": objects,
+                "placement_candidate_pool": placement_candidate_pool,
+            },
+        )
+
+    def _apply_static_initial_poses(
```
cvolkcvolk: Could we move this into a free function?

isaaclab_arena/relations/arena_env_relation_solver.py
cvolkcvolk: Maybe to `placement_events.py` like the functions below?

isaaclab_arena/environments/arena_env_builder.py
cvolkcvolk, line 1: Nice cleanup in here!

isaaclab_arena/environments/arena_env_builder.py
xyao-nv: Gotcha. Makes sense. Updated

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: done

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: Sure

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: yeah I like that

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: sure

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: sure

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: sure

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: i was thinking when embodiment will be added to the solver, how it will work. But I will make it simple in this MR to match parity

xyao-nv:
Updates based on last round of reviews:
- Keep original `PooledObjectPlacer : it should stay object-only: it owns ObjectPlacer, produces object layout candidates, and samples from that pool.
- Remove `ArenaRelationSolver` and remove `PlacementCandidatePool`.
- `solve_and_apply_relation_placement(...)` remains the orchestration boundary and `arena_env_builder.py` to call it directly.

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: I think it's lil overly complicated, dropped it.

isaaclab_arena/relations/placement_candidate_pool.py
xyao-nv: retain the original name PooledObjectPlacer

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: good call, removed it

isaaclab_arena/relations/arena_env_relation_solver.py
xyao-nv: sure

isaaclab_arena/relations/pooled_object_placer.py
```diff
@@ -78,20 +82,38 @@ def _solve_and_store(self, num_layouts: int) -> None:
...
+            self._layouts.extend(all_layouts)
+
+    def _validate_sampled_layouts(self, layouts: list[PlacementResult]) -> None:
+        """Run the optional validation callback on each sampled layout."""
+        if self._layout_validator is None:
+            return
+        for layout in layouts:
+            self._layout_validator(layout)
+
+    def _ensure_layouts_available(self, count: int) -> None:
```
alexmillane: nit: `_refill_pool_via_solve_if_required`

let's just make is doubly clear that this triggers a resolve 😅

isaaclab_arena/relations/pooled_object_placer.py
```diff
@@ -78,20 +82,38 @@ def _solve_and_store(self, num_layouts: int) -> None:
...
-            self._layouts.extend(all_results)
+            print("Warning: No layouts passed strict validation. Accepting best-loss layouts as fallback.")
+            self._layouts.extend(all_layouts)
+
+    def _validate_sampled_layouts(self, layouts: list[PlacementResult]) -> None:
+        """Run the optional validation callback on each sampled layout."""
+        if self._layout_validator is None:
+            return
+        for layout in layouts:
+            self._layout_validator(layout)
```
alexmillane: What's the effect of the layout validator?

To me, it looks as if it's not doing anything? given that it doesn't return anything? Does the validator modifies the layout? Or maybe throw a warning?

Is this a hook for some future feature, for example IK-validation? If that's the case, I would recommend that we remove this feature here and add it when it's functional?

isaaclab_arena/relations/pooled_object_placer.py
```diff
@@ -78,20 +82,38 @@ def _solve_and_store(self, num_layouts: int) -> None:
...
+
+    def _ensure_layouts_available(self, count: int) -> None:
+        """Refill the pool when fewer than *count* layouts are available."""
+        if self.remaining < count:
+            self._solve_and_store(max(self._pool_size, count))
```
alexmillane, line 99: What's the idea of solving less than `pool_size`?

Perhaps naively, I would have thought that we always refill the pool to it's maximum capacity.

Currently, we
- on creation of the pool fill it up,
- on subsequent solves, only fill it up enough to cover the requested number of layouts.

I thought the idea of the pool is that we solve for _more_ solutions than we currently need. I would have expected that we fill the pool when required, irrespective of how many solutions are currently being requested.

This is probably a question for @zhx06, as this behaviour existed before this MR. @xyao-nv feel free to ignore this comment for the purposes of merging this MR, as this is not a _change_ in behaviour.

isaaclab_arena/relations/pooled_object_placer.py
```diff
@@ -101,30 +123,30 @@ def sample_without_replacement(self, count: int) -> list[PlacementResult]:
...
+        return layouts
 
     @property
     def remaining(self) -> int:
         """Number of layouts not yet consumed by :meth:`sample_without_replacement`."""
         return len(self._layouts) - self._next_idx
+
+    @property
+    def pool_size(self) -> int:
+        """Number of layouts requested for each pool refill solve."""
```
alexmillane: See comment above. I think that this is actually not an accurate description of the current behaviour. But let's leave that to a future MR and @zhx06 to fix.

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -0,0 +1,190 @@
...
+        objects=objects,
+        placer_params=object_placer_params,
+        pool_size=num_envs * object_placer_params.min_unique_layouts_per_env,
+    )
+    return _apply_relation_placement_result(
+        objects=objects,
+        object_placer_params=object_placer_params,
+        placement_pool=placement_pool,
+        num_envs=num_envs,
+    )
```
alexmillane, line 68: Nice! clean!!!

isaaclab_arena/environments/relation_solver_interface.py
alexmillane, line 1: File looks clean! Great improvement. Thank you @xyao-nv 

I'm thinking that given the number of similarly named files in this folder, what do you think about moving this file into `isaaclab_arena/environments`, under the name `relation_solver_interface.py` to indicate that this file is bridging the IsaacLab env, with the underlying solver.

Does that sounds reasonable to you?

isaaclab_arena/relations/relation_placement.py
```diff
@@ -0,0 +1,190 @@
...
+
+    from isaaclab_arena.assets.object_base import ObjectBase
+
+
+def solve_and_apply_relation_placement(
+    objects: list[ObjectBase],
+    num_envs: int,
+    placement_seed: int | None = None,
+    resolve_on_reset: bool | None = None,
+    object_placer_params: ObjectPlacerParams | None = None,
```
alexmillane: I scanned the code and to my eye, this parameter is never used. I would recommend removing it because it complicates matters below (because we do branching based on whether it is passed).

isaaclab_arena/relations/relation_placement.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+    else:
+        object_placer_params = deepcopy(object_placer_params)
+        object_placer_params.apply_positions_to_objects = False
+        if placement_seed is not None:
+            object_placer_params.placement_seed = placement_seed
+
+    if resolve_on_reset is not None:
+        object_placer_params.resolve_on_reset = resolve_on_reset
+    return object_placer_params
```
alexmillane: See comment above, `object_placer_params` is never used, so we can get rid of this function (because we don't need the logic of branching on its presence).

alexmillane (approved):
Thanks for the clean up! Looks great. I had a few more small comments.

@zhx06 One of the comments if for you :)

isaaclab_arena/relations/pooled_object_placer.py
zhx06: Yeah, this is not an accurate description. We want the `pool_size` to be big so that we don't need to run the solver again every time we want to obtain a new layout.  A better description could be: `pool_size` is the minimum refill batch size; when the pool runs low, we solve at least this number of layouts so future samples can reuse the buffer.

xyao-nv: gotcha, will update in this func.
Thanks

isaaclab_arena/environments/relation_solver_interface.py
xyao-nv: sounds like a good idea!

isaaclab_arena/relations/pooled_object_placer.py
xyao-nv: gotcha updated

isaaclab_arena/relations/pooled_object_placer.py
xyao-nv: sure I will remove it to match parity.

isaaclab_arena/relations/pooled_object_placer.py
zhx06: Solving with less than `batch_size < pool_size` means we do not use an excessively large batch size that could exceed GPU memory. In this case, we may need to run the solver multiple times before the simulation environment starts. This behavior is still open to discussion: if we have access to a good GPU, we could also set `batch_size > pool_size` and run the solver only once.

xyao-nv: This is expected behavior, when batch size is smaller than pool capacity (pool size), we will up to its max capacity(pool size)

isaaclab_arena/relations/relation_placement.py
xyao-nv: Change to inline construct `ObjectPlacerParams` then

isaaclab_arena/relations/relation_placement.py
xyao-nv: removed

isaaclab_arena/tests/test_relation_solver_interface.py
```diff
@@ -0,0 +1,80 @@
...
+    result = solver.solve()
+
+    assert result.object_placer_params is not None
+    assert result.object_placer_params.resolve_on_reset is False
+    assert result.placement_candidate_pool is not None
+    assert result.placement_event_cfg is None
+
+    initial_pose = box.get_initial_pose()
+    assert isinstance(initial_pose, PosePerEnv)
+    assert len(initial_pose.poses) == 2
```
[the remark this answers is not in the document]
xyao-nv: will leave it @zhx06 in future MRs

isaaclab_arena/relations/pooled_object_placer.py
zhx06, line 41: This should be `object_placer_params`  since we have changed the name

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -0,0 +1,190 @@
...
+        if obj in anchor_objects_set:
+            continue
+        rotation_xyzw = get_rotation_xyzw(obj)
+        poses = []
+        for env_idx in range(num_envs):
+            pos = layouts[env_idx].positions.get(obj)
+            if pos is None:
+                raise RuntimeError(f"Placement layout is missing a solved position for '{obj.name}'.")
+            poses.append(Pose(position_xyz=pos, rotation_xyzw=rotation_xyzw))
+        obj.set_initial_pose(PosePerEnv(poses=poses))
```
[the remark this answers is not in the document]
xyao-nv: good catch. revert it

isaaclab_arena/relations/pooled_object_placer.py
xyao-nv: updated the docstring

PR #722, opened 2026-05-26
[release/0.2.1] Change the prompt for the finetuning

pulkitg01:
## Summary
Align static apple prompt

## Detailed description
- Reason: align the simulated static apple workflow with the same prompt used for the real setup.
- Changed the GR00T LeRobot training config prompt to `move the apple to the plate`.
- Updated the closed-loop evaluation config to use the same prompt.
- Updated the static apple policy-training docs snippet so users finetune with the matching prompt.
- Impact: training and evaluation now use consistent language conditioning across sim and real workflows.


shanereetz:
LGTM :rocket:

xyao-nv (approved):

PR #724, opened 2026-05-26
Add EnvGraphSpec to ArenaEnv method

xyao-nv:
## Summary
Convert EnvGraph to ArenaEnv

## Detailed description
- Adds `ArenaEnvGraphSpec.to_arena_env()` so YAML graph specs can be converted into `IsaacLabArenaEnvironment`.
- Resolves task classes at runtime and converts multi-task specs into a sequential composite task using all task entries.
- Includes nodes, state, task validation in to_yaml().


isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,481 @@
...
+        ArenaEnvGraphTaskSpec,
+    )
+
+
+def build_arena_env_from_graph_spec(
+    spec: ArenaEnvGraphSpec,
+    state_spec_id: str | None = None,
+) -> Any:
+    """Build an IsaacLabArenaEnvironment from the parsed graph spec."""
+    _validate_graph_spec_for_conversion(spec, state_spec_id)
```
qianl-nv: I feel the validations here are a bit mixed between basic spec sanity checks (_validate_initial_state_supported_for_conversion, _validate_spatial_constraint_shapes ect),  and conversion related checks (_validate_initial_state_supported_for_conversion). 

logically the basic spec sanity checks should already happen when we construct the EnvGraphSpec object from yaml, and only the conversion related checks should need to run at the spec.to_arena_env().

I'd suggest moving basic spec sanity checks into isaaclab_arena/environments/graph_spec_utils.py, and add an api spec.validate() to always call during from_yaml(). wdyt?

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,481 @@
...
+    state_spec = _select_state_spec(spec, spec.tasks, state_spec_id)
+    if state_spec is not None:
+        _validate_initial_state_supported_for_conversion(state_spec)
+
+
+def _validate_spatial_constraint_shapes(spec: ArenaEnvGraphSpec) -> None:
+    """Validate relationship endpoints and arity."""
+    from isaaclab_arena.environments.arena_env_graph_spec import ArenaEnvGraphSpatialConstraintType
+
+    child_required_types = {
```
qianl-nv: should we build a map from ArenaEnvGraphSpatialConstraintType to the actual relation class (in the relation.py files), and for the relation classes to have a static method in_unary(). would be easier to keep track then keeping a map here.
this might also solve the not_supported types in _validate_initial_state_supported_for_conversion cleaner.

qianl-nv (commented):
Thanks @xyao-nv for the draft. Overall looks ok, but isaaclab_arena/environments/arena_env_graph_conversion_utils.py is a long file and could potentially be clean up to make it simpler. Especially in terms of spec validation, many checks should happen at the spec construction step, not at the spec->arena env conversion step.

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
xyao-nv: I was hesitating if sanity checks shall be added here as users we want to create their own yaml, e.g. mimic next.
But I like your suggestion, updated.

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
xyao-nv: agree, updated

xyao-nv:
Changes made:
- Add `is_unary()` to `Relation` and `RelationBase`
- Add a spatial constraints lookup table to auto populate Relation class objects
- Move some validation related steps in graph_spec_utils.py

isaaclab_arena/relations/relations.py
```diff
@@ -34,7 +30,10 @@ class RelationBase:
     in its relations list.
     """
 
-    pass
+    @staticmethod
```
qianl-nv: this is a wrong update. should be added to the next class UnaryRelation

isaaclab_arena/relations/relations.py
```diff
@@ -205,6 +209,12 @@ def to_pose_range_centered_at(
         Returns:
             PoseRange spanning ± half-extents around the position and rotation.
         """
+        import torch
```
qianl-nv: any particular reason we are moving the import inside? ditto for the import change below on line 287

xyao-nv: ah, revert it

isaaclab_arena/relations/relations.py
xyao-nv: updated

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
```diff
@@ -0,0 +1,206 @@
...
+
+    Returns None if there are no specs, a single task instance if there's one,
+    or a SequentialTaskBase wrapping all of them in order.
+    """
+    if not task_specs:
+        return None
+    subtasks = [_build_task(task_spec, assets_by_id) for task_spec in task_specs]
+    if len(subtasks) == 1:
+        return subtasks[0]
+    return SequentialTaskBase(subtasks=subtasks, desired_subtask_success_state=[True for _ in subtasks])
```
xyao-nv: self review: maybe it will change based on what subtask tracking brings

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,130 @@
...
+        scene=Scene(assets=scene_assets),
+        embodiment=embodiment,
+        task=build_task_or_sequence(spec.tasks, assets_by_id),
+    )
+
+
+def _instantiate_node_assets(nodes: list[ArenaEnvGraphNodeSpec], asset_registry: Any) -> dict[str, Any]:
+    """Create concrete asset entities for graph nodes and wire object-reference nodes.
+
+    Upstream contract: nodes are ordered so an OBJECT_REFERENCE appears after its parent —
```
xyao-nv: self review: think about how to enforce those contract in upstream

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    task_args = _resolve_task_args(task_spec.task_args, assets_by_id)
+    return task_cls(**task_args)
+
+
+def _resolve_task_class(task_type: str) -> Any:
+    """Resolve a YAML ``type:`` string to a concrete task class.
+
+    Two forms are supported:
+      * Qualified import path — ``pkg.module:Class`` or ``pkg.module.Class``.
+      * Bare class name — searched for in the tasks package by exact ``__name__``.
```
xyao-nv: self review: I wonder if it shall be enforced to only use bare class name, as import path may be changed

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,113 @@
...
+
+    embodiment = None
+    scene_assets: list[Any] = []
+    # TODO(xinjieyao, 2026-05-26): include lighting later
+    for node in spec.nodes:
+        if node.type == ArenaEnvGraphNodeType.EMBODIMENT:
+            assert embodiment is None, "Only one embodiment node can be converted to an IsaacLabArenaEnvironment"
+            embodiment = assets_by_id[node.id]
+        elif node.type in (ArenaEnvGraphNodeType.OBJECT, ArenaEnvGraphNodeType.OBJECT_REFERENCE):
+            scene_assets.append(assets_by_id[node.id])
```
alexmillane: Shall we add a final `else`?

Perhaps right now the left over nodes are not handled? We could add a warning for now?

isaaclab_arena/environments/arena_env_graph_types.py
```diff
@@ -0,0 +1,137 @@
+# Copyright (c) 2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
xyao-nv, line 1: note: purely mv stuffs around, so this file only contains data, no orchestration/method.

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -183,6 +103,15 @@ def tasks_by_id(self) -> dict[str, ArenaEnvGraphTaskSpec]:
...
         return {state_spec.id: state_spec for state_spec in self.state_specs}
 
+    def to_arena_env(self) -> "IsaacLabArenaEnvironment":
+        """Convert this graph spec into an `IsaacLabArenaEnvironment`.
+
+        The first ``state_spec`` is used as the scene's initial state.
+        """
+        # TODO(xinjieyao, 2026-05-26): once `build_arena_env_from_graph_spec` aggregates across all state_specs,
+        # this wrapper stays single-arg — no caller-side selection is needed.
+        return build_arena_env_from_graph_spec(self)
```
xyao-nv, line 118: note: there are too many detailed fun calls, so I avoid spamming the class method.

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,111 @@
...
+            assets_by_id[node.id] = ObjectReference(
+                name=node.name,
+                prim_path=node.prim_path,
+                parent_asset=assets_by_id[node.parent],
+                object_type=node.object_type,
+                **node.params,
+            )
+            continue
+        asset_cls = asset_registry.get_asset_by_name(node.name)
+        assets_by_id[node.id] = asset_cls(**node.params)
```
alexmillane: nit: suggestion to use if..else, rather than if...continue.

For some reason claude likes to do this, but I (perhaps personally) feel that if...else is a far more normal construction.

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+def _instantiate_node_assets(nodes: list[ArenaEnvGraphNodeSpec], asset_registry: Any) -> dict[str, Any]:
+    """Create concrete asset entities for graph nodes and wire object-reference nodes.
+
+    Upstream contract:
+      * Nodes are ordered so an OBJECT_REFERENCE appears after its parent — a single pass is
+        enough; the parent lookup in `assets_by_id` would `KeyError` otherwise.
+      * `node.params` is emitted in the asset constructor's exact kwarg form, including any
+        `instance_name` needed to disambiguate duplicate assets.
+    """
```
alexmillane: Suggestion to check this (probably in a function).

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+    embodiment = None
+    scene_assets: list[Any] = []
+    # TODO(xinjieyao, 2026-05-26): include lighting later
+    for node in spec.nodes:
+        if node.type == ArenaEnvGraphNodeType.EMBODIMENT:
+            assert embodiment is None, "Only one embodiment node can be converted to an IsaacLabArenaEnvironment"
+            embodiment = assets_by_id[node.id]
+        elif node.type in (ArenaEnvGraphNodeType.OBJECT, ArenaEnvGraphNodeType.OBJECT_REFERENCE):
+            scene_assets.append(assets_by_id[node.id])
```
alexmillane: nit: move to a function so this function (`assets_to_scene_and_embodiment`)

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    # picking one. For now we just take the first state_spec, which is the initial state
+    # for the first task — this matches the previous default behavior.
+    state_spec = spec.state_specs[0] if spec.state_specs else None
+
+    assets_by_id = _instantiate_node_assets(spec.nodes, AssetRegistry())
+    if state_spec is not None:
+        _apply_state_spatial_constraints(state_spec, assets_by_id)
+
+    embodiment = None
+    scene_assets: list[Any] = []
```
alexmillane: Can we `Any` -> `Asset`?

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,111 @@
...
+        # Note(xinjieyao, 2026-05-26): at_pose constraint is a special case that is handled differently.
+        # Ideally it shall be handled within the placer module.
+        if constraint.type == ArenaEnvGraphSpatialConstraintType.AT_POSE:
+            params = dict(constraint.params)
+            position_xyz = params.pop("position_xyz", None)
+            rotation_xyzw = params.pop("rotation_xyzw", (0.0, 0.0, 0.0, 1.0))
+            assert position_xyz is not None, f"at_pose constraint '{constraint.id}' requires params.position_xyz"
+            assert not params, f"Unsupported at_pose params for constraint '{constraint.id}': {sorted(params)}"
+            parent_asset.set_initial_pose(Pose(position_xyz=position_xyz, rotation_xyzw=rotation_xyzw))
+            continue
```
alexmillane: prefer if...else over if...continue.

isaaclab_arena/environments/graph_spec_utils.py
```diff
@@ -129,5 +141,140 @@ def assert_references_exist(nodes: list[Any], tasks: list[Any], state_specs: lis
...
+
+    return {
+        ArenaEnvGraphSpatialConstraintType.IS_ANCHOR: IsAnchor,
+        ArenaEnvGraphSpatialConstraintType.NEXT_TO: NextTo,
+        ArenaEnvGraphSpatialConstraintType.ON: On,
+        ArenaEnvGraphSpatialConstraintType.AT_POSITION: AtPosition,
+        ArenaEnvGraphSpatialConstraintType.POSITION_LIMITS: PositionLimits,
+        ArenaEnvGraphSpatialConstraintType.RANDOM_AROUND_SOLUTION: RandomAroundSolution,
+        ArenaEnvGraphSpatialConstraintType.ROTATE_AROUND_SOLUTION: RotateAroundSolution,
+    }
```
alexmillane: Is there a reason to not make this a module level constant? I believe that would be more normal

```
SPATIAL_CONSTRAINT_RELATION_CLASSES={
    ArenaEnvGraphSpatialConstraintType.IS_ANCHOR: IsAnchor,
    ...
}
```

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    return SequentialTaskBase(subtasks=subtasks, desired_subtask_success_state=[True for _ in subtasks])
+
+
+def _build_task(task_spec: ArenaEnvGraphTaskSpec, assets_by_id: dict[str, Any]) -> Any:
+    """Build one task instance, swapping node-id strings in its args for the live asset objects.
+
+    Upstream contract: `task_args` keys are emitted in the task constructor's exact parameter
+    names — no loose / case-insensitive matching is performed here.
+    """
+    task_cls = _resolve_task_class(task_spec.type)
```
alexmillane: Elsewhere in arena we deal with the "string to object" problem via a registry.

We have a few registries in `registries.py`.

If you compare your code for decoding the scene to the code here for decoding the tasks, there's a big difference in complexity. Furthermore, the code here is not extendable by a user who wants to add tasks from outside the source-tree, because we only search for tasks effectively in a single folder.

I suggest we move to a registry for tasks. Effectively this just associates a string name to each task, and gives us a nice interface for looking them up.

What do you think?

isaaclab_arena/environments/arena_env_graph_types.py
```diff
@@ -0,0 +1,137 @@
...
+    OBJECT = "object"
+    OBJECT_REFERENCE = "object_reference"
+    LIGHTING = "lighting"
+
+
+class ArenaEnvGraphSpatialConstraintType(Enum):
+    IS_ANCHOR = "is_anchor"
+    NEXT_TO = "next_to"
+    ON = "on"
+    AT_POSE = "at_pose"  # through set_initial_pose()
```
alexmillane, line 37: At some point we'll have to clean this up and add a relation for this 😅 

Let's do that later.

isaaclab_arena/environments/graph_spec_utils.py
```diff
@@ -129,5 +141,140 @@ def assert_references_exist(nodes: list[Any], tasks: list[Any], state_specs: lis
...
+    )
+
+    return {
+        ArenaEnvGraphSpatialConstraintType.IS_ANCHOR: IsAnchor,
+        ArenaEnvGraphSpatialConstraintType.NEXT_TO: NextTo,
+        ArenaEnvGraphSpatialConstraintType.ON: On,
+        ArenaEnvGraphSpatialConstraintType.AT_POSITION: AtPosition,
+        ArenaEnvGraphSpatialConstraintType.POSITION_LIMITS: PositionLimits,
+        ArenaEnvGraphSpatialConstraintType.RANDOM_AROUND_SOLUTION: RandomAroundSolution,
+        ArenaEnvGraphSpatialConstraintType.ROTATE_AROUND_SOLUTION: RotateAroundSolution,
```
alexmillane: One issue here is that it requires a user to maintain the relationship between constraints in the relations folder and this static struct here.

That's not ideal, although not a huge problem for us. The issue is that if someone wants to extend  arena with new relations. They can do that on the relations side, but they can't do it here, because this mapping is hard coded.

I think the correction solution here is to have a constraint registry. Such a registry provides a mapping between a string name, and a constrain object. Such a registry is extensible by the user. It also eliminates the need to maintain this mapping between a name and an object - constraint declare their names.

alexmillane (commented):
A couple of nits, but my main comment is that we should add registries for tasks and constraints to avoid some stuff in here that makes the graph decoding not extensible out-of-source.

isaaclab_arena/environments/graph_spec_utils.py
xyao-nv: Good idea on the registry. Updated.

isaaclab_arena/environments/graph_spec_utils.py
xyao-nv: I don't like those mapping, so stick to your registry idea

isaaclab_arena/environments/arena_env_graph_types.py
xyao-nv: yeah it's on my todo after features for M1 are done

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
xyao-nv: so with registry, nothing to worry. Updated

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
xyao-nv: Agree on your suggestion. Updated.

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
xyao-nv: yes updated.

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
xyao-nv: aha thx for pointing out this pattern. updated.

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
xyao-nv: updated

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
xyao-nv: updated

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
xyao-nv: done

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
xyao-nv: Added a ordering check to validate()

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,150 @@
...
+    embodiment, scene_assets = _partition_nodes_into_embodiment_and_scene(graph_spec.nodes, assets_by_node_id)
+
+    # 4. Resolve task specs against the same assets_by_node_id so task args bind to the
+    #    actual asset instances created in step 1 (not duplicates).
+    return IsaacLabArenaEnvironment(
+        name=graph_spec.env_name,
+        scene=Scene(assets=scene_assets),
+        embodiment=embodiment,
+        task=build_task_from_specs(graph_spec.tasks, assets_by_node_id),
+    )
```
alexmillane, line 58: Nice! Super clean!

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        subtasks=task_instances,
+        desired_subtask_success_state=[True] * len(task_instances),
+    )
+
+
+def _build_task_from_spec(task_spec: ArenaEnvGraphTaskSpec, assets_by_node_id: dict[str, Any]) -> Any:
+    """Look up the task class by name, resolve any Asset-typed kwargs, instantiate."""
+    task_class = TaskRegistry().get_task_by_name(task_spec.type)
+    task_init_kwargs = _resolve_node_refs_in_task_args(task_class, task_spec.task_args, assets_by_node_id)
+    return task_class(**task_init_kwargs)
```
alexmillane, line 52: 👍 looking much cleaner and more extensible.

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
```diff
@@ -0,0 +1,131 @@
...
+) -> dict[str, Any]:
+    """Swap node-id strings for live assets on Asset / list[Asset] params; pass others through.
+
+    Example — for ``PickAndPlaceTask(pick_up_object: Asset, ..., episode_length_s: float)``::
+
+        raw_task_args     = {"pick_up_object": "cube", ..., "episode_length_s": 5.0}
+        assets_by_node_id = {"cube": <Object>, ...}
+        # -> {"pick_up_object": <Object: cube>, ..., "episode_length_s": 5.0}
+
+    Misspelled / non-string node ids raise AssertionError instead of silently passing through.
```
alexmillane: Slightly weird comment. Weird to assume that non-existing node names would just pass through. Suggestion:

-> "Misspelled / non-string node ids raise AssertionError"

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
```diff
@@ -0,0 +1,131 @@
...
+            # list[Asset]-typed param: resolve each element to its live asset.
+            resolved_task_kwargs[param_name] = [
+                _lookup_asset_by_node_id(raw_node_id, assets_by_node_id, task_class, param_name)
+                for raw_node_id in raw_param_value
+            ]
+        else:
+            # Asset-typed param: resolve the single node id to its live asset.
+            resolved_task_kwargs[param_name] = _lookup_asset_by_node_id(
+                raw_param_value, assets_by_node_id, task_class, param_name
+            )
```
alexmillane: Having three switch values on `is_collection` is a bit weird because "is" implies a `True` or `False` answer.

It's a nit but, suggestion to have two flags:

```
is_collection = is_collection_by_param_name.get(param_name)
is_node_ref = is_collection == None
```

alexmillane (approved):
Could of nits.

LGTM!

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
xyao-nv: I guess the pass thru is for genially valid non-asset argument. Updated.

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
xyao-nv: updated

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -0,0 +1,154 @@
...
+            embodiment = assets_by_node_id[node_spec.id]
+        elif node_spec.type in (
+            ArenaEnvGraphNodeType.BACKGROUND,
+            ArenaEnvGraphNodeType.OBJECT,
+            ArenaEnvGraphNodeType.OBJECT_REFERENCE,
+        ):
+            scene_assets.append(assets_by_node_id[node_spec.id])
+        else:
+            raise ValueError(f"Unsupported node type: {node_spec.type}")
+    return embodiment, scene_assets
```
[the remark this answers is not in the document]
xyao-nv: updated

isaaclab_arena/relations/relations.py
```diff
@@ -12,10 +12,10 @@
 from isaaclab.utils.math import euler_xyz_from_quat
 
 from isaaclab_arena.assets.register import register_object_relation
-from isaaclab_arena.utils.pose import PoseRange
 
 if TYPE_CHECKING:
     from isaaclab_arena.assets.object_base import ObjectBase
+    from isaaclab_arena.utils.pose import PoseRange
```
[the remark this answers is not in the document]
xyao-nv: true bug, updated

isaaclab_arena/environments/arena_env_graph_task_conversion_utils.py
```diff
@@ -0,0 +1,133 @@
...
+                resolved_task_kwargs[param_name] = [
+                    _lookup_asset_by_node_id(raw_node_id, assets_by_node_id, task_class, param_name)
+                    for raw_node_id in raw_param_value
+                ]
+            else:
+                # Asset-typed param: resolve the single node id to its live asset.
+                #   e.g. "pick_up_object": "cube"  ->  "pick_up_object": <Object: cube>
+                resolved_task_kwargs[param_name] = _lookup_asset_by_node_id(
+                    raw_param_value, assets_by_node_id, task_class, param_name
+                )
```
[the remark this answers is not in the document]
xyao-nv: will do later

PR #727, opened 2026-05-27
Document second-viewport teleop monitoring workflow

rwiltz:
## Summary
Doc two-person teleop with 2nd viewport

## Detailed description
- Per Edith's feedback, demos where the trajectory drifts outside the robot's recording-camera FOV silently degrade policy training. A second operator watching a viewport bound to the recording camera can catch this live and avoid a re-record.
- Adds a new "Step 2b: Monitor Recording with a Second Viewport (Optional)" between Steps 2 and 3 in `docs/pages/example_workflows/static_apple/step_2_teleoperation.rst`, with two screenshots: enabling Viewport 2 from the **Window** menu, and pointing it at `RobotHeadCam`.
- Notes that `RobotHeadCam` is only spawned when `--enable_cameras` is set — present by default in Step 4's `record_demos.py`, absent in the Step 2 `teleop.py` smoke test — so readers know when the camera will appear in the selector.
- Docs only; no code changes.


shanereetz:
LGTM, I think this is a reasonable workaround for now

xyao-nv (approved):

PR #734, opened 2026-05-28
Disable RTX scene ambient in Arena env config

cvolkcvolk:
## Summary

Zeroing out all USD lights (or sweeping a dome light down to `light_intensity=0`) still produced a lit scene in the policy cameras instead of black. 

Screenshot from 2026-05-29 16-35-53

Override the carb settings on every Arena eval via `RenderCfg.carb_settings` on the default `SimulationCfg`.

## Detailed description

- The RTX renderer has its own scene-ambient term that is completely independent of any UsdLuxLight prim. The carb setting  /rtx/sceneDb/ambientLightIntensity defaults to 1.0 with color [0.1, 0.1, 0.1] in the IsaacLab kit. That adds a hidden ~10%-gray ambient floor to every rendered frame.


isaaclab_arena/environments/isaaclab_arena_manager_based_env_cfg.py
```diff
@@ -70,7 +70,19 @@ class IsaacLabArenaManagerBasedRLEnvCfg(ManagerBasedRLEnvCfg):
...
+    # Override the RTX renderer's built-in scene ambient (carb /rtx/sceneDb/ambientLightIntensity, default 1.0 with
+    # color [0.1, 0.1, 0.1]) so that USD light prims fully control scene illumination. Without this, Arena scenes
+    # carry a ~10%-gray ambient floor independent of any UsdLuxLight, which silently confounds vision-policy evals.
+    sim: SimulationCfg = SimulationCfg(
+        dt=1 / 200,
+        render_interval=2,
+        render=RenderCfg(
+            carb_settings={
+                "/rtx/sceneDb/ambientLightIntensity": 0.0,
+                "/rtx/sceneDb/ambientLightColor": [0.0, 0.0, 0.0],
```
peterd-NV: Minor comment: Is the ambientLightColor override necessary? Once intensity is set to 0.0 then the colour should not matter no?

peterd-NV (approved):

isaaclab_arena/environments/isaaclab_arena_manager_based_env_cfg.py
cvolkcvolk: Good point. Let me confirm

cvolkcvolk: Verified its not needed. Removed.

PR #735, opened 2026-05-28
Fix object set on termination test.

alexmillane:
## Summary
Fix object set on termination test.

## Detailed description
- We had several issues with the `test_object_set_on_termination.py`
- Lowered threshold and fixed bug where parameters we're passed in test.

### DIFF
--- a/isaaclab_arena/tests/test_object_set_on_termination.py
+++ b/isaaclab_arena/tests/test_object_set_on_termination.py
@@ -16,7 +16,7 @@ NUM_STEPS = 100
 # sampling sometime you don't get both objects. This caused us to not
 # reliably catch issues when we had a bug.
 NUM_ENVS = 10
-HEADLESS = False
+HEADLESS = True
 
 
 def _test_object_set_on_destination_termination(simulation_app) -> bool:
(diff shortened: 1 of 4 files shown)


isaaclab_arena/tests/test_object_set_on_termination.py
```diff
@@ -9,16 +9,18 @@
 
 from isaaclab_arena.tests.utils.subprocess import run_simulation_app_function
 
-NUM_STEPS = 50
-# This should stay at 2 to test both objects in the object set.
-NUM_ENVS = 2
+# Turned this up to 200 to ensure all objects fall below the velocity threshold.
+NUM_STEPS = 100
+# This should stay >2 to test both objects in the object set.
+# Note(alexmillane, 2026-05-28): I realized that with 2 plus random
```
xyao-nv, line 15: Hmm,` random_choice` shall be set to off in object_set by default https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena/assets/object_set.py#L30.

Maybe this API has been deprecated in Lab 3.0. I will take a closer look later.

xyao-nv (approved):
Thx for fixing!

isaaclab_arena/tests/test_object_set_on_termination.py
alexmillane: Weirdly I seemed to still be getting randomness... I can't explain that. Maybe something on the lab side?

PR #736, opened 2026-05-28
Add ObjectRelation & Task Registries

xyao-nv:
## Summary
Add ObjectRelation Registry & Task Registry

## Detailed description
- Why: When constructing `ArenaEnv` from `EnvGraphSpec`, it requires matching spatial relationship & task to its class name. Without this registry, new relation/task need to be added in two places manually.
- What: Added `@register_object_relation` decorator backed by the added `ObjectRelationLibraryRegistry`. Added `@register_task` decorator backed by the added `TaskRegistry`.
- Leave TODO for `AT_POSE` and `IN` registry supports

### Note
- Added a Bool to block re-entry when using registration decorator. Without it, it triggers circular deps ImportError. It's caused by object relation registry.

### DIFF
--- a/isaaclab_arena/assets/registries.py
+++ b/isaaclab_arena/assets/registries.py
@@ -17,6 +17,7 @@ if TYPE_CHECKING:
     from isaaclab_arena.assets.teleop_device_base import TeleopDeviceBase
     from isaaclab_arena.policy.policy_base import PolicyBase
     from isaaclab_arena.relations.relations import RelationBase
+    from isaaclab_arena.tasks.task_base import TaskBase
 
 
 # Have to define all classes here in order to avoid circular import.
@@ -52,6 +53,7 @@ class Registry(metaclass=SingletonMeta):
                 PolicyRegistry,
                 HDRImageRegistry,
                 ObjectRelationLibraryRegistry,
+                TaskRegistry,
             ),
         ):
             ensure_assets_registered()
@@ -76,6 +78,7 @@ class Registry(metaclass=SingletonMeta):
                 PolicyRegistry,
                 HDRImageRegistry,
                 ObjectRelationLibraryRegistry,
+                TaskRegistry,
             ),
         ):
             ensure_assets_registered()
@@ -98,6 +101,7 @@ class Registry(metaclass=SingletonMeta):
                 PolicyRegistry,
                 HDRImageRegistry,
                 ObjectRelationLibraryRegistry,
+                TaskRegistry,
             ),
         ):
             ensure_assets_registered()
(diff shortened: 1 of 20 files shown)


isaaclab_arena/assets/registries.py
```diff
@@ -42,7 +44,18 @@ def is_registered(self, key: str) -> bool:
...
+        if isinstance(
+            self,
+            (
+                AssetRegistry,
+                DeviceRegistry,
+                RetargeterRegistry,
+                PolicyRegistry,
+                HDRImageRegistry,
+                ObjectRelationLibraryRegistry,
+                TaskRegistry,
```
alexmillane: It's not your code (it's my bad), but shall we take this chance to clean this up and define

```
REGISTRIES = (
    AssetRegistry,
    DeviceRegistry,
    RetargeterRegistry,
    PolicyRegistry,
    HDRImageRegistry,
    ObjectRelationLibraryRegistry,
    TaskRegistry,
)
```

then

```
        if isinstance(self, REGISTRIES):
            ensure_assets_registered()
```

isaaclab_arena/assets/registries.py
```diff
@@ -233,5 +306,9 @@ def ensure_assets_registered():
         import isaaclab_arena.assets.retargeter_library  # noqa: F401
         import isaaclab_arena.embodiments  # noqa: F401
         import isaaclab_arena.policy  # noqa: F401
+        import isaaclab_arena.relations.relations  # noqa: F401
+        import isaaclab_arena.tasks  # noqa: F401
 
         _assets_registered = True
+    finally:
+        _registration_in_progress = False
```
alexmillane, line 306: Nice. This issue has been hanging around since Arena day 1.

Thanks for fixing it.

isaaclab_arena/tasks/__init__.py
```diff
@@ -2,3 +2,21 @@
...
+    no_task,
+    open_door_task,
+    pick_and_place_task,
+    place_upright_task,
+    press_button_task,
+    rotate_revolute_joint_task,
+    sequential_task_base,
+    sorting_task,
+    turn_knob_task,
+)
```
alexmillane: 👍

isaaclab_arena/tests/test_task_registry.py
```diff
@@ -0,0 +1,37 @@
...
+        "TurnKnobTask",
+        "AssemblyTask",
+        "SortMultiObjectTask",
+    ]
+    registry = TaskRegistry()
+    for name in expected:
+        cls = registry.get_task_by_name(name)
+        assert cls.__name__ == name, f"{name} -> {cls.__name__}"
+    return True
+
```
alexmillane, line 33: We have these two objects that mirror each other:
- subclasses of `RelationBase`, that are registered with the `RelationRegistry`
- members of the `ArenaEnvGraphSpatialConstraintType` `enum`

Right now we have this constraint that all registered relations should have a corresponding enum item with the same name.

I feel it makes sense to test that that condition holds true over time. Could you add a test for that (obviously we'll have to exclude (manually) `at_pose` and `in` (for now)).

alexmillane (approved):
Looks great. Thanks for fixing an long-standing issue while doing this!

A few small comments. One requested test addition. Merge when you think you've addressed those.

isaaclab_arena/tests/test_task_registry.py
xyao-nv: updated

isaaclab_arena/assets/registries.py
xyao-nv: updated again based on what CI failed earlier

isaaclab_arena/assets/registries.py
xyao-nv: updated

PR #746, opened 2026-06-01
Add variation system base. Part 1/3, base classes, compiler, and two concrete variations

alexmillane:
## Summary
Adds the basic machinery for variations.

## Detailed description
- Variations are a structured API for introducing randomization into the simulation run.
- This change introduces two types of variation: run-time variation, and build-time variations:
  - **run-time variations:** Randomize per reset (implemented as a reset term)
  -  **build-time variations:** Randomize during construction of the environment (implemented as a mutation on an Arena object)
- Variations are implemented as a combination of
  - **Sampler:** Generates the random value
  - **Mutator:** Applies the random value to affect the change.
  - This design is chosen to allow use to tap in and record the random value, as it is applied.
- Concrete Variations
  - **Camera extrinsics:** Modifies a camera position by a sampled vector in the camera frame.
  - **HDR Image**: Selects from a passed list of HDR images.

## Not implemented (in follow up MRs)
- Recording the variations
- Control of the variations from the command line.


isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -72,6 +74,65 @@ def _solve_relations(self) -> None:
...
+        """
+        by_asset: dict[str, list[VariationBase]] = dict(self.arena_env.scene.get_asset_variations())
+        embodiment = self.arena_env.embodiment
+        if embodiment is not None:
+            embodiment_variations = embodiment.get_variations()
+            if embodiment_variations:
+                assert embodiment.name not in by_asset, (
+                    f"Embodiment name '{embodiment.name}' collides with a scene asset that also has variations; "
+                    "rename the scene asset or the embodiment so variation namespaces stay unique."
+                )
```
alexmillane: Remove long message. Infact remove message altogher to keep this on one line.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -72,6 +74,65 @@ def _solve_relations(self) -> None:
...
+                by_asset[embodiment.name] = list(embodiment_variations)
```
alexmillane: No need for `list`. `get_variations` already returns a list.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -72,6 +74,65 @@ def _solve_relations(self) -> None:
...
+        return by_asset
+
+    def _compose_variations_event_cfg(self) -> Any | None:
+        """Build a configclass with one :class:`EventTermCfg` per enabled run-time variation.
+
+        Returns ``None`` when no run-time variation is enabled. Build-time
+        variations are handled separately by :meth:`_apply_build_time_variations`.
+        """
+        fields: list[tuple[str, type, EventTermCfg]] = []
+        seen: set[str] = set()
```
alexmillane: `seen` -> `added_event_names`

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -72,6 +74,65 @@ def _solve_relations(self) -> None:
...
+                    "rename the scene asset or the embodiment so variation namespaces stay unique."
+                )
+                by_asset[embodiment.name] = list(embodiment_variations)
+        return by_asset
+
+    def _compose_variations_event_cfg(self) -> Any | None:
+        """Build a configclass with one :class:`EventTermCfg` per enabled run-time variation.
+
+        Returns ``None`` when no run-time variation is enabled. Build-time
+        variations are handled separately by :meth:`_apply_build_time_variations`.
```
alexmillane: Remove the comment about Build-time variations, that has nothing to do with this function.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -95,6 +156,10 @@ def compose_manager_cfg(self) -> IsaacLabArenaManagerBasedRLEnvCfg:
         if self.args.solve_relations:
             self._solve_relations()
 
+        # Apply build-time variations now: they mutate asset configs (e.g.
+        # DomeLight.spawner_cfg) and must run before scene_cfg is materialised.
```
alexmillane: Shorten comment to:

```
 # Apply build-time variations now, before scene_cfg is materialised.
```

isaaclab_arena/variations/__init__.py
alexmillane, line 1: Please simplify this whole file.

Let's just drop the module level imports. Don't support importing the whole variations folder.

isaaclab_arena/variations/camera_decalibration.py
```diff
@@ -0,0 +1,157 @@
...
+from isaaclab_arena.variations.uniform_sampler import UniformSamplerCfg
+from isaaclab_arena.variations.variation_base import RunTimeVariationBase, VariationBaseCfg
+
+if TYPE_CHECKING:
+    from isaaclab.envs import ManagerBasedEnv
+
+
+@configclass
+class CameraDecalibrationVariationCfg(VariationBaseCfg):
+    """Configuration for :class:`CameraDecalibrationVariation`.
```
alexmillane: -> Configuration for CameraDecalibrationVariation

isaaclab_arena/variations/camera_extrinsics_variation.py
alexmillane, line 1: Change file name to `camera_decalibration_variation.py`

isaaclab_arena/variations/camera_decalibration.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Attributes:
+        mode: Event mode. ``"reset"`` resamples each episode; ``"prestartup"``
+            picks a stable per-env offset that persists for the run.
+        sampler: 3D translation distribution in the camera's optical frame, axes
+            ``(x_right, y_down, z_forward)`` (ROS convention). Defaults to a
+            uniform over ``±5 mm`` per axis.
+    """
+
+    mode: str = "reset"
```
alexmillane: Let's remove the `mode` member from this and the other classes deriving from `VariationBaseCfg`. The reason is, at the moment all variations fire on reset. We can add this member back if we find that it's needed.

isaaclab_arena/variations/camera_decalibration.py
```diff
@@ -0,0 +1,157 @@
...
+@configclass
+class CameraDecalibrationVariationCfg(VariationBaseCfg):
+    """Configuration for :class:`CameraDecalibrationVariation`.
+
+    Attributes:
+        mode: Event mode. ``"reset"`` resamples each episode; ``"prestartup"``
+            picks a stable per-env offset that persists for the run.
+        sampler: 3D translation distribution in the camera's optical frame, axes
+            ``(x_right, y_down, z_forward)`` (ROS convention). Defaults to a
+            uniform over ``±5 mm`` per axis.
```
alexmillane: Can we remove this description of the attributes in the class-level docstring and replace them with docstrings per-member underneath?

isaaclab_arena/variations/camera_decalibration.py
```diff
@@ -0,0 +1,157 @@
...
+            high=[0.005, 0.005, 0.005],
+        )
+    )
+
+
+class CameraDecalibrationVariation(RunTimeVariationBase):
+    """Decalibrate a camera by adding a small offset to its nominal local position.
+
+    Only the camera's local transform is touched, so wrist-mounted cameras keep
+    tracking their parent body. Realised as a run-time event term.
```
alexmillane: remove: "Realised as a run-time event term."

isaaclab_arena/variations/camera_decalibration.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Args:
+        camera_name: Scene-entity name of the target camera. Must resolve to a
+            :class:`Camera` or :class:`TiledCamera` in ``env.scene``.
```
alexmillane: Remove: "Must resolve to a :class:`Camera` or :class:`TiledCamera` in ``env.scene``."

isaaclab_arena/variations/camera_decalibration.py
```diff
@@ -0,0 +1,157 @@
...
+        cfg: Tunable parameters. Defaults to a ``±5 mm`` per-axis uniform
+            sampler resampled on every reset.
```
alexmillane: Remove: "Defaults to a ``±5 mm`` per-axis uniform sampler resampled on every reset."

isaaclab_arena/variations/camera_decalibration.py
```diff
@@ -0,0 +1,157 @@
...
+            f"scene['{asset_cfg.name}']; got {type(camera).__name__}."
+        )
+        assert tuple(sampler.event_shape) == (3,), (
+            "decalibrate_camera_from_sampler expects a sampler with event_shape (3,) over XYZ; "
+            f"got {tuple(sampler.event_shape)}."
+        )
+
+        self._camera = camera
+        # Snapshotted on first ``__call__`` (the camera's view is only
+        # initialised by the sensor's lifecycle hooks at that point).
```
alexmillane: Simply comment to: "Snapshotted on first ``__call__``"

isaaclab_arena/variations/camera_decalibration.py
```diff
@@ -0,0 +1,157 @@
...
+
+        camera = env.scene[asset_cfg.name]
+        assert isinstance(camera, (Camera, TiledCamera)), (
+            "decalibrate_camera_from_sampler expects a Camera or TiledCamera at "
+            f"scene['{asset_cfg.name}']; got {type(camera).__name__}."
+        )
+        assert tuple(sampler.event_shape) == (3,), (
+            "decalibrate_camera_from_sampler expects a sampler with event_shape (3,) over XYZ; "
+            f"got {tuple(sampler.event_shape)}."
+        )
```
alexmillane: change `Sampler.event_shape` -> `Sampler.shape`

isaaclab_arena/variations/categorical_sampler.py
```diff
@@ -0,0 +1,42 @@
...
+    """Uniform categorical sampler returning items drawn from a per-call ``choices`` sequence.
+
+    Each call to :meth:`sample` returns a ``list`` of length ``num_samples``,
+    where each item is uniformly drawn (with replacement) from ``choices``.
+    """
+
+    def _sample(self, num_samples: int, choices: Sequence[Any], **kwargs) -> list[Any]:  # noqa: ARG002
+        assert num_samples >= 0, f"num_samples must be non-negative; got {num_samples}."
+        assert len(choices) >= 1, "CategoricalSampler requires a non-empty 'choices' sequence."
+        indices = torch.randint(low=0, high=len(choices), size=(num_samples,), dtype=torch.long)
```
alexmillane: Remove: `dtype=torch.long`

isaaclab_arena/variations/hdr_image_variation.py
```diff
@@ -0,0 +1,90 @@
...
+@configclass
+class HDRImageVariationCfg(VariationBaseCfg):
+    """Configuration for :class:`HDRImageVariation`.
+
+    Attributes:
+        hdr_names: Registered HDR names to sample from; empty means sample over
+            every registered HDR.
+        sampler: Categorical distribution over the resolved HDR pool. The
+            default works out of the box (the pool is supplied at
+            :meth:`HDRImageVariation.apply` time).
```
alexmillane: Simplify comment to: "Categorical distribution over the resolved HDR pool."

alexmillane: Move attribute descriptions to below, attached to each attribute.

isaaclab_arena/variations/hdr_image_variation.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    hdr_names: list[str] = field(default_factory=list)
+    sampler: CategoricalSamplerCfg = field(default_factory=CategoricalSamplerCfg)
```
alexmillane: `sampler` -> `sampler_cfg`

isaaclab_arena/variations/camera_decalibration.py
```diff
@@ -0,0 +1,157 @@
...
+    Attributes:
+        mode: Event mode. ``"reset"`` resamples each episode; ``"prestartup"``
+            picks a stable per-env offset that persists for the run.
+        sampler: 3D translation distribution in the camera's optical frame, axes
+            ``(x_right, y_down, z_forward)`` (ROS convention). Defaults to a
+            uniform over ``±5 mm`` per axis.
+    """
+
+    mode: str = "reset"
+    sampler: UniformSamplerCfg = field(
```
alexmillane: `sampler` -> `sampler_cfg`

isaaclab_arena/variations/hdr_image_variation.py
```diff
@@ -0,0 +1,90 @@
...
+        from isaaclab_arena.assets.registries import HDRImageRegistry  # noqa: PLC0415
+
+        registry = HDRImageRegistry()
+        if self.cfg.hdr_names:
+            for name in self.cfg.hdr_names:
+                assert registry.is_registered(name), (
+                    f"HDRImageVariation: HDR name '{name}' is not registered. "
+                    f"Registered HDRs: {sorted(registry.get_all_keys())}."
+                )
+            hdr_names = list(self.cfg.hdr_names)
```
alexmillane: why do we need `list`. The cfg contains a list for this field. Suggestion to remove `list`.

isaaclab_arena/variations/hdr_image_variation.py
```diff
@@ -0,0 +1,90 @@
...
+        else:
+            hdr_names = registry.get_all_keys()
+            assert hdr_names, "HDRImageVariation: no HDRs are registered; cannot sample."
+
+        assert self.sampler is not None, "HDRImageVariation: sampler not set."
+        # Pass HDR *names* (not indices) as the categorical sampler's choices,
+        # so the recorder logs the chosen HDR by name instead of an opaque
+        # index.
+        [chosen_name] = self.sampler.sample(num_samples=1, choices=hdr_names)
```
alexmillane: -> hdr_name = self.sampler.sample(num_samples=1, choices=hdr_names)[0]

isaaclab_arena/variations/sampler_base.py
```diff
@@ -0,0 +1,81 @@
...
+
+"""Sampler abstract base for the variation system.
+
+A :class:`SamplerBase` is a stateless description of how values are drawn,
+paired with a declarative :class:`SamplerBaseCfg` that ``build()`` s into it.
+
+:meth:`SamplerBase.sample` returns ``Any``: continuous samplers return a
+``torch.Tensor`` of shape ``(num_samples, *event_shape)``; categorical samplers
+return a ``list``. Listeners should handle either shape.
+"""
```
alexmillane: Remove

isaaclab_arena/variations/sampler_base.py
```diff
@@ -0,0 +1,81 @@
...
+    def add_listener(self, listener: Callable[[Any], None]) -> None:
+        """Register ``listener`` to receive every sample drawn from this sampler.
+
+        Called synchronously inside :meth:`sample` with the raw value (no copy).
+        """
+        self._listeners.append(listener)
+
+    def remove_listener(self, listener: Callable[[Any], None]) -> None:
+        """Remove a previously-registered ``listener``."""
+        self._listeners.remove(listener)
```
alexmillane: Remove stuff to do with listeners. This will be handled in a follow up MR.

isaaclab_arena/variations/sampler_base.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class SamplerBase(ABC):
+    """Abstract distribution over values.
+
+    Stateless: holds distribution parameters but no RNG state. Subscribe via
+    :meth:`add_listener` to observe every value drawn, or
+    :meth:`~isaaclab_arena.variations.variation_base.VariationBase.add_sample_listener`
+    to survive sampler swaps.
+    """
```
alexmillane: Replace docstring with:

"""Baseclass for all samplers. Stateless"""

isaaclab_arena/variations/sampler_base.py
```diff
@@ -0,0 +1,81 @@
...
+        Args:
+            num_samples: Number of independent samples to draw.
+            **kwargs: Extra per-call arguments forwarded to the concrete sampler
+                (e.g. ``choices`` for a categorical sampler).
+
+        Returns:
+            A tensor of shape ``(num_samples, *event_shape)`` for tensor
+            samplers, or a ``list`` of length ``num_samples`` for categorical
+            samplers.
+        """
```
alexmillane: Remove stuff about listeners which is being removed from this MR.

isaaclab_arena/variations/sampler_base.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def build(self) -> SamplerBase:
+        """Return the live :class:`SamplerBase` described by this cfg."""
+        raise NotImplementedError(
+            f"{type(self).__name__}.build() is not implemented; every concrete SamplerBaseCfg "
+            "subclass must provide a build() that returns its live SamplerBase."
+        )
+
+
+class SamplerBase(ABC):
```
alexmillane, line 28: Move the `event_shape` method into this base class. Just call it `SamplerBase.shape() -> int`, an abstract method.

isaaclab_arena/variations/uniform_sampler.py
```diff
@@ -0,0 +1,66 @@
...
+
+@configclass
+class UniformSamplerCfg(SamplerBaseCfg):
+    """Config for :class:`UniformSampler`.
+
+    Attributes:
+        low: Lower bound per dimension. Length determines the sampler's
+            ``event_shape``.
+        high: Upper bound per dimension. Must have the same length as ``low``
+            and be element-wise ``>= low``.
```
alexmillane: Move attribute comments below near the actual attributes.

isaaclab_arena/variations/uniform_sampler.py
```diff
@@ -0,0 +1,66 @@
...
+        low_t = torch.as_tensor(low, dtype=torch.float32)
+        high_t = torch.as_tensor(high, dtype=torch.float32)
+        assert (
+            low_t.shape == high_t.shape
+        ), f"UniformSampler low/high must have matching shape; got {tuple(low_t.shape)} vs {tuple(high_t.shape)}."
+        assert torch.all(
+            low_t <= high_t
+        ), f"UniformSampler requires low <= high elementwise; got low={low_t}, high={high_t}."
+        self.low = low_t
+        self.high = high_t
```
alexmillane: suggestion to remove `low_t` and `high_t` and just directly assign to `self.low` and `self.high` in the first lines, and then perform the checks on those.

isaaclab_arena/variations/uniform_sampler.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @property
+    def event_shape(self) -> torch.Size:
+        """Shape of a single sample."""
+        return self.low.shape
```
alexmillane, line 50: Call method `shape_per_sample` (and add to base class).

isaaclab_arena/variations/sampler_base.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def remove_listener(self, listener: Callable[[Any], None]) -> None:
+        """Remove a previously-registered ``listener``."""
+        self._listeners.remove(listener)
+
+    def sample(self, num_samples: int, **kwargs) -> Any:
+        """Draw ``num_samples`` values, notify listeners, and return them.
+
+        Args:
+            num_samples: Number of independent samples to draw.
```
alexmillane: Add: "typically the number of environments we're drawing a sample for."

isaaclab_arena/variations/variation_base.py
```diff
@@ -0,0 +1,171 @@
...
+from isaaclab_arena.variations.sampler_base import SamplerBase, SamplerBaseCfg
+
+
+@configclass
+class VariationBaseCfg:
+    """Base configclass for :class:`VariationBase` instances.
+
+    Attributes:
+        enabled: Whether the variation is applied. Defaults to ``False`` (opt in
+            via :meth:`VariationBase.enable` or a cfg override).
```
alexmillane: move attribute doc-string below near the attribute itself.

isaaclab_arena/variations/variation_base.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        self._sample_listeners.append(listener)
+        if self._sampler is not None:
+            self._sampler.add_listener(listener)
+
+    def remove_sample_listener(self, listener: Callable[[Any], None]) -> None:
+        """Unsubscribe a previously-registered ``listener``."""
+        self._sample_listeners.remove(listener)
+        if self._sampler is not None:
+            self._sampler.remove_listener(listener)
```
alexmillane: Remove everything to do with listeners. Will be added in follow up MR.

isaaclab_arena/variations/variation_base.py
```diff
@@ -0,0 +1,171 @@
...
+    #: Short, unique identifier for this variation kind (e.g. ``"color"``).
+    name: ClassVar[str]
+
+    #: The configclass instance holding this variation's tunable parameters.
+    cfg: VariationBaseCfg
+
+    def __init__(self, cfg: VariationBaseCfg):
+        self.cfg = cfg
+        self._sampler: SamplerBase | None = None
+        self._sample_listeners: list[Callable[[Any], None]] = []
```
alexmillane: remove

isaaclab_arena/variations/variation_base.py
```diff
@@ -0,0 +1,171 @@
...
+
+    enabled: bool = False
+
+
+class VariationBase(ABC):
+    """Abstract variation binding an asset, a sampler, and a realisation hook.
+
+    Concrete subclasses declare a class-level :attr:`name`, pair with a
+    :class:`VariationBaseCfg` subclass, and inherit one of the flavored bases
+    (:class:`RunTimeVariationBase` or :class:`BuildTimeVariationBase`).
```
alexmillane: Remove the longer docstring and just leave first line.

isaaclab_arena/variations/variation_base.py
```diff
@@ -0,0 +1,171 @@
...
+        ``super().apply_cfg(cfg)`` first.
+
+        Args:
+            cfg: A cfg of the :class:`VariationBaseCfg` subclass this variation
+                accepts.
+        """
+        self.cfg = cfg
+        sampler_cfg = getattr(cfg, "sampler", None)
+        if isinstance(sampler_cfg, SamplerBaseCfg):
+            self.set_sampler(sampler_cfg)
```
alexmillane: This indicates that `VariationBaseCfg` needs a compulsory sampler field.

isaaclab_arena/variations/variation_base.py
```diff
@@ -0,0 +1,171 @@
...
+
+    Use when the underlying property can be flipped during simulation (e.g.
+    visual color, initial pose, mass).
+    """
+
+    @abstractmethod
+    def build_event_cfg(self) -> tuple[str, EventTermCfg]:
+        """Return the ``(name, cfg)`` event term that realises this variation.
+
+        The name must be unique across all enabled variations in the scene.
```
alexmillane: remove.

alexmillane (commented):
Self-review. First cursory pass.

isaaclab_arena/environments/isaaclab_arena_manager_based_env_cfg.py
```diff
@@ -66,9 +64,6 @@ class IsaacLabArenaManagerBasedRLEnvCfg(ManagerBasedRLEnvCfg):
 
     metrics: object | None = None
 
-    # Isaaclab Arena Env. Held as a member to allow use of internal functions
-    isaaclab_arena_env: IsaacLabArenaEnvironment | None = None
```
xyao-nv, line 70: by accident?

isaaclab_arena/variations/uniform_sampler.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @property
+    def shape_per_sample(self) -> torch.Size:
+        """Shape of a single sample."""
+        return self.low.shape
+
+    def _sample(self, num_samples: int, **kwargs) -> torch.Tensor:  # noqa: ARG002
+        assert num_samples >= 0, f"num_samples must be non-negative; got {num_samples}."
+        shape = (num_samples, *self.shape_per_sample)
+        u = torch.rand(shape)
```
xyao-nv, line 64: Do we assume it's using global RNG state set by the placer?
Shall we unify all rand seed in our stack such that it uses a system-wide seed for everything?

isaaclab_arena/variations/camera_extrinsics_variation.py
```diff
@@ -0,0 +1,162 @@
...
+from isaaclab_arena.variations.sampler_base import SamplerBase
+from isaaclab_arena.variations.uniform_sampler import UniformSamplerCfg
+from isaaclab_arena.variations.variation_base import RunTimeVariationBase, VariationBaseCfg
+
+if TYPE_CHECKING:
+    from isaaclab.envs import ManagerBasedEnv
+
+
+@configclass
+class CameraExtrinsicsVariationCfg(VariationBaseCfg):
```
xyao-nv, line 37: Not sure if you plan it for the future, to add a variation registry.
Is there any way for users to view "what variations exist at run time or build time" and choose accordingly?

isaaclab_arena/variations/camera_extrinsics_variation.py
```diff
@@ -0,0 +1,162 @@
...
+    Only the camera's local transform is touched, so wrist-mounted cameras keep
+    tracking their parent body.
+
+    Args:
+        camera_name: Scene-entity name of the target camera.
+        cfg: Tunable parameters. Override the translation distribution via
+            ``cfg.sampler_cfg``.
+    """
+
+    name = "camera_extrinsics"
```
xyao-nv: My understanding is your `Asset.add_variation()` stores it as `self.variations[variation.name] = variation`.
Does it mean a robot with `num_cams` (`num_cams`>1) cannot carry `num_cams` extrinsic variations at the same time?

But `event_name = f"{self.camera_name}_extrinsics_variation"` is per-camera. Anything I missed?

isaaclab_arena/variations/camera_extrinsics_variation.py
```diff
@@ -0,0 +1,162 @@
...
+        env: ManagerBasedEnv,  # noqa: ARG002
+        env_ids: torch.Tensor,
+        asset_cfg: SceneEntityCfg,  # noqa: ARG002
+        sampler: SamplerBase,
+    ):
+        view = self._camera._view
+        assert view is not None, "Camera view was not initialized."
+        if self._t_parent_C_in_parent is None:
+            # NOTE(alexmillane): in the version of Isaac Lab that we're using, the get_local_poses()
+            # method claims to return w,x,y,z, but it actually returns x,y,z,w.
```
xyao-nv, line 142: ... amazing

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -177,8 +235,6 @@ def compose_manager_cfg(self) -> IsaacLabArenaManagerBasedRLEnvCfg:
             task.get_commands_cfg(),
         )
 
-        isaaclab_arena_env = self.arena_env
```
xyao-nv, line 181: by accident?

isaaclab_arena/variations/hdr_image_variation.py
```diff
@@ -0,0 +1,79 @@
...
+                    f"Registered HDRs: {sorted(registry.get_all_keys())}."
+                )
+            hdr_names = self.cfg.hdr_names
+        else:
+            hdr_names = registry.get_all_keys()
+            assert hdr_names, "HDRImageVariation: no HDRs are registered; cannot sample."
+
+        assert self.sampler is not None, "HDRImageVariation: sampler not set."
+        # Pass HDR names as the categorical sampler's choices.
+        hdr_name = self.sampler.sample(num_samples=1, choices=hdr_names)[0]
```
xyao-nv, line 78: Will this hdr_name be saved into some log for eval results?
Basically user can know Run1 -> Use HDR-A, Run 2: Used HDR-B. So if they want to repro, they can directly select the exact HDR for that run.

xyao-nv (approved):

isaaclab_arena/embodiments/franka/franka.py
```diff
@@ -39,8 +39,9 @@
 from isaaclab_arena.embodiments.embodiment_base import EmbodimentBase
 from isaaclab_arena.embodiments.franka.observations import gripper_pos
 from isaaclab_arena.utils.pose import Pose
+from isaaclab_arena.variations.camera_extrinsics_variation import CameraExtrinsicsVariation
 
-_DEFAULT_CAMERA_OFFSET = Pose(position_xyz=(0.11, -0.031, -0.074), rotation_xyzw=(0.0, 0.0, -0.66262, -0.74896))
+_DEFAULT_CAMERA_OFFSET = Pose(position_xyz=(0.11, -0.031, -0.074), rotation_xyzw=(0.0, 0.0, 0.70711, 0.70711))
```
cvolkcvolk, line 44: Is this an intentional change?

isaaclab_arena/variations/camera_extrinsics_variation.py
```diff
@@ -0,0 +1,162 @@
...
+            # I am testing this in the test test_isaaclab_bug_get_local_poses.py.
```
cvolkcvolk, line 143: Thanks for guarding this

isaaclab_arena/environments/arena_env_builder.py
cvolkcvolk: Looks like this PR is based on [this refactor](https://github.com/isaac-sim/IsaacLab-Arena/pull/754/changes)

isaaclab_arena/assets/asset.py
```diff
@@ -17,3 +24,24 @@ def __init__(self, name: str, tags: list[str] | None = None, **kwargs):
...
         self.tags = tags
+        self.variations: dict[str, VariationBase] = {}
+
+    def add_variation(self, variation: VariationBase) -> None:
+        """Attach a variation under its class-level ``name``, replacing any existing one.
+
+        Subclasses call this from their ``__init__`` to declare the variations
+        they support.
+        """
+        self.variations[variation.name] = variation
```
cvolkcvolk, line 38: `variation.name` is a class level name. Assuming we have two variations of the same class, for example

```
add_variation(CameraExtrinsicsVariation("wrist_cam")) 
add_variation(CameraExtrinsicsVariation("external_cam"))
```
We would silently overwrite the first one? 
Suggestion to assert in that case

isaaclab_arena/variations/uniform_sampler.py
```diff
@@ -0,0 +1,59 @@
...
+        assert torch.all(
+            self.low <= self.high
+        ), f"UniformSampler requires low <= high elementwise; got low={self.low}, high={self.high}."
+
+    @property
+    def shape_per_sample(self) -> torch.Size:
+        """Shape of a single sample."""
+        return self.low.shape
+
+    def _sample(self, num_samples: int, **kwargs) -> torch.Tensor:  # noqa: ARG002
```
cvolkcvolk: I feel like we are putting too much on the base sampler class: we're forcing two samplers with different different inputs and different outputs, requiring a pretty vaque `sample(num_samples, **kwargs) -> Any`. 
Which we then need to carry over into the child classes (the `**kwargs` argument)... And then supress the linter on the unused argument with `noqa: ARG002` again.
For example `choices` is passed through `**kwargs` in the `CategoricalSampler` while `UniformSampler._sample`, `**kwargs` is never used, but to match the base signature. 
And on the return type `sample()` returns `Any`, a tensor for UniformSampler and a list for CategoricalSampler.


What if we split the base sampler into two, one for each family it covers, continous or discrete like

```
  class ContinuousSampler(ABC):
      """Draws continuous numeric values from a distribution (uniform, normal, ...)."""
      def sample(self, num_samples: int) -> torch.Tensor: ...
      @property
      def shape_per_sample(self) -> torch.Size: ...

  class ChoiceSampler(ABC, Generic[T]):
      """Picks discrete items from a per-call pool."""
      def sample(self, num_samples: int, choices: Sequence[T]) -> list[T]: ...
```

This way we could stay descriptive on the expected input (and output).

cvolkcvolk (approved):
Looks great!

isaaclab_arena/variations/uniform_sampler.py
alexmillane: So. I hear you. And this is correct, in the scope of this MR. I agree with you.

The reason to push everything through a single base class is based on Part 3/3 of this MR series - the recording of the samples.

In part 3/3, the base class takes care of recording the samples to a ledger, which is then used to output the results of sampling. Forcing all calls through a single `sample` function, and delegating the actual sampling to subclasses is how we achieve consistent recording across all samplers.

Perhaps though, there's a nicer way to implement this and retain the descriptiveness of the base classes you suggest. Maybe a decorator rather than a base class function is more appropriate.

I'll make the change that you suggest. And we can deal with the recording later.

isaaclab_arena/environments/arena_env_builder.py
alexmillane: Yea I separated out that refactor into its own MR. Thanks for catching this though.

isaaclab_arena/variations/camera_extrinsics_variation.py
alexmillane: It was very painful to catch that :sweat:

isaaclab_arena/environments/isaaclab_arena_manager_based_env_cfg.py
alexmillane: see below, I separated this change into a now-merged MR. Good catch.

isaaclab_arena/variations/uniform_sampler.py
alexmillane: Hmm good catch. I feel like your right, we need to centralize our randomness.

A brief search of the code shows that we're setting the random seed in some places and using generators in others.

We have this function `set_seed` which I guess is our attempt at centralization.

If you don't mind I'd suggest we rely on `set_seed` here for now, and try to centralize all calls to rand in a later MR?

isaaclab_arena/variations/camera_extrinsics_variation.py
alexmillane: Good point.

In part 2/3 of this series I'm adding the CLI part of variations. I think I will add a flag that dumps all the available variations to the CLI. Does that address what you're thinking?

isaaclab_arena/variations/camera_extrinsics_variation.py
alexmillane: great catch!

isaaclab_arena/variations/hdr_image_variation.py
alexmillane: Part 3/3 of this series is dealing with recording the samples :)

isaaclab_arena/embodiments/franka/franka.py
alexmillane: It was. But maybe I should back it out.

I used it while debugging the wxyz vs. xyzw bug to make the numbers a little easier to interpret.

I left it in, because I'm not sure why the quaternion is slightly off what would be a normal angle.

This indicates a slight rotation around the camera optical axis. 8 degrees off 90 degrees (which is what I change it to)

I can't see how that could make sense, motivating the change.

isaaclab_arena/variations/camera_extrinsics_variation.py
alexmillane: no worries.

isaaclab_arena/assets/asset.py
alexmillane: Mmm this is a bug. Good catch 😬

PR #747, opened 2026-06-01
Guidance on docstrings in AGENTS.md

alexmillane:
## Summary
Add docstring guidelines to `AGENTS.md`

## Detailed description
- Copies in guidelines from Google style (modified slightly to match our style).


AGENTS.md
```diff
@@ -40,6 +40,57 @@ Use the `dev-container` skill for build, start, attach, and exec inside the cont
 - Prefer `assert condition, "message"` over `if not condition: raise ValueError("message")` for internal invariant checks. (Formatting, imports, and typing are enforced by `pre-commit` — see `.pre-commit-config.yaml`.)
 - PR bodies follow `.github/pull_request_template.md` — a one-line Summary plus 2–5 detail bullets. Resist the agent default of long, multi-section descriptions.
 
+## Coding Style
```
cvolkcvolk: We already have a coding style section just above. Suggestion to remove.

AGENTS.md
```diff
@@ -40,6 +40,57 @@ Use the `dev-container` skill for build, start, attach, and exec inside the cont
...
+
+### Checks/Asserts
+
+In Arena we prefer asserts over if...raises in most cases where such an check represents a coding error
```
cvolkcvolk: Looks like a duplicate of what we aleady state above on line 40, eg 
`- Prefer `assert condition, "message"` over `if not condition: raise ValueError("message")` for internal invariant checks. (Formatting, imports, and typing are enforced by `pre-commit` — see `.pre-commit-config.yaml`.)`

AGENTS.md
```diff
@@ -40,6 +40,57 @@ Use the `dev-container` skill for build, start, attach, and exec inside the cont
...
+(i.e) something that shouldn't be recovered from. Asserts are briefer than the exception based counterpart.
+The raises-based approach is more verbose (for example it always uses multiple lines), and because
+this type of error is not intended to be recovered from, it offers no advantage.
+
+### Docstrings
```
cvolkcvolk: I feel like the follwing is a pretty verbose paragraph to describe how to write (good) docstrings. Slightly ironic given the existing doc explicitly states "`Resist the agent default of long, multi-section descriptions.`" :) 
It's generally recommended to keep the agents.md at or below 150-300 lines of text as it might add to general context rot in other areas. This section alone would add a considerably large part.

Did you have any luck in adding a concise section to the coding style section like
```
- Docstrings: Google style (https://google.github.io/styleguide/pyguide.html#383-functions-and-methods).
- Prefer one line; a 2–3 line paragraph may follow if needed.
- Document `Args` and `Returns`, but **not** `Raises`. Omit `Returns` when it only returns None
    or the summary already covers it.
```
maybe with a short example?

cvolkcvolk: Eg we refer to the google styleguide (Edit: maybe without linking the URL) and then only state our deltas explicitely

AGENTS.md
```diff
@@ -40,6 +40,57 @@ Use the `dev-container` skill for build, start, attach, and exec inside the cont
...
+
+Prefer short docstrings. Where a single line is not sufficient, a short (i.e. 2-3 line)
+paragraph can follow. Document args and returns, but not raises.
+We follow google-style for docstrings which is repeated below:
```
xyao-nv: google-style -- URL?

xyao-nv (approved):

AGENTS.md
alexmillane: I think it's best to avoid the URL, in case the agent decides to load the whole webpage into context. For now I've removed references to google style (I can't think of a way of including it with a small amount of text) and summarized the guidance contained in google style.

AGENTS.md
alexmillane: I've adopted something short and similar to the above. no URL.

AGENTS.md
alexmillane: good catch.

PR #719, opened 2026-05-26
Generate html from EnvGraphSpec yaml for user review

qianl-nv:
## Summary
First part of the LLM env gen pipeline: add static html view of the LLM-generated initial scene graph proposal

## Detailed description
- Reason: Provide user with an intuitive way to visualize the env graph yaml
- Changed: Add a review_graph which takes yaml input and generate html with graph viz and node snapshots
- Impact: Adds the user facing component for the agentic env gen part 1

To run:
python -m isaaclab_arena.llm_env_gen.review_graph  --yaml isaaclab_arena/tests/test_data/pick_and_place_maple_table_env_graph.yaml --render-thumbnails

PR #753, opened 2026-06-03
Resolve initial state specs to full state specs in env graph

xyao-nv:
## Summary
Resolve initial state specs to full state specs in env graph

## Detailed description

- Why:
1. upstream produces only nodes, the task list, and the initial state estimation(state_spec_0)
2. Populating it to intermediate & final states to reflect how Env State changes along the task execution
3. Note that those Env State changes may pose constraints on the initial Env placement.

- What:

1. Add a declarative `TaskTransition` model such that it describes what effect a task causes to the Env if success condition is met. E.g. `Relocate` to another spatial relation, or `SetState` to another state for embodiment/object itself. Each task class has an optional `success_state_transition` and can be access thru the `TaskRegistry`.
2. Add `resolve_constraints` method in `ArenaEnvGraphSpec`: it chains each task's declared success condition onto the previous state to derive the missing states and wire every task. Topology is implicit in the sequential list, so N tasks yield N+1 state specs. 
3. Add a constraint update mechanism, such that unrelated ones are kept, related ones are updated from dropped based on the task state transition. Also include reachability of both a state's postcondition and precondition target.

## APIs
```
unresolved_graph_spec  = ArenaEnvGraphSpec.from_yaml('<unresolved_yaml_path>', is_task_wiring_enabled=False)
full_graph_spec = unresolved_graph_spec.resolve_constraints()
```

## Known limitation/ TODOs 
- Only work for sequential tasks so far
- Only implemented `Relocate` realization for PnP task using "hardcode" object ON destination


isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -101,6 +104,25 @@ def validate(self) -> None:
         assert_references_exist(self.nodes, self.tasks, self.state_specs)
         assert_spatial_constraint_shapes(self.state_specs)
 
+    def to_dict(self) -> dict[str, Any]:
```
qianl-nv, line 76: dataclass has a .asdict method. you need to implement special handing for enum, but others can be handled automatically. can avoid manual serialization with _node_to_dict/_task_to_dict

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -129,6 +151,50 @@ def to_arena_env(self) -> "IsaacLabArenaEnvironment":
         return build_arena_env_from_graph_spec(self)
 
 
+@dataclass
```
qianl-nv: do we really need a separate graph / task spec specifically for unresolved graph?

when the agent outputs the task list, the # of intermediate states and how they are connected by the task is already known (given sequential tasks). what is "unresolved" is the complete set of constrains on each state (including the initial state). 

thus "resolving" does not change the graph topology, it only modifies the constraints on each state spec. 

I feel it's simpler to just have a .resolve_constraints method on the ArenaEnvGraphSpec class. Defining a new UnresolvedArenaEnvGraphSpec / UnresolvedArenaEnvGraphTaskSpec class feels overkill for this.

isaaclab_arena/agentic_environment_generation/state_spec_resolver.py
```diff
@@ -0,0 +1,182 @@
...
+        num_tasks = len(tasks_in)
+        for i, task in enumerate(tasks_in):
+            new_state_id = f"state_spec_{i + 1}"
+            is_terminal = i == num_tasks - 1
+            if is_terminal:
+                # The terminal state of the chain reaches the last task's reach_target_on_success.
+                reach_target = _transition_for(task).reach_target_on_success
+            else:
+                # While task i+1 exists, the next state of the chain reaches its subject.
+                reach_target = _transition_for(tasks_in[i + 1]).subject
```
qianl-nv: I think we also need to check reach_target_on_success on intermediate states, not just the terminal state. if we have (s0) -> open door (s1) -> pick & place (s2) -> close door (3), we need to check the place target is reachable at the (s2) state, not (s3) state? Hope I'm not missing any logic here.

qianl-nv (commented):
Thanks for making the change.

I've been playing with unifying IntentSpec with the GraphSpec today and realize the bigger blocker is actually pydantic vs dataclass. Pydantic BaseModel provides much bette default methods for validation/schema generation and is a lot more LLM-interface friendly.

I make a MR to move the whole ArenaEnvGraphSpec to pydantic https://github.com/isaac-sim/IsaacLab-Arena/pull/757/changes

It should also simply a lot of your serialization code in this MR.

isaaclab_arena/environments/arena_env_graph_spec.py
xyao-nv: this is removed after rebase

isaaclab_arena/environments/arena_env_graph_spec.py
xyao-nv: yeah I agree. Updated to reuse  ArenaEnvGraphSpec class while relaxing task wiring related logics in loading.

isaaclab_arena/agentic_environment_generation/state_spec_resolver.py
xyao-nv: added. you are correct

qianl-nv: thanks!

qianl-nv (approved):
Thanks for rebasing on the pydantic refacotor MR. This change is now much simpler. LGTM!

isaaclab_arena/agentic_environment_generation/state_spec_resolver.py
```diff
@@ -0,0 +1,218 @@
...
+        # destination), and -- when a next task exists -- that task's subject (the next thing to act on).
+        reach_targets_postcondition = [transitions[i].reach_target_on_success]
+        reach_targets_precondition = []
+        # final state has no precondition.
+        if not is_final_state:
+            reach_targets_precondition = [transitions[i + 1].subject]
+        reach_targets = reach_targets_postcondition + reach_targets_precondition
+        states.append(
+            _get_next_state_spec(
+                prev_state=states[-1],
```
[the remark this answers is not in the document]
xyao-nv: it's fixed

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -129,6 +153,18 @@ def write_yaml(self, path: str | Path) -> None:
         with Path(path).open("w", encoding="utf-8") as f:
             yaml.safe_dump(self.to_dict(), f, sort_keys=False)
 
+    def resolve_constraints(self, env_name: str | None = None) -> ArenaEnvGraphSpec:
```
xyao-nv: I dont like this. To be changed

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -129,6 +154,64 @@ def write_yaml(self, path: str | Path) -> None:
...
+        """Chain this unresolved graph's task success conditions into its full state specs.
+
+        The upstream emits a partially-wired graph: nodes, the ordered task list, and only the initial
```
alexmillane: This is a super general, and perhaps embarrassingly basic, question: but what is our graph here?

Is the graph:
- **nodes:** objects
- **edges:** spatial relations.

In what sense is this graph partially wired?

As we move through the time/tasks, are we generating more graphs, one for each (sub)task, or are we referring to the whole thing as a single higher-order graph?

In general, is there some way we can document (or better visualize), the graph we're referring to here.

These comments are great, thank you for putting so much effort into explaining things. I think with a diagram(?) to explain the graph (i.e. the basics) this will all make more sense (to me).

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -129,6 +154,64 @@ def write_yaml(self, path: str | Path) -> None:
...
+
+        Returns a fully-wired, strict-validated ``ArenaEnvGraphSpec``.
+        """
+        assert len(self.state_specs) == 1, (
+            f"unresolved graph must define exactly the initial state (state_spec_0); got {len(self.state_specs)} state"
+            " specs"
+        )
+        embodiment_id = _get_embodiment_id_from_nodes(self.nodes)
+
+        transitions = [_get_task_state_transition(task) for task in self.tasks]
```
alexmillane: consider a type annotation here as you've done in the lines below. That's super helpful..

isaaclab_arena/tasks/task_transition.py
```diff
@@ -0,0 +1,62 @@
...
+  a knob level, or the embodiment opens/closes gripper).
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any
+
+
+@dataclass()
```
alexmillane: nit: suggestion to remove the trailing open-close bracket. We don't dont do that elsewhere.

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -155,3 +238,131 @@ def to_arena_env(self) -> IsaacLabArenaEnvironment:
...
+                    id=f"{new_state_id}_{embodiment_id.split('_')[0]}_reach_{reach_target}",
+                    type=ArenaEnvGraphTaskConstraintType.REACH,
+                    parent=embodiment_id,
+                    child=reach_target,
+                )
+            )
+    return ArenaEnvGraphStateSpec(
+        id=new_state_id, spatial_constraints=spatial_constraints, task_constraints=task_constraints
+    )
+
```
alexmillane: Suggestion to break this long function up into smaller parts. It's composed of three for loops so perhaps three functions?

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -132,6 +134,19 @@ def get_viewer_cfg(self) -> ViewerCfg:
...
+    def success_state_transition(cls, task_args: Mapping[str, Any]) -> TaskTransition:
+        """Success (``object_on_destination``): the picked object ends up a relation with the destination."""
+        pick_object = task_args["pick_up_object"]
+        destination_location = task_args["destination_location"]
+        # Note: with the current AABB-based object solver, placing an object ``on`` an open container
+        # and letting it fall is equivalent to it being ``in`` the container, so a single ``on``
+        # relation covers both surfaces and containers.
+        return TaskTransition(
+            subject=pick_object,
+            effects=(Relocate(subject=pick_object, relation="on", target=destination_location),),
```
alexmillane: This string "on" is constrained to equal `On.name`?

Suggestion to return `On.name`

isaaclab_arena/relations/relations.py
```diff
@@ -35,7 +35,15 @@ class RelationBase:
...
-    pass
+    @staticmethod
+    def is_spawn_pose_constraint() -> bool:
+        """Whether this only sets an object's spawn pose at init.
+
+        Structural relations carry into the success states, while spawn-pose constraints do not.
+        Overridden True by the placement constraints that set an object's spawn pose at init
+        (AtPosition, PositionLimits, RandomAroundSolution, RotateAroundSolution).
+        """
+        return False
```
alexmillane: I'm not sure I get the distinction between a spawn constraint and a non-spawn constraint, and what we're using that for.

For example, if we spawn the apple and the bowl _on_ the table. That seems to be a spawn constraint? I.e. a constraint that applies at spawn time.

IMO all constraints that effect the graph a step 0 are spawn constraints.

I'm wondering why we need this concept?

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -155,3 +238,131 @@ def to_arena_env(self) -> IsaacLabArenaEnvironment:
...
+    Returns:
+        The next state spec.
+    """
+    relocations = _get_task_relocations(transition)
+    moved_objects_ids = {relocation.subject for relocation in relocations}
+    spatial_constraints: list[ArenaEnvGraphSpatialConstraintSpec] = []
+
+    for prev_spatial_constraint in prev_state_spec.spatial_constraints:
+        # Spawn pose constraints are those that are set at reset time. e.g."at position", "position_limits", etc.
+        constraint_is_spawn_pose = spatial_constraint_is_spawn_pose(prev_spatial_constraint.type)
```
alexmillane: I'm not sure I get this distiction. Aren't all constraints that the user applies the to graph at t=0 applied at reset time? For example if the user requests the apple to be on the table, that relation is `False` according to this distiction, but fires at reset time and constrains the spawn pose.

I'm not sure I get the distiction here?

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -155,3 +238,131 @@ def to_arena_env(self) -> IsaacLabArenaEnvironment:
...
+        # A moved object's old placement is replaced by the new placement
+        # Case 1: when the object is the constraint's child. (e.g. "cube on table" -> "cube on shelf")
+        # Case 2: when the object is the owner of its own (unary) spawn-pose constraint. (e.g. "cube at_position A" -> "cube at_position B")
+        child_is_relocated = prev_spatial_constraint.child in moved_objects_ids
+        parent_is_spawn_pose_owner = (prev_spatial_constraint.parent in moved_objects_ids) and constraint_is_spawn_pose
+        constraint_is_replaced = child_is_relocated or parent_is_spawn_pose_owner
+
+        # Because the spawn pose is at reset time, it is not affected by the success condition.
+        # e.g. a never-moved bowl's "position_limits".
+        constraint_is_dropped = is_final_state and constraint_is_spawn_pose
```
alexmillane: I think I need help understanding why we drop a bunch of constraints at the final state, but not at the intermediate states?

isaaclab_arena/tests/test_data/pick_and_place_maple_table_init_env_graph.yaml
```diff
@@ -0,0 +1,104 @@
...
+        params:
+          x: 0.65
+          y: 0.25
+          z: 0.85
+
+    task_constraints:
+      - id: state_spec_0_droid_reach_rubiks_cube_hot3d_robolab
+        type: "reach"
+        parent: droid_abs_joint_pos
+        child: rubiks_cube_hot3d_robolab
```
alexmillane: I'm wondering why we introduce a reachability constraint to the Rubik's Cube but not the mug. Perhaps because we set the mugs position manually (and assume that the user has satisfied that constraint).

One thing that is odd to me, is that this MR effectly introduces a mechanism for translating tasks into spatial constraints that affect the graph over time).

Is this reachability constraint not also implied by the task. Could the task not also add a reachability constraint at t=0?

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -132,6 +134,19 @@ def get_viewer_cfg(self) -> ViewerCfg:
             offset=np.array([-1.5, -1.5, 1.5]),
         )
 
+    @classmethod
+    def success_state_transition(cls, task_args: Mapping[str, Any]) -> TaskTransition:
```
alexmillane: any reason we're using `Mapping` here? Perhaps it's actually more correct, but I think else where we use `dict[str, Any]`

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -132,6 +134,19 @@ def get_viewer_cfg(self) -> ViewerCfg:
...
+        """Success (``object_on_destination``): the picked object ends up a relation with the destination."""
+        pick_object = task_args["pick_up_object"]
+        destination_location = task_args["destination_location"]
```
alexmillane: Is it possible to move `"pick_up_object"` and `"destination_location"` into the function arguments?

It seems weird to get a pretty loosely specified dict (`str` to `any`), pick out two string members.

We could constrain things by requiring these things as arguments.

alexmillane (commented):
Thanks for this. Looks pretty good.

It's quite complex, what's going on. Let's have a call so I can fully understand the intent?

But it's close I think!

isaaclab_arena/tests/test_data/pick_and_place_maple_table_init_env_graph.yaml
xyao-nv: Well I guess it should not be here... The resolve method shall take care of populating this.

isaaclab_arena/environments/arena_env_graph_spec.py
xyao-nv: `EnvGraph` is coded above where nodes are anything USDs `obejct/object_reference/embodiment,` edges are time/state-dependent, containing each state's spatial and task related constraints, encapsulated in the `state_spec.`

The request to document it down sounds to me like a Doc MR, and to visualize it will through Qian's upcoming HTML viz [MR](https://github.com/isaac-sim/IsaacLab-Arena/pull/719)

The graph is partially wired as no intermediate and final state constraints are added into the spec. Given how agent's capability (spatial understanding and context lengh) varies, we cannot fully count on agent to derive if constraints can be capture fully along the time

As we move through time/tasks, it essentially updates edges from previous state and adds edges from current task as the current state. Given what we aligned this morning, intermediate & final state spec in EnvGraph will only contains the diff from previous state.

isaaclab_arena/environments/arena_env_graph_spec.py
xyao-nv: Updated as this var could be un-necessary. But I will add type annotations for other vars. THanks!

isaaclab_arena/environments/arena_env_graph_spec.py
xyao-nv: I see the confusion, and I agree having this spawn pose logic is causing the complexities. It was designed to help decide what constraint to keep and what to remove (i.e. Apple `SetPose`,  PnP Apple cause `SetPose` to be removed, and `OnBowl` to be added).

In theory, I expect solver shall take all `spatial_constraints_t0 + NOT(delta_spatial_constraints_ t1...tN)` + `task_constraints_t0...tN`.

This spawn pose logic has been removed after we decide to only track state **diffs** in constraints along the task progression. It simplifies the logics here. Take the above example, it will just have OnBowl to be added.

isaaclab_arena/tasks/task_transition.py
xyao-nv: updated.

isaaclab_arena/tasks/pick_and_place_task.py
xyao-nv: yeah dict shall be. thx for catching

isaaclab_arena/tasks/pick_and_place_task.py
xyao-nv: I like your suggestion. Done

isaaclab_arena/tasks/pick_and_place_task.py
xyao-nv: I like your suggestion. Done

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -129,6 +154,49 @@ def write_yaml(self, path: str | Path) -> None:
...
+                    update={"initial_state_spec_id": f"state_spec_{i}", "success_state_spec_id": new_state_id}
+                )
+            )
+
+        return ArenaEnvGraphSpec(
+            env_name=env_name or self.env_name,
+            nodes=self.nodes,
+            tasks=out_tasks,
+            state_specs=states,
+        )
```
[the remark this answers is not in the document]
xyao-nv: added. thx

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -129,6 +154,49 @@ def write_yaml(self, path: str | Path) -> None:
...
+        moved objects land), not a full snapshot. ``N`` tasks therefore yield ``N+1`` state specs.
+
+        Returns a fully-wired, strict-validated ``ArenaEnvGraphSpec``.
+        """
+        assert len(self.state_specs) == 1, (
+            f"unresolved graph must define exactly the initial state (state_spec_0); got {len(self.state_specs)} state"
+            " specs"
+        )
+        # state_spec_0 is the given initial state, and the transitions chain the rest off the task list.
+        states: list[ArenaEnvGraphStateSpec] = [self.state_specs[0]]
```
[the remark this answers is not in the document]
xyao-nv: updated

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -70,10 +74,27 @@ def _parse_nodes(cls, nodes: Any) -> list[Any]:
...
+
+    def _run_validation(self, is_task_wiring_enabled: bool) -> None:
+        """Check unique ids, cross-references, constraint shapes, and CLI override specs."""
         assert_unique_ids(self.nodes, self.tasks, self.state_specs)
         assert_references_exist(self.nodes, self.tasks, self.state_specs)
+        if is_task_wiring_enabled:
+            assert_task_wiring(self.tasks, self.state_specs)
         assert_spatial_constraint_shapes(self.state_specs)
         assert_cli_override_specs_reference_nodes(self.nodes, self.cli_override_specs)
         return self
```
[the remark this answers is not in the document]
xyao-nv: removed

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -129,6 +154,64 @@ def write_yaml(self, path: str | Path) -> None:
...
+                    embodiment_id=embodiment_id,
+                    reach_targets=reach_targets,
+                    is_final_state=is_final_state,
+                )
+            )
+            out_tasks.append(
+                task.model_copy(
+                    update={"initial_state_spec_id": f"state_spec_{i}", "success_state_spec_id": new_state_id}
+                )
+            )
```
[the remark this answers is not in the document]
xyao-nv: updated

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -155,3 +238,131 @@ def to_arena_env(self) -> IsaacLabArenaEnvironment:
...
+    return ArenaEnvGraphStateSpec(
+        id=new_state_id, spatial_constraints=spatial_constraints, task_constraints=task_constraints
+    )
+
+
+def _reprefix_id(old_id: str, old_prefix: str, new_prefix: str) -> str:
+    """Swap a constraint id's ``{state_id}`` prefix so carried constraints stay uniquely named."""
+    assert old_id.startswith(old_prefix), f"constraint id {old_id!r} is not prefixed by its state id {old_prefix!r}"
+    return new_prefix + old_id[len(old_prefix) :]
```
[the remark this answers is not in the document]
xyao-nv: outpdated

isaaclab_arena/tests/test_arena_env_graph_spec.py
```diff
@@ -309,6 +311,67 @@ def test_arena_env_graph_spec_rejects_invalid_data():
...
+    with pytest.raises(ValidationError):
+        ArenaEnvGraphSpec.from_yaml(_INIT_GRAPH)
+
+
+def test_chain_wires_each_success_state_as_next_initial_state():
+    """task[i].success_state_spec_id == task[i+1].initial_state_spec_id (a single chain)."""
+    resolved = ArenaEnvGraphSpec.from_yaml(_INIT_GRAPH, is_task_wiring_enabled=False).resolve_constraints()
+    ordered = sorted(resolved.tasks_by_id.values(), key=lambda t: t.id)
+    for earlier, later in zip(ordered, ordered[1:]):
+        assert earlier.success_state_spec_id == later.initial_state_spec_id
```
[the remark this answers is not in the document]
xyao-nv: done

isaaclab_arena/environments/graph_spec_utils.py
```diff
@@ -96,6 +90,23 @@ def assert_references_exist(nodes: list[Any], tasks: list[Any], state_specs: lis
...
+    """
+    state_spec_ids = {state_spec.id for state_spec in state_specs}
+    for task in tasks:
+        for label, state_spec_id in (
+            ("initial_state_spec_id", task.initial_state_spec_id),
+            ("success_state_spec_id", task.success_state_spec_id),
+        ):
+            assert (
+                state_spec_id in state_spec_ids
+            ), f"Task '{task.id}' references unknown state spec '{state_spec_id}' for '{label}'"
```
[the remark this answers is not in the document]
xyao-nv: not this time

isaaclab_arena/relations/relations.py
xyao-nv: Agree it's confusing and vaguely defined. Dropped it and limited to state delta.

isaaclab_arena/environments/arena_env_graph_spec.py
xyao-nv: Agree it's confusing and vaguely defined. Dropped it and limited to state delta.

isaaclab_arena/environments/arena_env_graph_spec.py
xyao-nv: Now it's only one for.

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -89,13 +109,17 @@ def _load_yaml_dict(path: str | Path) -> dict[str, Any]:
...
-        return cls.from_dict(cls._load_yaml_dict(path))
+    def from_yaml(cls, path: str | Path, is_task_wiring_enabled: bool = True) -> ArenaEnvGraphSpec:
+        return cls.from_dict(cls._load_yaml_dict(path), is_task_wiring_enabled=is_task_wiring_enabled)
 
     @classmethod
-    def from_dict(cls, data: dict[str, Any]) -> ArenaEnvGraphSpec:
+    def from_dict(cls, data: dict[str, Any], is_task_wiring_enabled: bool = True) -> ArenaEnvGraphSpec:
+        """Validate and build a spec from a mapping.
+
+        Pass ``is_task_wiring_enabled=False`` to load an *unresolved* graph.
```
alexmillane: Suggestion to add a bit more information about what an unresolved graph is.

I would guess maybe(?) and unresolved graph is one that:
- only contains the initial state implied by the (initial) spatial constraints. Is that true?

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -155,3 +224,52 @@ def to_arena_env(self) -> IsaacLabArenaEnvironment:
...
+
+def _get_next_state_spec(new_state_id: str, transition: TaskTransition) -> ArenaEnvGraphStateSpec:
+    """Build the success state as the *delta* of the task's transition -- only what it changed.
+
+    The chain focuses on the delta: ``state_spec_0`` holds the initial state, and each derived state
+    records only the effects the task established, not a fresh snapshot of the whole scene (followed by is_delta=True).
+
+    E.g. task "place mug on bowl" (mug is the moved object) yields a state holding only::
+
+        mug on bowl  # the effect from the PnP task's relocation
```
alexmillane: Is it also true that

~~mug on table~~

I.e. does a change in relation, break the relations that already exist on that object?

isaaclab_arena/tasks/rotate_revolute_joint_task.py
```diff
@@ -71,6 +76,16 @@ def get_metrics(self) -> list[MetricBase]:
...
 
+    @classmethod
+    def success_state_transition(cls, openable_object: Asset) -> TaskTransition:
+        """Success is the openable object's joint state change.
+
+        The env-graph spec has not yet defined joint openness, so no effects are
+        emitted today — only the subject node (the openable object) is recorded.
+        # TODO(xinjieyao, 2026.06.02): allow env graph spec to define node state change
+        """
+        return TaskTransition(subject=openable_object.name)
```
alexmillane: Looking at the assets passed in here, they are empty except for the name, which is loaded from `success_transition_asset_args`

There's a bit of a round trip here. What do you think about removing `success_transition_asset_args` and doing:

```
classmethod
    def success_state_transition(cls, openable_object_name: str, **kwags) -> TaskTransition:
        """Success is the openable object's joint state change.

        The env-graph spec has not yet defined joint openness, so no effects are
        emitted today — only the subject node (the openable object) is recorded.
        # TODO(xinjieyao, 2026.06.02): allow env graph spec to define node state change
        """
        return TaskTransition(subject=openable_object_name)
```

And then the call site goes from

```
asset_args = {}
for name in task_cls.success_transition_asset_args:
    assert name in task.task_args, f"task {task.id} ({task.type}) is missing required task_arg '{name}'"
    asset_args[name] = Asset(name=task.task_args[name])
return task_cls.success_state_transition(**asset_args)
```

to

```
return task_cls.success_state_transition(**task_cls.success_transition_asset_args)
```

This should reduce a bit of the coupling between the run-time task definition and the graph loading (the run-time task doesn't have to store some dict keys used in some remote part of the code).

isaaclab_arena/tasks/task_base.py
```diff
@@ -11,10 +11,18 @@
...
 
 
 class TaskBase(ABC):
 
+    success_transition_asset_args: tuple[str, ...] = ()
+    """Which task args the success transition needs (e.g. ``pick_up_object``).
+
+    Resolution passes only these to ``success_state_transition`` and ignores the rest (scene,
+    episode length, ...). Override this alongside ``success_state_transition``.
+    """
```
alexmillane: see comment above. Suggestion to remove.

alexmillane (approved):
Looking much better.

Thanks for working on this. I have a few comments:
- reducing coupling between the run-time tasks and the graphs.
- do we handle prior removing constraints when we get a transition.

One broader question that I have is how these future constraints are actually used. Presumably we will fold thesefuture constraints back to something that operates on the initial layout of the scene?

alexmillane (approved):
Feel free to merge when you address the above

isaaclab_arena/environments/arena_env_graph_spec.py
xyao-nv: thx for suggestion, udpated

isaaclab_arena/environments/arena_env_graph_spec.py
xyao-nv: Yes, and that's why in the beginning it was mainly the full snapshots.
But the purpose of this, is to NOT ops on constraints implied intermediate and final states. 
Imagine doing a Push mug To a circular zone on the table
So the solver will see Mug On Table + Not (Mug in Circle).

isaaclab_arena/environments/arena_env_graph_spec.py
xyao-nv: it's removed

isaaclab_arena/tasks/rotate_revolute_joint_task.py
xyao-nv: yeah agree it's cleaner. updated.

isaaclab_arena/environments/arena_env_graph_spec.py
```diff
@@ -155,3 +225,49 @@ def to_arena_env(self) -> IsaacLabArenaEnvironment:
...
+
+# --- Constraint resolution helpers (used by ArenaEnvGraphSpec.resolve_constraints) ---
+
+
+def _get_task_state_transition(task: ArenaEnvGraphTaskSpec) -> TaskTransition:
+    """Look up the task class via ``TaskRegistry`` and return its declared success transition."""
+    task_cls = TaskRegistry().get_task_by_name(task.type)
+    assert task_cls is not None, f"task {task.type} not found in TaskRegistry"
+    asset_args = {name: task.task_args[name] for name in task_cls.success_transition_asset_args}
+    return task_cls.success_state_transition(**asset_args)
```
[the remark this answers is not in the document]
xyao-nv: outdated

isaaclab_arena/tasks/task_base.py
xyao-nv: done

PR #733, opened 2026-05-28
Refactor: Metrics in IsaacLabArenaManagerBasedRLEnvCfg

alexmillane:
## Summary
Refactors how metrics enter our `IsaacLabArenaManagerBasedRLEnvCfg` to match Isaac Lab convention.

## Detailed description
- We were attaching metrics runtime objects to the `IsaacLabArenaManagerBasedRLEnvCfg`
- Isaac Lab expects only `configclass`es in `*Cfg` objects.
- This changes our design to match more closely with Isaac Lab.
  - Metrics now enter the env through `MetricsCfg`s
  - Run-time metrics are then built at env build time by `IsaacLabArenaManagerBasedRLEnv`
  - This matches how things are done in IsaacLab - Cfgs are spcified by the user, and then run-time objects are built by the system based on those configs.

PR #754, opened 2026-06-03
Move isaaclab_arena_environment out of IsaacLabArenaManagerBasedRLEnvCfg

alexmillane:
## Summary
Part of a refactor started in #733 to get run-time objects out of `IsaacLabArenaManagerBasedRLEnvCfg`

## Detailed description
- IsaacLab expects only configuration objects in this struct.
- Our non-adherence to this has started to cause problems in later commits.
- This removes the last run-time objects on `main`


cvolkcvolk (approved):
Nice. Makes sense. Thanks!

PR #758, opened 2026-06-03
Add dataclass and state machine for fine grained subtask tracking

peterd-NV:
## Summary
This PR adds foundational machinery required for subtask progress tracking in Arena.

- Adds a new ProgressObjective data class used to specify predicate groups for subtask tracking. 
- Adds a state machine (ProgressTrakcer) and runners (ProgressObjectiveRunner) for tracking predicate progression in Arena tasks.
- Updates CompositeTaskBase and TaskBase with new interfaces to support subtask tracking
- Adds test cases for the new data class and tracker


cvolkcvolk:
Thanks for adding this! Going to be super useful once we have the full pipeline in.
One highler level thing that made it a little hard for me to follow at first: I think we've ended up overloading the word "subtask" a bit. Right now it means a few different things:

  - the composed tasks in CompositeTaskBase.subtasks (with SequentialTaskBase,...)
  - the Mimic datagen configs (SubTaskConfig)
  - and now the new progress chains (`FineGrainedSubtask`, `FineGrainedSubtaskTerminationsCfg`, `FineGrainedSubtaskEventsCfg`, ...)

So when reading "subtask" anywhere its not clear which one is meant if not already familiar.


If I read it correctly we've got two systems now that do opposite jobs (but read like siblings in the code):
  - The success termination: Decides whether the episode is done/passed. 
  - The new FineGrainedSubtask tracking:  Just watches and reports progress (grab → lift → place). Returns all-False every step, _never ends the episode_.

Currently "FineGrained" names the "resolution" of the Task, not the part that it has control over (Can it end the episode or not?)
And both (`get_fine_grained_subtask_termination_cfg` & actual termination configs) are registered as termination terms and sit side by side and look the same.

My proposal is to name by that question _Can it end the episode or not?_:
  - **authoritative** (can end the episode) → stays success / termination. No change.
  - **observational** (only tracks, never ends) → call it progress.

So the renames could be roughly:
  - `FineGrainedSubtask → ProgressObjective`
  - `FineGrainedSubtaskTrackingStateMachine → ProgressTracker`
  - `get_fine_grained_subtasks() → get_progress_objectives()`
  - `env.extras["fine_grained_subtask"] → env.extras["progress"]`
  - ...

Two side benefits:

1. "subtask" goes back to meaning one thing: the composition unit in CompositeTaskBase.subtasks.
2. Right now the tracking runs as a termination `term. get_fine_grained_subtask_termination_cfg()`, returning all-False every step, so it never terminates anything... Once we're calling it "progress," it's more clear it shouldn't live in terminations at all. We could move it to a recorder (like the existing SubtaskSuccessStateRecorder) so where it lives also tells you it's observational.

isaaclab_arena/tasks/fine_grained_progress_tracker.py
```diff
@@ -0,0 +1,372 @@
...
+            [{"step": int, "fine_grained_subtask": str, "group": str,
+              "predicate_index": int, "predicate_name": str,
+              "score_delta": float}, ...],
+            ...                                        # one list per env
+        ],
+    }
+    """
+
+    env.extras["fine_grained_subtask"] = {
+        "states": sm.get_state(),
```
cvolkcvolk: As this is a termination event, this runs on every step. so get_state() and get_events() rebuild the full per-env view each step even when nothing is reading those extras. Could we move this to a recorder term so it only runs when recording is active?

cvolkcvolk: Touches on the renaming and suggestion to move this state tracking out of termination and into a recorder

isaaclab_arena/tasks/fine_grained_progress_tracker.py
```diff
@@ -0,0 +1,372 @@
...
+        "events": sm.get_events(),
+    }
+
+    # Return all-False so it does not contribute to termination.
+    return torch.zeros(env.num_envs, dtype=torch.bool, device=env.device)
```
cvolkcvolk: Since this always returns all-False/zeros it never contributes to termination, so it's really just being used as a "runs every step" hook. `RecorderTerm.record_post_step` could potentially give the same per-step call without adding a zero-return statement to every progress task's terminations config.

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -0,0 +1,372 @@
...
+
+        # Otherwise return True only for envs whose current
+        # parent-subtask index matches this FineGrainedSubtask's parent_subtask_idx.
+        if torch.is_tensor(current_idx):
+            ci = current_idx.to(self.device)
+        else:
+            ci = torch.as_tensor(current_idx, device=self.device)
+        return ci == int(self.fine_grained_subtask.parent_subtask_idx)
+
+    def step(self, env, step_index: torch.Tensor | None) -> list[dict]:
```
cvolkcvolk: This function is quite long and handles gating,  groups and chain positions. Could we move parts out into functions and split this up into something like

```
  def step(self, env, step_index) -> list[dict]:
      gating_mask = self._compute_composite_task_gating_mask(env)
      if not bool(gating_mask.any().item()):
          return []
      events = []
      for group_name, predicate_chain in self.fine_grained_subtask.canonical_predicate_groups.items():
          events += self._step_group(env, group_name, predicate_chain, gating_mask, step_index)
      return events

```

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -0,0 +1,372 @@
...
+
+    def reset(self, env_ids) -> None:
+        """Reset the runners for the provided envs."""
+
+        for runner in self.runners:
+            runner.reset(env_ids)
+        for eid in env_ids:
+            self._events[eid] = []
+
+    def get_state(self) -> list[dict]:
```
cvolkcvolk: Same here. Claude had a remark and a suggestion (obviously untested :) ). But would be nice to split this up as well into digestible junks.


Ai suggestion: "Per-env loop currently recomputes whole tensors per env (`runner.is_complete()[env_idx]`, `runner.overall_score_per_env()[env_idx]` are called once per `env_idx`, but each call builds the full (num_envs,) tensor — so it's `O(num_envs²))`. Extracting fixes both. Add a method on the runner that owns its own state:

```
  # on FineGrainedSubtaskRunner
  def state_for_env(self, env_idx, is_complete, score) -> dict:
      """Per-env view of this runner. is_complete/score passed in so the
      full-tensor computations happen once, outside the env loop."""
      fgs = self.fine_grained_subtask
      active_predicates = {}
      completed_groups = 0
      for group_name in fgs.group_names:
          chain = fgs.canonical_predicate_groups[group_name]
          cur = int(self.current_index[group_name][env_idx].item())
          if cur >= len(chain):
              active_predicates[group_name] = None
              completed_groups += 1
          else:
              active_predicates[group_name] = _predicate_repr(chain[cur][0])
      return {
          "completed_groups": completed_groups,
          "total_groups": len(fgs.group_names),
          "score": float(score),
          "is_complete": bool(is_complete),
          "active_predicates": active_predicates,
      }
```

  Then get_state hoists the tensor computations out of the loop and just assembles:

```
  def get_state(self) -> list[dict]:
      # Compute the per-runner tensors once, not once per env.
      completeness = [r.is_complete() for r in self.runners]
      scores = [r.overall_score_per_env() for r in self.runners]

      output = []
      for env_idx in range(self.num_envs):
          fgs_states = {
              r.fine_grained_subtask.name: r.state_for_env(env_idx, completeness[i][env_idx], scores[i][env_idx])
              for i, r in enumerate(self.runners)
          }
          overall_score = sum(
              r.fine_grained_subtask.score * fgs_states[r.fine_grained_subtask.name]["score"]
              for r in self.runners
          )
          all_complete = all(s["is_complete"] for s in fgs_states.values())
          output.append({
              "fine_grained_subtasks": fgs_states,
              "overall_score": overall_score,
              "all_complete": all_complete,
          })
      return output
```
"

isaaclab_arena/tasks/fine_grained_progress_tracker.py
peterd-NV: Refactored to use RecorderManager instead of TerminationManager for driving state machine step. TerminationTermCfg has been replaced with a recorder term that runs record_post_step. Since we are using the mechanism to "drive the state machine step" and not actually record data, the recorder term returns (None, None). 

One note: The progress tracking should be run at every step, regardless if recording is active or not. The events are complex dictionaries with many fields and thus returned to the user via env.extras, and not via a RecorderTerm.

isaaclab_arena/tasks/fine_grained_progress_tracker.py
peterd-NV: Updated to user recorder manager/term.

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: Updated to simply `step` and move `_step_group` to a separate function.

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: Applied the change to separate this into getting state per_env and split up into different chunks

isaaclab_arena/tasks/fine_grained_progress_tracker.py
```diff
@@ -0,0 +1,372 @@
...
+        return events
+
+    def reset(self, env_ids) -> None:
+        """Reset the state machine runner for the provided envs."""
+
+        for group_name in self.fine_grained_subtask.group_names:
+            for eid in env_ids:
+                self.current_index[group_name][eid] = 0
+                self.group_score[group_name][eid] = 0.0
+                self.group_complete[group_name][eid] = False
```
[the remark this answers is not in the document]
peterd-NV: fixed

isaaclab_arena/tasks/task_base.py
```diff
@@ -62,3 +67,18 @@ def get_episode_length_s(self) -> float | None:
...
+        fine_grained_subtasks = self.get_fine_grained_subtasks()
+        if not fine_grained_subtasks:
+            return None
+        return make_fine_grained_subtask_events_cfg(fine_grained_subtasks)
+
+    def get_fine_grained_subtask_termination_cfg(self) -> Any:
+        fine_grained_subtasks = self.get_fine_grained_subtasks()
+        if not fine_grained_subtasks:
+            return None
+        return make_fine_grained_subtask_termination_cfg(fine_grained_subtasks)
```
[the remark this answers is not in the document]
peterd-NV: Does not cause the issue mentioned as _ensure_progress_tracker caches the list. However, there are indeed separate lists. Added a step in TaskBase to resolve objective lists only once and then cache.

isaaclab_arena/tasks/fine_grained_progress_tracker.py
```diff
@@ -0,0 +1,391 @@
...
+        )
+        setattr(env, _STATE_MACHINE_ATTR, sm)
+    return sm
+
+
+class FineGrainedProgressRecorder(RecorderTerm):
+    """Per-step hook that ticks the FineGrainedProgressTracker. Records nothing.
+
+    Registered as a recorder term so it runs once per env.step via
+    record_post_step. It advances the state nachine and publishes the per-step state/events to
```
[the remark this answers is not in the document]
peterd-NV: fixed

peterd-NV:
Thanks @cvolkcvolk for the review. I've gone ahead renamed from "FineGrainedSubtasks" to "FineGrainedProgressObjective" and "FineGrainedProgressTracker".

I kept the "FineGrained" terminology since I wanted to make it clear that the predicate chains we give tasks here can be made to be more detailed and include this notion of multiple "subtasks" in the progress (i.e.  we can have multiple predicates or predicate groups for a single TaskBase).

isaaclab_arena/progress_tracking/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
cvolkcvolk: Could remove 2025 as its a new module

isaaclab_arena/progress_tracking/fine_grained_progress_objective.py
```diff
@@ -0,0 +1,177 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
cvolkcvolk:
```suggestion
# Copyright (c) 2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```

isaaclab_arena/progress_tracking/fine_grained_progress_tracker.py
```diff
@@ -0,0 +1,419 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
cvolkcvolk:
```suggestion
# Copyright (c) 2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```

isaaclab_arena/progress_tracking/fine_grained_progress_tracker.py
```diff
@@ -0,0 +1,419 @@
...
+                self.current_index[group_name],
+            )
+            # Update the group score for the envs that were advanced.
+            self.group_score[group_name] = self.group_score[group_name] + advance_mask.float() * float(score_weight)
+            # Update the advanced mask for the envs that were advanced.
+            advanced = advanced | advance_mask
+
+            # Emit an event for each env where a predicate was advanced.
+            pred_name = _predicate_repr(predicate)
+            for eid in torch.nonzero(advance_mask, as_tuple=False).flatten().tolist():
```
cvolkcvolk: Suggestion to use the full name `env_index`

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -360,6 +362,31 @@ def get_metrics(self) -> list[MetricBase]:
 
         return subtask_metrics
 
+    def get_own_fine_grained_progress_objectives(self) -> list[FineGrainedProgressObjective]:
+        """Composite-level FineGrainedProgressObjectives.
+
+        These are added on top of whatever FGPOs the child subtasks declare and are not gated.
```
cvolkcvolk: Suggestion to use the full name instead here as "FGPO" has not been introduced

isaaclab_arena/progress_tracking/fine_grained_progress_objective.py
```diff
@@ -0,0 +1,177 @@
...
+    name: str
+    predicate_groups: PredicateGroups
+    score: float = 1.0
+    logical: Literal["all", "any", "choose"] = "all"
+    K: int | None = None
+    description: str | None = None
+
+    canonical_predicate_groups: dict[str, list[tuple[Callable, float]]] = field(init=False, repr=False)
+
+    # Index of the parent TaskBase this recipe belongs to. Set automatically by
```
cvolkcvolk: What is a recipe? Is it a _progress_objective_? Suggestion to rename or remove

isaaclab_arena/progress_tracking/fine_grained_progress_tracker.py
```diff
@@ -0,0 +1,419 @@
...
+from dataclasses import MISSING
+from typing import Any
+
+from isaaclab.managers import EventTermCfg
+from isaaclab.managers.recorder_manager import RecorderManagerBaseCfg, RecorderTerm, RecorderTermCfg
+from isaaclab.utils import configclass
+
+from isaaclab_arena.progress_tracking.fine_grained_progress_objective import FineGrainedProgressObjective
+
+_STATE_MACHINE_ATTR = "_fine_grained_progress_tracker"
```
cvolkcvolk: Could we unify on the explicit `tracker`  `runner` names and drop "state machine" and `sm`? 
I think we've got "state machine" terminology (_STATE_MACHINE_ATTR, the `sm` vriable, and the class docstrings) sitting alongside the tracker and runner names interchangeably. 
(eg `sm` below is a `FineGrainedProgressTracker`)

cvolkcvolk:
> Thanks @cvolkcvolk for the review. I've gone ahead renamed from "FineGrainedSubtasks" to "FineGrainedProgressObjective" and "FineGrainedProgressTracker".
> 
> I kept the "FineGrained" terminology since I wanted to make it clear that the predicate chains we give tasks here can be made to be more detailed and include this notion of multiple "subtasks" in the progress (i.e. we can have multiple predicates or predicate groups for a single TaskBase).

Yeah fully agree on the explicit naming. However "Fine-grained" doesn't actually encode the concept "multiple predicate groups" or "subtasks" either :-) 
Another thing with `fine_grained_*` and `FineGrained*` might suggest there's a _coarse ProgressObjective_ somewhere, which there isn't (And it doesn't match the `progress_tracking/` package name after our refactor anymore)

isaaclab_arena/tasks/task_base.py
```diff
@@ -64,6 +69,27 @@ def get_episode_length_s(self) -> float | None:
     def get_task_description(self) -> str | None:
         return self.task_description
 
+    def get_fine_grained_progress_objectives(self) -> list[FineGrainedProgressObjective]:
+        return []
+
+    def _resolve_fine_grained_progress_objectives(self) -> list[FineGrainedProgressObjective]:
```
cvolkcvolk: Do we need those methods in `TaskBase`? The other functions above are single methods (get_events_cfg, get_recorder_term_cfg, etc...) each returning a `cfg` or `None` that the builder then combines. 

Could we drop these from `TaskBase` and let the `arena_env_builder` handle the cfg creation? Something like 

```
progress_objectives = task.get_fine_grained_progress_objectives()
events_cfg = combine_configclass_instances(
    ...,
    make_fine_grained_progress_objective_events_cfg(progress_objectives) if progress_objectives else None,
)
recorder_manager_cfg = combine_configclass_instances(
    make_fine_grained_progress_objective_recorder_cfg(progress_objectives) if progress_objectives else None,
)
```

`_resolve_fine_grained_progress_objectives()` shouldn't be needed then either

isaaclab_arena/progress_tracking/progress_objective.py
```diff
@@ -0,0 +1,177 @@
...
+    Canonical form: ``dict[group_name: list[(callable, score)]]``.
+
+    Accepted input shapes:
+      1. func (single callable)                one group with one predicate
+      2. [func, func, ...]                     one group, sequential chain
+      3. [(func, score), ...]                  one group, sequential chain, weighted
+      4. {group: func}                         multiple groups, one predicate each
+      5. {group: [func, ...]}                  multiple groups, sequential chains
+      6. {group: [(func, score), ...]}         multiple groups, sequential chains, weighted
+    """
```
alexmillane: Suggestion to expand description to say what "canonical form" means. Potentially a description of the input and output arguments would help here.

alexmillane: Also, is this only called internally? Suggestion to prefix with an underscore.

isaaclab_arena/progress_tracking/fine_grained_progress_objective.py
```diff
@@ -0,0 +1,177 @@
...
+      5. {group: [func, ...]}                  multiple groups, sequential chains
+      6. {group: [(func, score), ...]}         multiple groups, sequential chains, weighted
+    """
+
+    if callable(predicate_groups):
+        return {DEFAULT_GROUP_NAME: [(predicate_groups, 1.0)]}
+
+    if isinstance(predicate_groups, list):
+        if len(predicate_groups) == 0:
+            raise ValueError("FineGrainedProgressObjective.predicate_groups list cannot be empty")
```
alexmillane: style in Arena is to prefer `assert` over `if...raise` in cases where an error is unrecoverable.

isaaclab_arena/progress_tracking/progress_objective.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+    if len(value) == 0:
+        raise ValueError(f"Predicate chain for group '{group_name}' cannot be empty")
+
+    first = value[0]
+    if isinstance(first, tuple):
+        chain = []
+        for i, item in enumerate(value):
+            if not (isinstance(item, tuple) and len(item) == 2):
+                raise TypeError(f"Group '{group_name}' index {i}: expected (callable, score) tuple, got {item!r}")
```
alexmillane: Here and elsewhere: `assert`

isaaclab_arena/progress_tracking/progress_objective.py
```diff
@@ -0,0 +1,177 @@
...
+            for group_name, value in predicate_groups.items()
+        }
+
+    raise TypeError(
+        "FineGrainedProgressObjective.predicate_groups must be a callable, list, or dict; got"
+        f" {type(predicate_groups).__name__}"
+    )
+
+
+def _format_group_chain(value, group_name: str) -> list[tuple[Callable, float]]:
```
alexmillane: nit: Suggestion to move these internal long functions to another file. Because they're kinda long and complex, they drown out the actual external facing class in the file `FineGrainedProgressObjective`.

isaaclab_arena/progress_tracking/progress_objective.py
```diff
@@ -0,0 +1,177 @@
...
+def format_predicate_groups(predicate_groups: PredicateGroups) -> dict[str, list[tuple[Callable, float]]]:
+    """Format predicate_groups into the canonical form.
+
+    Canonical form: ``dict[group_name: list[(callable, score)]]``.
+
+    Accepted input shapes:
+      1. func (single callable)                one group with one predicate
+      2. [func, func, ...]                     one group, sequential chain
+      3. [(func, score), ...]                  one group, sequential chain, weighted
+      4. {group: func}                         multiple groups, one predicate each
```
alexmillane: Is a "group" here a collection of non-sequential composite predicate?

isaaclab_arena/progress_tracking/progress_objective.py
alexmillane: Suggestion to add a docstring here. Not quite clear what this function does

isaaclab_arena/progress_tracking/fine_grained_progress_objective.py
```diff
@@ -0,0 +1,177 @@
...
+        self.canonical_predicate_groups = normalized
+
+        # Validate the logical and K parameters.
+        num_groups = len(self.canonical_predicate_groups)
+        if self.logical == "choose":
+            if self.K is None:
+                raise ValueError(f"FineGrainedProgressObjective '{self.name}': K is required when logical='choose'")
+            if not (1 <= self.K <= num_groups):
+                raise ValueError(
+                    f"FineGrainedProgressObjective '{self.name}': K={self.K} but must be in [1, {num_groups}]"
```
alexmillane: Here and elsewhere, `assert`

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -0,0 +1,419 @@
...
+from isaaclab.managers import EventTermCfg
+from isaaclab.managers.recorder_manager import RecorderManagerBaseCfg, RecorderTerm, RecorderTermCfg
+from isaaclab.utils import configclass
+
+from isaaclab_arena.progress_tracking.fine_grained_progress_objective import FineGrainedProgressObjective
+
+_STATE_MACHINE_ATTR = "_fine_grained_progress_tracker"
+
+
+def _predicate_repr(pred) -> str:
```
alexmillane: can we put a type on `pred` to avoid all the `getattr`s below? Do we know something about the type here?

isaaclab_arena/progress_tracking/progress_objective.py
```diff
@@ -0,0 +1,177 @@
...
+            Can be "all", "any", or "choose"
+        K: Required when logical == "choose". Specifies the number of groups that must be completed
+            to consider the FineGrainedProgressObjective complete.
+        description: An optional description of the FineGrainedProgressObjective.
+    """
+
+    name: str
+    predicate_groups: PredicateGroups
+    score: float = 1.0
+    logical: Literal["all", "any", "choose"] = "all"
```
alexmillane: elsewhere we use enums. Any reason not to use an enum here?

isaaclab_arena/progress_tracking/fine_grained_progress_tracker.py
```diff
@@ -0,0 +1,419 @@
...
+
+        # Initialize the state machine's internal state.
+        self.current_index: dict[str, torch.Tensor] = {}
+        self.group_score: dict[str, torch.Tensor] = {}
+        self.group_complete: dict[str, torch.Tensor] = {}
+
+        for group_name in fine_grained_progress_objective.group_names:
+            self.current_index[group_name] = torch.zeros(num_envs, dtype=torch.long, device=device)
+            self.group_score[group_name] = torch.zeros(num_envs, dtype=torch.float32, device=device)
+            self.group_complete[group_name] = torch.zeros(num_envs, dtype=torch.bool, device=device)
```
alexmillane: Maybe expand the comment here. So we have an index per group, per env.

I guess `current_index` corresponds to index _within_ a group? Suggestion to expand the comments, and maybe even expand the names to be more descriptive (if that would help).

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -0,0 +1,419 @@
...
+
+        # If no env._current_subtask_idx -> composite task is not sequential (returns all True).
+        current_idx = getattr(env, "_current_subtask_idx", None)
+        if current_idx is None:
+            return torch.ones(self.num_envs, dtype=torch.bool, device=self.device)
+
+        # Otherwise return True only for envs whose current
+        # parent-subtask index matches this FineGrainedProgressObjective's parent_subtask_idx.
+        if torch.is_tensor(current_idx):
+            ci = current_idx.to(self.device)
```
alexmillane: `ci` -> `current_idx_tensor`

isaaclab_arena/progress_tracking/fine_grained_progress_tracker.py
peterd-NV: done

isaaclab_arena/progress_tracking/fine_grained_progress_tracker.py
peterd-NV: Yes, dropping all usage of state_machine/sm and using tracker/runner instead (note: runner = an instance for each ProgressObjective object, tracker = overall "state machine" which may have multiple runners).

isaaclab_arena/progress_tracking/fine_grained_progress_objective.py
peterd-NV: Yes, renaming to "progress_objective"

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: Done

isaaclab_arena/tasks/task_base.py
peterd-NV: Yup, dropped these from TaskBase to let it keep the current pattern of returning cfg or None. I was originally trying to match the pattern in arena_env_builder where we can directly call something like task.get_events_cfg but I agree that keeping the existing pattern in TaskBase is cleaner.

peterd-NV:
> > Thanks @cvolkcvolk for the review. I've gone ahead renamed from "FineGrainedSubtasks" to "FineGrainedProgressObjective" and "FineGrainedProgressTracker".
> > I kept the "FineGrained" terminology since I wanted to make it clear that the predicate chains we give tasks here can be made to be more detailed and include this notion of multiple "subtasks" in the progress (i.e. we can have multiple predicates or predicate groups for a single TaskBase).
> 
> Yeah fully agree on the explicit naming. However "Fine-grained" doesn't actually encode the concept "multiple predicate groups" or "subtasks" either :-) Another thing with `fine_grained_*` and `FineGrained*` might suggest there's a _coarse ProgressObjective_ somewhere, which there isn't (And it doesn't match the `progress_tracking/` package name after our refactor anymore)

Looking at it again I think we can remove the FineGrained terminology entirely now that we've moved the "progress tracking" to it's own separate module and decoupled from "tasks." I think we can rename everything to `ProgressObjective` and `ProgressTracker`, and it's clear that each can have multiple predicates without the need for the term "fine-grained."

isaaclab_arena/progress_tracking/fine_grained_progress_tracker.py
peterd-NV: done

isaaclab_arena/progress_tracking/fine_grained_progress_objective.py
peterd-NV: done

isaaclab_arena/progress_tracking/__init__.py
peterd-NV: done

isaaclab_arena/progress_tracking/progress_objective.py
```diff
@@ -0,0 +1,173 @@
...
+    return out
+
+
+@dataclass
+class ProgressObjective:
+    """Configuration object that defines a scored predicate sequence to track progress within a task.
+
+    A ProgressObjective specifies what the progress tracker (ProgressTracker) should track.
+    Each ProgressObjective holds one or more sequential predicate chains (groups).
+    Within a group, predicates run in order. Across groups, predicates run in parallel.
```
alexmillane, line 38: Shall we add something about how a `ProgressObjective` is completed? I would guess that the objective is completed: when a sufficient number (controlled by K) of parallel predicate groups contain completed chains.

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -0,0 +1,413 @@
...
+
+        # Initialize the runner's internal state.
+        self.current_index: dict[str, torch.Tensor] = {}
+        self.group_score: dict[str, torch.Tensor] = {}
+        self.group_complete: dict[str, torch.Tensor] = {}
+
+        for group_name in progress_objective.group_names:
+            self.current_index[group_name] = torch.zeros(num_envs, dtype=torch.long, device=device)
+            self.group_score[group_name] = torch.zeros(num_envs, dtype=torch.float32, device=device)
+            self.group_complete[group_name] = torch.zeros(num_envs, dtype=torch.bool, device=device)
```
alexmillane: Suggestion to expand the comment describing these members. For example does `current_idx` mean the index of the currently active predicate, for each group, for each env_id? Is group score the currently accumulated score for completed predicates? etc.

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -0,0 +1,413 @@
...
+        else:
+            ci = torch.as_tensor(current_idx, device=self.device)
+        return ci == int(self.progress_objective.parent_subtask_idx)
+
+    def step(self, env, step_index: torch.Tensor | None) -> list[dict]:
+        """Step the runner for a single env.step.
+
+        Advance each group's predicate chain by at most one position per env and return a
+        transition event for every env/group that advanced this step.
+        """
```
alexmillane: Suggestion to describe the returned value. `list[dict]` is a bit ambiguous the return actually contains.

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -0,0 +1,413 @@
...
+            if not bool(at_position.any().item()):
+                continue
+
+            # Evaluate the predicate for all envs, reshaped to a flat (num_envs,) bool tensor.
+            result = torch.as_tensor(predicate(env), dtype=torch.bool, device=self.device).reshape(-1)
+            if result.shape[0] != self.num_envs:
+                raise RuntimeError(
+                    f"Predicate {_predicate_repr(predicate)} returned shape {tuple(result.shape)};"
+                    f" expected ({self.num_envs},)"
+                )
```
alexmillane: assert

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -0,0 +1,413 @@
...
+            for env_idx in torch.nonzero(advance_mask, as_tuple=False).flatten().tolist():
+                events.append({
+                    "env_idx": int(env_idx),
+                    "step": int(step_index[env_idx].item()) if step_index is not None else -1,
+                    "progress_objective": self.progress_objective.name,
+                    "group": group_name,
+                    "predicate_index": chain_idx,
+                    "predicate_name": pred_name,
+                    "score_delta": float(score_weight),
+                })
```
alexmillane: Do events always contain the same data? What do you think about making a dataclass for this. This will make it much easier to interpret, for example, the return types of this and other functions. A `list[PredicateEvent]` is much more descriptive than `list[dict]`

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -0,0 +1,413 @@
...
+        return stacked.sum(dim=1) >= int(self.progress_objective.K or 1)
+
+    def overall_score_per_env(self) -> torch.Tensor:
+        """Compute mean group score within this ProgressObjective (in [0, 1])."""
+
+        groups = self.progress_objective.group_names
+        stacked = torch.stack([self.group_score[g] for g in groups], dim=1)
+        return stacked.mean(dim=1)
+
+    def get_state_for_env(self, env_idx: int, is_complete, score) -> dict:
```
alexmillane: What do you think about making a type (dataclass) for the return value here? Right now this puts the responsibility on the caller to know what's inside, rather than describing that by using the type system.

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -0,0 +1,413 @@
...
+    if progress_tracker is None:
+        progress_tracker = ProgressTracker(
+            progress_objectives=progress_objectives, num_envs=env.num_envs, device=env.device
+        )
+        setattr(env, _PROGRESS_TRACKER_ATTR, progress_tracker)
+    return progress_tracker
+
+
+class ProgressTrackingRecorder(RecorderTerm):
+    """Per-step hook that ticks the ProgressTracker. Records nothing.
```
alexmillane, line 369: 😬 

I presume we're doing this because there's no good alternative in Isaac Lab? Seems weird to be commandeering the recorder manager to achieve per-step callbacks.

I guess there's no better way?

Note that now (since a couple of weeks) we have our own `env` wrapper: https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena/environments/isaaclab_arena_manager_based_env.py

Could you override something in there to get this per step? What about override `env.step` and then inserting your stuff either before or after `self.super().step()`

alexmillane: _Edit: I see below that we are actually recording something. We're just recording the `env.extras` rather than to disk._

Still, I feel that we're mixing concerns a bit here, by using a recorder to do the updates. Perhaps we can separate these things? Do the updating in `env.step` and the recording here.

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -360,6 +362,31 @@ def get_metrics(self) -> list[MetricBase]:
 
         return subtask_metrics
 
+    def get_own_progress_objectives(self) -> list[ProgressObjective]:
+        """Composite-level ProgressObjectives.
+
+        These are added on top of whatever progress objectives the child subtasks declare and are not gated.
+        """
+        return []
```
alexmillane: Do we forsee needing this?

If we don't have a clear need for it, I would remove it for now and add it when we need it.

alexmillane (approved):
Looks great! I just left a bunch of nits. Approved!

isaaclab_arena/progress_tracking/progress_objective.py
peterd-NV: Not quite, a "group" here is a collection of predicates that are evaluated in sequential order. When there are multiple groups, each group is run in parallel but the predicates within a group must execute in order. I've added a comment to explain that a bit more clearly in the code.

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: I think we can annotate it as `pred: Callable`, but we can't fully remove the getattrs since a predicate can be a partial (which would have those) but it can also be a basic function or callable. 

I can clean it up a bit though by splitting this up into an `isistance(pref, functools.partial)` and else branch. It gets rid of most of the get attrs and makes it bit more explicit.

isaaclab_arena/progress_tracking/progress_objective.py
peterd-NV: I used this to be the same form as Robolab, but yea I agree we use enums in other places like the ArmMode so no particular reason why we need to keep the robolab format for this. Changed it to enum.

isaaclab_arena/progress_tracking/fine_grained_progress_tracker.py
peterd-NV: Done

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: done

isaaclab_arena/progress_tracking/progress_objective.py
peterd-NV: Done: expanded comment and prefixed with underscore

isaaclab_arena/progress_tracking/fine_grained_progress_objective.py
peterd-NV: done

isaaclab_arena/progress_tracking/progress_objective.py
peterd-NV: done

isaaclab_arena/progress_tracking/progress_objective.py
peterd-NV: done

isaaclab_arena/progress_tracking/fine_grained_progress_objective.py
peterd-NV: done

isaaclab_arena/progress_tracking/progress_objective.py
peterd-NV: Yes that is correct. Added it to the docstring

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: Done, same as above comment for the older named file.

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: done

isaaclab_arena/progress_tracking/progress_objective.py
peterd-NV: done

isaaclab_arena/tasks/composite_task_base.py
peterd-NV: Removed and will add back in if we have a concrete need for it later on.

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: Yea I had a chat about this earlier with Clemens since we were hooking into the Terminations manager and then moved it to the recorder manager since it made more sense. Overriding env.step and inserting before or after super().step() probably won't work since we need this somewhere in the middle of the step function. For example if we placed it after the step then any env resets will cause us to lose the final predicate that is tied to task success since the reset will have occurred. Using the managers has the benefit that they automatically fire at the right time since they are placed at intentional positions within env.step.

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: Make's sense, added a state dataclass in a similar way to the PredicateEvent dataclass.

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: Added a PredicateEvent dataclass

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: Created a dataclass for return type to replace generic dict

PR #760, opened 2026-06-04
Derive Docker container name per clone

cvolkcvolk:
## Summary
Per-clone Docker container names. Enables working in several IsaacLab-Arena clones in parallel.

## Detailed description
- `run_docker.sh` defaulted every clone to the bare container name `isaaclab_arena-latest`. With a container already running, launching a second clone execed into the first clone's container instead of mounting its own tree, so _tests/builds silently ran the wrong code_.

- Auto-derive the container suffix from the repo directory name when `-s` is not given: 
`~/IsaacLab-Arena` keeps the bare name, `~/IsaacLab-Arena_<x>` becomes `isaaclab_arena-latest-<x>`.

## Example

**Before this PR — both clones compute the same name `isaaclab_arena-latest`:**

```
~/workspaces/IsaacLab-Arena    $ ./docker/run_docker.sh
   # name not running -> docker run, mounts . (IsaacLab-Arena) at /workspaces/isaaclab_arena

~/workspaces/IsaacLab-Arena_B  $ ./docker/run_docker.sh
   # name isaaclab_arena-latest is ALREADY running (clone A's)
   #   -> takes the "attach" branch: docker exec -it isaaclab_arena-latest ...
   #   -> the `-v .` mount line never runs
   #   -> inside, /workspaces/isaaclab_arena is still IsaacLab-Arena, NOT _B
```

**After this PR — the name is derived from the directory, so the clones never collide:**

```
~/workspaces/IsaacLab-Arena    $ ./docker/run_docker.sh   # -> isaaclab_arena-latest     (mounts IsaacLab-Arena)
~/workspaces/IsaacLab-Arena_B  $ ./docker/run_docker.sh   # -> isaaclab_arena-latest-B   (mounts IsaacLab-Arena_B)
```

Each clone gets its own container mounting its own tree, and the two run in parallel.

Agents/docs resolve their container by mount instead of a hardcoded name:

```
ARENA_CONTAINER=$(docker ps --filter "volume=$(git rev-parse --show-toplevel)" --format '{{.Names}}' | head -1)
```

### DIFF
--- a/.agents/skills/dev-container/SKILL.md
+++ b/.agents/skills/dev-container/SKILL.md
@@ -6,7 +6,19 @@ allowed-tools: Bash(./docker/run_docker.sh *) Bash(docker exec *) Bash(docker im
 
 # Dev Container
 
-Arena uses a single Docker container (`isaaclab_arena-latest`) as the dev, test, training, and eval environment. There is no separate dev container.
+Arena uses a single Docker container as the dev, test, training, and eval environment. There is no separate dev container.
+
+Each clone gets its own container so multiple clones run in parallel. The image (`isaaclab_arena:latest`) is shared; only the container name varies: `isaaclab_arena-latest` for the `IsaacLab-Arena` clone, `isaaclab_arena-latest-<suffix>` for an `IsaacLab-Arena_<suffix>` clone. `run_docker.sh` derives the suffix from the clone directory automatically.
+
+## Discover this clone's container
+
+Never hardcode the name. Resolve the container that mounts the current clone, then reuse `$ARENA_CONTAINER`:
+
+```bash
+ARENA_CONTAINER=$(docker ps --filter "volume=$(git rev-parse --show-toplevel)" --format '{{.Names}}' | head -1)
+```
+
+Empty result means no container is running for this clone — start one (below).
 
 ## Start or attach
 
@@ -14,7 +26,7 @@ Arena uses a single Docker container (`isaaclab_arena-latest`) as the dev, test,
 ./docker/run_docker.sh
 ```
 
-Idempotent: builds the image if it does not exist, starts the container if it is not running, then attaches.
+Idempotent: builds the (shared) image if it does not exist, starts this clone's container if it is not running, then attaches. The container name is auto-derived from the clone directory; pass `-s <suffix>` to override it.
 
 ## Common flag combinations
 
@@ -35,7 +47,7 @@ Example with custom mounts:
 ## Run a command in the already-running container
 
 ```bash
-docker exec isaaclab_arena-latest bash -c \
+docker exec "$ARENA_CONTAINER" bash -c \
   "cd /workspaces/isaaclab_arena && <command>"
 ```
 
@@ -50,7 +62,7 @@ Inside the container, `python` is aliased to `/isaac-sim/python.sh`. Both forms 
 A container is up and importable when:
 
 ```bash
-docker exec isaaclab_arena-latest bash -c \
+docker exec "$ARENA_CONTAINER" bash -c \
   "/isaac-sim/python.sh -c 'import isaaclab_arena; print(isaaclab_arena.__file__)'"
 ```
 
--- a/.agents/skills/dev-container/SKILL.md
+++ b/.agents/skills/dev-container/SKILL.md
@@ -8,17 +8,17 @@ allowed-tools: Bash(./docker/run_docker.sh *) Bash(docker exec *) Bash(docker im
 
 Arena uses a single Docker container as the dev, test, training, and eval environment. There is no separate dev container.
 
-Each clone gets its own container so clones run in parallel: the image (`isaaclab_arena:latest`) is shared, but the name is `isaaclab_arena-latest` for `IsaacLab-Arena` and `isaaclab_arena-latest-<suffix>` for `IsaacLab-Arena_<suffix>` (`run_docker.sh` derives the suffix automatically).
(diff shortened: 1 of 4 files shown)


.agents/skills/dev-container/SKILL.md
```diff
@@ -6,15 +6,27 @@ allowed-tools: Bash(./docker/run_docker.sh *) Bash(docker exec *) Bash(docker im
 
 # Dev Container
 
-Arena uses a single Docker container (`isaaclab_arena-latest`) as the dev, test, training, and eval environment. There is no separate dev container.
+Arena uses a single Docker container as the dev, test, training, and eval environment. There is no separate dev container.
+
+Each clone gets its own container so clones run in parallel: the image (`isaaclab_arena:latest`) is shared, but the name is `isaaclab_arena-latest` for `IsaacLab-Arena` and `isaaclab_arena-latest-<suffix>` for `IsaacLab-Arena_<suffix>` (`run_docker.sh` derives the suffix automatically).
```
alexmillane: "Each clone of the repo on the host machine gets its own..."

Each clone of the repo on the host machine gets its own container, so separate clones run can run in parallel. The image (`isaaclab_arena:latest`) is shared, but the container name is `isaaclab_arena-latest` for the folder named `IsaacLab-Arena` and `isaaclab_arena-latest-<suffix>` for folders named `IsaacLab-Arena_<suffix>` (`run_docker.sh` derives the suffix automatically).

.agents/skills/dev-container/SKILL.md
```diff
@@ -6,15 +6,27 @@ allowed-tools: Bash(./docker/run_docker.sh *) Bash(docker exec *) Bash(docker im
...
+
+## Discover this clone's container
+
+Never hardcode the name. Resolve the container mounting this clone (empty result = none running, so start one below):
+
+```bash
+docker ps --filter "volume=$(git rev-parse --show-toplevel)" --format '{{.Names}}' | head -1
+```
+
+Substitute that name for `$ARENA_CONTAINER` in the commands below — a shell variable won't carry across separate `docker exec` calls, so use the literal name (or set `ARENA_CONTAINER` in the same shell as each command).
```
alexmillane: Do we want the the agent to have to do this _every_ time it runs a command? This instruction implies that... Maybe that's OK?

Personally I woul have said, "at the start of a session discover the container name...".

.agents/skills/run-tests/SKILL.md
```diff
@@ -14,12 +14,20 @@ allowed-tools: Bash(docker exec *)
 
 The three phases run mutually exclusive sets of tests (selected via different pytest marker filters) and must be invoked separately. For a full pre-PR run, every phase must pass before moving on to the next.
 
+All commands run inside the already-running container via `docker exec`. Each clone has its own container, so first resolve the one mounting this clone (empty = none running, use the `dev-container` skill first):
```
alexmillane: each clone of the repo on the host system has...

.agents/skills/run-tests/SKILL.md
```diff
@@ -14,12 +14,20 @@ allowed-tools: Bash(docker exec *)
...
+
+```bash
+docker ps --filter "volume=$(git rev-parse --show-toplevel)" --format '{{.Names}}' | head -1
+```
+
+Substitute that name for `$ARENA_CONTAINER` in the commands below — a shell variable won't carry across separate `docker exec` calls, so use the literal name (or set `ARENA_CONTAINER` in the same shell as each command).
```
alexmillane: This seems to imply that you set the variable `ARENA_CONTAINER`. Should we add that to the command?

i.e. start the command with:

```
ARENA_CONTAINER=$(docker ps --filter "volume=$(git rev-parse --show-toplevel)" --format '{{.Names}}' | head -1)
```

or something

AGENTS.md
```diff
@@ -20,19 +20,19 @@ pre-commit install    # on the host — registers git pre-commit hooks
 
 ## Docker environment
 
-Commands that touch Isaac Sim or Arena's package code (tests, training, evaluation, runtime scripts) run inside the `isaaclab_arena-latest` Docker container. The repo root is mounted at `/workspaces/isaaclab_arena`. Inside the container, `python` is aliased to `/isaac-sim/python.sh` — prefer the explicit path in `docker exec` invocations from outside the container, where the alias is not active.
+Commands that touch Isaac Sim or Arena's package code (tests, training, evaluation, runtime scripts) run inside this clone's Docker container. The repo root is mounted at `/workspaces/isaaclab_arena`. Inside the container, `python` is aliased to `/isaac-sim/python.sh` — prefer the explicit path in `docker exec` invocations from outside the container, where the alias is not active.
```
alexmillane: this clone -> "the clone of the repo"

alexmillane: I feel like "this clone" is a bit ambiguous.

alexmillane (approved):
Thanks for fixing this!

.agents/skills/dev-container/SKILL.md
cvolkcvolk: Updated

.agents/skills/dev-container/SKILL.md
cvolkcvolk: (Only) after the skill is invoked it is loaded into the agents sessions memory. The resolved name then is in the session context and should get reused on later turns. At least it should from my understanding of how session contexts work.
Reworded slightly though. I think it does not add any complexity.

.agents/skills/run-tests/SKILL.md
cvolkcvolk: Done

.agents/skills/run-tests/SKILL.md
cvolkcvolk: Done. Updated the other skill the same way

.agents/skills/dev-container/SKILL.md
cvolkcvolk: Actually you had a good comment below. At least for using the literal string, I prefer a direct assignment via
` ARENA_CONTAINER=$(docker ps --filter "volume=$(git rev-parse --show-toplevel)" --format '{{.Names}}' | head -1)
 `

PR #761, opened 2026-06-04
Feature/nvblox next datagen

viiik-inside:
## Summary

## Detailed description
- What has been changed?


isaaclab_arena/evaluation/policy_runner.py
```diff
@@ -84,41 +209,48 @@ def rollout_policy(
...
-                    num_episodes_completed += completed_episodes
-                    if hasattr(env.unwrapped.cfg, "metrics") and env.unwrapped.cfg.metrics is not None:
-                        metrics = env.unwrapped.compute_metrics()
-                        tqdm.tqdm.write(
-                            f"[Rank {get_local_rank()}/{get_world_size()}] Metrics:"
-                            f" {metrics_to_plain_python_types(metrics)}"
+        if collector is not None:
+            # Datagen path: env auto-reset is disabled, so we drive episode boundaries
+            # and resets explicitly (see _run_datagen_rollout / record_demos.py).
+            _run_datagen_rollout(
```
david-tingdahl-nvidia, line 215: We should use the same loop with and without collector. not hijack the flow with separate function.
I believe current reason is due to different reset criteria being in use (fix stale fist-frame). needs revisit

isaaclab_arena_datagen/audit_dynamic_extents.py
david-tingdahl-nvidia, line 1: Let's revisit dataset auditing in a separate PR

isaaclab_arena/evaluation/episode_outcome.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Map the termination term that ended a datagen episode to an outcome label."""
+
+from __future__ import annotations
+
+
+def classify_outcome(ended_by: str | None) -> str:
```
david-tingdahl-nvidia, line 10: Switch to enum instead of magic strings

isaaclab_arena_datagen/collection/collector.py
```diff
@@ -0,0 +1,337 @@
...
+        output_dir: Directory where per-episode ``episode_NNNN/dataset.h5`` files are written.
+        cameras: Explicit camera trajectories. When ``None`` (default), the
+            collector uses the environment class's ``get_default_cameras`` if it
+            has one, else falls back to a single default view.
+        width: Image width in pixels, shared by all cameras.
+        height: Image height in pixels, shared by all cameras.
+        dynamic_translation_eps: Per-step translation threshold (metres).
+        dynamic_rotation_eps: Per-step rotation threshold (radians).
+        mesh_sample_spacing: Mesh surface sample spacing (metres).
+        camera_sampler: Optional callable returning a fresh fixed-length
```
david-tingdahl-nvidia, line 62: Can we just pass the trajectories as optional instead? why need a function handle?

isaaclab_arena_datagen/collection/collector.py
```diff
@@ -0,0 +1,337 @@
...
+        """Re-pose the existing camera sensors to *trajectories* (look-at) in place."""
+        assert len(trajectories) == len(self._camera_setups), (
+            f"camera_sampler returned {len(trajectories)} cameras but "
+            f"{len(self._camera_setups)} were spawned; the count must stay fixed."
+        )
+        for cam, traj in zip(self._camera_setups, trajectories):
+            cam.handler.set_world_pose(resolve_coord(traj.position, 0), resolve_coord(traj.target, 0))
+            cam.trajectory = traj
+
+    def resample_cameras(self) -> None:
```
david-tingdahl-nvidia, line 221: There is quite some logic required to handle first-frame issues. I wonder if we should simply skip recording the first (few) frames instead. Even with these fixes in place, they're still unreliable (DLSS noise, gravity dropping stationary objects into place ,...)

isaaclab_arena_datagen/collection/collector.py
```diff
@@ -0,0 +1,337 @@
...
+from isaaclab_arena_datagen.utils.camera_utils import resolve_coord
+from isaaclab_arena_datagen.utils.constants import DEFAULT_ROTATION_EPS_RAD, DEFAULT_TRANSLATION_EPS_M
+
+# Extra capacity over max_episode_length when pre-allocating an episode file,
+# guarding against off-by-one between the env's horizon and recorded frames.
+_CAPACITY_MARGIN = 2
+
+
+@dataclasses.dataclass
+class DatagenCollectorConfig:
```
david-tingdahl-nvidia, line 49: This class does three main things:
1. Dumps data to nvblox_next dataset
2. Owns and setups observation cameras
3. Keep tracks of dynamic objects (using a class for this)

I wonder if it make sense to break this functionality up and make this class a pure orchestrator. E.g. an idnependent  `ObservationCameras` class might be useful elsewhere in Arena.

 Let's discuss this offline

isaaclab_arena_datagen/manifest.py
david-tingdahl-nvidia, line 1: port to nvblox_next?

david-tingdahl-nvidia: Another option is to keep the data writer in arena instead of nvblox_next, if such functionality is desired in Arena. Not sure if it harmonizes well with any existing writers. 

Note that we're also on extending the nvblox_next writer to include robot state - perhaps we can find a common ground

isaaclab_arena_datagen/io/hdf5_keys.py
david-tingdahl-nvidia, line 1: to nvblox_next

isaaclab_arena_datagen/environments/miscellaneous/ball_and_box_environment.py
david-tingdahl-nvidia, line 1: I think this whole directory with synthetic scenes can be removed from Arena

isaaclab_arena_datagen/geometry/__init__.py
david-tingdahl-nvidia, line 1: Let's adapt the code that stays in Arena to not rely on these helpers.
If anything is remaining - try to align with any existing geometry helpers

isaaclab_arena_datagen/__init__.py
david-tingdahl-nvidia, line 1: This whole module is copied from nvblox_next and adapted for IsaacLab 3.0

Some modules have been added, most notably `collector.py` 

We have to decide what to do with it. 

IMO we have two options:
* **Keep datagen in nvblox_next:** we'd then have to implement an Arena-only dataset writer to be used in `eval_runner.py`. nvblox_next would contain utils for converting the dataset to nvblox_next's format
* **Keep datagen here and deprecate in nvblox_next:** If Arena can find use of the tools here, this would be a simpler solution. nvblox would then use the dataset format specified by Arena only.

isaaclab_arena_environments/eval_jobs_configs/pick_and_place_datagen_jobs_config.json
david-tingdahl-nvidia, line 1: the job configs better go to nvblox_next eventually

david-tingdahl-nvidia:
I left some comments, mostly for myself. I'll look into migrating this

PR #762, opened 2026-06-04
Run in-container agent commands as the host user

alexmillane:
## Summary
Agents now run in container as the host user, not root

## Detailed description
- We were running as root which could cause bugs where assets were downloaded as root, and therefore inaccessable by human users.
- We now instruct agents to run as the user.


cvolkcvolk (approved):

xyao-nv (approved):
Thanks for fixing it in the right way!

PR #766, opened 2026-06-05
Synchronized Multi-env Visualization

zhx06:
## Summary
Add synchronized multi-env visualization utility

## Detailed description
- Added `SynchronizedVisualizer`: global auto-framed camera + per-env `TiledCamera` grid anchored to env_origins, with mp4/gif output
- Added related tests

### DIFF
--- a/isaaclab_arena/utils/synchronized_visualizer.py
+++ b/isaaclab_arena/utils/synchronized_visualizer.py
@@ -20,9 +20,8 @@ offset from each env origin and the global view is framed from the env-origin
 bounding box. GlobalView takes world-frame eye/target (or None to auto-frame);
 EnvView takes eye/target offsets relative to each env origin.
 
-Both views read what env.step already rendered, so frames stay in sync. Under
---enable_cameras, env.render() reuses that render product instead of running its
-own sim.render().
+env.render() retrieves what env.step already rendered, so the two views stay in
+sync without an extra physics step.
 
 Usage:
 
@@ -54,6 +53,11 @@ if TYPE_CHECKING:
     import gymnasium
 
     from isaaclab.sensors.camera import Camera
+
+# Horizontal aperture [mm] shared by the per-env and global pinhole cameras. Paired
+# with focal_length to set the horizontal FOV; keep both cameras on one value so
+# their framing matches.
+HORIZONTAL_APERTURE_MM = 20.955
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


isaaclab_arena/utils/synchronized_visualizer.py
```diff
@@ -0,0 +1,642 @@
...
+                    update_period=0,
+                    width=self.env_view.width,
+                    height=self.env_view.height,
+                    data_types=["rgb"],
+                    spawn=_pinhole(self.env_view.focal_length),
+                )
+            )
+
+        sim_utils.create_prim(f"{self.prim_root}/GlobalCam", "Xform")
+        self._global_camera = Camera(
```
xyao-nv: Can the "global cam" just be the viewport camera? Your logics of setting its pose and viewing angle will still be needed. From perf perspective, it's better we can reuse rather than adding another one.

isaaclab_arena/utils/synchronized_visualizer.py
```diff
@@ -0,0 +1,642 @@
...
+        ``render_mode=None`` the env is not rendered automatically by
+        ``env.step``, so an explicit ``sim.render()`` is required first. Cameras
+        are updated here.
+        """
+        assert self._initialized, "Call initialize() before capture()."
+        assert self._env_camera is not None and self._global_camera is not None
+        dt = float(getattr(self.unwrapped, "step_dt", 0.0))
+
+        self._env_camera.update(dt=dt)
+        self._global_camera.update(dt=dt)
```
xyao-nv: Is it really necessary to call sim.render() and cam.update() in a separate process than env.step()? Touching the rendering frame buffer is easy to result in frame out-of-sync bugs.

xyao-nv:
Can we attach the camera sensor to the Scene, like how contact sensor is added to Asset?

`Scene.add_env_cam(env_view, name=...)` that attaches a `TiledCamCfg` to SceneCfg.
Or even  `Scene.add_sensor(sensor_cfg, name=...)`?

isaaclab_arena/utils/synchronized_visualizer.py
zhx06: Addressed. Now it is a viewport camera, where we have an extra camera prim to make it work correctly in headless mode.

isaaclab_arena/utils/synchronized_visualizer.py
zhx06: No, we don't do that anymore. This is removed.

PR #768, opened 2026-06-08
Drop reload_arena_modules from eval job loop

cvolkcvolk:
## Summary
Remove per-job module reload from eval_runner.

## Detailed description
- **What `reload_arena_modules()` is for:** It deletes the cached bytecode and
  re-imports every `isaaclab_arena*` module, so a running Python process picks up
  source edits without a restart. The intended use case is the interactive dev
  notebooks (`example_env_notebook.py`, `compile_env_notebook.py`).

- **Why it is not needed here:** `eval_runner` runs a fixed list of jobs from a
  config and never edits source between them. Reloading every module on each job
  recompiles unchanged code for no benefit, and each reload re-runs the module
  bodies, re-creating the gym registries and env-config dataclasses. Those pile up
  in the process and are never reclaimed (~5 MB plus config/tensor churn per job).

- **Impact:** No behavior change for eval, since source is not edited mid-run.


xyao-nv:
aha thanks for explaining and fixing!

xyao-nv (approved):

PR #771, opened 2026-06-09
Mesh-based Non-collision Constraints 

zhx06:
## Summary

Add mesh-based non-collision constraints via sphere-to-SDF, unified with the existing AABB path

## Detailed description

- Introduces `CollisionMode.MESH` as an alternative to AABB for no-overlap constraints, using greedy sphere decomposition + differentiable Warp SDF queries against actual collision geometry.
- Solver falls back to AABB for pairs where either object lacks a mesh. Validator mirrors this.
- Unifies naming and computational flow between AABB and mesh modes: both use subject/obstacle terminology, forward/reverse directed pairs, and broadphase gating.
- `MeshPairCache` dataclass and `MeshPairEntry` NamedTuple give the mesh path the same collect-then-batch structure as the AABB vectorized path.
- Environment-level `placer_params` is the sole configuration source for `collision_mode` and `random_yaw_init` (the `--random_yaw_init` CLI flag is removed; these are environment-level decisions, not runtime toggles).
- Shared yaw rotation utilities (`rotate_points_by_yaw`, `centers_in_target_frame`) in `utils/pose.py` serve both the GPU solver and CPU validator, eliminating duplicate rotation math.

### DIFF
--- a/isaaclab_arena/assets/dummy_object.py
+++ b/isaaclab_arena/assets/dummy_object.py
@@ -5,14 +5,11 @@
 from __future__ import annotations
 
 import torch
-from typing import TYPE_CHECKING
+import trimesh
 
 from isaaclab_arena.relations.relations import IsAnchor, Relation, RelationBase, UnaryRelation
 from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox, quaternion_to_90_deg_z_quarters
 from isaaclab_arena.utils.pose import Pose
-
-if TYPE_CHECKING:
-    import trimesh
 
 
 class DummyObject:
--- a/isaaclab_arena/assets/dummy_object.py
+++ b/isaaclab_arena/assets/dummy_object.py
@@ -2,17 +2,21 @@
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
-import torch
+from __future__ import annotations
 
-from isaaclab_arena.relations.relations import Relation, RelationBase, UnaryRelation
+import torch
+from typing import TYPE_CHECKING
+
+from isaaclab_arena.relations.relations import IsAnchor, Relation, RelationBase, UnaryRelation
 from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox, quaternion_to_90_deg_z_quarters
 from isaaclab_arena.utils.pose import Pose
 
+if TYPE_CHECKING:
+    import trimesh
+
 
 class DummyObject:
-    """
-    Dummy object for testing purposes without Isaac Sim dependencies.
-    """
+    """Dummy object for testing purposes without Isaac Sim dependencies."""
 
     def __init__(
         self,
@@ -20,6 +24,7 @@ class DummyObject:
         bounding_box: AxisAlignedBoundingBox,
         initial_pose: Pose | None = None,
         relations: list[RelationBase] = [],
+        collision_mesh: trimesh.Trimesh | None = None,
         **kwargs,
     ):
         self.name = name
@@ -27,6 +32,7 @@ class DummyObject:
         self.bounding_box = bounding_box
         assert self.bounding_box is not None
         self.relations = list(relations)
+        self._collision_mesh = collision_mesh
 
     def add_relation(self, relation: RelationBase) -> None:
         self.relations.append(relation)
@@ -63,3 +69,11 @@ class DummyObject:
 
     def is_initial_pose_set(self) -> bool:
         return self.initial_pose is not None
+
+    @property
+    def is_anchor(self) -> bool:
+        return any(isinstance(r, IsAnchor) for r in self.relations)
+
+    def get_collision_mesh(self) -> trimesh.Trimesh | None:
+        """Return the collision mesh, or None to fall back to AABB."""
+        return self._collision_mesh
--- a/isaaclab_arena/assets/object.py
+++ b/isaaclab_arena/assets/object.py
@@ -78,7 +78,7 @@ class Object(ObjectBase):
         return self.bounding_box
 
     def get_collision_mesh(self) -> trimesh.Trimesh | None:
-        """Return None; collision mesh is not available at the object level."""
+        """Collision mesh is unavailable for USD-backed objects; subclasses with preloaded geometry may override."""
 
     def get_world_bounding_box(self) -> AxisAlignedBoundingBox:
         """Get bounding box in world coordinates (local bbox rotated and translated).
--- a/isaaclab_arena/assets/object.py
+++ b/isaaclab_arena/assets/object.py
@@ -78,7 +78,7 @@ class Object(ObjectBase):
         return self.bounding_box
 
     def get_collision_mesh(self) -> trimesh.Trimesh | None:
-        """Collision mesh is unavailable for USD-backed objects; subclasses with preloaded geometry may override."""
+        """Return None: USD-backed objects expose no preloaded collision mesh."""
 
     def get_world_bounding_box(self) -> AxisAlignedBoundingBox:
         """Get bounding box in world coordinates (local bbox rotated and translated).
--- a/isaaclab_arena/assets/object.py
+++ b/isaaclab_arena/assets/object.py
@@ -2,8 +2,13 @@
 # All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
 import torch
-from typing import Any
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    import trimesh
 
 from isaaclab.assets import ArticulationCfg, AssetBaseCfg, RigidObjectCfg
 from isaaclab.sensors.contact_sensor.contact_sensor_cfg import ContactSensorCfg
@@ -20,9 +25,7 @@ from isaaclab_arena.utils.usd_helpers import compute_local_bounding_box_from_usd
 
 
 class Object(ObjectBase):
-    """
-    Encapsulates the pick-up object config for a pick-and-place environment.
-    """
+    """Pick-up object config for a pick-and-place environment."""
 
     def __init__(
         self,
@@ -73,6 +76,9 @@ class Object(ObjectBase):
         if self.bounding_box is None:
             self.bounding_box = compute_local_bounding_box_from_usd(self.usd_path, self.scale)
         return self.bounding_box
+
+    def get_collision_mesh(self) -> trimesh.Trimesh | None:
+        """Return None; collision mesh is not available at the object level."""
 
     def get_world_bounding_box(self) -> AxisAlignedBoundingBox:
         """Get bounding box in world coordinates (local bbox rotated and translated).
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -7,9 +7,13 @@ from __future__ import annotations
 
 import torch
 from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
 
 import warp as wp
 from isaaclab.assets import ArticulationCfg, AssetBaseCfg, RigidObjectCfg
+
+if TYPE_CHECKING:
+    import trimesh
 from isaaclab.envs import ManagerBasedEnv
 from isaaclab.managers import EventTermCfg, SceneEntityCfg
 from isaaclab.sensors.contact_sensor.contact_sensor_cfg import ContactSensorCfg
@@ -72,6 +76,9 @@ class ObjectBase(Asset, ABC):
     def get_world_bounding_box(self) -> AxisAlignedBoundingBox:
         """Get bounding box in world coordinates (local bbox rotated and translated)."""
         ...
+
+    def get_collision_mesh(self) -> trimesh.Trimesh | None:
+        """Return collision mesh, or None to fall back to AABB overlap."""
 
     def _get_initial_pose_as_pose(self) -> Pose | None:
         """Return a single ``Pose`` suitable for *init_state* and bounding-box calculations.
@@ -169,6 +176,13 @@ class ObjectBase(Asset, ABC):
     def get_relations(self) -> list[RelationBase]:
         """Get all relations for this object."""
         return self.relations
+
+    @property
+    def is_anchor(self) -> bool:
+        """True if this object has an IsAnchor relation."""
+        from isaaclab_arena.relations.relations import IsAnchor
+
+        return any(isinstance(r, IsAnchor) for r in self.relations)
 
     def get_spatial_relations(self) -> list[RelationBase]:
         """Get only spatial relations (On, NextTo, AtPosition, etc.), excluding markers like IsAnchor."""
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -26,7 +26,7 @@ from isaaclab_arena.assets.asset import Asset
 # while pure-Python spec modules can import from `object_type` directly without
 # pulling in isaaclab/omni/pxr at module-load time.
 from isaaclab_arena.assets.object_type import ObjectType
-from isaaclab_arena.relations.relations import Relation, RelationBase, UnaryRelation
+from isaaclab_arena.relations.relations import IsAnchor, Relation, RelationBase, UnaryRelation
 from isaaclab_arena.terms.events import set_object_pose, set_object_pose_per_env
 from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox
 from isaaclab_arena.utils.pose import Pose, PosePerEnv, PoseRange
@@ -180,8 +180,6 @@ class ObjectBase(Asset, ABC):
     @property
     def is_anchor(self) -> bool:
         """True if this object has an IsAnchor relation."""
-        from isaaclab_arena.relations.relations import IsAnchor
-
         return any(isinstance(r, IsAnchor) for r in self.relations)
 
     def get_spatial_relations(self) -> list[RelationBase]:
--- a/isaaclab_arena/cli/isaaclab_arena_cli.py
+++ b/isaaclab_arena/cli/isaaclab_arena_cli.py
@@ -77,27 +77,10 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
         ),
     )
     arena_group.add_argument(
-        "--random_yaw_init",
-        action="store_true",
-        default=False,
-        help=(
-            "Randomly rotate objects (except anchors) around the Z-axis for scene variety. "
-            "Collisions use a larger enclosing box; the solver won't optimize this rotation. "
-            "Only affects objects positioned by the placement solver; manually-placed objects are unaffected."
-        ),
-    )
-    arena_group.add_argument(
         "--list-variations",
         action="store_true",
         default=False,
         help="Print Hydra-configurable variations for the selected environment and exit.",
-    )
-    arena_group.add_argument(
-        "--collision_mode",
-        type=str,
-        choices=["bbox", "mesh"],
-        default="bbox",
-        help="Collision detection mode: 'bbox' (AABB, default) or 'mesh' (sphere-to-SDF, requires Warp).",
     )
 
 
--- a/isaaclab_arena/cli/isaaclab_arena_cli.py
+++ b/isaaclab_arena/cli/isaaclab_arena_cli.py
@@ -92,6 +92,13 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
         default=False,
         help="Print Hydra-configurable variations for the selected environment and exit.",
     )
+    arena_group.add_argument(
+        "--collision_mode",
+        type=str,
+        choices=["bbox", "mesh"],
+        default="bbox",
+        help="Collision detection mode: 'bbox' (AABB, default) or 'mesh' (sphere-to-SDF, requires Warp).",
+    )
 
 
 def add_env_graph_spec_cli_args(parser: argparse.ArgumentParser) -> None:
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -36,11 +36,13 @@ from isaaclab_arena.progress_tracking.progress_tracker import (
 )
 from isaaclab_arena.recording.common_terms import CoreEpisodeRecorderTermCfg, VariationEpisodeRecorderTermCfg
 from isaaclab_arena.recording.episode_recorder_manager import EpisodeRecorderTermCfg
+<<<<<<< HEAD
 from isaaclab_arena.recording.progress_terms import ProgressEpisodeRecorderTermCfg
 from isaaclab_arena.relations.collision_mode import CollisionMode
+=======
+>>>>>>> a690c8f59 (address comments)
 from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
 from isaaclab_arena.relations.placement_events import PLACEMENT_RESET_EVENT_NAME
-from isaaclab_arena.relations.relation_solver_params import RelationSolverParams
 from isaaclab_arena.tasks.no_task import NoTask
 from isaaclab_arena.utils.configclass import combine_configclass_instances, make_configclass
 from isaaclab_arena.utils.isaaclab_utils.simulation_app import reapply_viewer_cfg
@@ -87,20 +89,9 @@ class ArenaEnvBuilder:
         """
         objects_with_relations = self.arena_env.scene.get_objects_with_relations()
 
-        # Prefer env-level placer_params; fall back to CLI-constructed defaults.
         placer_params = self.arena_env.placer_params
         if placer_params is None:
-            collision_mode_str = getattr(self.args, "collision_mode", "bbox")
-            mode = CollisionMode.MESH if collision_mode_str == "mesh" else CollisionMode.BBOX
-            placer_params = ObjectPlacerParams(
-                placement_seed=self.args.placement_seed,
-                random_yaw_init=self.args.random_yaw_init,
-                solver_params=RelationSolverParams(
-                    collision_mode=mode,
-                    save_position_history=False,
-                    verbose=False,
-                ),
-            )
+            placer_params = ObjectPlacerParams(placement_seed=self.args.placement_seed)
             if self.args.resolve_on_reset is not None:
                 placer_params.resolve_on_reset = self.args.resolve_on_reset
         self._placement_event_cfg = solve_and_apply_relation_placement(
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -36,11 +36,7 @@ from isaaclab_arena.progress_tracking.progress_tracker import (
 )
 from isaaclab_arena.recording.common_terms import CoreEpisodeRecorderTermCfg, VariationEpisodeRecorderTermCfg
 from isaaclab_arena.recording.episode_recorder_manager import EpisodeRecorderTermCfg
-<<<<<<< HEAD
 from isaaclab_arena.recording.progress_terms import ProgressEpisodeRecorderTermCfg
-from isaaclab_arena.relations.collision_mode import CollisionMode
-=======
->>>>>>> a690c8f59 (address comments)
 from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
 from isaaclab_arena.relations.placement_events import PLACEMENT_RESET_EVENT_NAME
 from isaaclab_arena.tasks.no_task import NoTask
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -37,7 +37,10 @@ from isaaclab_arena.progress_tracking.progress_tracker import (
 from isaaclab_arena.recording.common_terms import CoreEpisodeRecorderTermCfg, VariationEpisodeRecorderTermCfg
 from isaaclab_arena.recording.episode_recorder_manager import EpisodeRecorderTermCfg
 from isaaclab_arena.recording.progress_terms import ProgressEpisodeRecorderTermCfg
+from isaaclab_arena.relations.collision_mode import CollisionMode
+from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
 from isaaclab_arena.relations.placement_events import PLACEMENT_RESET_EVENT_NAME
+from isaaclab_arena.relations.relation_solver_params import RelationSolverParams
 from isaaclab_arena.tasks.no_task import NoTask
 from isaaclab_arena.utils.configclass import combine_configclass_instances, make_configclass
 from isaaclab_arena.utils.isaaclab_utils.simulation_app import reapply_viewer_cfg
@@ -83,12 +86,27 @@ class ArenaEnvBuilder:
           events restore the same layout every time.
         """
         objects_with_relations = self.arena_env.scene.get_objects_with_relations()
+
+        # Prefer env-level placer_params; fall back to CLI-constructed defaults.
+        placer_params = self.arena_env.placer_params
+        if placer_params is None:
+            collision_mode_str = getattr(self.args, "collision_mode", "bbox")
+            mode = CollisionMode.MESH if collision_mode_str == "mesh" else CollisionMode.BBOX
+            placer_params = ObjectPlacerParams(
+                placement_seed=self.args.placement_seed,
+                random_yaw_init=self.args.random_yaw_init,
+                solver_params=RelationSolverParams(
+                    collision_mode=mode,
+                    save_position_history=False,
+                    verbose=False,
+                ),
+            )
+            if self.args.resolve_on_reset is not None:
+                placer_params.resolve_on_reset = self.args.resolve_on_reset
         self._placement_event_cfg = solve_and_apply_relation_placement(
             objects_with_relations,
             num_envs=self.args.num_envs,
-            placement_seed=self.args.placement_seed,
-            resolve_on_reset=self.args.resolve_on_reset,
-            random_yaw_init=self.args.random_yaw_init,
+            placer_params=placer_params,
         )
 
     def get_all_variations(self) -> dict[str, list[VariationBase]]:
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -39,6 +39,7 @@ from isaaclab_arena.recording.episode_recorder_manager import EpisodeRecorderTer
 from isaaclab_arena.recording.progress_terms import ProgressEpisodeRecorderTermCfg
 from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
 from isaaclab_arena.relations.placement_events import PLACEMENT_RESET_EVENT_NAME
+from isaaclab_arena.relations.relation_solver_params import RelationSolverParams
 from isaaclab_arena.tasks.no_task import NoTask
 from isaaclab_arena.utils.configclass import combine_configclass_instances, make_configclass
 from isaaclab_arena.utils.isaaclab_utils.simulation_app import reapply_viewer_cfg
@@ -87,7 +88,10 @@ class ArenaEnvBuilder:
 
         placer_params = self.arena_env.placer_params
         if placer_params is None:
-            placer_params = ObjectPlacerParams(placement_seed=self.args.placement_seed)
+            placer_params = ObjectPlacerParams(
+                placement_seed=self.args.placement_seed,
+                solver_params=RelationSolverParams(verbose=False, save_position_history=False),
+            )
             if self.args.resolve_on_reset is not None:
                 placer_params.resolve_on_reset = self.args.resolve_on_reset
         self._placement_event_cfg = solve_and_apply_relation_placement(
--- a/isaaclab_arena/environments/isaaclab_arena_environment.py
+++ b/isaaclab_arena/environments/isaaclab_arena_environment.py
@@ -13,6 +13,7 @@ if TYPE_CHECKING:
     from isaaclab_arena.embodiments.embodiment_base import EmbodimentBase
     from isaaclab_arena.environments.isaaclab_arena_manager_based_env_cfg import IsaacLabArenaManagerBasedRLEnvCfg
     from isaaclab_arena.recording.episode_recorder_manager import EpisodeRecorderTermCfg
+    from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
     from isaaclab_arena.scene.scene import Scene
     from isaaclab_arena.tasks.task_base import TaskBase
 
@@ -31,6 +32,7 @@ class IsaacLabArenaEnvironment:
         rl_framework_entry_point: str | None = None,
         rl_policy_cfg: str | None = None,
         episode_recorder_terms: dict[str, EpisodeRecorderTermCfg] | None = None,
+        placer_params: ObjectPlacerParams | None = None,
     ):
         """
         Args:
@@ -50,6 +52,8 @@ class IsaacLabArenaEnvironment:
                 ``"my_module:RLPolicyCfg"``.
             episode_recorder_terms: Additional per-episode recorder terms to record alongside the
                 built-in ones, keyed by name.
+            placer_params: Object placement configuration. When set, used as-is
+                (CLI flags are ignored). When None, params are built from CLI flags.
         """
         self.name = name
         self.scene = scene
@@ -62,3 +66,4 @@ class IsaacLabArenaEnvironment:
         self.rl_framework_entry_point = rl_framework_entry_point
         self.rl_policy_cfg = rl_policy_cfg
         self.episode_recorder_terms = episode_recorder_terms or {}
+        self.placer_params = placer_params
--- a/isaaclab_arena/environments/isaaclab_arena_environment.py
+++ b/isaaclab_arena/environments/isaaclab_arena_environment.py
@@ -52,8 +52,8 @@ class IsaacLabArenaEnvironment:
                 ``"my_module:RLPolicyCfg"``.
             episode_recorder_terms: Additional per-episode recorder terms to record alongside the
                 built-in ones, keyed by name.
-            placer_params: Object placement configuration. When set, used as-is
-                (CLI flags are ignored). When None, params are built from CLI flags.
+            placer_params: Object placement configuration. When None, default
+                ObjectPlacerParams are used.
         """
         self.name = name
         self.scene = scene
--- a/isaaclab_arena/environments/relation_solver_interface.py
+++ b/isaaclab_arena/environments/relation_solver_interface.py
@@ -5,12 +5,12 @@
 
 from __future__ import annotations
 
+import copy
 from typing import TYPE_CHECKING
 
 from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
 from isaaclab_arena.relations.placement_events import get_rotation_xyzw, solve_and_place_objects
 from isaaclab_arena.relations.pooled_object_placer import PooledObjectPlacer
-from isaaclab_arena.relations.relation_solver_params import RelationSolverParams
 from isaaclab_arena.relations.relations import get_anchor_objects
 from isaaclab_arena.utils.pose import Pose, PosePerEnv, rotate_quat_by_yaw
 
@@ -23,39 +23,19 @@ if TYPE_CHECKING:
 def solve_and_apply_relation_placement(
     objects: list[ObjectBase],
     num_envs: int,
-    placement_seed: int | None = None,
-    resolve_on_reset: bool | None = None,
-    random_yaw_init: bool = False,
+    placer_params: ObjectPlacerParams | None = None,
 ) -> EventTermCfg | None:
-    """Solve relation placement and apply the result to object reset/static state.
-
-    Args:
-        objects: Objects with spatial predicates that should be relation-solved.
-        num_envs: Number of environments to prepare placements for.
-        placement_seed: Optional random seed for reproducible object placement.
-        resolve_on_reset: Optional override for whether to draw fresh layouts from
-            the placement pool on reset. When ``False``, fixed per-environment
-            initial poses are applied immediately.
-        random_yaw_init: If True, randomly rotates non-anchor objects around the vertical (Z)
-            axis at startup to add visual variety to the scene.
-
-    Returns:
-        Reset event config to attach to the environment when placement should be
-        resolved on reset. Returns ``None`` when no reset event is needed.
-    """
+    """Solve relation placement and return a reset EventTermCfg (or None if no objects)."""
     objects = list(objects)
     if not objects:
         print("No objects with relations found in scene. Skipping relation solving.")
         return None
 
-    placer_params = ObjectPlacerParams(
-        placement_seed=placement_seed,
-        apply_positions_to_objects=False,
-        solver_params=RelationSolverParams(save_position_history=False, verbose=False),
-        random_yaw_init=random_yaw_init,
-    )
-    if resolve_on_reset is not None:
-        placer_params.resolve_on_reset = resolve_on_reset
+    if placer_params is None:
+        placer_params = ObjectPlacerParams()
+    else:
+        placer_params = copy.copy(placer_params)
+    placer_params.apply_positions_to_objects = False
 
     # TODO(xinjieyao, 2026-05-22): Add joint object/embodiment placement once task-dependent
     # reachability constraints are available. For now this always uses the object-only placer.
--- a/isaaclab_arena/environments/relation_solver_interface.py
+++ b/isaaclab_arena/environments/relation_solver_interface.py
@@ -12,7 +12,7 @@ from isaaclab_arena.relations.object_placer_params import ObjectPlacerParams
 from isaaclab_arena.relations.placement_events import get_rotation_xyzw, solve_and_place_objects
 from isaaclab_arena.relations.pooled_object_placer import PooledObjectPlacer
 from isaaclab_arena.relations.relations import get_anchor_objects
-from isaaclab_arena.utils.pose import Pose, PosePerEnv, rotate_quat_by_yaw
+from isaaclab_arena.utils.pose import Pose, PosePerEnv, rotate_quat_by_yaw, yaw_from_quat_xyzw
 
 if TYPE_CHECKING:
     from isaaclab.managers import EventTermCfg
@@ -107,8 +107,10 @@ def _apply_dynamic_spawn_pose(
         pos = layout.positions.get(obj)
         if pos is None:
             continue
-        yaw = layout.orientations.get(obj, 0.0)
-        rot = rotate_quat_by_yaw(get_rotation_xyzw(obj), yaw)
+        base_rot = get_rotation_xyzw(obj)
+        marker_yaw = yaw_from_quat_xyzw(base_rot)
+        total_yaw = layout.orientations.get(obj, marker_yaw)
+        rot = rotate_quat_by_yaw(base_rot, total_yaw - marker_yaw)
         object_cfg = getattr(obj, "object_cfg", None)
         assert object_cfg is not None, f"Object '{obj.name}' must have object_cfg initialized before placement."
         object_cfg.init_state.pos = pos
@@ -143,8 +145,9 @@ def _apply_static_initial_poses(
             if pos is None:
                 missing_envs.append(env_idx)
             else:
-                yaw = layouts[env_idx].orientations.get(obj, 0.0)
-                rotation_xyzw = rotate_quat_by_yaw(base_rotation_xyzw, yaw)
+                marker_yaw = yaw_from_quat_xyzw(base_rotation_xyzw)
+                total_yaw = layouts[env_idx].orientations.get(obj, marker_yaw)
+                rotation_xyzw = rotate_quat_by_yaw(base_rotation_xyzw, total_yaw - marker_yaw)
                 poses.append(Pose(position_xyz=pos, rotation_xyzw=rotation_xyzw))
         if missing_envs:
             print(
--- a/isaaclab_arena/relations/mesh_pair_cache.py
+++ b/isaaclab_arena/relations/mesh_pair_cache.py
@@ -1,0 +1,88 @@
+# Copyright (c) 2025-2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Typed container for precomputed mesh-collision pair data."""
+
+from __future__ import annotations
+
+import torch
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import warp as wp
+
+    from isaaclab_arena.assets.object_base import ObjectBase
+
+
+@dataclass(slots=True)
+class MeshPairCache:
+    """Precomputed per-pair collision data for the vectorized multi-mesh kernel."""
+
+    all_centers_local: torch.Tensor
+    """(S, 3) sphere centers in each child's local frame, concatenated across pairs."""
+
+    all_radii: torch.Tensor
+    """(S,) sphere radii, concatenated across pairs."""
+
+    pair_child_objs: list[ObjectBase]
+    """Per-pair child object reference."""
+
+    pair_parent_objs: list[ObjectBase]
+    """Per-pair parent/target object reference."""
+
+    pair_is_anchor: list[bool]
+    """Per-pair flag: True if parent is a static anchor."""
+
+    pair_anchor_pos: list[torch.Tensor | None]
+    """Per-pair world position of anchors (None for non-anchor parents)."""
+
+    pair_anchor_yaw: list[float]
+    """Per-pair anchor yaw in radians (0.0 for non-anchor parents)."""
+
+    pair_c_bbox_min: torch.Tensor
+    """(P, B, 3) child bbox min corners for broadphase."""
+
+    pair_c_bbox_max: torch.Tensor
+    """(P, B, 3) child bbox max corners for broadphase."""
+
+    pair_p_bbox_min: torch.Tensor
+    """(P, B, 3) parent bbox min corners for broadphase."""
+
+    pair_p_bbox_max: torch.Tensor
+    """(P, B, 3) parent bbox max corners for broadphase."""
+
+    pair_max_r: torch.Tensor
+    """(P,) max sphere radius per pair (broadphase margin)."""
+
+    sphere_pair_id: torch.Tensor
+    """(S,) maps each sphere to its pair index for segment reduction."""
+
+    sphere_mesh_idx: torch.Tensor
+    """(S,) per-sphere index into mesh_id_array."""
+
+    pair_sphere_count: torch.Tensor
+    """(P,) number of spheres per pair (for mean reduction)."""
+
+    mesh_id_array: wp.array
+    """Warp uint64 array of mesh IDs for the multi-mesh kernel."""
+
+    num_pairs: int
+    """Total number of active object pairs."""
+
+    total_spheres: int
+    """Total number of sphere queries across all pairs."""
+
+    def __post_init__(self) -> None:
+        assert len(self.pair_child_objs) == self.num_pairs, "pair_child_objs length mismatch"
+        assert len(self.pair_parent_objs) == self.num_pairs, "pair_parent_objs length mismatch"
+        assert len(self.pair_is_anchor) == self.num_pairs, "pair_is_anchor length mismatch"
+        assert self.all_centers_local.shape[0] == self.total_spheres, "all_centers_local size mismatch"
+        assert self.all_radii.shape[0] == self.total_spheres, "all_radii size mismatch"
+        assert self.sphere_pair_id.shape[0] == self.total_spheres, "sphere_pair_id size mismatch"
+        assert self.sphere_mesh_idx.shape[0] == self.total_spheres, "sphere_mesh_idx size mismatch"
+        assert int(self.pair_sphere_count.sum().item()) == self.total_spheres, "pair_sphere_count sum mismatch"
+        for i, (is_anchor, pos) in enumerate(zip(self.pair_is_anchor, self.pair_anchor_pos)):
+            assert not is_anchor or pos is not None, f"pair {i}: is_anchor=True but anchor_pos is None"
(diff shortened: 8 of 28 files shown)


isaaclab_arena/assets/object.py
```diff
@@ -74,6 +84,31 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            if k == "_collision_mesh":
+                setattr(result, k, None)
+            else:
+                setattr(result, k, copy.deepcopy(v, memo))
+        return result
```
alexmillane: What do you think about moving the mesh stuff out of `Object` and into the solver?

I guess we have it here because we want caching behaviour. We could instead have something like `CollisionShapeManager` in the solver which manages the caching.

We should also probably move the bounding boxes there too.

isaaclab_arena/cli/isaaclab_arena_cli.py
```diff
@@ -92,6 +92,13 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
         default=False,
         help="Print Hydra-configurable variations for the selected environment and exit.",
     )
+    arena_group.add_argument(
+        "--collision_mode",
+        type=str,
+        choices=["bbox", "mesh"],
+        default="bbox",
+        help="Collision detection mode: 'bbox' (AABB, default) or 'mesh' (sphere-to-SDF, requires Warp).",
+    )
```
alexmillane: I feel like the mesh vs. bounding-box decision should be made on the environment level. For example, an environment with a shelf we'd like to place something on qill require mesh-mode. Here we're leaving it up to the user a run time.

That seems like an odd choice to me.

Can you see a way we could configure this in each environment?

alexmillane: To me the obvious way would be to make the `ObjectPlacerParams`/`RelationSolverParams` an (optional) part of `IsaacLabArenaEnvironment`. If specified in the environment they're passed along to the solver.

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -26,6 +26,7 @@ def solve_and_apply_relation_placement(
     placement_seed: int | None = None,
     resolve_on_reset: bool | None = None,
     random_yaw_init: bool = False,
+    collision_mode: str = "bbox",
```
alexmillane: Related to the comment above.

We're seeing a reflection of the members of `ObjectPlacerParams` on the interface to this function.

I think the correct way to get out of this issue (that we have to copy members of `ObjectPlacerParams` onto the interface), is to make this function accept a full `ObjectPlacerParams` instance, which is specified per-environment in `IsaacLabArenaEnvironment`.

We should probably make this change in an MR that preceeds this one.

isaaclab_arena/relations/object_placer.py
```diff
@@ -68,6 +70,19 @@ def __init__(self, params: ObjectPlacerParams | None = None):
...
+        import copy
+
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            if k in ("_cpu_mesh_manager", "_warned_no_mesh"):
+                continue
+            setattr(result, k, copy.deepcopy(v, memo))
+        return result
```
alexmillane: This is a bit ugly and we'd like to avoid it. Or at worst, we could isolate it.

What the motivation here?

Could we solve this in the `WarpMeshManager`. Potentially by overriding its deep-copy operation?

Aside: why do we need special treatment of `_warned_no_mesh`? That appears just to be a `set[str]`?

isaaclab_arena/relations/object_placer.py
```diff
@@ -598,35 +616,211 @@ def _validate_no_overlap(
...
+    def _get_cpu_mesh_manager(self):
+        """Lazily create a CPU WarpMeshManager, cached across validation calls."""
+        if not hasattr(self, "_cpu_mesh_manager"):
+            from isaaclab_arena.relations.warp_mesh_manager import WarpMeshManager
+
+            self._cpu_mesh_manager = WarpMeshManager(
+                num_spheres=self.params.solver_params.num_spheres,
+                device="cpu",
+            )
+        return self._cpu_mesh_manager
```
alexmillane: Out of interest, why do we do this lazily?

It seems to me that there's no cost to creating one of these guys, as they just create empty dicts. What's the motivation here?

isaaclab_arena/relations/object_placer.py
```diff
@@ -598,35 +616,211 @@ def _validate_no_overlap(
...
+        Skips On pairs, anchor-anchor, and mesh-less objects.
+        """
+        from isaaclab_arena.relations.warp_sdf_kernels import (
+            _check_sdf_sentinel,
+            has_sdf_sentinel,
+            mesh_sdf,
+            reset_sdf_sentinel_warning,
+        )
+
+        reset_sdf_sentinel_warning()
```
alexmillane: This looks like it touches some global / has some global side-effect.

What's the motivation there? Can we store whatever global variable this is touching as a member on some class?

isaaclab_arena/relations/object_placer.py
```diff
@@ -598,35 +616,211 @@ def _validate_no_overlap(
...
+
+        on_pairs: set[tuple] = set()
+        anchor_ids: set[int] = set()
+        for obj in positions:
+            for rel in obj.get_relations():
+                if isinstance(rel, On) and rel.parent in positions:
+                    on_pairs.add((id(obj), id(rel.parent)))
+                    on_pairs.add((id(rel.parent), id(obj)))
+            if any(isinstance(r, IsAnchor) for r in obj.get_relations()):
+                anchor_ids.add(id(obj))
```
alexmillane: This is shared with `_validate_no_overlap`

Suggestion to move to a function.

isaaclab_arena/relations/object_placer.py
```diff
@@ -598,35 +616,211 @@ def _validate_no_overlap(
...
+
+        clearance_m = self.params.solver_params.clearance_m
+        tolerance = max(0.0, clearance_m - 1e-6)
+        manager = self._get_cpu_mesh_manager()
+
+        if not hasattr(self, "_warned_no_mesh"):
+            self._warned_no_mesh = set()
+        warned_no_mesh = self._warned_no_mesh
```
alexmillane: This `hasattr(self, "_warned_no_mesh"):` is ugly. How often is this function called? Can we just initialize this set on each call to this function, remove the member variable, and warn a bit more frequently?

isaaclab_arena/relations/object_placer.py
```diff
@@ -598,35 +616,211 @@ def _validate_no_overlap(
...
+                sdf_b = mesh_sdf(centers_b_in_a, warp_a)
+                _check_sdf_sentinel(sdf_b)
+                if has_sdf_sentinel(sdf_b):
+                    return False
+                if (sdf_b < spheres_b[:, 3] + tolerance).any():
+                    if self.params.verbose:
+                        print(f"  Mesh overlap between '{b.name}' and '{a.name}'")
+                    return False
+
+        return True
```
alexmillane, line 794: This function is very long and deeply nested.

Long functions are considered "code-smells" and should be avoided. There are many references on this but, for example, number 7 on this list: https://opsera.ai/blog/what-is-code-smell/

The general approach to breaking this is to break a function into smaller functions.

isaaclab_arena/relations/object_placer.py
```diff
@@ -598,35 +616,211 @@ def _validate_no_overlap(
...
+                        a_bbox = a_bbox.rotated_around_z(a_yaw)
+                    if b_yaw != 0.0:
+                        b_bbox = b_bbox.rotated_around_z(b_yaw)
+                    a_world = a_bbox.translated(a_pos)
+                    b_world = b_bbox.translated(b_pos)
+                    if a_world.overlaps(b_world, margin=tolerance).item():
+                        if self.params.verbose:
+                            print(f"  AABB overlap between '{a.name}' and '{b.name}' (mesh unavailable)")
+                        return False
+                    continue
```
alexmillane: Move out to another function. Ideally this function would be shared with `_validate_no_overlap`

isaaclab_arena/relations/object_placer.py
```diff
@@ -598,35 +616,211 @@ def _validate_no_overlap(
...
+            for rel in obj.get_relations():
+                if isinstance(rel, On) and rel.parent in positions:
+                    on_pairs.add((id(obj), id(rel.parent)))
+                    on_pairs.add((id(rel.parent), id(obj)))
+            if any(isinstance(r, IsAnchor) for r in obj.get_relations()):
+                anchor_ids.add(id(obj))
+
+        clearance_m = self.params.solver_params.clearance_m
+        tolerance = max(0.0, clearance_m - 1e-6)
+        manager = self._get_cpu_mesh_manager()
```
alexmillane: `manager` -> `mesh_manager`

isaaclab_arena/relations/object_placer.py
```diff
@@ -598,35 +616,211 @@ def _validate_no_overlap(
...
+                warp_a = manager.get_warp_mesh(a_mesh, obj=a)
+                centers_b_in_a = self._centers_in_target_frame(spheres_b[:, :3], b, a, b_pos, a_pos, orientations)
+                sdf_b = mesh_sdf(centers_b_in_a, warp_a)
+                _check_sdf_sentinel(sdf_b)
+                if has_sdf_sentinel(sdf_b):
+                    return False
+                if (sdf_b < spheres_b[:, 3] + tolerance).any():
+                    if self.params.verbose:
+                        print(f"  Mesh overlap between '{b.name}' and '{a.name}'")
+                    return False
```
alexmillane: This section has two repeated blocks of code. First we do a check on a, then a check on b (I think). Could we make a function to get rid of this duplication?

isaaclab_arena/relations/object_placer.py
```diff
@@ -598,35 +616,211 @@ def _validate_no_overlap(
...
+        """Resolve effective Z-yaw: orientations dict, then initial_pose for anchors only.
+
+        Mirrors the solver's rule: non-anchors absent from the dict have yaw=0;
+        anchors fall back to initial_pose (which the placer never overwrites).
+        """
+        if orientations is not None and obj in orientations:
+            return orientations[obj]
+        # Only anchors carry a meaningful initial_pose yaw at validation time.
+        if not any(isinstance(r, IsAnchor) for r in obj.get_relations()):
+            return 0.0
```
alexmillane: We seem to repeat this check to determine if an object is an anchor all over the place. Perhaps we could add a method to object: `object.is_anchor` to save repeatedly writing this check/loop.

isaaclab_arena/relations/object_placer.py
```diff
@@ -649,7 +845,10 @@ def _apply_poses(
                 poses = [
                     Pose(
                         position_xyz=positions_per_env[env_idx][obj],
-                        rotation_xyzw=rotate_quat_by_yaw(base_rotation, orientations_per_env[env_idx].get(obj, 0.0)),
+                        rotation_xyzw=rotate_quat_by_yaw(
+                            base_rotation,
+                            orientations_per_env[env_idx].get(obj, marker_yaw) - marker_yaw,
```
alexmillane: What's this `orientations_per_env[env_idx].get(obj, marker_yaw) - marker_yaw`

Suggestion to separate out this expression. An because it's kind of weird looking, I would add a comment.

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -412,49 +420,96 @@ def compute_loss(
...
+                    child_obj,
+                    child_mesh,
+                    parent_pos,
+                    parent_obj,
+                    parent_mesh,
+                    child_yaw=child_yaw,
+                    parent_yaw=parent_yaw,
+                )
+            for obj, mesh in [(child_obj, child_mesh), (parent_obj, parent_mesh)]:
+                name = getattr(obj, "name", "?")
```
alexmillane: why is the `getattr` needed here? I think all `Assets` are constrained to have a name? or is there something I'm not seeing.

In general, try to watch out for the agents' overuse of `getattr`. It makes the code more difficult to read. The agents are overly defensive here.

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -412,49 +420,96 @@ def compute_loss(
...
+                if mesh is None and name not in self._warned_no_mesh:
+                    self._warned_no_mesh.add(name)
+                    print(f"  [NoCollision] MESH mode: '{name}' has no collision mesh, falling back to AABB")
+
+        return self._compute_aabb_loss(clearance_m, child_pos, child_bbox, parent_world_bbox)
```
alexmillane: When does this fallback happen? 

Are there legitimate cases where a user asks for a mesh-based loss, and we can't deliver and it makes sense to fall back to an AABB loss?

Potentially this is happening because we're (in this MR) requesting meshes for ALL objects, rather than a few selected objects.

In general, if this is an abnormal situation, where we request meshes and we dont get them. I would prefer to check-fail (assert).

However, maybe this is a legitimate reason here?

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -496,6 +548,88 @@ def compute_loss(
...
+    ) -> torch.Tensor:
+        """Per-pair sphere-to-SDF penetration loss."""
+        from isaaclab_arena.relations.warp_sdf_kernels import sphere_penetration_loss
+
+        single_input = child_pos.dim() == 1
+        if single_input:
+            child_pos = child_pos.unsqueeze(0)
+
+        if parent_pos is None:
+            from isaaclab_arena.utils.pose import Pose
```
alexmillane: move the import to the top of the file.

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -496,6 +548,88 @@ def compute_loss(
...
+        child_obj: ObjectBase,
+        child_mesh: trimesh.Trimesh,
+        parent_pos: torch.Tensor | None,
+        parent_obj: ObjectBase,
+        parent_mesh: trimesh.Trimesh,
+        child_yaw: float = 0.0,
+        parent_yaw: float = 0.0,
+    ) -> torch.Tensor:
+        """Per-pair sphere-to-SDF penetration loss."""
+        from isaaclab_arena.relations.warp_sdf_kernels import sphere_penetration_loss
```
alexmillane: import to top

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -496,6 +548,88 @@ def compute_loss(
...
+            from isaaclab_arena.utils.pose import Pose
+
+            pose = parent_obj.get_initial_pose()
+            assert pose is not None, f"parent_pos=None but '{getattr(parent_obj, 'name', '?')}' has no initial pose"
+            assert isinstance(
+                pose, Pose
+            ), f"Anchor '{getattr(parent_obj, 'name', '?')}' must have a fixed Pose for mesh collision"
+            parent_pos = torch.tensor(pose.position_xyz, dtype=child_pos.dtype, device=child_pos.device)
+            if parent_yaw == 0.0:
+                from isaaclab_arena.utils.pose import yaw_from_quat_xyzw
```
alexmillane: import to top

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -496,6 +548,88 @@ def compute_loss(
...
+
+                parent_yaw = yaw_from_quat_xyzw(pose.rotation_xyzw)
+
+        assert parent_pos is not None
+        parent_pos_resolved: torch.Tensor = parent_pos
+        if parent_pos_resolved.dim() == 1:
+            parent_pos_resolved = parent_pos_resolved.unsqueeze(0)
+
```
alexmillane: What does resolved mean here?

Is it only that we've expanded along the batch dimension?

Suggestion to create a new local variable with the name `parent_pos`

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -496,6 +548,88 @@ def compute_loss(
...
+        for b in range(batch_size):
+            offset = child_pos[b] - parent_pos_resolved[b]
+            # Rotate offset into the parent's local frame.
+            if parent_yaw != 0.0:
+                cos_p = math.cos(-parent_yaw)
+                sin_p = math.sin(-parent_yaw)
+                ox = offset[0] * cos_p - offset[1] * sin_p
+                oy = offset[0] * sin_p + offset[1] * cos_p
+                offset = torch.stack([ox, oy, offset[2]])
+            centers_in_parent = centers_local + offset
```
alexmillane: Two suggestions:
- Part of this code represents transforming vectors in one frame to another. Under a transformation (that happens to only have a yaw component). I'm wondering if we can use more general tools for this. For example, can we use: `isaaclab.utils.math` for example `quat_apply` for applying rotations, rather than recoding the code for rotations here.
- It is useful if you use structured notation. For example `q_B_A` represents the rotation from the frame `A` to the frame `B`.

I just wrote a good example of doing this type of this [here](https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena/variations/camera_extrinsics_variation.py#L131)

isaaclab_arena/relations/warp_mesh_manager.py
```diff
@@ -0,0 +1,191 @@
...
+        num_spheres: Maximum number of output spheres.
+        sphere_radius: Inflation added to tangent sphere radii (safety margin).
+        n_candidates: Number of candidate sphere centers sampled.
+        n_surface: Number of surface points for coverage tracking.
+        seed: RNG seed for reproducible surface sampling.
+
+    Returns:
+        (K, 4) array of [cx, cy, cz, radius] in mesh-local frame. K <= num_spheres.
+    """
+    import trimesh as _trimesh
```
alexmillane: why the underscore here?

isaaclab_arena/relations/warp_mesh_manager.py
```diff
@@ -0,0 +1,191 @@
...
+
+    n_candidates = max(num_spheres, n_candidates)
+    n_surface = max(n_candidates, n_surface)
+
+    rng = np.random.default_rng(seed)
```
alexmillane, line 54: This might destroy the seeding we do elsewhere.

I'm wondering if we should avoid setting the seed anywhere in our codebase, except for the one place we set it in Isaac Lab.

@xyao-nv @peterd-NV What do you think? Should we outlaw setting seeds in our code?

isaaclab_arena/relations/warp_mesh_manager.py
```diff
@@ -0,0 +1,191 @@
...
+    ):
+        self._num_spheres = num_spheres
+        self._sphere_radius = sphere_radius
+        self._device = device
+        self._warp_mesh_cache: dict[tuple, wp.Mesh] = {}
+        self._sphere_cache: dict[tuple, torch.Tensor] = {}
+
+    def __deepcopy__(self, memo):
+        """Deep copy, dropping lazy Warp caches (unpicklable C pointers); rebuilt on demand."""
+        import copy
```
alexmillane: import to top

alexmillane (commented):
First partial review.

Looks good. I haven't got to the warp mesh based stuff.

isaaclab_arena/relations/object_placer.py
zhx06: This is to avoid loading warp when we only use bbox

isaaclab_arena/relations/warp_mesh_manager.py
zhx06: Addressed.

isaaclab_arena/relations/warp_mesh_manager.py
zhx06: Fixed

isaaclab_arena/relations/relation_loss_strategies.py
zhx06: Addressed. It is in `parent_pos`

isaaclab_arena/relations/relation_loss_strategies.py
zhx06: Addressed.

isaaclab_arena/relations/relation_loss_strategies.py
zhx06: Addressed

isaaclab_arena/relations/object_placer.py
zhx06: Addressed. Extracted to `_sampled_yaw_delta(env_idx)`

isaaclab_arena/relations/object_placer.py
zhx06: Addressed.

isaaclab_arena/relations/object_placer.py
zhx06: Addressed. It is now initialized in `__init__`, no more `hasattr`

isaaclab_arena/relations/object_placer.py
zhx06: Addressed. It is now `_collect_skip_pairs()`

isaaclab_arena/cli/isaaclab_arena_cli.py
```diff
@@ -92,6 +92,13 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
         default=False,
         help="Print Hydra-configurable variations for the selected environment and exit.",
     )
+    arena_group.add_argument(
+        "--collision_mode",
```
alexmillane: Same point as my earlier comment: the mesh-vs-bbox decision is being made at runtime via a CLI flag. To me this belongs at the environment level — an environment with, say, a shelf we place onto inherently requires mesh-mode, it's not a per-run user choice.

I sugest we make `ObjectPlacerParams`/`RelationSolverParams` an optional member of `IsaacLabArenaEnvironment`, and pass it through to the solver from there.

_Edit: We spoke on the phone about maybe doing this in a follow up MR. Is there any reason no to do it in this one?_

isaaclab_arena/relations/relation_solver.py
```diff
@@ -153,15 +177,30 @@ def _compute_no_overlap_loss(
         Returns:
             Per-environment loss tensor of shape (batch_size,).
         """
+        if self.params.collision_mode == CollisionMode.MESH:
+            mesh_loss = self._compute_no_overlap_loss_mesh(state, debug)
+            aabb_loss = self._compute_no_overlap_loss_aabb(state, debug, skip_mesh_pairs=True)
```
alexmillane: I guess there are really instances when we can't get a mesh?

isaaclab_arena/relations/warp_sdf_kernels.py
```diff
@@ -0,0 +1,278 @@
...
+    Returns:
+        (N,) signed distance values. Negative = penetrating.
+    """
+    return _MeshSDFFunction.apply(points, warp_mesh)
+
+
+# Warp returns ~1e6 when a BVH query finds no enclosing face; 1e5 catches these
+# while staying safely below any realistic SDF magnitude.
+_SDF_SENTINEL = 1.0e5
+_sentinel_warned_this_solve = False
```
alexmillane: Could this be moved into a class?

isaaclab_arena/assets/object.py
```diff
@@ -74,6 +84,31 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
             self.bounding_box = compute_local_bounding_box_from_usd(self.usd_path, self.scale)
         return self.bounding_box
 
+    def get_collision_mesh(self) -> trimesh.Trimesh | None:
+        """Lazily extract collision mesh from USD. Cached after first call."""
+        if not hasattr(self, "_collision_mesh"):
```
alexmillane: suggestion to initialize `_collision_mesh` in `__init__` (to `None`) and drop the `hasattr` check? It reads much more clearly.

isaaclab_arena/relations/relation_solver.py
```diff
@@ -223,11 +279,375 @@ def _compute_no_overlap_loss(
...
+        self,
+        state: RelationSolverState,
+        on_pairs: set[tuple[int, int]],
+    ) -> None:
+        """Precompute static per-pair mesh collision data (called once per solve)."""
+        from isaaclab_arena.relations.warp_mesh_manager import WarpMeshManager
+
+        device = state.device
+        device_str = str(device)
+        if not hasattr(self, "_mesh_manager") or self._mesh_manager.device != device_str:
```
alexmillane: Suggest initializing `_mesh_manager = None` in the constructor and checking `is None` here.

isaaclab_arena/relations/warp_mesh_manager.py
```diff
@@ -0,0 +1,190 @@
...
+        return result
+
+    @property
+    def device(self) -> str:
+        """Target Warp device string (e.g. 'cuda:0', 'cpu')."""
+        return self._device
+
+    def _cache_key(self, mesh: trimesh.Trimesh, obj=None) -> tuple:
+        """Compute cache key. Uses (usd_path, scale) for USD objects, content hash otherwise."""
+        usd_path = getattr(obj, "usd_path", None) if obj is not None else None
```
alexmillane, line 175: defensive `getattr(obj, ...)`

isaaclab_arena/assets/object.py
```diff
@@ -74,6 +84,31 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+            if self.usd_path is not None:
+                try:
+                    self._collision_mesh = extract_trimesh_from_usd(self.usd_path, self.scale)
+                except (ValueError, OSError, AttributeError, TypeError, IndexError) as e:
+                    print(f"  [MeshManager] Could not extract mesh for '{self.name}': {type(e).__name__}: {e}")
+        return self._collision_mesh
+
+    def __deepcopy__(self, memo):
+        """Exclude _collision_mesh from deepcopy (trimesh has unpicklable C pointers)."""
+        import copy
```
alexmillane: Local stdlib import. `import copy` (and `import math`/`import numpy as np` elsewhere in this PR) should move to module top — there's no init-order reason to defer stdlib/numpy. The `warp`/`trimesh` deferred imports may be intentional to avoid importing before Sim start; if so, fine, but the stdlib ones aren't.

isaaclab_arena/assets/object.py
alexmillane: Although as I have commented elsewhere, I think the right thing to do is separate the objects from their bounding-box/collision mesh computation.

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -38,20 +39,25 @@ def solve_and_apply_relation_placement(
             initial poses are applied immediately.
         random_yaw_init: If True, randomly rotates non-anchor objects around the vertical (Z)
             axis at startup to add visual variety to the scene.
+        collision_mode: Collision detection mode: "bbox" or "mesh".
 
     Returns:
         Reset event config to attach to the environment when placement should be
         resolved on reset. Returns ``None`` when no reset event is needed.
     """
+    from isaaclab_arena.relations.relation_solver_params import CollisionMode
```
alexmillane: local import required?

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -38,20 +39,25 @@ def solve_and_apply_relation_placement(
...
+
     objects = list(objects)
     if not objects:
         print("No objects with relations found in scene. Skipping relation solving.")
         return None
 
+    assert collision_mode in ("bbox", "mesh"), f"Invalid collision_mode '{collision_mode}', expected 'bbox' or 'mesh'"
+    mode = CollisionMode.MESH if collision_mode == "mesh" else CollisionMode.BBOX
```
alexmillane: I guess we're doing this here because we get a string on the command line? I think with argparse you can do the conversion during parsing?

```
class Color(Enum):
    RED = "red"
    GREEN = "green"
    BLUE = "blue"

    @classmethod
    def from_string(cls, s: str) -> "Color":
        try:
            return cls(s.lower())
        except ValueError:
            raise ValueError(
                f"Invalid color '{s}'. Expected one of: "
                f"{', '.join(c.value for c in cls)}"
            )
```

Then:

```
parser.add_argument("--color", type=Color.from_string)
```

But maybe this is not required if we don't break this parameter out to the CLI (see other comments)>

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -412,49 +421,95 @@ def compute_loss(
...
+        self._warned_no_mesh: set[str] = set()
+        self._mesh_managers: dict[str, WarpMeshManager] = {}
+
+        if self._mode == CollisionMode.MESH:
+            try:
+                import warp  # noqa: F401
+            except ImportError as e:
+                raise ImportError(
+                    "CollisionMode.MESH requires the 'warp' package. Install it with: pip install warp-lang"
+                ) from e
```
alexmillane: What's the idea here?

I think `warp` is a dependency of Isaac Lab. So I think we're safe to assume it's available and just import it without the safeties.

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -496,6 +548,81 @@ def compute_loss(
 
         return total_loss.squeeze(0) if single_input else total_loss
 
+    def _get_mesh_manager(self, device: str = "cuda:0") -> WarpMeshManager:
+        """Return a cached WarpMeshManager for the given device."""
+        if device not in self._mesh_managers:
+            from isaaclab_arena.relations.warp_mesh_manager import WarpMeshManager
```
alexmillane: import to top?

isaaclab_arena/relations/relation_solver.py
```diff
@@ -42,10 +43,30 @@ def __init__(
...
+        )
         self._last_loss_history: list[float] = []
         self._last_position_history: list = []
         self._last_loss_per_env: torch.Tensor | None = None
+        self._mesh_orientations: list[dict[ObjectBase, float]] | None = None
+        self._warned_no_mesh: set[str] = set()
+
+    def __deepcopy__(self, memo):
+        """Deep copy, dropping lazy Warp caches (unpicklable C pointers); rebuilt on next solve."""
+        import copy
```
alexmillane: import to top

isaaclab_arena/relations/relation_solver.py
```diff
@@ -42,10 +43,30 @@ def __init__(
...
+
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        skip = {"_mesh_manager", "_mesh_cache_fwd", "_mesh_cache_rev", "_mesh_orientations"}
+        for k, v in self.__dict__.items():
+            if k in skip:
+                continue
+            setattr(result, k, copy.deepcopy(v, memo))
+        return result
```
alexmillane: Implementing these copy functions for numerous classes is a bit ugly.

Could you help me understand why this is necessary?

isaaclab_arena/relations/relation_solver.py
```diff
@@ -223,11 +279,375 @@ def _compute_no_overlap_loss(
 
         return total_loss
 
+    def _prepare_mesh_collision_cache(
+        self,
+        state: RelationSolverState,
+        on_pairs: set[tuple[int, int]],
+    ) -> None:
+        """Precompute static per-pair mesh collision data (called once per solve)."""
+        from isaaclab_arena.relations.warp_mesh_manager import WarpMeshManager
```
alexmillane: import to top

isaaclab_arena/relations/relation_solver.py
```diff
@@ -223,11 +279,375 @@ def _compute_no_overlap_loss(
...
+    ) -> dict | None:
+        """Build vectorized pair cache for one direction.
+
+        Returns None if no valid pairs exist for this direction.
+        """
+        import numpy as np
+
+        import warp as wp
+
+        from isaaclab_arena.utils.pose import Pose, yaw_from_quat_xyzw
```
alexmillane: imports to top

isaaclab_arena/relations/relation_solver.py
```diff
@@ -223,11 +279,375 @@ def _compute_no_overlap_loss(
...
+
+    def _compute_no_overlap_loss_mesh(
+        self,
+        state: RelationSolverState,
+        debug: bool,
+    ) -> torch.Tensor:
+        """Differentiable mesh no-overlap loss from precomputed pair cache."""
+        import warp as wp
+
+        from isaaclab_arena.relations.warp_sdf_kernels import _check_sdf_sentinel, multi_mesh_sdf
```
alexmillane: imports to top

isaaclab_arena/relations/relation_solver.py
```diff
@@ -223,11 +279,375 @@ def _compute_no_overlap_loss(
...
+            "pair_p_bbox_min": torch.stack(pair_p_bbox_min),
+            "pair_p_bbox_max": torch.stack(pair_p_bbox_max),
+            "pair_max_r": torch.tensor(pair_max_r, device=device),
+            "sphere_pair_id": sphere_pair_id,
+            "sphere_mesh_idx": torch.tensor(mesh_idx_per_sphere, dtype=torch.int32, device=device),
+            "pair_sphere_count": pair_sphere_count,
+            "mesh_id_array": wp.array(np.array(mesh_id_values, dtype=np.uint64), dtype=wp.uint64, device=wp_device),
+            "num_pairs": len(pair_slices),
+            "total_spheres": offset,
+        }
```
alexmillane: Can we make a dataclass out of this?

isaaclab_arena/relations/relation_solver.py
```diff
@@ -223,11 +279,375 @@ def _compute_no_overlap_loss(
...
+            return None
+
+        wp_device = str(device)
+        # Maps each sphere to its pair — enables vectorized segment reduction.
+        pair_sphere_count = torch.tensor([e - s for s, e in pair_slices], dtype=torch.float32, device=device)
+        sphere_pair_id = torch.repeat_interleave(
+            torch.arange(len(pair_slices), device=device), pair_sphere_count.long()
+        )
+
+        return {
```
alexmillane: This is a super long function that creates a huge dictionary.

Let's remove this from the file in two ways:
- replace the dictionary with an object (maybe a dataclass)?
- move it's computation to another module that just returns the cache object here.

isaaclab_arena/relations/relation_solver.py
```diff
@@ -223,11 +279,375 @@ def _compute_no_overlap_loss(
...
+                    local_centers = torch.stack([rx, ry, local_centers[:, 2]], dim=-1)
+
+                    pair_offsets = offsets[active_sphere_pair_id]
+                    p_yaws = parent_yaws[active_sphere_pair_id]
+                    cos_p = torch.cos(-p_yaws)
+                    sin_p = torch.sin(-p_yaws)
+                    ox = pair_offsets[:, 0] * cos_p - pair_offsets[:, 1] * sin_p
+                    oy = pair_offsets[:, 0] * sin_p + pair_offsets[:, 1] * cos_p
+                    rotated_offsets = torch.stack([ox, oy, pair_offsets[:, 2]], dim=-1)
+                    active_centers = local_centers + rotated_offsets
```
alexmillane: Would be great if we can use established functions for doing transforms, rather than recoding them every time

isaaclab_arena/relations/relation_solver.py
```diff
@@ -223,11 +279,375 @@ def _compute_no_overlap_loss(
...
+            "pair_max_r": torch.tensor(pair_max_r, device=device),
+            "sphere_pair_id": sphere_pair_id,
+            "sphere_mesh_idx": torch.tensor(mesh_idx_per_sphere, dtype=torch.int32, device=device),
+            "pair_sphere_count": pair_sphere_count,
+            "mesh_id_array": wp.array(np.array(mesh_id_values, dtype=np.uint64), dtype=wp.uint64, device=wp_device),
+            "num_pairs": len(pair_slices),
+            "total_spheres": offset,
+        }
+
+    def _compute_no_overlap_loss_mesh(
```
alexmillane: Suggestion to add a long comment at the start of this function to describe what it's doing. It appears quite non-trivial.

isaaclab_arena/relations/relation_solver.py
```diff
@@ -223,11 +279,375 @@ def _compute_no_overlap_loss(
...
+                    rotated_offsets = torch.stack([ox, oy, pair_offsets[:, 2]], dim=-1)
+                    active_centers = local_centers + rotated_offsets
+                else:
+                    active_centers = local_centers + offsets[active_sphere_pair_id]
+                active_radii = cache["all_radii"][active_idx]
+                active_mesh_idx = cache["sphere_mesh_idx"][active_idx].contiguous()
+
+                active_mesh_indices_wp = wp.from_torch(active_mesh_idx, dtype=wp.int32)
+                sdf_values = multi_mesh_sdf(active_centers, cache["mesh_id_array"], active_mesh_indices_wp)
+                _check_sdf_sentinel(sdf_values)
```
alexmillane: This function looks suspicious because it has no output but is called "check". Usually a check returns result.

Is this a check-fail (i.e. an assert)?

isaaclab_arena/relations/warp_mesh_manager.py
```diff
@@ -0,0 +1,190 @@
...
+        """Deep copy, dropping lazy Warp caches (unpicklable C pointers); rebuilt on demand."""
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            if k in ("_warp_mesh_cache", "_sphere_cache"):
+                setattr(result, k, {})
+            else:
+                setattr(result, k, copy.deepcopy(v, memo))
+        return result
```
alexmillane: suspicious deep copy.

isaaclab_arena/relations/warp_mesh_manager.py
```diff
@@ -0,0 +1,190 @@
...
+
+    if not selected:
+        print("  [SphereDecomp] Set-cover selected 0 spheres — using uniform fallback")
+        pts = points[:num_spheres]
+        return np.column_stack([pts, np.full(len(pts), sphere_radius)])
+
+    return np.column_stack([centers[selected], radii[selected]])
+
+
+class WarpMeshManager:
```
alexmillane: `WarpMeshAndSphereCache`

alexmillane (commented):
Another partial review

isaaclab_arena/relations/warp_mesh_manager.py
zhx06: Done. It is renamed as `WarpMeshAndSphereCache`

isaaclab_arena/relations/relation_solver.py
zhx06: It is now removed.

isaaclab_arena/relations/relation_solver.py
zhx06: Added a two-line docstring describing the batched sphere-to-SDF approach.

isaaclab_arena/relations/relation_loss_strategies.py
zhx06: Done.

isaaclab_arena/relations/relation_solver.py
zhx06: Done. Ir now uses `quat_apply(q_net_z, local_centers)` and `quat_apply_inverse(q_parent_z, offsets)` from `isaaclab.utils.math`

isaaclab_arena/relations/relation_solver.py
zhx06: Done.

isaaclab_arena/relations/relation_solver.py
zhx06: Done.

isaaclab_arena/relations/relation_solver.py
zhx06: Done.

isaaclab_arena/relations/relation_solver.py
zhx06: Done.

isaaclab_arena/relations/relation_solver.py
zhx06: Done. This is now `MeshPairCache`.

isaaclab_arena/relations/relation_loss_strategies.py
zhx06: Removed `warp` is imported at module level now.

isaaclab_arena/assets/object.py
zhx06: Done. This is removed.

isaaclab_arena/environments/relation_solver_interface.py
zhx06: Done. `CollisionMode` is removed.

isaaclab_arena/assets/object.py
zhx06: Done. Mesh extraction moved out of `Object` class.

isaaclab_arena/relations/relation_loss_strategies.py
zhx06: Done.

isaaclab_arena/relations/relation_solver.py
zhx06: Done.

isaaclab_arena/cli/isaaclab_arena_cli.py
zhx06: Done in this MR. `IsaacLabArenaEnvironment` now takes an optional `placer_params: ObjectPlacerParams`

isaaclab_arena/relations/relation_solver.py
zhx06: Yes, objects without a USD asset  can't produce a mesh

isaaclab_arena/relations/warp_sdf_kernels.py
zhx06: Done. It is now in `WarpMeshAndSphereCache`

isaaclab_arena/relations/relation_loss_strategies.py
zhx06: The reason is that objects without USD can't produce a mesh, and a fallback keeps those pairs covered.

isaaclab_arena/relations/object_placer.py
zhx06: Done, it now has `ObjectBase.is_anchor`.

isaaclab_arena/relations/object_placer.py
zhx06: Done. We now have a helper for this.

isaaclab_arena/relations/relation_loss_strategies.py
zhx06: Done. We now use  `obj.name` directly

isaaclab_arena/relations/object_placer.py
zhx06: Done. The code has been extracted to `_pair_aabb_overlaps()` and `_spheres_penetrate_mesh()` helper

isaaclab_arena/relations/object_placer.py
zhx06: Done. It is now split into `_pair_aabb_overlaps()`, `_spheres_penetrate_mesh()`, and `_non_skip_pairs()`

isaaclab_arena/relations/object_placer.py
zhx06: Done. It is moved to `WarpMeshAndSphereCache.reset_sentinel_warning()`

isaaclab_arena/relations/object_placer.py
zhx06: This is removed so it is same to copy `ObjectPlacer` now.

isaaclab_arena/assets/object.py
zhx06: Done. Mesh extraction and caching have been moved to `WarpMeshAndSphereCache` so `Object` class does not have meshes.

isaaclab_arena/environments/relation_solver_interface.py
zhx06: It is done in this MR. The function now accepts `placer_params: ObjectPlacerParams` directly.

isaaclab_arena/cli/isaaclab_arena_cli.py
zhx06: Done.  `IsaacLabArenaEnvironment` accepts `placer_params: ObjectPlacerParams`.

isaaclab_arena/relations/relation_solver.py
zhx06: It is only in  `RelationSolver` now. The reason is that `warp.Mesh` handles can't be pickled. Hydra calls deepcopy on env configs, so we need to skip these fields and let them rebuild on the next` solve()`

isaaclab_arena/assets/dummy_object.py
```diff
@@ -2,31 +2,37 @@
...
 import torch
+from typing import TYPE_CHECKING
 
-from isaaclab_arena.relations.relations import Relation, RelationBase, UnaryRelation
+from isaaclab_arena.relations.relations import IsAnchor, Relation, RelationBase, UnaryRelation
 from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox, quaternion_to_90_deg_z_quarters
 from isaaclab_arena.utils.pose import Pose
 
+if TYPE_CHECKING:
+    import trimesh
```
alexmillane: out of interest, why do we guard trimesh?

With Isaac Sim we have to do this because we need Isaac Sim started up to do the import, but here we don't have that kind of restriction do we?

isaaclab_arena/cli/isaaclab_arena_cli.py
alexmillane: Part my comment was about not deciding about mesh vs. bounding box globally. But deciding it on a per-object level.

But I think we can delay this to a future MR.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -82,12 +85,27 @@ def _solve_relations(self) -> None:
           events restore the same layout every time.
         """
         objects_with_relations = self.arena_env.scene.get_objects_with_relations()
+
+        # Prefer env-level placer_params; fall back to CLI-constructed defaults.
+        placer_params = self.arena_env.placer_params
+        if placer_params is None:
+            collision_mode_str = getattr(self.args, "collision_mode", "bbox")
```
alexmillane: I don't think we need the `getattr` here. `args.collision_mode` is always defined. It's defined with the rest of the args.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -82,12 +85,27 @@ def _solve_relations(self) -> None:
...
+                placement_seed=self.args.placement_seed,
+                random_yaw_init=self.args.random_yaw_init,
+                solver_params=RelationSolverParams(
+                    collision_mode=mode,
+                    save_position_history=False,
+                    verbose=False,
+                ),
+            )
+            if self.args.resolve_on_reset is not None:
+                placer_params.resolve_on_reset = self.args.resolve_on_reset
```
alexmillane, line 96: Hmm this seems a bit dangerous. This means users' CLI args will be ignored.

I think we should revert this.

What do you think about NOT exposing `random_yaw_init`, `collision_mode` on the CLI at all? These seem like they should be coded into the environment? Like, how often do we want to run the same environment with a different value for `collision_mode`, decided at run time. I think not very often. This is likely to be decided once, while putting together the environment.

isaaclab_arena/environments/isaaclab_arena_environment.py
```diff
@@ -50,6 +52,8 @@ def __init__(
                 ``"my_module:RLPolicyCfg"``.
             episode_recorder_terms: Additional per-episode recorder terms to record alongside the
                 built-in ones, keyed by name.
+            placer_params: Object placement configuration. When set, used as-is
+                (CLI flags are ignored). When None, params are built from CLI flags.
```
alexmillane: See comment above. We should never ignore the CLI flags. Either:
- remove the flags and always rely on the params member here.
- have the flags and they take precedence.

isaaclab_arena/relations/mesh_pair_cache.py
```diff
@@ -0,0 +1,88 @@
...
+"""Typed container for precomputed mesh-collision pair data."""
+
+from __future__ import annotations
+
+import torch
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import warp as wp
```
alexmillane: why are we guarding warp here?

To hazard a guess, you're guarding these things (warp and trimesh) because they are, right now, not in our container. If that's the case, that's the wrong approach. We should add them to the container and make unguarded imports.

If that's not the reason, let's discuss what the reason is.

isaaclab_arena/relations/mesh_pair_cache.py
```diff
@@ -0,0 +1,88 @@
...
+    """(S, 3) sphere centers in each subject's local frame, concatenated across pairs."""
+
+    all_radii: torch.Tensor
+    """(S,) sphere radii, concatenated across pairs."""
+
+    pair_subject_objs: list[ObjectBase]
+    """Per-pair subject (sphere source) object reference."""
+
+    pair_obstacle_objs: list[ObjectBase]
+    """Per-pair obstacle (mesh target) object reference."""
```
alexmillane: I think some dimensions would help here.

Do we have `N` subjects and `N` obstacles? What is the size of the radii? `N` (one sphere per subject and obstacle (which are the same)?

isaaclab_arena/relations/mesh_pair_cache.py
```diff
@@ -0,0 +1,88 @@
...
+    """Per-pair flag: True if the obstacle is a static anchor."""
+
+    pair_anchor_pos: list[torch.Tensor | None]
+    """Per-pair world position for anchor obstacles (None for non-anchor obstacles)."""
+
+    pair_anchor_yaw: list[float]
+    """Per-pair anchor yaw in radians (0.0 for non-anchor obstacles)."""
+
+    pair_subject_bbox_min: torch.Tensor
+    """(P, B, 3) subject bbox min corners for broadphase."""
```
alexmillane: What's `P` and what is `B`. Maybe `P` is the number of pairs? what's `B`.

Suggestion to add shapes to all members in this object (you've done a good job with _some_ of the members). And then add a definition of the symbols to the docstring on the dataclass.

isaaclab_arena/relations/object_placer.py
```diff
@@ -222,16 +226,19 @@ def _place_ranked(
                 self._generate_initial_orientations(objects, anchor_objects_set, generator)
             )
 
-        # Bake each candidate's yaw into a conservative enclosing bbox; no-op when yaw is disabled.
-        # The solver and validation then treat the rotated object as an axis-aligned box.
+        # Bake each candidate's total yaw into a conservative enclosing bbox (AABB broadphase).
```
alexmillane: what does "broadphase" mean here?

isaaclab_arena/relations/object_placer.py
alexmillane: """A scored solver result ~~used for ranking inside ObjectPlacer~~."""

docstrings should describe what a thing is, not how other classes use it.

Agent's love to do this.

alexmillane (commented):
Partial review

isaaclab_arena/relations/mesh_pair_cache.py
zhx06: Yeah, that was a historical issue. When I first developed the mesh support, I needed to install these libraries myself. We should use unguarded imports as you suggested.

isaaclab_arena/relations/object_placer.py
zhx06: It is the first coarse check for the collision. Changed into overlap check for better explaination.

isaaclab_arena/assets/dummy_object.py
zhx06: TYPE_CHECKING removed. This was a historical issue. Switch to unguarded imports.

isaaclab_arena/environments/arena_env_builder.py
zhx06: both `random_yaw_init` and `collision_mode` are  inside `RelationSolverParams`. Remove them from CLI args.

isaaclab_arena/environments/isaaclab_arena_environment.py
zhx06: Done. Same as above.

isaaclab_arena/relations/mesh_pair_cache.py
zhx06: Docstring updated.

isaaclab_arena/relations/object_placer.py
zhx06: Done. Now docstrings simply describe what it is.

isaaclab_arena/relations/mesh_pair_cache.py
zhx06: Same as above. Docstring updated.

isaaclab_arena/assets/object_base.py
```diff
@@ -170,6 +177,13 @@ def get_relations(self) -> list[RelationBase]:
         """Get all relations for this object."""
         return self.relations
 
+    @property
+    def is_anchor(self) -> bool:
+        """True if this object has an IsAnchor relation."""
+        from isaaclab_arena.relations.relations import IsAnchor
```
alexmillane: local import

isaaclab_arena/relations/object_placer.py
```diff
@@ -590,38 +597,30 @@ def _validate_on_relations(
...
-        The margin is derived from the solver's clearance_m parameter (with a
-        small float tolerance subtracted to avoid rejecting solutions that are
-        within solver residual).
-
-        Args:
-            positions: Solved positions for each object.
-            env_bboxes: Per-object bboxes for the current env, each with shape (1, 3).
-        """
+    ) -> tuple[set[tuple], set[int]]:
+        """Build On-pair skip set and anchor ID set from positioned objects."""
```
alexmillane: Add a docstring about the form of the output.

isaaclab_arena/relations/object_placer.py
```diff
@@ -723,21 +740,176 @@ def _not_next_to_margin(self, relation: NotNextTo) -> float:
...
+        b_bbox = b.get_bounding_box()
+        a_yaw = ObjectPlacer._effective_yaw(a, orientations)
+        b_yaw = ObjectPlacer._effective_yaw(b, orientations)
+        if a_yaw != 0.0:
+            a_bbox = a_bbox.rotated_around_z(a_yaw)
+        if b_yaw != 0.0:
+            b_bbox = b_bbox.rotated_around_z(b_yaw)
+        a_world = a_bbox.translated(a_pos)
+        b_world = b_bbox.translated(b_pos)
+        return a_world.overlaps(b_world, margin=margin).item()
```
alexmillane: We seem to have two methods for doing aabb bounding box checks. The first is `_validate_no_overlap` and the second is `_pair_aabb_overlaps` (this one).

Could we unify these functions?

isaaclab_arena/relations/relation_solver.py
```diff
@@ -37,8 +46,28 @@ class NoOverlapPair:
...
 
+class MeshPairEntry(NamedTuple):
+    """One directed sphere-to-mesh collision pair (subject spheres vs obstacle mesh).
+
+    Dimensions: S = sphere count for this pair's subject, B = batch_size.
+    """
+
+    subject: ObjectBase
+    obstacle: ObjectBase
+    is_anchor: bool
```
alexmillane: Does this mean that the subject is an archor? (or the obstacle)?

isaaclab_arena/relations/relation_solver.py
```diff
@@ -244,11 +291,329 @@ def _compute_no_overlap_loss(
...
+        self,
+        state: RelationSolverState,
+        on_pairs: set[tuple[int, int]],
+    ) -> None:
+        """Precompute static per-pair mesh collision data."""
+        device = state.device
+        device_str = str(device)
+        if self._mesh_manager is None or self._mesh_manager.device != device_str:
+            self._mesh_manager = WarpMeshAndSphereCache(num_spheres=self.params.num_spheres, device=device_str)
+        manager = self._mesh_manager
```
alexmillane: we can probably use the member variable `self._mesh_manager` below.

isaaclab_arena/relations/relation_solver.py
```diff
@@ -244,11 +291,329 @@ def _compute_no_overlap_loss(
...
+    ) -> torch.Tensor:
+        """Per-env sphere-to-SDF penetration loss."""
+        device = state.device
+        total_loss = torch.zeros(state.batch_size, device=device, dtype=torch.float32)
+        clearance_m = self.params.clearance_m
+        slope = self._no_collision_strategy.slope
+
+        # Per-env loop (not batched like AABB): per-env yaw and active-pair masking produce a different sphere subset per env.
+        for b in range(state.batch_size):
+            for cache in (self._mesh_cache_forward, self._mesh_cache_reverse):
```
alexmillane: Do we ever use `self._mesh_cache_forward` and `self._mesh_cache_reverse` separately. Can we just have a single cache combining both directions of pairs?

isaaclab_arena/relations/relation_solver.py
```diff
@@ -244,11 +291,329 @@ def _compute_no_overlap_loss(
...
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Return the AABB enclosing a Z-rotated bbox (bbox_min/max: (N,3), yaws: (N,))."""
+        cos_y = torch.cos(yaws).abs().unsqueeze(1)
+        sin_y = torch.sin(yaws).abs().unsqueeze(1)
+        half = (bbox_max - bbox_min) / 2.0
+        center = (bbox_max + bbox_min) / 2.0
+        new_hx = half[:, 0:1] * cos_y + half[:, 1:2] * sin_y
+        new_hy = half[:, 0:1] * sin_y + half[:, 1:2] * cos_y
+        new_half = torch.cat([new_hx, new_hy, half[:, 2:3]], dim=1)
+        return center - new_half, center + new_half
```
alexmillane: This kind of function, where we rotate vectors by a yaw is repeated a number of times throughout this MR. Is there any way we can call out to a single, more general function in the util file somewhere?

isaaclab_arena/relations/relation_solver.py
```diff
@@ -244,11 +291,329 @@ def _compute_no_overlap_loss(
...
+                pair_sum = torch.zeros(num_pairs, device=device, dtype=penetration.dtype)
+                pair_sum.index_add_(0, active_sphere_pair_id, penetration)
+                pair_mean = pair_sum / cache.pair_sphere_count
+                active_pair_idx = active_pair.nonzero(as_tuple=True)[0]
+                total_loss[b] = total_loss[b] + slope * pair_mean[active_pair_idx].sum()
+
+        if debug:
+            print(f"  [NoOverlap MESH] total_loss={total_loss.tolist()}")
+
+        return total_loss
```
alexmillane: Here we've added 300+ lines of mesh code to the relation solver.

I'm thinking that we should move all collision code out of the solve into dedicated files. One for computing the mesh-based collision cost, and one for computing the AABB-based collision cost.

This is similar to how the other relations compute their costs: the cost computation is defered to the relation in question.

This would help to reduce the complexity of the solver. The solver just calls out to other files to get the costs, and then mimimizes them.

Can we do this?

isaaclab_arena/relations/warp_mesh_manager.py
```diff
@@ -0,0 +1,213 @@
...
+    def get_collision_mesh(self, obj: ObjectBase) -> trimesh.Trimesh | None:
+        """Return the cached collision mesh, extracting from USD on first access."""
+        # ObjectBase doesn't guarantee usd_path; only Object subclasses set it.
+        usd_path = getattr(obj, "usd_path", None)
+        if usd_path is None:
+            return obj.get_collision_mesh()
+        scale = tuple(getattr(obj, "scale", (1.0, 1.0, 1.0)))
+        key = (usd_path, scale)
+        if key not in self._trimesh_cache:
+            from isaaclab_arena.utils.usd_helpers import extract_trimesh_from_usd
```
alexmillane: local import required?

alexmillane (approved):
It's looking a lot better.

I have a few comments.

I have approved so once you solve them, feel free to merge.

isaaclab_arena/relations/warp_mesh_manager.py
zhx06: Yes, `extract_trimesh_from_usd` imports `pxr`. It cannot be run without Isaac Sim.

isaaclab_arena/relations/relation_solver.py
zhx06: Done.  Collision logic now is in `no_overlap_mesh`.py and `no_overlap_aabb.py`

isaaclab_arena/relations/relation_solver.py
zhx06: Done. `rotate_points_by_yaw` (scalar) and `rotate_points_by_yaw_batch` (per-element tensor) live in `utils/pose.py `

isaaclab_arena/relations/relation_solver.py
zhx06: Yes, it is unified into one `MeshPairCache`.

isaaclab_arena/relations/relation_solver.py
zhx06: Done. Now the solver references `self._mesh_manager` directly

isaaclab_arena/relations/object_placer.py
zhx06: Done. Now `_pair_aabb_overlaps` is now the single rotate→translate→overlap check used by both `_validate_no_overlap` and the mesh validator.

isaaclab_arena/relations/relation_solver.py
zhx06: It is the obstacle. It has docstring with it.

isaaclab_arena/assets/object_base.py
zhx06: Editted.

isaaclab_arena/relations/object_placer.py
zhx06: Docstring added.

isaaclab_arena/environments/arena_env_builder.py
zhx06: Resolved.

isaaclab_arena/relations/relation_solver.py
zhx06: Addressed.

PR #776, opened 2026-06-10
Extend eval runner: per-env camera recording, metrics file, and episode boundaries

aiguldzh-nvidia:
## Summary                                                                                                                                                                                                                                                    
                                                                  
Extend eval runner with per-env camera recording, metrics file output, timestamped run directories, and per-job episode record JSON                                                                                                                           
  
## Detailed description                                                                                                                                                                                                                                       
                                                                
- `CameraObsVideoRecorder` now records all parallel envs (one file per env per camera) instead of env 0 only; each run gets a timestamped subdirectory under `--video_dir` to avoid stale files accumulating across runs                                      
- `eval_runner` gains `--camera_video`, `--metrics_file`, and `--episode_record_dir` flags; `save_metrics_to_file()` and camera video recording existed but were never wired up from the eval runner
- `--episode_record_dir` writes one JSON file per job after it completes (or fails), capturing: `task_name`, `embodiment`, `env_params`, `policy_type`, `policy_config`, `language_instruction`, `hdf5_path`, `video_paths`, `metrics`, `episode_boundaries`, 
`wall_time_seconds`                                                                                                                                                                                                                                           
- `rollout_policy()` now returns `(metrics, episode_boundaries)` tracking `{env_idx, start_step, end_step}` per completed episode; frame index in `CameraObsVideoRecorder` output equals step index so boundaries can be used to slice per-env videos into    
individual episode clips                                                                                                                                                                                                                                      
- `Job` now carries `task_name`, `embodiment`, and `env_params` extracted from `arena_env_args` before it is converted to a CLI list
- HDF5 dataset filename now includes a run timestamp to prevent `EAGAIN` file-lock conflicts when a previous run crashed without releasing the lock

### DIFF
--- a/isaaclab_arena/evaluation/eval_runner.py
+++ b/isaaclab_arena/evaluation/eval_runner.py
@@ -173,9 +173,18 @@ def _split_episodes_across_rebuilds(num_episodes: int | None, num_rebuilds: int,
 
 
 def _stamp_path(path: str, run_ts: str) -> str:
-    """Insert run_ts between the parent dir and the final segment of path."""
+    """Insert run_ts so all run artifacts share a common timestamped ancestor.
+
+    ``parent/name``      -> ``parent/<ts>/name``   (file or directory with a parent)
+    ``name``             -> ``name/<ts>``           (bare directory name)
+    ``name.ext``         -> ``<ts>/name.ext``       (bare filename — ts prepended)
+    """
     d, final = os.path.split(path.rstrip(os.sep))
-    return os.path.join(d, run_ts, final) if d else os.path.join(final, run_ts)
+    if d:
+        return os.path.join(d, run_ts, final)
+    if os.path.splitext(final)[1]:
+        return os.path.join(run_ts, final)
+    return os.path.join(final, run_ts)
 
 
 def _extract_hdf5_path(env) -> str | None:
--- a/isaaclab_arena/evaluation/eval_runner.py
+++ b/isaaclab_arena/evaluation/eval_runner.py
@@ -201,8 +201,7 @@ def _write_job_record(
 
     language_instruction = job.language_instruction
     if language_instruction is None and env is not None:
-        with contextlib.suppress(Exception):
-            language_instruction = env.unwrapped.cfg.task_description
+        language_instruction = env.unwrapped.cfg.task_description
 
     plain_metrics = metrics_to_plain_python_types(metrics) if metrics else {}
     plain_metrics.pop("subtask_success_rate", None)
--- a/isaaclab_arena/evaluation/eval_runner.py
+++ b/isaaclab_arena/evaluation/eval_runner.py
@@ -15,7 +15,6 @@ import sys
 import tempfile
 import torch
 import traceback
-from datetime import datetime, timezone
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -187,33 +186,14 @@ def _stamp_path(path: str, run_ts: str) -> str:
     return os.path.join(final, run_ts)
 
 
-def _extract_hdf5_path(env) -> str | None:
-    """Return the HDF5 output path from env's recorder config, or None."""
-    with contextlib.suppress(Exception):
-        cfg = env.unwrapped.cfg
-        if hasattr(cfg, "recorders") and cfg.recorders is not None:
-            return str(Path(cfg.recorders.dataset_export_dir_path) / (cfg.recorders.dataset_filename + ".hdf5"))
-    return None
-
-
 def _write_job_record(
     record_dir: str,
     job: Job,
     metrics: "MetricsDataCollection | None",
-    video_dir: str | None,
     env=None,
-    hdf5_path: str | None = None,
 ) -> str:
     """Write a JSON record for one completed or failed job."""
     os.makedirs(record_dir, exist_ok=True)
-
-    video_paths: list[str] = []
-    if video_dir:
-        job_video_dir = os.path.join(video_dir, job.name)
-        if os.path.isdir(job_video_dir):
-            video_paths = sorted(
-                os.path.join(job_video_dir, n) for n in os.listdir(job_video_dir) if n.endswith(".mp4")
-            )
 
     wall_time = None
     if job.start_time is not None and job.end_time is not None:
@@ -223,9 +203,6 @@ def _write_job_record(
     if language_instruction is None and env is not None:
         with contextlib.suppress(Exception):
             language_instruction = env.unwrapped.cfg.task_description
-
-    if hdf5_path is None and env is not None:
-        hdf5_path = _extract_hdf5_path(env)
 
     plain_metrics = metrics_to_plain_python_types(metrics) if metrics else {}
     plain_metrics.pop("subtask_success_rate", None)
@@ -239,9 +216,7 @@ def _write_job_record(
     record = {
         "schema_version": "1.0",
         "job_name": job.name,
-        "task_name": job.task_name,
-        "embodiment": job.embodiment,
-        "env_params": dict(job.env_params),
+        "arena_env_args": job.arena_env_args_dict,
         "policy_type": job.policy_type,
         "policy_config": dict(job.policy_config_dict),
         "num_envs": job.num_envs,
@@ -251,10 +226,7 @@ def _write_job_record(
         "status": job.status.value,
         "metrics": plain_metrics,
         "per_episode_metrics": per_episode_metrics,
-        "hdf5_path": hdf5_path,
-        "video_paths": video_paths,
         "wall_time_seconds": wall_time,
-        "created_at": datetime.now(timezone.utc).isoformat(),
     }
 
     path = os.path.join(record_dir, f"{job.name}.json")
@@ -368,7 +340,6 @@ def main():
             policy = None
 
             metrics_per_run: list[MetricsDataCollection] = []
-            last_hdf5_path: str | None = None
 
             # num_episodes is the total across rebuilds, so split it over the rebuilds.
             num_episodes_per_rebuild = _split_episodes_across_rebuilds(job.num_episodes, job.num_rebuilds, job.name)
@@ -415,7 +386,6 @@ def main():
                         language_instruction=job.language_instruction,
                     )
 
-                    last_hdf5_path = _extract_hdf5_path(env)
                     job_manager.complete_job(job, metrics=metrics, status=Status.COMPLETED)
 
                     # users may not specify metrics for a task, although it's not recommended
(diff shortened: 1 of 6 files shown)


isaaclab_arena/evaluation/camera_video.py
```diff
@@ -3,24 +3,22 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""Gym wrapper that records one mp4 per camera in ``obs['camera_obs']``.
+"""Gym wrapper that records one mp4 per camera per env in ``obs['camera_obs']``.
```
alexmillane: Out of interest, what was the prior behaviour... Perhaps I'll find out if I read on.

isaaclab_arena/evaluation/camera_video.py
```diff
@@ -91,34 +91,46 @@ def step(self, action):
...
+                if k not in self.buffers:
+                    self.buffers[k] = [[] for _ in range(n_envs)]
+                for env_idx in range(n_envs):
+                    self.buffers[k][env_idx].append(_to_uint8(frames[env_idx]))
+
+            if self.buffers and all(
+                len(env_frames) >= self.video_length
+                for env_frame_lists in self.buffers.values()
+                for env_frames in env_frame_lists
+            ):
```
alexmillane: This compound statement is hard to read—suggestion to add a comment (or split it over multiple lines).

Maybe the comment is something like: "flush if all the videos in the buffer exceed the video length.

isaaclab_arena/evaluation/camera_video.py
```diff
@@ -91,34 +91,46 @@ def step(self, action):
...
+            for env_idx, frames in enumerate(env_frame_lists):
+                if not frames:
+                    continue
+                path = os.path.join(
+                    self.video_folder,
+                    f"{self.name_prefix}-env{env_idx}-{_sanitize_cam_key(cam)}-step-{self.recording_start_step}.mp4",
+                )
+                clip = ImageSequenceClip(list(frames), fps=self.fps)
+                clip.write_videofile(path, logger=None, audio=False)
+                del clip
```
alexmillane: Am I correct that this could take some time to write? What do you think about adding some logging here to indicate it's started. Maybe just a print at the start of the function?

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -253,8 +349,30 @@ def main():
                     if metrics is not None:
                         metrics_logger.append_job_metrics(job.name, metrics)
 
+                    if args_cli.episode_record_dir:
+                        rec_path = _write_episode_record(
```
alexmillane: Rollout policies above records for multiple episodes, so this function actually writes per-_job_ right?

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -127,6 +134,73 @@ def _close_job_resources(policy: "PolicyBase | None", env) -> None:
...
+        "schema_version": "1.0",
+        "job_name": job.name,
+        "task_name": job.task_name,
+        "embodiment": job.embodiment,
+        "env_params": dict(job.env_params),
+        "policy_type": job.policy_type,
+        "policy_config": dict(job.policy_config_dict),
+        "num_envs": job.num_envs,
+        "num_steps": job.num_steps,
+        "num_episodes": job.num_episodes,
```
alexmillane, line 224: All of these `job.*` fields are present in the config file that is _input_ to the `eval_runner.py` via the config file. Here, we re-extract the information and reexport it.

Is there a reason that we can't just use the input file (for example see [here](https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena_environments/eval_jobs_configs/zero_action_jobs_config.json)), rather than reading and then re-exporting the information?

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -127,6 +134,73 @@ def _close_job_resources(policy: "PolicyBase | None", env) -> None:
...
+        "language_instruction": language_instruction,
+        "status": status,
+        "metrics": plain_metrics,
+        "hdf5_path": hdf5_path,
+        "video_paths": video_paths,
```
alexmillane: Can the downstream consumer just read the video and hdf5 paths from the output directory?

That seems simpler to me.

Also this approach has the issue that if a user renames the output directory, the paths stored in the episode record are now wrong. Whereas if the user just searches the folder for the files they need, there's no recorded path to invalidate.

Does that make sense?

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -225,21 +309,33 @@ def main():
...
+                    if args_cli.camera_video:
+                        print(f"[INFO] Recording camera video for job '{job.name}' -> {job_video_dir}")
+                        env = CameraObsVideoRecorder(
+                            env,
+                            video_folder=job_video_dir,
+                            step_trigger=lambda step: step == 0,
+                            video_length=video_length,
+                        )
+
+                    metrics, episode_boundaries = rollout_policy(
```
alexmillane: Instead of keeping track of episode boundaries, can we modify `CameraObsVideoRecorder` to instead slice videos at the episode boundaries? That would seem like a more natural behaviour to me.

Like this, we introduce a new output to our policy runner, effectively to cover for the fact that video recorder, maybe accidentally, combines all episodes into a single video.

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -127,6 +134,73 @@ def _close_job_resources(policy: "PolicyBase | None", env) -> None:
...
+            )
+
+    wall_time = None
+    if job.start_time is not None and job.end_time is not None:
+        wall_time = round(job.end_time - job.start_time, 3)
+
+    language_instruction = job.language_instruction
+    if language_instruction is None and env is not None:
+        with contextlib.suppress(Exception):
+            language_instruction = env.unwrapped.cfg.task_description
```
alexmillane: Why do we need the suppress exception here?

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -127,6 +134,73 @@ def _close_job_resources(policy: "PolicyBase | None", env) -> None:
...
+
+    hdf5_path = None
+    if env is not None:
+        with contextlib.suppress(Exception):
```
alexmillane: Why do we need the suppress exception here?

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -253,8 +349,30 @@ def main():
                     if metrics is not None:
                         metrics_logger.append_job_metrics(job.name, metrics)
 
+                    if args_cli.episode_record_dir:
+                        rec_path = _write_episode_record(
+                            args_cli.episode_record_dir,
+                            job,
+                            metrics,
+                            "completed",
```
alexmillane: `Job` has a status. Can we use that one?

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -253,8 +349,30 @@ def main():
...
+                        print(f"[INFO] Episode record written to: {rec_path}")
+
                 except Exception as e:
                     job_manager.complete_job(job, metrics={}, status=Status.FAILED)
+                    if args_cli.episode_record_dir:
+                        _write_episode_record(
+                            args_cli.episode_record_dir,
+                            job,
+                            {},
+                            "failed",
```
alexmillane: Failed is written to the job status 5 lines above here. Consider using `Job.status`

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -270,6 +388,9 @@ def main():
 
         job_manager.print_jobs_info()
         metrics_logger.print_metrics()
+        if args_cli.metrics_file is not None:
+            metrics_logger.save_metrics_to_file()
+            print(f"[INFO] Metrics saved to: {metrics_logger.metrics_file}")
```
alexmillane, line 429: Thanks for adding that!

isaaclab_arena/evaluation/job_manager.py
```diff
@@ -28,6 +28,9 @@ def __init__(
         policy_config_dict: dict = None,
         status: Status = None,
         language_instruction: str = None,
+        task_name: str = None,
+        embodiment: str = None,
```
alexmillane: These two variables are overdetermined. They're contained in the config `.json` inside the environment.

I guess what you want here is some serialization (i.e. a string) representing the Arena environment? Is that correct? 

I think there is a better way to do this. Let's talk about it.

isaaclab_arena/evaluation/job_manager.py
```diff
@@ -28,6 +28,9 @@ def __init__(
...
+        env_params: dict = None,
```
alexmillane: These are already stored in the class as `arena_env_args`

isaaclab_arena/evaluation/policy_runner.py
```diff
@@ -63,7 +63,17 @@ def rollout_policy(
     num_steps: int | None,
     num_episodes: int | None,
     language_instruction: str | None = None,
-) -> dict[str, Any]:
+) -> tuple[dict[str, Any] | None, list[dict]]:
```
alexmillane: See comment above. Suggestion to revert this and change our camera recorder to record per-episode videos.

alexmillane (commented):
Thanks for adding this!

I have a few comments about the requirements for some of these things. I think we should be able to clean things up and simplify drastically.

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -199,18 +273,28 @@ def main():
...
-        metrics_logger = MetricsLogger()
+        metrics_logger = MetricsLogger(metrics_file=args_cli.metrics_file or "metrics.json")
 
         job_manager.print_jobs_info()
 
-        if args_cli.video:
-            os.makedirs(args_cli.video_dir, exist_ok=True)
-            print(f"[INFO] Video recording enabled. Videos will be saved to: {args_cli.video_dir}")
+        if args_cli.video or args_cli.camera_video or args_cli.metrics_file or args_cli.episode_record_dir:
+            run_ts = datetime.now().strftime("%Y%m%dT%H%M%S")
```
alexmillane: Let's just unconditionally query the date, and then we get rid of this quadruple if statement and de-indent the following if statements by one level.

isaaclab_arena/evaluation/camera_video.py
aiguldzh-nvidia: Previously, only environment 0 was recorded, and frames from other parallel environments were silently ignored. Now, all environments are recorded, producing one MP4 file per (env, camera) pair.

isaaclab_arena/evaluation/eval_runner.py
aiguldzh-nvidia: Yes, correct, it writes one json per job, not per episode, the name is misleading, I'll rename it

PR #777, opened 2026-06-10
Unify env random seeding using ManagerBasedRLEnv seed

peterd-NV:
## Summary

Arena was previously setting the seed using a custom `set_seed` function in Arena. The seeding was inconsistent and not always being set (for example it was not used for batched eval). This gave non-deterministic results.

This PR removes the inconsistent env seeding and uses Isaac Lab's built in env random seeding mechanism. Any time a user provides a seed via CLI, it is passed to the arena environment builder which sets it in the env_cfg. Isaac Lab then handles setting all global seeds. 

Note:  Isaac Lab's seed mechanism sets all of the seeds that Arena was setting plus additional seeds for replicator and warp. It is more comprehensive than what Arena had.

### DIFF
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -316,7 +316,7 @@ class ArenaEnvBuilder:
             env_cfg = self.arena_env.env_cfg_callback(env_cfg)
 
         # Set seed for Isaac Lab env.
-        env_cfg.seed = getattr(self.args, "seed", None)
+        env_cfg.seed = self.args.seed
 
         # Apply the --presets CLI flag (e.g. --presets newton).
         # This runs after the callback so the user's CLI choice is the final authority.
--- a/isaaclab_arena/environments/arena_env_builder.py
+++ b/isaaclab_arena/environments/arena_env_builder.py
@@ -310,13 +310,13 @@ class ArenaEnvBuilder:
                 viewer=viewer_cfg,
             )
 
-        # Set seed for Isaac Lab env.
-        env_cfg.seed = getattr(self.args, "seed", None)
-
         # Apply the environment configuration callback if it is set
         # This can be used to modify the simulation configuration, etc.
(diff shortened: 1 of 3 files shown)


xyao-nv (approved):
Thanks for fixing!

xyao-nv:
How about for the seed you modify here, we call it `runtime_seed`? To differentiate it from `placement_seed`.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -297,6 +297,9 @@ def compose_manager_cfg(self) -> IsaacLabArenaManagerBasedRLEnvCfg:
         if self.arena_env.env_cfg_callback is not None:
             env_cfg = self.arena_env.env_cfg_callback(env_cfg)
 
+        # Set seed for Isaac Lab env.
+        env_cfg.seed = getattr(self.args, "seed", None)
```
alexmillane: We have a default seed of 42, so I think we can skip the `getattr` no?

claude likes to chuck these everywhere, but in my opinion that hurt the readablity.

alexmillane (approved):
Looks good to me!!! Thanks for doing this!

We still have this variable `placement_seed` around. I wonder if we can also simplify by getting rid of that? But that's a question for another time.

xyao-nv:
> Looks good to me!!! Thanks for doing this!
> 
> We still have this variable `placement_seed` around. I wonder if we can also simplify by getting rid of that? But that's a question for another time.

I think Peter's idea is placement seed is at build time, and this seed is at run time. Leave it for future if unifying these two explicitly is needed.

isaaclab_arena/environments/arena_env_builder.py
xyao-nv: Double checked and agreed.
Lab handles it in configure_seed():

>     if seed is None or seed == -1:
>         seed = 42 if torch_deterministic else random.randint(0, 10000)
>

PR #778, opened 2026-06-11
Add script for e2e agentic scene gen

qianl-nv:
## Summary
Add environment_generation_runner for running the agentic environment generation end to end

## Detailed description
We need an entry point to run the full agentic env gen pipeline for integration testing.

- What has been changed?
A new script is added to run the agentic env gen. It does the following steps:
  - Takes user prompt from CLI, pass through EnvGenAgent to generate EnvIntentSpec
  - EnvIntentSpec compiles to UnresolvedArenaEnvGraphSpec (asset match into registry), then ArenaEnvGraphSpec (constraints populated). The final ArenaEnvGraphSpec serialized to yaml.
  - ArenaEnvGraphSpec is loaded back from yaml, and build into an ArenaEnvironment
  - Run zero_action policy simulation for visualizing output (with RelationSolver initializing layouts)

  - The env generation agent is harden to reduce server side failure rate from ~30% to <1%.
  - Small fixes are added to the agent: add anchor relation, and add lights, and the generate arena env is valid & can be visualized correctly in sim.
  - Known limitations: no support for ObjectSet or ObjectReference. No user in the loop editing.

To run:
python -m isaaclab_arena.agentic_environment_generation.environment_generation_runner --num_envs 16 --env_spacing 1.5 --num-steps 2000 --viz kit --prompt "franka pick up avocado from the maple table and place it into a bowl on the table. there are other veggies on the table as distractor."

2026-4-14-agentic-arena-env-generation-design-google-docs (1)


qianl-nv:
Refactored into separate MRs. Last outstanding one is this https://github.com/isaac-sim/IsaacLab-Arena/pull/785
closing

PR #784, opened 2026-06-11
Support anchor in agentic env gen

qianl-nv:
## Summary
EnvGenAgent adds the is_anchor relation.

## Detailed description
  is_anchor is currently missing from agent generated spec.
- What has been changed?
  -- agent prompt is updated to include request for is_anchor in system prompt.
  -- reformat system prompt for better readibility.
  -- pose is required for anchor object. set pose during conversion to arena env if missing.
  -- Agent generated env spec should now work with relation solver.

### DIFF
--- a/isaaclab_arena/agentic_environment_generation/environment_generation_agent.py
+++ b/isaaclab_arena/agentic_environment_generation/environment_generation_agent.py
@@ -299,6 +299,7 @@ class EnvironmentGenerationAgent:
             "  Do NOT hallucinate values — the resolver tolerates nulls; it cannot fix invented data.\n"
             "- For binary relations (e.g. on), subject is the placed object and reference is "
             "the surface it is relative to (typically the background name).\n"
+            "- REQUIRED: include an is_anchor (unary) relation for the surface other objects rest on.\n"
             "- Articulated objects (microwave, fridge, cabinet) still need an "
             "'on' relation in initial_state_graph (subject=object, reference=background) "
             "to anchor them.\n"
--- a/isaaclab_arena/agentic_environment_generation/environment_generation_agent.py
+++ b/isaaclab_arena/agentic_environment_generation/environment_generation_agent.py
@@ -288,24 +288,23 @@ class EnvironmentGenerationAgent:
         return spec, text
 
     def _system_prompt(self) -> str:
-        return (
-            "You are an env-generation parser for robot manipulation tasks.\n"
-            "Convert a natural-language prompt into an EnvironmentIntentSpec.\n\n"
-            "GUIDANCE:\n"
-            "- Follow the per-field ``description`` strings in the schema for what each field expects.\n"
-            "- Use only asset names from EMBODIMENTS / BACKGROUNDS / OBJECTS, relation "
-            "kinds from RELATIONS, and task kinds from TASKS in the user message.\n"
-            "- If the prompt does not specify a value for an optional field, output null.\n"
-            "  Do NOT hallucinate values — the resolver tolerates nulls; it cannot fix invented data.\n"
-            "- For binary relations (e.g. on), subject is the placed object and reference is "
-            "the surface it is relative to (typically the background name).\n"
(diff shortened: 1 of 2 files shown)


isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -125,6 +126,11 @@ def _attach_spatial_constraints_to_assets(
         # is passed as the Relation constructor's first arg — matches how add_relation is wired.
         if relation_class.is_unary():
             subject_asset.add_relation(relation_class(**spatial_constraint.params))
+            # An is_anchor asset must have a fixed initial pose for the placement solver.
+            # If the asset class does not declare one, default to the world origin so
+            # LLM-generated specs (which never set an explicit pose) work out of the box.
+            if spatial_constraint.kind == "is_anchor" and subject_asset.get_initial_pose() is None:
+                subject_asset.set_initial_pose(Pose.identity())
```
xyao-nv, line 133: Does it mean all anchor objs are placed at the same spot? And it will not be optimized by the solver's non-overlapping loss. 

How about uniform sampling from the target area? At least it's not biased towards a fix location.

xyao-nv: Regarding whether it's a hack, in general my thinking process for spec item is,
if user does not provide it, then it is our job to ensure we can provide valid items, and ensure it can vary each run;
else, we shall provide what user exactly asks for.

xyao-nv (approved):

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
qianl-nv: > it will not be optimized by the solver's non-overlapping loss
relation solve do not move the anchor subject position, it's always assume fixed. this does not introduce any bias, as other objects (maybe embodiment) are placed relative to  (and sampled uniformly w.r.t.) anchor subject by the solver.

isaaclab_arena/environments/arena_env_graph_conversion_utils.py
```diff
@@ -125,6 +126,11 @@ def _attach_spatial_constraints_to_assets(
         # is passed as the Relation constructor's first arg — matches how add_relation is wired.
         if relation_class.is_unary():
             subject_asset.add_relation(relation_class(**spatial_constraint.params))
+            # An is_anchor asset must have a fixed initial pose for the placement solver.
+            # If the asset class does not declare one, default to the world origin so
+            # LLM-generated specs (which never set an explicit pose) work out of the box.
+            if spatial_constraint.kind == "is_anchor" and subject_asset.get_initial_pose() is None:
```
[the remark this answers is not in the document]
qianl-nv: is_anchor is a special marker for the solver. your suggestion for designing a constant right before using it in the next line doesn't seem an improvement. skipping it.

isaaclab_arena/agentic_environment_generation/environment_generation_agent.py
```diff
@@ -288,23 +288,23 @@ def generate_spec(
...
+
+GUIDANCE:
+- Follow the per-field ``description`` strings in the schema for what each field expects.
+- Use only asset names from EMBODIMENTS / BACKGROUNDS / OBJECTS, relation kinds from \
+RELATIONS, and task kinds from TASKS in the user message.
+- If the prompt does not specify a value for an optional field, output null.
+  Do NOT hallucinate values — the resolver tolerates nulls; it cannot fix invented data.
+- For binary relations (e.g. on), subject is the placed object and reference is \
+the surface it is relative to (typically the background name).
+- REQUIRED: include an is_anchor (unary) relation for the surface other objects rest on.
```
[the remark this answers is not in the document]
qianl-nv: instruction highlighted this is unary relation, and in the RelationSpec field description we explicitly explains for unary relation only subject is assigned and reference is null. I think the LLM should be able to figure this out.

PR #785, opened 2026-06-11
Add e2e script for agentic env gen

qianl-nv:
## Summary
Add environment_generation_runner for running the agentic environment generation end to end

## Detailed description
  We need an entry point to run the full agentic env gen pipeline for integration testing.

- What has been changed?
A new script is added to run the agentic env gen. It does the following steps:
  - Takes user prompt from CLI, pass through EnvGenAgent to generate EnvIntentSpec
  - EnvIntentSpec compiles to UnresolvedArenaEnvGraphSpec (asset match into registry), then ArenaEnvGraphSpec (constraints populated). The final ArenaEnvGraphSpec serialized to yaml.
  - ArenaEnvGraphSpec is loaded back from yaml, and build into an ArenaEnvironment
  - Run zero_action policy simulation for visualizing output (with RelationSolver initializing layouts)

  - Known limitations: no support for ObjectSet or ObjectReference. No user in the loop editing.

To run: python -m isaaclab_arena_examples.agentic_environment_generation.environment_generation_runner --num_envs 16 --env_spacing 1.5 --num-steps 2000 --viz kit [--prompt <custom prompt>]

2026-4-14-agentic-arena-env-generation-design-google-docs (1)

### DIFF
--- a/isaaclab_arena_examples/agentic_environment_generation/environment_generation_runner.py
+++ b/isaaclab_arena_examples/agentic_environment_generation/environment_generation_runner.py
@@ -1,0 +1,217 @@
+# Copyright (c) 2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""End-to-end agentic environment generation and execution.
+
+Pipeline:
+
+  1. Call :class:`EnvironmentGenerationAgent` on the prompt →
+     :class:`~isaaclab_arena.agentic_environment_generation.environment_intent_spec.EnvironmentIntentSpec`.
+  2. :class:`~isaaclab_arena.agentic_environment_generation.intent_compiler.IntentCompiler` →
+     :class:`~isaaclab_arena.environments.arena_env_graph_spec.ArenaEnvInitialGraphSpec`.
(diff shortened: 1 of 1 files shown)


isaaclab_arena_examples/agentic_environment_generation/environment_generation_runner.py
```diff
@@ -0,0 +1,217 @@
...
+        help="Directory for the generated YAML files (default: isaaclab_arena_environments/llm_generated).",
+    )
+
+
+def main() -> int:
+    parser = get_isaaclab_arena_cli_parser()
+    _add_runner_cli_args(parser)
+    args_cli = parser.parse_args()
+
+    with SimulationAppContext(args_cli):
```
xyao-nv, line 100: Is there anything involving SimApp before step 5? 
In EnvGraph building process, I purposely involve SimApp only at the build arena env stage. So steps before it is clean, and fast given iterations will be expected.
I'm assuming with your html, you don't want to invoke SimApp too as it's slow and costly.
Anything I miss?

qianl-nv: I have to move step 1-5 inside the SimApp context, otherwise there's pxr import error when sim starting. I believe SimApp crashes if the pxr modules are reimported after step 1-5 imported them first (probably from registry).
LMK if you have idea how to get around it.

qianl-nv:
A refactored version of this script is merging in https://github.com/isaac-sim/IsaacLab-Arena/pull/796
closing this one

PR #791, opened 2026-06-13
Add asset snapshot in browser-based spec reviewer tool

qianl-nv:
## Summary
Spin up a sim app and render snapshots for each node to visualize in the browser.

## Detailed description
- What has been changed?


qianl-nv:
Splitted the impl and working on this one instead https://github.com/isaac-sim/IsaacLab-Arena/pull/804

PR #792, opened 2026-06-15
split camera recordings per episode instead of one long video

aiguldzh-nvidia:
## Summary                                                                                                                                                                                                                                                  
Split camera recordings per episode
                                                                                                                                                                                                                                                            
## Detailed description                                         
- `CameraObsVideoRecorder` recorded one long video per `(env, camera)` spanning all episodes, requiring `episode_boundaries` metadata to slice it afterward. It also only recorded env 0, silently dropping all other envs.                                 
- Replaced with per-episode flushing: each env's buffer is flushed independently on reset, producing one mp4 per `(env, camera, episode)` named `<prefix>-env<N>-<camera_name>-episode-<E>.mp4`                                                             
- Removed `step_trigger` and `video_length` from the interface                                                                                                                                                                                              
- Partial episodes cut off by `--num_steps` are discarded on close; only episodes that end with a real reset produce output files                                                                                                                                                                                                                                                             
- Added `--camera_video` to `eval_runner`
- Fixed `policy_runner` to set `enable_cameras=True` both before `SimulationAppContext` and after the full arg parse, covering both `--camera_video` and `--camera-video` aliases                                                                           
- Added 7 unit tests covering flush-on-termination, per-env episode counters, sequential filenames, partial episode drop, empty buffer handling, and post-reset frame discard - no Isaac Sim or GPU required; real file-writing test skipped without ffmpeg

### DIFF
--- a/isaaclab_arena/evaluation/camera_video.py
+++ b/isaaclab_arena/evaluation/camera_video.py
@@ -9,7 +9,7 @@ Frames are flushed to disk each time an environment resets (terminated or
 truncated), so each output file corresponds to exactly one complete episode.
 Partial episodes cut off by ``num_steps`` are discarded on ``close()``.
 
-Output filename: ``<name_prefix>-env<N>-<cam>-episode-<E>.mp4``
+Output filename: ``<name_prefix>-env<N>-<camera_name>-episode-<E>.mp4``
 
 policy_runner.py wraps the env with this alongside ``RecordVideo`` so
 the kit viewport mp4 (third-person scene view) and the embodiment-
@@ -17,9 +17,10 @@ mounted camera mp4s (what the policy actually sees) are written
 together when ``--video --camera_video`` is set.
 
 Memory note: each env buffers raw uint8 frames for its current episode before
-encoding.  Peak RAM is N×L×H×W×C bytes where L is max episode length, not the
-full rollout.  For 10 envs, 500-step episodes, 512×512×3 frames that is ~3.8 GB
-of raw frames — the encoded mp4s are far smaller (H.264 compresses ~100:1).
+encoding.  Buffers are cleared after each episode is written to disk, so peak
+RAM is N×L×H×W×C bytes where L is max episode length, not the full rollout.
+For 10 envs, 500-step episodes, 512×512×3 frames that is ~3.8 GB of raw frames
+— the encoded mp4s are far smaller (H.264 compresses ~100:1).
 """
 
 from __future__ import annotations
@@ -46,9 +47,9 @@ def _to_uint8(frame: torch.Tensor | np.ndarray) -> np.ndarray:
     return frame.astype(np.uint8)
 
 
-def _sanitize_cam_key(key: str) -> str:
-    """Strip path separators so a camera key can't escape video_folder."""
-    return key.replace("/", "_").replace(os.sep, "_")
+def _sanitize_cam_key(camera_name: str) -> str:
+    """Strip path separators so a camera name can't escape video_folder."""
+    return camera_name.replace("/", "_").replace(os.sep, "_")
 
 
 class CameraObsVideoRecorder(gym.Wrapper):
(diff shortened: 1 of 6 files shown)


isaaclab_arena/evaluation/camera_video.py
```diff
@@ -3,24 +3,23 @@
...
 
-Multi-env note: Arena batches camera observations as
-``[N_envs, H, W, C]``. This wrapper records env 0 only (``frame[0]``);
-other envs are silently dropped. ``RecordVideo`` for the kit viewport
-is unaffected — the viewport shows the full scene regardless of
-``num_envs``.
+Memory note: each env buffers raw uint8 frames for its current episode before
+encoding.  Peak RAM is N×L×H×W×C bytes where L is max episode length, not the
+full rollout.  For 10 envs, 500-step episodes, 512×512×3 frames that is ~3.8 GB
+of raw frames — the encoded mp4s are far smaller (H.264 compresses ~100:1).
```
alexmillane: Do we clear memory after an episode when the frames are written to disk? I think so. Maybe just make that clear in the comment.

isaaclab_arena/evaluation/camera_video.py
```diff
@@ -54,71 +52,95 @@ def _sanitize_cam_key(key: str) -> str:
...
+            n_envs = frames_sample.shape[0]
+
+            if self._n_envs is None:
+                self._n_envs = n_envs
+                self.episode_counts = [0] * n_envs
+
+            # Determine done envs before appending frames. Isaac Lab auto-resets on
+            # termination, so the obs returned for a done env is the post-reset first
+            # frame of the new episode — discard it so it doesn't contaminate the
+            # current episode or create a phantom one-frame episode on close().
```
alexmillane: Do we loose this frame from the next episode? Maybe that's not a huge problem? Maybe clarify this in the comment.

isaaclab_arena/evaluation/camera_video.py
```diff
@@ -54,71 +52,95 @@ def _sanitize_cam_key(key: str) -> str:
...
+            if isinstance(terminated, torch.Tensor):
+                done_envs = (terminated | truncated).nonzero().flatten().tolist()
+            else:
+                done_envs = [0] if (terminated or truncated) else []
```
alexmillane: What's going on here? In what case do we not get a tensor back?

From reading the code, is this the case when we only have a single env or something?

Might be worth a comment.

isaaclab_arena/evaluation/camera_video.py
```diff
@@ -54,71 +52,95 @@ def _sanitize_cam_key(key: str) -> str:
...
+            done_set = set(done_envs)
+
+            for k, frames in cam_obs.items():
+                if k not in self.buffers:
+                    self.buffers[k] = [[] for _ in range(n_envs)]
```
alexmillane: What is k here? Is this a `str` or an `int`

Suggestion to call this `camera_name` or `camera_idx` (or something completely different that gives a clue on what it is.

alexmillane (approved):
Thanks for doing this!

I left a couple of nits.

What do you think about adding a test? Run an environment and check that a video file apprears?

Feel free to merge when you've addressed.

isaaclab_arena/evaluation/camera_video.py
aiguldzh-nvidia: extended the comment

isaaclab_arena/evaluation/camera_video.py
aiguldzh-nvidia: extended the comment

isaaclab_arena/evaluation/camera_video.py
aiguldzh-nvidia: this was a fallback for gym envs, agree that it's redundant, removed it

isaaclab_arena/evaluation/camera_video.py
aiguldzh-nvidia: Renamed k -> camera_name

PR #795, opened 2026-06-15
Add X/Y margin to ON relation & access by intent compiler

xyao-nv:
## Summary

## Detailed description
- Why: An object placed On a surface could solve to a pose on the rim, causing it topple off the edge. There was no way to keep its footprint a safe distance inside the parent surface.
- What:
1. On has an `edge_margin_m` parameter (default 0.0) that insets the valid X/Y footprint band by the margin, so the whole footprint stays at least that far from every rim. 
2. A margin too large for the parent surface is now rejected in validation
3. `IntentCompiler.compile()` injects a default edge_margin_m (0.05 m) on every on relation built from an intent spec

- Impact: Agent-generated scenes keep objects clear of surface edges by default. 

## Results

optimization_margin_0p30
placement_margin_0p30

### With agentic env gen pipeline 
Before
Screenshot from 2026-06-15 12-17-15

After
Screenshot from 2026-06-15 11-02-57


isaaclab_arena/agentic_environment_generation/intent_compiler.py
```diff
@@ -53,19 +52,28 @@ def has_resolution_errors(self) -> bool:
...
         """Compile an :class:`EnvironmentIntentSpec` into an :class:`ArenaEnvInitialGraphSpec`.
 
         Args:
             spec: Agent-produced intent spec describing the scene, initial relations,
                 and task chain.
             env_name: Override for the graph's ``env_name`` field.  When ``None``
                 the name is derived as ``llm_gen_{background}_{first_task_kind}``.
+            on_edge_margin_m: Inward inset (meters) injected as ``edge_margin_m`` on every
+                ``on`` relation, keeping the subject's footprint off the edge of its support
+                surface.  Set to ``0.0`` to inject nothing (plain ``on`` behavior).
```
zhx06: For docstring, can we use `On` instead of `on`?

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -274,11 +274,26 @@ def compute_loss(
...
         parent_z_max = parent_world_bbox.max_point[:, 2]  # Top surface
 
-        # Compute valid position ranges such that child's entire footprint is within parent
-        valid_x_min = parent_x_min - child_bbox.min_point[:, 0]  # child's left at parent's left
-        valid_x_max = parent_x_max - child_bbox.max_point[:, 0]  # child's right at parent's right
-        valid_y_min = parent_y_min - child_bbox.min_point[:, 1]
-        valid_y_max = parent_y_max - child_bbox.max_point[:, 1]
+        # Compute valid position ranges such that child's entire footprint is within parent,
+        # with the parent's extent inset by edge_margin_m so the footprint stays off the rim.
+        m = relation.edge_margin_m
```
zhx06, line 279: Can we add similar logics to `_validate_on_relations` in `ObjectPlacer` to make it consistent?

isaaclab_arena/relations/relation_loss_strategies.py
```diff
@@ -274,11 +274,26 @@ def compute_loss(
...
+        valid_x_min = parent_x_min + m - child_bbox.min_point[:, 0]  # child's left at parent's left + margin
+        valid_x_max = parent_x_max - m - child_bbox.max_point[:, 0]  # child's right at parent's right - margin
+        valid_y_min = parent_y_min + m - child_bbox.min_point[:, 1]
+        valid_y_max = parent_y_max - m - child_bbox.max_point[:, 1]
+
+        # The lower and upper bounds can be inverted under 2 conditions:
+        # 1. The margin is too large for the support surface.
+        # 2. The child is oversized.
+        # Reject the case where margin is too large. Keep the case where child is oversized.
+        if m > 0.0:
```
zhx06: Similar to the previous comment, we probably should not do a validation step in the middle of optimization. If it is not feasible, we should have a high loss, instead of rejecting it inside optimization. It is better to check how `clearance_m` in applied in loss computation. The rejection should be done in a validation step.

isaaclab_arena/agentic_environment_generation/intent_compiler.py
xyao-nv: ofc

isaaclab_arena/relations/relation_loss_strategies.py
xyao-nv: good call, added

isaaclab_arena/relations/relation_loss_strategies.py
xyao-nv: Thanks for clarifying the original design. Updated to solving with high loss -> rejecting in validation.

zhx06 (approved):
LGTM

isaaclab_arena/agentic_environment_generation/intent_compiler.py
```diff
@@ -183,6 +193,18 @@ def _build_spatial_constraint(
...
+        """Override ``edge_margin_m`` in ``rel.params`` for ``On`` relations."""
+        rel.params["edge_margin_m"] = on_edge_margin_m
+        self.trace.append(
+            IntentResolutionTraceEvent(
+                "relation.initial.on_edge_margin",
+                rel.subject,
+                rel.reference,
+                note=f"edge_margin_m={on_edge_margin_m}",
+            )
+        )
```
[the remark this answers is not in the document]
xyao-nv: updated

isaaclab_arena/relations/object_placer.py
```diff
@@ -547,11 +547,34 @@ def _validate_on_relations(
...
+                            " this layout is rejected."
+                        )
+                        return False
                 if (
-                    child_world.min_point[0, 0] < parent_world.min_point[0, 0]
-                    or child_world.max_point[0, 0] > parent_world.max_point[0, 0]
-                    or child_world.min_point[0, 1] < parent_world.min_point[0, 1]
-                    or child_world.max_point[0, 1] > parent_world.max_point[0, 1]
+                    child_world.min_point[0, 0] < parent_world.min_point[0, 0] + m
+                    or child_world.max_point[0, 0] > parent_world.max_point[0, 0] - m
```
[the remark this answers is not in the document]
xyao-nv: updated

isaaclab_arena/agentic_environment_generation/intent_compiler.py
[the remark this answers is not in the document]
xyao-nv: corrected

isaaclab_arena/agentic_environment_generation/intent_compiler.py
```diff
@@ -173,14 +181,31 @@ def _build_spatial_constraint(
...
+    def _inject_on_edge_margin_in_params(
+        self, params: dict, subject: str, reference: str | None, on_edge_margin_m: float
+    ) -> None:
+        """Inject ``edge_margin_m`` into the constraint's ``params`` copy, leaving ``rel`` untouched."""
+        params["edge_margin_m"] = on_edge_margin_m
+        self.trace.append(
+            IntentResolutionTraceEvent(
+                "relation.initial.on_edge_margin",
+                rel.subject,
+                rel.reference,
```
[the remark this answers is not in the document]
xyao-nv: updated

isaaclab_arena/relations/object_placer.py
```diff
@@ -547,14 +547,40 @@ def _validate_on_relations(
...
+                            return False
                 if (
-                    child_world.min_point[0, 0] < parent_world.min_point[0, 0]
-                    or child_world.max_point[0, 0] > parent_world.max_point[0, 0]
-                    or child_world.min_point[0, 1] < parent_world.min_point[0, 1]
-                    or child_world.max_point[0, 1] > parent_world.max_point[0, 1]
+                    child_world.min_point[0, 0] < parent_world.min_point[0, 0] + m
+                    or child_world.max_point[0, 0] > parent_world.max_point[0, 0] - m
+                    or child_world.min_point[0, 1] < parent_world.min_point[0, 1] + m
+                    or child_world.max_point[0, 1] > parent_world.max_point[0, 1] - m
```
alexmillane, line 575: From my reading this function is a composition of three sequential tests. (you're add a new one here). The function is getting a bit long and difficult to read.

Suggestion to just add some comments above each of the three steps:

```
# 1) Checking that the parent is wide enough to place the child on top with enough margin
# 2) Checking that the child lies within the parent's xy
# 3) Checking that the child lies within an acceptable z-range. 
```

isaaclab_arena/relations/object_placer.py
```diff
@@ -547,14 +547,40 @@ def _validate_on_relations(
...
+                    free_x = (
+                        parent_world.max_point[0, 0]
+                        - parent_world.min_point[0, 0]
+                        - (child_world.max_point[0, 0] - child_world.min_point[0, 0])
+                    ).item()
+                    free_y = (
+                        parent_world.max_point[0, 1]
+                        - parent_world.min_point[0, 1]
+                        - (child_world.max_point[0, 1] - child_world.min_point[0, 1])
+                    ).item()
```
alexmillane: Suggestion to simplify, shorten, and make more readable these conditionals through the introduction of intermediate variables.

```
parent_size = parent_world.max_point - parent_world.min_point
child_size = parent_world.max_point - parent_world.min_point
if edge_margin_m > 0.0:
   freespace = parent_size - child_size
   if torch.any(freespace[0, :2] < 2 *edge_margin_m):
        ...
```

isaaclab_arena/agentic_environment_generation/intent_compiler.py
```diff
@@ -53,19 +52,28 @@ def has_resolution_errors(self) -> bool:
         """``True`` if the last :meth:`compile` call produced any error-stage trace events."""
         return bool(self.resolution_errors)
 
-    def compile(self, spec: EnvironmentIntentSpec, env_name: str | None = None) -> ArenaEnvInitialGraphSpec:
+    def compile(
+        self,
+        spec: EnvironmentIntentSpec,
+        env_name: str | None = None,
+        on_edge_margin_m: float = DEFAULT_ON_EDGE_MARGIN_M,
```
alexmillane: Instead of threading this through the intent compiler, what do you think about just making this distance the default for the relation.

Of course, maybe down the line we need to add a separate specification, for the agent(?), but honestly this margin just seems like a good thing to be using everywhere.

What do you think?

isaaclab_arena/agentic_environment_generation/intent_compiler.py
alexmillane: See above. Suggestion to just make this a default and then we don't have to pass this parameter around everywhere.

isaaclab_arena/agentic_environment_generation/intent_compiler.py
```diff
@@ -173,14 +181,31 @@ def _build_spatial_constraint(
...
+        """Inject ``edge_margin_m`` into the constraint's ``params`` copy for ``On`` relations."""
+        params["edge_margin_m"] = on_edge_margin_m
+        self.trace.append(
+            IntentResolutionTraceEvent(
+                "relation.initial.on_edge_margin",
+                subject,
+                reference,
+                note=f"edge_margin_m={on_edge_margin_m}",
+            )
         )
```
alexmillane: Suggestion to just make the edge margin a default and then we can get rid of all this complexity.

alexmillane (approved):
LGTM

A few nits but my main suggestion is to just make this margin a default then we can get rid of the complexity resulting from needing to thread the parameter around.

isaaclab_arena/relations/object_placer.py
xyao-nv: updated

isaaclab_arena/relations/object_placer.py
xyao-nv: good idea, thx. added

isaaclab_arena/agentic_environment_generation/intent_compiler.py
xyao-nv: Added to rel params using 0.05 as default. 
Removed the unnecessary threading.
For agent interface, will add it once we have a way to auto-derive it.

isaaclab_arena/agentic_environment_generation/intent_compiler.py
xyao-nv: done

isaaclab_arena/agentic_environment_generation/intent_compiler.py
xyao-nv: done

PR #806, opened 2026-06-18
DRAFT: Record per episode metadata and results

alexmillane:
## Summary
Record per-episode data.

## Detailed description
- Adds a per-episode data recorder to our `env` that accumulates per-episode data and optionally writes it to disk.
- Getting at the per environment success flag requires overriding internal an internal method of `IsaacLabManagerBasedRLEnv`


isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -310,6 +320,11 @@ def main():
                         language_instruction=job.language_instruction,
                     )
 
+                    # Request the per-episode results write into this job's output subdir
+                    # (the same directory the video recorders use), one file per rebuild.
+                    results_path = os.path.join(video_cfg.video_base_dir, f"episode_results_rebuild{rebuild_idx}.jsonl")
+                    env.unwrapped.episode_results_recorder.write(results_path)
```
aiguldzh-nvidia, line 325: write() is inside the try block, so if it raises (e.g. a permissions error) the except catches it and marks the job as failed, even though the rollout itself completed successfully

aiguldzh-nvidia (commented):
The per-episode recording contains almost all of the data we need. The only thing missing is some job-level metadata, but I don't think that's a blocker, the current approach looks good to me.

alexmillane:
Superceded by #810

PR #810, opened 2026-06-22
Episode Recorder

alexmillane:
## Summary
Introduce an `EpisodeRecorderManager` that records per-episode data.

## Detailed description
- Adds a per-episode data recorder to our `env` that accumulates per-episode data and optionally writes it to disk.
- Users can register custom `EpisodeRecorderTermCfgs` to record their custom data.
- Implemented 2 basic recorders.
  - core metadata recorder (env_id, episode_idx, success, etc.)
  - variations recorder (records samples for the variations)
- Added the recorded meta data into the evaluation report.

## Note
- Getting at the per environment success flag requires overriding internal an internal method of `IsaacLabManagerBasedRLEnv`

## Not done
- Gracefully handle rebuilds and `run_in_chunks`. Will be handled in later MRs.

PR #811, opened 2026-06-23
SensitivityDataset: read episode_results.jsonl directly

cvolkcvolk:
## Summary
Sensitivity analysis reads recorder output directly

## Detailed description
- **Why:** the per-episode recorder now emits an `episode_results.jsonl` carrying the realized variation draws, so the old hand-written `factors.yaml` schema plus bespoke nested `episode_summary.jsonl` pair is redundant.
- **Reader:** added `SensitivityDataset.from_episode_results`, which discovers the factors from each row's `variations` block — a number becomes continuous, a numeric vector splits into one factor per component, a string becomes categorical (choices learned from the data). `generate_report` takes a single `--episode_results` input, with `--factors` to analyze a subset.
- **Simpler data model:** removed the `factors.yaml` path (`from_files`), the `FactorSchema` wrapper, and `FactorSpec.dim`. It is now just `FactorSpec` (one factor = one theta column) and `SensitivityDataset` (factors + theta + x). Continuous-only sweeps fit NPE, mixed sweeps fit MNPE automatically.
- **Robustness:** factors that never varied are dropped (a constant categorical otherwise crashed the MNPE fit); an all-constant input errors clearly.


isaaclab_arena/analysis/sensitivity/dataset.py
```diff
@@ -225,103 +164,141 @@ def default_observation(self) -> torch.Tensor:
...
+
+    factors: list[FactorSpec] = []
+    columns: list[torch.Tensor] = []
+    for name in continuous_names:
+        values = factor_values[name]
+        factors.append(FactorSpec(name=name, type=FactorType.CONTINUOUS, range=[(min(values), max(values))]))
+        columns.append(torch.tensor(values, dtype=torch.float32).unsqueeze(1))
+    for name in categorical_names:
+        choices = sorted(set(factor_values[name]))
+        if len(choices) == 1:
```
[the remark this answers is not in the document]
cvolkcvolk: Adressed. Dropping non-varied factors.

isaaclab_arena/analysis/sensitivity/dataset.py
```diff
@@ -22,16 +22,15 @@ class FactorType(str, Enum):
 
 @dataclass
 class FactorSpec:
-    """One factor's schema as declared in factors.yaml.
+    """One factor's schema, occupying a single column of theta.
```
alexmillane: Potentially this "single column of theta" is a bit confusing without an explanation of what theta is? I would guess theta is a matrix whose rows are samples and whose collumns are variations?

isaaclab_arena/analysis/sensitivity/dataset.py
```diff
@@ -22,16 +22,15 @@ class FactorType(str, Enum):
...
-    factors carry choices (a list of string labels, integer-encoded by index in theta).
+    A continuous factor carries a range, one [low, high] pair. A categorical factor carries
+    choices, a list of string labels that are integer-encoded by their index in theta.
     """
 
     name: str
     type: FactorType
-    dim: int = 1
-    range: list[tuple[float, float]] | None = None  # one (low, high) pair per dim, continuous only
+    range: list[tuple[float, float]] | None = None  # a single (low, high) pair, continuous only
```
alexmillane: "# a single (low, high) pair, continuous only"

is this true? The type says there's a list of high, low pairs?

isaaclab_arena/analysis/sensitivity/dataset.py
```diff
@@ -42,163 +41,95 @@ def __post_init__(self) -> None:
...
-    The object is a pure container: it holds the schema and the two tensors, and exposes
-    the prior and column layouts an analyzer consumes. It can be built two ways:
+    """The varied factors paired with their per-episode theta (factor values) and x (outcomes).
 
-      - from_files — parse a factors.yaml / episode_summary.jsonl pair
-        (the path eval runs take).
-      - the constructor — wrap in-memory tensors directly (what a synthetic simulator or
-        a unit test takes). The tensors must already be in the layout factor_columns
-        describes: continuous columns first, then one integer-coded column per categorical.
+    A pure container holding the factor list, the two tensors, and the column layout an analyzer
```
alexmillane: I feel like the use of the work "column" here needs explanation.

isaaclab_arena/analysis/sensitivity/dataset.py
```diff
@@ -42,163 +41,95 @@ def __post_init__(self) -> None:
...
 
     @classmethod
-    def from_files(
+    def from_episode_results(
         cls,
-        factors_yaml: str | Path,
         jsonl_path: str | Path,
         outcome_names: list[str] | tuple[str, ...] = ("success",),
+        factor_names: list[str] | tuple[str, ...] | None = None,
     ) -> SensitivityDataset:
```
alexmillane: What do you think about moving this function and `_build_dataset_from_episode_rows`.

The presence of this function in this file couples the dataset definition to the write format of the episode recorder. That's necessary, but I feel it would be nicer to have it in another file so a reader just interested in a pure dataset in memory, doesn't have to see the file format ugliness.

isaaclab_arena/analysis/sensitivity/dataset.py
```diff
@@ -209,119 +140,202 @@ def num_episodes(self) -> int:
...
+        pairs.append((f"{key}[{component_index}]", float(component)))
+    return pairs
+
+
+def _build_dataset_from_episode_rows(
+    rows: list[dict],
+    outcome_names: list[str] | tuple[str, ...],
+    jsonl_path: str | Path,
+    factor_names: list[str] | tuple[str, ...] | None = None,
+) -> tuple[list[FactorSpec], torch.Tensor, torch.Tensor]:
```
alexmillane: Suggestion to have this guy just return `SensitivityDataset`

isaaclab_arena/analysis/sensitivity/dataset.py
```diff
@@ -209,119 +140,202 @@ def num_episodes(self) -> int:
...
+        f"All discovered factors in {jsonl_path} are constant (each took a single value across all "
+        "episodes); nothing to analyze. Vary at least one factor."
+    )
+
+    theta = torch.cat(columns, dim=1)
+    x = torch.cat(
+        [torch.tensor([float(row[name]) for row in rows], dtype=torch.float32).unsqueeze(1) for name in outcome_names],
+        dim=1,
+    )
+    return factors, theta, x
```
alexmillane: suggestion to break this function up into a handful of smaller (internal) functions. The top-level function should be a short composition of intenal functions.

isaaclab_arena/tests/test_sensitivity_analysis.py
alexmillane, line 1: We may also want to add an e2e smoke test. I.e. run something through the policy runner and then do a sensitivity analysis.

Given the loosely enforced contract (actually not enforced) between the episode recorder and the reading on the sensitivity side, such a test would be useful to ensure that this (unspoken) contract remains adhered to.

alexmillane (approved):
Looks great.

Only real comment is the request for a e2e smoke test.

isaaclab_arena/analysis/sensitivity/dataset.py
cvolkcvolk: yeah good idea

isaaclab_arena/tests/test_sensitivity_analysis.py
cvolkcvolk: Good idea. Adding it in a follow up once #810 is in

isaaclab_arena/analysis/sensitivity/dataset.py
cvolkcvolk: Good point. Done

isaaclab_arena/analysis/sensitivity/dataset.py
cvolkcvolk: Done

PR #815, opened 2026-06-24
Add missing robolab assets

qianl-nv:
## Summary
Add robolab assets used in pick and place tasks

## Detailed description
We want to use agent to regenerate pick and place tasks in robolab using their assets.
- What has been changed?
Add the list of missing assets
Fix scales to match robolab default scales for some of the assets
This only adds to asset registry. No other side effect.

### DIFF
--- a/isaaclab_arena/assets/object_library.py
+++ b/isaaclab_arena/assets/object_library.py
@@ -1785,13 +1785,6 @@ class WoodBlockYcbRobolab(LibraryObject):
     usd_path = f"{ISAACLAB_NUCLEUS_DIR}/Arena/assets/object_library/srl_robolab_assets/objects/ycb/wood_block.usd"
 
 
-@register_asset
-class TableMapleRobolab(LibraryObject):
-    name = "table_maple_robolab"
-    tags = ["background", "fixture", "robolab"]
-    usd_path = f"{ISAACLAB_NUCLEUS_DIR}/Arena/assets/object_library/srl_robolab_assets/fixtures/table_maple.usd"
-
-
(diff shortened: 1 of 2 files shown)


isaaclab_arena/assets/object_library.py
```diff
@@ -1786,10 +1786,230 @@ class WoodBlockYcbRobolab(LibraryObject):
...
+    tags = ["object", "container", "graspable", "robolab"]
+    usd_path = f"{ISAACLAB_NUCLEUS_DIR}/Arena/assets/object_library/srl_robolab_assets/objects/vomp/wireshelving_a01/wireshelving_a01.usd"
+
+
+@register_asset
+class GreyBinRobolab(LibraryObject):
+    name = "grey_bin_robolab"
+    tags = ["object", "container", "robolab"]
+    usd_path = f"{ISAACLAB_NUCLEUS_DIR}/Arena/assets/object_library/srl_robolab_assets/fixtures/grey_bin.usd"
+    scale = (0.007, 0.007, 0.007)
```
xyao-nv, line 1987: Can you add a note saying the original USD is too large with size in xx meters?

xyao-nv (approved):

isaaclab_arena/assets/object_library.py
qianl-nv: done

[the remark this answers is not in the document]
qianl-nv: this is in the original. IsaacLab spawner drops the USD default scale, so we have to set it here explicitly to match.

PR #822, opened 2026-06-25
Add unified success check termination function

peterd-NV:
## Summary
Adds a new unified termination function (`check_success`) that can be composed of multiple termination predicates. Allows tasks to compose multiple termination funcs together without needing to create a new func.

Individual termination predicates are composed one of 3 ways:

- ALL (all must be true)
- ANY (at least one must be true)
- CHOOSE (at least k must be true)  

Updates PickAndPlace task to use the `check_success`


qianl-nv (approved):
Thanks for the change Peter! I've checked that this indeed fix the FP success for my put-into-bin tasks (after adding objects_in_proximity predicate).

PR #825, opened 2026-06-25
Add objects_in_proximity to PickAndPlaceTask success termination

qianl-nv:
## Summary
Require proximity for pick-and-place success

## Detailed description
Contact-only success can fire when the object grazes the destination edge; add `objects_in_proximity` as a second predicate under `check_success` (ALL).

- What has been changed?
Expose `max_separation` on `PickAndPlaceTask`. Default to None so objects_in_proximity is skipped for backward compatibility; set to 0.1 m on three robolab yamls with bins as the destination.

Existing envs keep working with defaults unless center-to-destination distance exceeds the tolerance; tune per env via YAML if needed.

### DIFF
--- a/isaaclab_arena/tasks/pick_and_place_task.py
+++ b/isaaclab_arena/tasks/pick_and_place_task.py
@@ -24,7 +24,7 @@ from isaaclab_arena.metrics.success_rate import SuccessRateMetric
 from isaaclab_arena.tasks.common.mimic_default_params import MIMIC_DATAGEN_CONFIG_DEFAULTS
 from isaaclab_arena.tasks.task_base import TaskBase
 from isaaclab_arena.tasks.task_transition import Relocate, TaskTransition
-from isaaclab_arena.tasks.terminations import SuccessMode, check_success, object_on_destination
+from isaaclab_arena.tasks.terminations import SuccessMode, check_success, object_on_destination, objects_in_proximity
 from isaaclab_arena.utils.cameras import get_viewer_cfg_look_at_object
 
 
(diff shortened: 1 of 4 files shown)


isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -57,6 +57,9 @@ def __init__(
         task_description: str | None = None,
         force_threshold: float = 0.1,
         velocity_threshold: float = 0.1,
+        max_x_separation: float = 0.5,
+        max_y_separation: float = 0.5,
+        max_z_separation: float = 0.5,
```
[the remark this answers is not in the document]
qianl-nv: PR description updated. 0.5 is intentional to not break existing passing envs.

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -101,6 +107,18 @@ def make_termination_cfg(self):
                             "velocity_threshold": self.velocity_threshold,
                         },
                     ),
+                    # TODO(qianl): replace objects_in_proximity with object_centroid_in_container
```
peterd-NV: Nit about this plan: pick and place doesn't need to necessarily have a container, for example the destination could be a different table. I would probably make the future function something like: `object_centroids_in_proximity`

peterd-NV (approved):
lgtm

isaaclab_arena/tasks/pick_and_place_task.py
qianl-nv: I actually think we need more than in_proximity with (axis-align) x/y/z threshold. for a circular plate or a rotated object for example this might have edge case failures. we probably need something like "inside 2d pojected surface of destination object with xxx cm margin". I'll leave this as is for now

qianl-nv: actually decided to update it as you suggest.

PR #828, opened 2026-06-26
Switch default model for agentic env gen

qianl-nv:
## Summary
Current default model in env gen agent runs into frequent timeouts and flacky unit test results.


alexmillane (approved):
Thaaanks!

PR #835, opened 2026-06-29
Add Collision Check for Background Objects

zhx06:
## Summary

Add background collision checking to relation-based object placement

## Detailed description

- Background objects without relations were invisible to the placement solver, so placed objects could overlap furniture.
- Fixed, relation-free scene geometry is now treated as passive obstacles during placement; in `MESH` mode fixtures merge into one aggregate collision mesh, with AABB fallback for meshless objects.
- Placed objects now clear nearby fixtures (RoboCasa kitchen example below).

kitchen_place_and_view


isaaclab_arena/scene/scene.py
```diff
@@ -119,6 +120,43 @@ def get_objects_with_relations(self) -> list[Object | ObjectReference]:
...
+            if isinstance(asset, ObjectReference) and asset.parent_asset.usd_path is None:
+                continue
+            initial_pose = asset.get_initial_pose()
+            # A non-fixed pose on static furniture is almost certainly an authoring mistake.
+            if isinstance(initial_pose, (PoseRange, PosePerEnv)):
+                print(
+                    f"Warning: background object '{asset.name}' has a non-fixed pose "
+                    f"({type(initial_pose).__name__}); excluding it from collision obstacles."
+                )
+                continue
```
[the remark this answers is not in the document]
zhx06: We use print across the project

isaaclab_arena/scene/scene.py
```diff
@@ -119,6 +122,60 @@ def get_objects_with_relations(self) -> list[Object | ObjectReference]:
                 objects_with_relations.append(asset)
         return objects_with_relations
 
+    def get_collision_objects(self, combine_background_mesh: bool = False) -> list[Any]:
```
qianl-nv: rename to get_passive_collision_objects for clarify

isaaclab_arena/assets/object_base.py
```diff
@@ -58,6 +58,8 @@ def __init__(
         self.object_cfg = None
         self.event_cfg = None
         self.relations: list[RelationBase] = []
+        self.use_collision_mesh_as_is = False
```
qianl-nv: Put the inline comment on top of the field 

```
# If True, mesh collision uses this object's raw mesh without convex-hull repair.
self.use_collision_mesh_as_is = False
```

isaaclab_arena/relations/placement_events.py
```diff
@@ -23,6 +24,36 @@
...
+
+
+def register_placement_pool(placement_pool: PooledObjectPlacer) -> str:
+    """Register a runtime placement pool and return its lightweight key."""
+    key = f"placement_pool_{next(_PLACEMENT_POOL_KEY_COUNTER)}"
+    _PLACEMENT_POOL_REGISTRY[key] = placement_pool
+    return key
+
+
+def unregister_placement_pool(key: str) -> None:
```
qianl-nv: this is defined by not used in production. should we call it in close_environment?

isaaclab_arena/relations/background_collision_object.py
```diff
@@ -0,0 +1,141 @@
...
+    def get_world_bounding_box(self) -> AxisAlignedBoundingBox:
+        """Return the world-space bounds of the combined background mesh."""
+        return self._bounding_box
+
+    def get_collision_mesh(self) -> trimesh.Trimesh:
+        """Return the combined background mesh in world coordinates."""
+        return self._mesh
+
+
+def make_background_collision_object(objects: Sequence[ObjectBase]) -> BackgroundCollisionObject | None:
```
qianl-nv: make_background_collision_object is also not used in production path, only in test. 
consider dropping it if all we need is make_background_collision_objects

isaaclab_arena/relations/background_collision_object.py
```diff
@@ -0,0 +1,141 @@
...
+from isaaclab_arena.utils.pose import Pose
+
+if TYPE_CHECKING:
+    from isaaclab_arena.assets.object_base import ObjectBase
+
+
+BACKGROUND_COLLISION_OBJECT_NAME = "__background_collision_mesh__"
+
+
+class BackgroundCollisionObject:
```
qianl-nv: "Background" in the name is a bit misleading, since this includes assets both in Background and generic Object. 
Conisder renaming it "FixedCollisionObject"?

isaaclab_arena/scene/scene.py
```diff
@@ -119,6 +122,60 @@ def get_objects_with_relations(self) -> list[Object | ObjectReference]:
...
+            Passive collision objects for placement loss and validation.
+        """
+        collision_objects: list[Object | ObjectReference] = []
+        for asset in self.assets.values():
+            if not isinstance(asset, (Object, ObjectReference)):
+                continue
+            if isinstance(asset, Background) and not combine_background_mesh:
+                continue
+            if asset.get_relations():
+                continue
```
qianl-nv: we should exclude ObjectReference whose parent is already included in the collision_objects list
this might be relevant for ObejctReference that are used for pick and place destination, which don't get a relation

qianl-nv (approved):
Thanks Zihao, change is well structured and thorough.

A few nits, feel free to merge after fixing. 

One question: is the goal for v0.3 to switch to mesh mode by default?  Are we still keeping AABB mode then? 
This MR currently don't provide a user interface through CLI/config to switch to the mesh mode with policy_runner/eval_runner scripts. What's your plan for it?

isaaclab_arena/assets/dummy_object.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         assert self.bounding_box is not None
         self.relations = list(relations)
         self._collision_mesh = collision_mesh
+        self.use_collision_mesh_as_is = False
```
alexmillane: Shall we instead say what this variable does, rather than what it optionally doesn't do?

What about `self.do_collision_mesh_convex_hull_repair = True`

isaaclab_arena/assets/object_reference.py
```diff
@@ -73,14 +77,34 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+        world_position = self.get_initial_pose().position_xyz
+        parent_pose = self.parent_asset.initial_pose
+        if parent_pose is None:
+            return box.translated(world_position)
+        quarters = quaternion_to_90_deg_z_quarters(parent_pose.rotation_xyzw)
+        return box.rotated_90_around_z(quarters).translated(world_position)
+
+    def get_collision_mesh(self) -> trimesh.Trimesh | None:
+        """Return the referenced prim's collision mesh in its local frame, or None if unavailable."""
+        if not getattr(self, "_collision_mesh_loaded", False):
```
alexmillane: suggestion to get rid of the `getattr` by initializing the variable `_collision_mesh_loaded` in the constructor.

Just as a general rule, if you're having to use `getattr`, and you're not meta-programming, something is likely wrong.

isaaclab_arena/assets/object_reference.py
```diff
@@ -73,14 +77,34 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+            self._collision_mesh_loaded = True
+            try:
+                usd_path = getattr(self.parent_asset, "usd_path")
```
alexmillane: see comment above on `getattr`

isaaclab_arena/assets/object_reference.py
```diff
@@ -73,14 +77,34 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+                with open_stage(usd_path) as parent_stage:
+                    prim_path_in_usd = self.isaaclab_prim_path_to_original_prim_path(
+                        self.prim_path, self.parent_asset, parent_stage
+                    )
+                    self._collision_mesh = extract_trimesh_from_prim(parent_stage, prim_path_in_usd, self._parent_scale)
+            except (OSError, ValueError) as e:
+                print(f"Could not extract collision mesh for object reference '{self.name}': {e}")
+                return None
```
alexmillane, line 107: Suggestion to hide this stage stuff in a function somewhere. Perhaps alongside `extract_trimesh_from_prim`

isaaclab_arena/scene/scene.py
alexmillane: `combine_background_mesh` -> `include_background`

isaaclab_arena/scene/scene.py
```diff
@@ -119,6 +122,60 @@ def get_objects_with_relations(self) -> list[Object | ObjectReference]:
...
+                    f"Skipping background object '{asset.name}' as a collision obstacle: "
+                    f"needs a fixed pose but has {pose_kind}."
+                )
+                continue
+            collision_objects.append(asset)
+        if not collision_objects:
+            return []
+        if combine_background_mesh:
+            return make_background_collision_objects(collision_objects)
+        return collision_objects
```
alexmillane: Can we get rid of these different returns:

- for the first return, the list is already empty, so you can remove this condition no?
- For the second return, just do: `collision_objects make_background_collision_objects(collision_objects)`

isaaclab_arena/scene/scene.py
alexmillane: The return annotation could be specified further according to the type definition below.

isaaclab_arena/scene/scene.py
```diff
@@ -119,6 +122,60 @@ def get_objects_with_relations(self) -> list[Object | ObjectReference]:
...
+            combine_background_mesh: If True and mesh extraction succeeds, return one
+                collision-only object with extracted background meshes baked into world coordinates,
+                plus individual meshless non-Background objects for AABB fallback.
+                When False, whole-scene Background assets are skipped because their AABB spans
+                the full scene and would reject valid layouts.
+
+        Returns:
+            Passive collision objects for placement loss and validation.
+        """
+        collision_objects: list[Object | ObjectReference] = []
```
alexmillane: This disagrees with the return annotation.

isaaclab_arena/scene/scene.py
```diff
@@ -119,6 +122,60 @@ def get_objects_with_relations(self) -> list[Object | ObjectReference]:
...
+
+        A qualifying asset is an Object / ObjectReference that carries no relations (so it is
+        absent from get_objects_with_relations), exposes a USD path for geometry, and has a
+        single fixed initial Pose. Assets with a per-env / per-reset or unset pose are skipped,
+        since no constant world bounding box can be computed for them.
+
+        Args:
+            combine_background_mesh: If True and mesh extraction succeeds, return one
+                collision-only object with extracted background meshes baked into world coordinates,
+                plus individual meshless non-Background objects for AABB fallback.
```
alexmillane: Does this also bake in the non background objects into a single mesh?

isaaclab_arena/scene/scene.py
alexmillane: Suggestion to move this function out of the scene.

The scene is just a container for objects effectively and doesn't know anything about collisions. Suggestion to move this into a (potentially new) file that deals with collision stuff.

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -24,8 +28,22 @@ def solve_and_apply_relation_placement(
     objects: list[ObjectBase],
     num_envs: int,
     placer_params: ObjectPlacerParams | None = None,
+    collision_objects: list[ObjectBase] | None = None,
```
alexmillane: Is it possible that we move the object around as a single list and extract the relevant information inside functions.

If that's too much repeated code we could introduce a new class/container that encapsulates that functionality.

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -120,8 +139,10 @@ def _apply_dynamic_spawn_pose(
         func=solve_and_place_objects,
         mode="reset",
         params={
-            "objects": objects,
-            "placement_pool": placement_pool,
+            "object_names": [obj.name for obj in objects],
+            "anchor_object_names": [obj.name for obj in anchor_objects_set],
+            "base_rotations_by_name": {obj.name: get_rotation_xyzw(obj) for obj in objects},
```
alexmillane: What's a base rotation?

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -120,8 +139,10 @@ def _apply_dynamic_spawn_pose(
...
+            "placement_pool_key": register_placement_pool(placement_pool),
```
alexmillane: What's happened here?

We appear to have go from two abstract inputs to the event term, to passing 4 arguments.

That doesn't appear related to the mesh-background subject of this MR?

isaaclab_arena/relations/background_collision_object.py
```diff
@@ -0,0 +1,141 @@
...
+from typing import TYPE_CHECKING
+
+from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox
+from isaaclab_arena.utils.pose import Pose
+
+if TYPE_CHECKING:
+    from isaaclab_arena.assets.object_base import ObjectBase
+
+
+BACKGROUND_COLLISION_OBJECT_NAME = "__background_collision_mesh__"
```
alexmillane: Just put this variable inline in the constructor.

Why use double underscores? `__`. Usually that is reserved for python internal stuff. Any reason to use it here?

isaaclab_arena/relations/background_collision_object.py
```diff
@@ -0,0 +1,141 @@
...
+    if skipped_objects:
+        aabb_fallback_objects = [obj for obj in skipped_objects if not isinstance(obj, Background)]
+        skipped_backgrounds = [obj for obj in skipped_objects if isinstance(obj, Background)]
+        if skipped_backgrounds:
+            print(
+                "Background mesh extraction failed for whole-scene Background assets "
+                f"{[obj.name for obj in skipped_backgrounds]}; skipping their room-sized AABB fallback."
+            )
+        if not aabb_fallback_objects:
+            return collision_objects
```
alexmillane: Can we get rid of the early return by relying on the list being empty, so the extend call below is a no-op

isaaclab_arena/relations/background_collision_object.py
```diff
@@ -0,0 +1,141 @@
...
+        mesh = manager.get_collision_mesh(obj)
+        if mesh is None:
+            skipped_objects.append(obj)
+            continue
+        pose = obj.get_initial_pose()
+        assert isinstance(pose, Pose), f"Background object '{obj.name}' must have a fixed Pose."
+        meshes.append(_mesh_in_world_frame(mesh, pose))
+    if not meshes:
+        return None, skipped_objects
+    return trimesh.util.concatenate(meshes), skipped_objects
```
alexmillane: This function has nothing to do with `background`s right? It just combines meshes right?

Suggestion to remove the Background designation.

isaaclab_arena/relations/placement_events.py
```diff
@@ -23,6 +24,36 @@
...
+
+def _get_registered_placement_pool(key: str) -> PooledObjectPlacer:
+    """Return a registered placement pool or raise a reset-friendly error."""
+    try:
+        return _PLACEMENT_POOL_REGISTRY[key]
+    except KeyError as exc:
+        raise RuntimeError(
+            f"Placement pool '{key}' is not registered in this process. "
+            "Pooled placement reset events must be created and used in the same Python process."
+        ) from exc
```
alexmillane: What's driving the requirement to introduce the placement pool store?

This seems tangential to the subject of this MR which is introducing background mesh support for the placer.

isaaclab_arena/relations/placement_events.py
```diff
@@ -79,15 +113,33 @@ def write_layout_to_sim(
...
+        base_rotations_by_name={obj.name: rotation for obj, rotation in base_rotations.items()},
+    )
+
+
+def _write_layout_to_sim_by_name(
+    env: ManagerBasedEnv,
+    env_id: int,
+    result: PlacementResult,
+    anchor_object_names: set[str],
+    base_rotations_by_name: dict[str, tuple[float, float, float, float]],
```
alexmillane: why "base" rotations? Is this just a rotation?

isaaclab_arena/relations/relation_solver.py
```diff
@@ -141,8 +141,13 @@ def _compute_no_overlap_loss(
         state: RelationSolverState,
         debug: bool = False,
     ) -> torch.Tensor:
-        """Compute pairwise no-overlap loss, skipping On-linked pairs."""
+        """Compute pairwise no-overlap loss, skipping On-linked pairs.
+
+        - Non-anchor vs fixed obstacle (anchor or background): gradient flows to the non-anchor only.
+        - Non-anchor vs non-anchor mesh pairs are checked in every available SDF direction.
```
alexmillane: What is "every available sdf direction?

Do we just mean the non-anchor vs non-anchor the gradient flows both ways?

alexmillane (commented):
Review #1. 

The results look great!

I have a few comments. I'll have another review after you take a look.

isaaclab_arena/scene/scene.py
zhx06: Done. Renamed to `get_passive_collision_objects`.

isaaclab_arena/assets/object_base.py
zhx06: Done. Replaced the negative `use_collision_mesh_as_is` with a positive field: `repair_collision_mesh_non_watertight = True`, with the comment above the field.

isaaclab_arena/relations/placement_events.py
zhx06: Done. The placement pool is now unregistered when the env closes.

isaaclab_arena/relations/background_collision_object.py
zhx06: Done. Removed the test-only singular factory and kept the production plural path.

isaaclab_arena/relations/background_collision_object.py
zhx06: Done. Renamed to `FixedCollisionObject`.

isaaclab_arena/scene/scene.py
zhx06: Done.` Scene.get_passive_collision_objects() `now removes `ObjectReferences` whose parent asset is already included, independent of asset iteration order.

isaaclab_arena/assets/dummy_object.py
zhx06: Done. Renamed the flag to `repair_collision_mesh_non_watertight`

isaaclab_arena/assets/object_reference.py
zhx06: Done. `_collision_mesh_loaded` is now initialized in `ObjectReference.__init__`

isaaclab_arena/assets/object_reference.py
zhx06: Done. `getattr ` removed.

isaaclab_arena/assets/object_reference.py
zhx06: Done. `get_collision_mesh()` now delegates the USD-stage/path-resolution work to `_extract_collision_mesh()`

isaaclab_arena/scene/scene.py
zhx06: Done. The arg is now `include_background`

isaaclab_arena/scene/scene.py
zhx06: Done. Removed the empty-list return and now use one final return after optional aggregation.

isaaclab_arena/scene/scene.py
zhx06: Done. The internal list is now typed as `list[ObjectBase]`

isaaclab_arena/scene/scene.py
zhx06: Yes, non-Background objects can also be baked into the fixed collision mesh.

isaaclab_arena/environments/relation_solver_interface.py
zhx06: Agreed, this was too much for this MR. I reverted this reset-event param refactor,

isaaclab_arena/environments/relation_solver_interface.py
zhx06: Done. Renamed  to `rotations_by_name`

isaaclab_arena/relations/background_collision_object.py
zhx06: Done. Changed to `fixed_collision_mesh`.

isaaclab_arena/relations/background_collision_object.py
zhx06: Done. Removed the early return.

isaaclab_arena/relations/background_collision_object.py
zhx06: Done. We make it fixed-collision focus.

isaaclab_arena/relations/relation_solver.py
zhx06: Yes, I have polished the docstring

isaaclab_arena/relations/placement_events.py
zhx06: It stores each object’s rotation before applying the random yaw init.

isaaclab_arena/scene/scene.py
zhx06: Addressed. It returns `list[ObjectBase | FixedCollisionObject]`

zhx06: Addressed. It now moves to `isaaclab_arena.relations`.

isaaclab_arena/environments/relation_solver_interface.py
zhx06: So currently we leave it there because the behaviour of these objects are very different. `Objects`  are foreground objects are go into the Placement Pool, while `collision_objects` are mostly background objects. A new class/container is doable, but this will make the scope of this MR bigger. We could have a future MR for this. WDYT?

isaaclab_arena/relations/placement_events.py
zhx06: Agree. The changes are reverted.

isaaclab_arena/assets/object_reference.py
```diff
@@ -73,14 +77,45 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
...
+        world_position = self.get_initial_pose().position_xyz
+        parent_pose = self.parent_asset.initial_pose
+        if parent_pose is None:
+            return box.translated(world_position)
+        quarters = quaternion_to_90_deg_z_quarters(parent_pose.rotation_xyzw)
+        return box.rotated_90_around_z(quarters).translated(world_position)
+
+    def get_collision_mesh(self) -> trimesh.Trimesh | None:
+        """Return the referenced prim's collision mesh in its local frame, or None if unavailable."""
+        if not self._collision_mesh_loaded:
```
alexmillane, line 100: nit: Do we need a separate flag for this of could we detect `self._collision_mesh == None`.

Maybe there's a good reason not to.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -95,10 +101,40 @@ def _solve_relations(self) -> None:
...
+            include_background=self._should_include_background_mesh(
+                objects_with_relations, placer_params.solver_params.collision_mode
+            ),
+        )
         self._placement_event_cfg = solve_and_apply_relation_placement(
             objects_with_relations,
             num_envs=self.cfg.num_envs,
             placer_params=placer_params,
+            collision_objects=collision_objects,
+        )
```
alexmillane: Here we make two calls to `solve_and_apply_relation_placement` which have the same form. They only differ in that the first call calls with `collision_objects=[]` and the second call calls with some collision_objects.

Can we collapse these two code paths into a single call to `solve_and_apply_relation_placement`.

Also it's not clear to me what these two code paths represent. I feel we need a comment here. The first code path is `objects_with_relations=None` and `collision_objects=[]`... What is actually happening then? 😆

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -95,10 +101,40 @@ def _solve_relations(self) -> None:
...
+        """Return True when the default mode or any object/Background override resolves to MESH."""
+        if default_collision_mode == CollisionMode.MESH:
+            return True
+        if any(get_object_collision_mode(obj, default_collision_mode) == CollisionMode.MESH for obj in objects):
+            return True
+        return any(
+            isinstance(asset, Background)
+            and get_object_collision_mode(asset, default_collision_mode) == CollisionMode.MESH
+            for asset in self.arena_env.scene.assets.values()
         )
```
alexmillane: Suggestion to hide all of this outside the `arena_env_builder`. Let's return this function to what it was before this MR, and move the reasoning about which objects to pass into the called function.

Suggestion to just take the whole scene into the first arugment of `solve_and_apply_relation_placement` and do the reasoning about which objects go where inside, so it's not visible in the EnvironmentCompiler.

isaaclab_arena/relations/background_collision_object.py
```diff
@@ -0,0 +1,145 @@
...
+    transformed.apply_transform(transform)
+    return transformed
+
+
+def _bounding_box_from_mesh(mesh: trimesh.Trimesh) -> AxisAlignedBoundingBox:
+    bounds = mesh.bounds
+    return AxisAlignedBoundingBox(
+        min_point=tuple(float(v) for v in bounds[0]),
+        max_point=tuple(float(v) for v in bounds[1]),
+    )
```
alexmillane: Shall we move these to `isaaclab_arena.utils.trimesh`

alexmillane (approved):
A few more small comments. Feel free to merge once addressed.

isaaclab_arena/environments/arena_env_builder.py
zhx06: Done. The builder now makes one call.

isaaclab_arena/environments/arena_env_builder.py
zhx06: Addressed.

isaaclab_arena/relations/background_collision_object.py
zhx06: Done. Both helpers now live in `isaaclab_arena/utils/trimesh.py`

isaaclab_arena/assets/object_reference.py
zhx06: The flag is intentional because `None` is a valid cached result for meshless prims.

PR #836, opened 2026-06-29
Ignore Isaac Sim core dumps in Docker build context

xyao-nv:
## Summary
IsaacSim SimApp Crash core dumps (core.*, ~3GB each) accumulate in the repo root and were pulled into the image by 'COPY *.*', bloating it by tens of GB. Exclude them via .dockerignore.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,5 @@
 /.git
 /.agents
 /.claude
+# Isaac Sim crash core dumps — ~3GB each; otherwise pulled into the image by `COPY *.*`
+core.*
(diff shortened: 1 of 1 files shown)


.dockerignore
```diff
@@ -3,3 +3,5 @@
 /.git
 /.agents
 /.claude
+# Isaac Sim crash core dumps — ~3GB each; otherwise pulled into the image by `COPY *.*`
+core.*
```
peterd-NV: Seems like a reasonable suggestion above

peterd-NV (approved):
minor comment, see if you want to include

PR #837, opened 2026-06-30
Add Curobo-IK reachability post-validation check (Part 1/3)

xyao-nv:
## Summary
Placement post-validation: Curobo IK reachability tests. Limit to DROID, in-Sim, top-down grasp in this MR.

## Detailed description
- Why? Pooled placement layouts could place objects where the DROID arm cannot actually grasp them, with no
  feasibility gate to catch it.
- What?

1. Curobo-Lab handles all obstacles in robot base frame, thru observing the world. Adds a robot-base-frame obstacle transformation call.
2. Adds `isaaclab_arena_curobo` pkg including the planner for droid, a batched top-down-grasp IK checker, and a pool validator to stamp P/F per layout.
3. Curobo related DROID cfg are added to S3 as they are converted from USD by @njawale42 
4. Enables the cuRobo image (`./docker/run_docker.sh -c`) and a CI job for curobo related test. Removed GR00T related docker machinery.

- Impact: it records verdicts and does not alter runtime placement; all cuRobo code is gated to the cuRobo image.

## Example Outputs
```
/isaac-sim/python.sh isaaclab_arena_curobo/scripts/run_placement_pool_ik_validation.py --headless --num_envs 10 --env_graph_spec_yaml isaaclab_arena_environments/robolab/two_bin_linked.yaml 
```
```
....
9 episode 0: PASS [no_overlap=True, on_relation=True, next_to=True, not_next_to=True, ik_reachable=True]
env 9 episode 1: PASS [no_overlap=True, on_relation=True, next_to=True, not_next_to=True, ik_reachable=False]
env 9 episode 2: PASS [no_overlap=True, on_relation=True, next_to=True, not_next_to=True, ik_reachable=True]
env 9 episode 3: PASS [no_overlap=True, on_relation=True, next_to=True, not_next_to=True, ik_reachable=True]
env 9 episode 4: PASS [no_overlap=True, on_relation=True, next_to=True, not_next_to=True, ik_reachable=True]
Summary: 33/50 layouts IK-reachable.
```

## TODO
1. Part 2/3: Connect with placer validation
2. Part 3/3: Skip spawning SimApp (P0) + batch env IK (P1)


isaaclab_arena_curobo/curobo_planner_utils.py
```diff
@@ -0,0 +1,226 @@
...
+                obj_pos_w = wp.to_torch(obj.data.root_pos_w)[self.env_id, :3]
+                obj_quat_w = wp.to_torch(obj.data.root_quat_w)[self.env_id, :4]
+                # Re-express the object pose in the robot base frame so it matches the IK goal frame.
+                pos_robot = (r_w2r @ (obj_pos_w - robot_pos_w).unsqueeze(-1)).squeeze(-1)
+                quat_robot = math_utils.quat_mul(robot_quat_inv.unsqueeze(0), obj_quat_w.unsqueeze(0))[0]
+                pos_c = self._to_curobo_device(pos_robot)
+                quat_c = self._to_curobo_device(quat_robot)
+                # cuRobo's world-model writer wants a flat [x, y, z, qw, qx, qy, qz] list.
+                pose_list = [
+                    float(pos_c[0]),
```
qianl-nv: pose_list = torch.cat((pos_c, quat_c)).tolist()
this should work and more compact?

isaaclab_arena_curobo/curobo_planner_utils.py
```diff
@@ -0,0 +1,226 @@
...
+from isaaclab_mimic.motion_planners.curobo.curobo_planner import CuroboPlanner
+from isaaclab_mimic.motion_planners.curobo.curobo_planner_cfg import CuroboPlannerCfg
+
+if TYPE_CHECKING:
+    from isaaclab.envs import ManagerBasedEnv
+
+# Top-down grasp orientation (gripper pointing -Z) expressed in the robot base frame, (w, x, y, z).
+DOWN_FACING_QUAT_WXYZ = torch.tensor([0.0, 1.0, 0.0, 0.0], dtype=torch.float32)
+
+_ROBOT_ASSET_DIR = f"{ISAACLAB_NUCLEUS_DIR}/Arena/assets/robot_library/droid/droid_fixed_mimic_joint"
```
qianl-nv: can we move the curobo configs for driod into the embodiment class, and make these utils read from the embodiment (and error out if the embodiment is missing curobo config)?
would be cleaner than the current way of hard coding droid into everything.

isaaclab_arena_curobo/ik_utils.py
```diff
@@ -0,0 +1,142 @@
...
+        quaternion=PoseUtils.quat_from_matrix(rotations),
+    )
+
+    ik_seed = None
+    if seed_config is not None:
+        ik_seed = planner._to_curobo_device(seed_config)
+        while ik_seed.dim() < 3:
+            ik_seed = ik_seed.unsqueeze(0)
+
+    ik_result = planner.motion_gen.ik_solver.solve_batch(goal_pose, seed_config=ik_seed)
```
qianl-nv, line 62: do we also need to check ik_result.success besides the pose residual?

docker/Dockerfile.isaaclab_arena
```diff
@@ -97,25 +94,9 @@ RUN (type -p wget >/dev/null || (apt update && apt install wget -y)) \
     apt update && \
     apt install -y gh
 
-###############################
```
qianl-nv: can we split out removing gr00t into a separate MR and merge that first? mixing adding curobo with removing gr00t makes the current MR hard to read.

qianl-nv (commented):
Thanks for adding the curobo deps!
Would be nice to split the change removing Gr00t from docker to a separate MR (I imagine we can merge that one quickly) to keep this clean.

docker/Dockerfile.isaaclab_arena
alexmillane: Agreed here. This will also make the history cleaner, should we need to look back at the gr00t removal, which is a possiblity.

docker/Dockerfile.isaaclab_arena
```diff
@@ -155,6 +126,25 @@ COPY docs ${WORKDIR}/docs
 # Install IsaacLab Arena (runtime + dev deps declared in setup.py)
 RUN /isaac-sim/python.sh -m pip install -e "${WORKDIR}/[dev]"
 
+################################
+# Install cuRobo (optional)
+################################
+COPY docker/setup/install_cuda.sh /tmp/install_cuda_curobo.sh
```
alexmillane: nit: `COPY docker/setup/install_cuda.sh /tmp/install_cuda_curobo.sh`

I feel like its a bit confusing to change the name of the script. Any reason to do that?

.github/workflows/ci.yml
```diff
@@ -327,11 +359,42 @@ jobs:
...
+      - *cleanup_step
+      - *checkout_step
+      - *install_docker_step
+      - *docker_login_ngc_step
+
+      # cuRobo-enabled image (:curobo tag). Compiles CUDA extensions, so it adds build time.
+      - name: Build & push cuRobo image to NGC
+        run: |
+          chmod +x ./docker/push_to_ngc* || true
+          ./docker/push_to_ngc.sh -c -p
```
alexmillane, line 398: Out of interest, how do we envision this working in the end? Will Arena depend on cuRobo, or will it not?

I feel like we need to make a decision there one way or another. It makes things a bit more complicated if we have an optional dependency, as we do now.

Is there a reason we're making the dependency optional at the moment?

docker/Dockerfile.isaaclab_arena
```diff
@@ -155,6 +126,25 @@ COPY docs ${WORKDIR}/docs
...
+      export PATH=${CUDA_HOME}/bin:${PATH} && \
+      export LD_LIBRARY_PATH=${CUDA_HOME}/lib64:${LD_LIBRARY_PATH:-} && \
+      export TORCH_CUDA_ARCH_LIST="8.0+PTX;8.6+PTX;8.9+PTX;9.0+PTX" && \
+      /isaac-sim/python.sh -m pip install --no-build-isolation \
+        "nvidia-curobo @ git+https://github.com/NVlabs/curobo.git@ebb71702f3f70e767f40fd8e050674af0288abe8" && \
+      echo 'export CUDA_HOME=/usr/local/cuda-12.8' > /etc/profile.d/cuda.sh && \
+      echo 'export PATH=${CUDA_HOME}/bin:${PATH}' >> /etc/profile.d/cuda.sh && \
+      echo 'export LD_LIBRARY_PATH=${CUDA_HOME}/lib64:${LD_LIBRARY_PATH:-}' >> /etc/profile.d/cuda.sh; \
+    else echo "Skipping cuRobo installation"; fi && \
+    rm -f /tmp/install_cuda_curobo.sh
```
alexmillane: Just to raise a flag here, I think this might make our `pip` installation more complex. Because we now have a dependency on CUDA.

Not sure what to do about that though.

I guess we just have to proceed for now and work it out later.

isaaclab_arena_curobo/curobo_planner_utils.py
alexmillane: Agreed. I feel that each embodiment should have an optional method for getting files needed for curobo.

For embodiments that don't implement these functions, IK validation is not available.

isaaclab_arena_curobo/curobo_planner_utils.py
```diff
@@ -0,0 +1,226 @@
...
+_ROBOT_CFG_TEMPLATE = f"{_ROBOT_ASSET_DIR}/franka_robotiq_2f_85_zero_curobo.yml"
```
alexmillane: Out of interest, what is in this file?

isaaclab_arena_curobo/curobo_planner_utils.py
```diff
@@ -0,0 +1,226 @@
...
+_ROBOT_URDF = f"{_ROBOT_ASSET_DIR}/urdf/franka_robotiq_2f_85_zero.urdf"
+
+
+def pose_from_pos_quat(pos_xyz: torch.Tensor, quat_wxyz: torch.Tensor) -> torch.Tensor:
+    """Build a 4x4 homogeneous transform from a position and a (w, x, y, z) quaternion."""
+    rot = math_utils.matrix_from_quat(quat_wxyz.unsqueeze(0))[0]
+    return math_utils.make_pose(pos_xyz, rot)
```
alexmillane: Does this function exist elsewhere in the codebase? Seems like quite a basic function to live here.

One option would be to add a method to our `Pose` class to return a transformation matrix?

isaaclab_arena_curobo/curobo_planner_utils.py
```diff
@@ -0,0 +1,226 @@
...
+# Top-down grasp orientation (gripper pointing -Z) expressed in the robot base frame, (w, x, y, z).
+DOWN_FACING_QUAT_WXYZ = torch.tensor([0.0, 1.0, 0.0, 0.0], dtype=torch.float32)
+
+_ROBOT_ASSET_DIR = f"{ISAACLAB_NUCLEUS_DIR}/Arena/assets/robot_library/droid/droid_fixed_mimic_joint"
+_ROBOT_CFG_TEMPLATE = f"{_ROBOT_ASSET_DIR}/franka_robotiq_2f_85_zero_curobo.yml"
+_ROBOT_URDF = f"{_ROBOT_ASSET_DIR}/urdf/franka_robotiq_2f_85_zero.urdf"
+
+
+def pose_from_pos_quat(pos_xyz: torch.Tensor, quat_wxyz: torch.Tensor) -> torch.Tensor:
+    """Build a 4x4 homogeneous transform from a position and a (w, x, y, z) quaternion."""
```
alexmillane: Just double checking: `(w, x, y, z)` is intended? Elsewhere in Arena we have `(x, y, z, w)`. Probably curobo uses `(w, x, y, z)`

isaaclab_arena_curobo/curobo_planner_utils.py
```diff
@@ -0,0 +1,226 @@
...
+        retreat_distance=retreat_distance,
+        time_dilation_factor=time_dilation_factor,
+        collision_activation_distance=0.05,
+        motion_step_size=None,
+        trajopt_tsteps=42,
+        visualize_plan=False,
+        visualize_spheres=False,
+        debug_planner=debug_planner,
+        world_ignore_substrings=None,
+    )
```
alexmillane, line 88: Hmmm, this is quite DROID specific. I feel we may need a cuRobo adaptor per embodiment.

That could live in its own file (and maybe even a registry), separating it from the generic (embodiment un-specific tools).

docker/Dockerfile.isaaclab_arena
xyao-nv: done

isaaclab_arena_curobo/curobo_planner_utils.py
alexmillane, line 88: Could we store the non-robot related parameters as defaults somewhere.

isaaclab_arena_curobo/curobo_planner_utils.py
```diff
@@ -0,0 +1,226 @@
...
+        # USD prim path per scene object, used as the key into cuRobo's collision world model.
+        object_mappings = self._get_object_mappings()
+        world_model = self.motion_gen.world_coll_checker.world_model
+        rigid_objects = self.env.scene.rigid_objects
+        # Robot base pose in world frame (warp arrays -> torch before indexing in this Newton build).
+        robot_pos_w = wp.to_torch(self.robot.data.root_pos_w)[self.env_id, :3]
+        robot_quat_w = wp.to_torch(self.robot.data.root_quat_w)[self.env_id, :4]
+        # World->robot rotation: transpose of the base rotation matrix; quat inverse for orientations.
+        r_w2r = math_utils.matrix_from_quat(robot_quat_w.unsqueeze(0))[0].T
+        robot_quat_inv = math_utils.quat_inv(robot_quat_w.unsqueeze(0))[0]
```
alexmillane: Suggestion to have these rotations follow the frame transforming convention above.

I believe this would be `q_r2w`

isaaclab_arena_curobo/curobo_planner_utils.py
```diff
@@ -0,0 +1,226 @@
...
+    def _sync_robot_base_frame(self):
+        # USD prim path per scene object, used as the key into cuRobo's collision world model.
+        object_mappings = self._get_object_mappings()
+        world_model = self.motion_gen.world_coll_checker.world_model
+        rigid_objects = self.env.scene.rigid_objects
+        # Robot base pose in world frame (warp arrays -> torch before indexing in this Newton build).
+        robot_pos_w = wp.to_torch(self.robot.data.root_pos_w)[self.env_id, :3]
+        robot_quat_w = wp.to_torch(self.robot.data.root_quat_w)[self.env_id, :4]
+        # World->robot rotation: transpose of the base rotation matrix; quat inverse for orientations.
+        r_w2r = math_utils.matrix_from_quat(robot_quat_w.unsqueeze(0))[0].T
```
alexmillane: elsewhere we have rotation matrices at capital `R`

isaaclab_arena_curobo/curobo_planner_utils.py
```diff
@@ -0,0 +1,226 @@
...
+        robot_quat_inv = math_utils.quat_inv(robot_quat_w.unsqueeze(0))[0]
+        updated_count = 0
+        for object_name, object_path in object_mappings.items():
+            # Skip objects flagged static in the planner config (e.g. the table) — they aren't re-synced.
+            static_objects = getattr(self.config, "static_objects", [])
+            if object_name in rigid_objects and not any(s in object_name.lower() for s in static_objects):
+                obj = rigid_objects[object_name]
+                # Object pose in world frame (warp -> torch).
+                obj_pos_w = wp.to_torch(obj.data.root_pos_w)[self.env_id, :3]
+                obj_quat_w = wp.to_torch(obj.data.root_quat_w)[self.env_id, :4]
```
alexmillane: Suggestion that we adopt a uniform convention for dealing with transformations. Suggestion to check out my function here: https://github.com/isaac-sim/IsaacLab-Arena/blob/main/isaaclab_arena/variations/camera_extrinsics_variation.py. This is the convention we used at ASL.

`q_B_A` takes a vector in frame `A` to frame `B` such that `t_B = q_B_A * t_A`

isaaclab_arena_curobo/tests/test_ik_reachability.py
```diff
@@ -0,0 +1,143 @@
...
+    pose = Pose(position_xyz=tuple(float(v) for v in xyz)).to_tensor(device=env.device).unsqueeze(0)
+    pose[0, :3] += env.scene.env_origins[0, :]
+    env.scene[name].write_root_pose_to_sim(pose, env_ids=env_ids)
+
+
+def _run_reachability_check(args_cli) -> bool:
+    """Place one object in-reach and one out-of-reach; return True iff IK agrees with that split."""
+    import torch
+
+    import warp as wp
```
alexmillane: do we need these imports local?

isaaclab_arena_curobo/tests/test_ik_reachability.py
```diff
@@ -0,0 +1,143 @@
...
+
+    from isaaclab_arena_curobo.curobo_planner_utils import (
+        make_curobo_planner_for_droid,
+        top_down_grasp_pose_in_robot_frame,
+    )
+    from isaaclab_arena_curobo.ik_utils import check_ik_feasibility_batch_goal_poses
```
alexmillane, line 211: Do we need these imports to be local?

isaaclab_arena_curobo/scripts/run_placement_pool_ik_validation.py
```diff
@@ -0,0 +1,98 @@
...
+    add_ik_validation_arguments(args_parser)
+    args_cli, _ = args_parser.parse_known_args()
+
+    with SimulationAppContext(args_cli):
+        from isaaclab_arena_curobo.curobo_planner_utils import make_curobo_planner_for_droid
+        from isaaclab_arena_curobo.placement_pool_ik_validation import print_ik_validation_results, validate_pool_ik
+        from isaaclab_arena_environments.cli import (
+            get_arena_builder_from_cli,
+            get_isaaclab_arena_environments_cli_parser,
+        )
```
alexmillane: Reason to have these imports local?

alexmillane (approved):
A few nits. Looks acceptable for trying some stuff out.

I feel we need to make some decisions long term about:
- **dependency on curobo:** In the long term I think we need to go either for taking a dependency or not. An optional dependency doesn't make sense IMO. We need to decide if our workflows will use it or not.
- How we structure the embodiment-specific curobo stuff

.github/workflows/ci.yml
qianl-nv: I agree with Alex here. Is the +10min additional build time for docker the main concern, or we have not yet validate if adding curobo breaks anything else? If the +10min build time only, I would vote for taking the hit and making the dependency default.

isaaclab_arena_curobo/curobo_planner_utils.py
xyao-nv: good point, updated

isaaclab_arena_curobo/curobo_planner_utils.py
xyao-nv: done

.github/workflows/ci.yml
xyao-nv: 10+min built time + 10Gb docker image increase.  Not all envs need to be checked by reachability and for eval purpose, it does not require curobo too.
At the point of this MR, I don't think it needs to be the default base docker.

docker/Dockerfile.isaaclab_arena
xyao-nv: Curobo uses cuda-libs to speed up their algo. We can work it out later.

xyao-nv: I think it's because of our base image does not have cuda installed as default, and we have to handle that with pip.

docker/Dockerfile.isaaclab_arena
xyao-nv: no reason, I can keep the name consistent

isaaclab_arena_curobo/curobo_planner_utils.py
xyao-nv: renamed

isaaclab_arena_curobo/curobo_planner_utils.py
xyao-nv: I'm more than happy to adopt that broadly.

isaaclab_arena_curobo/ik_utils.py
```diff
@@ -0,0 +1,142 @@
...
+        dof = ik_result.solution.shape[-1]
+        joint_solution = ik_result.solution.view(-1, dof)[best_idx].detach().clone()
+
+    planner.logger.debug(
+        f"IK feasibility: feasible={feasible}, pos_err={best_pos_err:.4f}m, rot_err={best_rot_err:.4f}rad"
+    )
+    return feasible, best_pos_err, best_rot_err, joint_solution
+
+
+def check_ik_feasibility_batch_goal_poses(
```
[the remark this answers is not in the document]
xyao-nv: this needs to be seeded. agree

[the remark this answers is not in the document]
xyao-nv: defer to later MR as it's complex

isaaclab_arena_curobo/curobo_embodiment_cfg.py
```diff
@@ -0,0 +1,51 @@
...
+    gripper_open_joint_pos: dict[str, float]
+    """Gripper joint targets for the open pose, merged over the robot config's locked joints."""
+
+    gripper_closed_joint_pos: dict[str, float]
+    """Gripper joint targets for the closed pose, merged over the robot config's locked joints."""
+
+    hand_link_names: list[str]
+    """Links forming the hand, excluded from self-collision against the grasped object."""
+
+    grasp_gripper_open_val: float = 10.0
```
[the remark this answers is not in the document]
xyao-nv: it's robot specific params for curobo

isaaclab_arena_curobo/tests/test_ik_reachability.py
xyao-nv: nah updated

isaaclab_arena_curobo/tests/test_ik_reachability.py
xyao-nv: nah updated

isaaclab_arena_curobo/scripts/run_placement_pool_ik_validation.py
xyao-nv: nah updated

.github/workflows/ci.yml
alexmillane: Personally I feed that we the basic install of Arena should work for all Arena environments.

Do you know if cuRobo relies on the cuda runtime (`libcudart.so`), or the compiler toolchain (`nvcc`).

If it's the former, that is actually inside the docker, inside the isaac-sim install, so we should be able to use that.

If it needs `nvcc` we need to do some thinking. It appears there is now a pip wheel for the compiler toolchain: https://pypi.org/project/nvidia-cuda-nvcc/#history. Ideally we'd depend on that form of CUDA, because then we can still make Arena pip installable.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -49,6 +49,8 @@ def __init__(
         self.mimic_env: Any | None = None
         self.xr: Any | None = None
         self.termination_cfg: Any | None = None
+        # Optional cuRobo description for motion planning.
+        self.curobo_config: Any | None = None
```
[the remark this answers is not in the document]
xyao-nv: updated

isaaclab_arena_curobo/curobo_planner_utils.py
xyao-nv: good catch. Update all of them

isaaclab_arena_curobo/curobo_planner_utils.py
xyao-nv: I like the idea of registry. Added it

isaaclab_arena_curobo/curobo_planner_utils.py
xyao-nv: done

isaaclab_arena_curobo/ik_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    Args:
+        planner: ``CuroboPlanner`` instance.
+
+    Returns:
+        Joint position tensor of shape ``(1, dof)``.
+    """
+    return planner._get_current_joint_state_for_curobo().position.detach().clone()
+
+
+def check_ik_feasibility_per_goal_pose(
```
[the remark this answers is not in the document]
xyao-nv: removed, done

isaaclab_arena_curobo/curobo_planner_utils.py
xyao-nv: yes done

isaaclab_arena_curobo/curobo_planner_utils.py
xyao-nv: done

.github/workflows/ci.yml
xyao-nv: Curobo needs nvcc at buildtime, and uses libcurdart.so at run-time.

According to claude, nvidiac-cuda-nvcc pip wheel is not enough as it requires a host c++ complier. And curobo does not have pre-built pip wheel.

isaaclab_arena_curobo/ik_utils.py
xyao-nv: right now, the grasp pose will trigger a collision with the robot itself.
If it's unreachable because of joint limit, it will reflect in the residual error.

isaaclab_arena_curobo/ik_utils.py
```diff
@@ -0,0 +1,142 @@
...
+Kept outside the IsaacLab submodule so changes can be committed directly
+to IsaacLab-Arena without forking the upstream planner.
+"""
+
+from __future__ import annotations
+
+import os
+import torch
+
+import isaaclab.utils.math as PoseUtils
```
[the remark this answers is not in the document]
xyao-nv: done

isaaclab_arena_curobo/ik_utils.py
```diff
@@ -0,0 +1,142 @@
...
+
+# cuRobo captures a CUDA graph on the IK solver's first solve and, by default, errors when a later
+# solve changes the "goal type". In this use case, the planner warms up with a single-goal
+# solve, then we issue a batched ``solve_batch``. On CUDA >= 12 cuRobo can reset the graph instead of
+# erroring, but only when this env var opts in. ``setdefault`` keeps any explicit user override.
+os.environ.setdefault("CUROBO_TORCH_CUDA_GRAPH_RESET", "1")
+
+
+def check_ik_feasibility_per_goal_pose(
```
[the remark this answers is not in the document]
xyao-nv: rm

isaaclab_arena/utils/pose.py
```diff
@@ -7,6 +7,8 @@
 import torch
 from dataclasses import dataclass
 
+import isaaclab.utils.math as math_utils
```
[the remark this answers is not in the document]
xyao-nv: udpated

isaaclab_arena/utils/pose.py
```diff
@@ -52,6 +54,11 @@ def to_tensor(self, device: torch.device) -> torch.Tensor:
     def multiply(self, other: "Pose") -> "Pose":
         return compose_poses(self, other)
 
+    def make_pose(self, device: torch.device) -> torch.Tensor:
```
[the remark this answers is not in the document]
xyao-nv: done

PR #843, opened 2026-06-30
Add DreamZero remote policy

TuurStuyck:
## Summary
Add support for DeamZero. DreamZero runs on Osmo while policy_runner runs locally with access to DreamZero through port forwarding. 

## Detailed description

1. Submit the OSMO job

osmo workflow submit isaaclab_arena_dreamzero/docker/dreamzero_inference_server.yaml --pool isaac-dev-h100-01

2. Wait for the server to be ready

3. Port-forward to your local machine

osmo workflow port-forward <workflow_id> serve --port 5000

This tunnels localhost:5000 to the server inside the OSMO task. Leave this terminal open while running the policy.

4. Run the policy_runner

/isaac-sim/python.sh isaaclab_arena/evaluation/policy_runner.py --policy_type isaaclab_arena_dreamzero.policy.dreamzero_remote_policy.DreamZeroRemotePolicy --enable_cameras --num_episodes 5 --viz kit --dreamzero_host localhost --dreamzero_port 5000 --dreamzero_num_arm_joints 7 --dreamzero_open_loop_horizon 24 --dreamzero_cam_exterior_left external_camera_rgb --dreamzero_cam_wrist wrist_camera_rgb --language_instruction 'Pick up the cube and place it in the bowl' pick_and_place_maple_table --embodiment droid_abs_joint_pos

https://github.com/user-attachments/assets/06c27539-0528-4ea0-8712-f65b17d1c441

PR #846, opened 2026-07-01
Add native uv installation

rafaelcathomen:
## Summary

Add an optional native `uv` installation path alongside Docker.

`uv sync` creates an isolated Python 3.12 environment containing Arena, Isaac Lab 3.0.0b2, Isaac Sim 6.0.0.1, and the matching PyTorch/CUDA stack from public wheels.

## Changes

- Migrate package metadata from `setup.py` to PEP 621 in `pyproject.toml`.
- Add an Arena-owned `uv.lock` with strict index priority and one explicit PyPI route for an NVIDIA index collision.
- Keep native simulator dependencies in a PEP 735 group so Docker continues using its pinned Isaac Lab submodule.
- Add compatibility fixes for public-wheel and Docker Isaac Lab API differences.
- Document installation and add lightweight lock and editable-package CI checks.

## Validation

- Resolved the final 381-package lock from scratch without resolver cache; offline lock check, sync check, and `uv pip check` pass.
- Passed repository-wide pre-commit and the editable-package/resource guard.
- At `60f6b0d9c`, completed a native 20-step `cube_goal_pose` policy rollout with report output and clean shutdown.
- Passed the real-MP4 camera-recorder test with the locked MoviePy 1.0.3 and FFmpeg.
- Full simulator installation is intentionally not repeated in CI due to its approximately 27 GB footprint.

## Scope

Related to #327, but focused on the current Isaac Sim 6 / Python 3.12 public-wheel workflow. Conda, PyCharm configuration, and editable Isaac Lab source installation remain separate potential follow-ups.


isaaclab_arena/variations/camera_extrinsics_variation.py
```diff
@@ -142,6 +142,10 @@ def __call__(
             # method claims to return w,x,y,z, but it actually returns x,y,z,w.
             # I am testing this in the test test_isaaclab_bug_get_local_poses.py.
             t_parent_C, q_parent_C_xyzw = view.get_local_poses()
+            if not isinstance(t_parent_C, torch.Tensor):
```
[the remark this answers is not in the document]
rafaelcathomen: Checked the current branch: there is only one production conversion site, while both test copies intentionally inspect the raw upstream return type. Routing those tests through a shared implementation helper would couple the regression coverage to the shim, so I am keeping the small conversions local. I also confirmed that `ARENA_NV_API_KEY` is configured and injected as `NV_API_KEY` in trusted Arena test jobs; the skip only applies to local or fork runs without credentials.

.github/workflows/ci.yml
```diff
@@ -95,6 +95,54 @@ jobs:
...
+          uv lock --check --offline --no-python-downloads
+          uv venv --python 3.12 /tmp/arena-package-venv
+          uv pip install --python /tmp/arena-package-venv/bin/python --no-deps --editable .
+          cd /tmp
+          /tmp/arena-package-venv/bin/python - <<'PY'
+          import importlib.util
+          from importlib import metadata
+          from pathlib import Path
+
+          package_roots = (
```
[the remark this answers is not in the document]
rafaelcathomen: I am keeping this as an explicit, independent test oracle. Deriving the expected roots from the same pyproject.toml package configuration would make the completeness check self-referential and would not catch a package accidentally omitted there. The short duplication is intentional; the leaf-path variable is now named discoverable_modules to state what that part checks.

isaaclab_arena/variations/camera_extrinsics_variation.py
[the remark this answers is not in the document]
rafaelcathomen: This duplicates the earlier thread: there is one production conversion site, while the two test copies deliberately inspect the raw upstream return type. Routing those tests through a shared implementation helper would weaken the regression coverage, so the small conversions remain local.

.github/workflows/ci.yml
```diff
@@ -95,6 +95,54 @@ jobs:
...
+          for runtime_resource in runtime_resources:
+              assert (repo_root / runtime_resource).is_file(), runtime_resource
+
+          version = metadata.version("isaaclab_arena")
+          print(f"editable package check passed: {version}")
+          PY
+
   test:
     name: Run tests
     runs-on: [self-hosted, gpu-arena]
```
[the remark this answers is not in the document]
rafaelcathomen: This CI step is intentionally a committed-lock consistency gate, not an upstream-release monitor. An online uv lock --check also does not perform an upgrade or provide the proposed drift guarantee; that would require a separate scheduled relock policy. The final manifest was additionally resolved online from scratch without cache during validation.

pyproject.toml
```diff
@@ -1,3 +1,175 @@
+[build-system]
+requires = ["setuptools>=61"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "isaaclab_arena"
+version = "1.0.0"
+description = "Isaac Lab - Arena. An Isaac Lab extension for robotic policy evaluation."
+readme = "README.md"
+requires-python = ">=3.12,<3.13"
```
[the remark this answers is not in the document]
rafaelcathomen: Confirmed with direct evidence: /isaac-sim/python.sh is Python 3.12.12, and exact-head CI installed this checkout and passed the main, subprocess, and GR00T jobs. This duplicates the earlier interpreter question.

isaaclab_arena/variations/camera_extrinsics_variation.py
```diff
@@ -163,4 +167,5 @@ def __call__(
         t_parent_Cnew_in_parent = self._t_parent_C_in_parent[env_ids] + t_C_Cnew_in_parent
 
         # Apply the the sim.
-        view.set_local_poses(translations=t_parent_Cnew_in_parent, orientations=None, indices=env_ids.tolist())
+        view_indices = env_ids.detach().to(device="cpu", dtype=torch.int32).contiguous()
```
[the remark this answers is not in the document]
rafaelcathomen: Confirmed on both supported stacks: the public wheel requires CPU int32 tensor indices, and the pinned Docker submodule accepts the same representation. The conversion explicitly records CPU, int32, and contiguous layout. This is an optional prose improvement rather than a behavioral fix, so it is deferred after exact-head CI.

pyproject.toml
```diff
@@ -1,3 +1,175 @@
...
+#     prescribes `pin-pink==3.1.0 daqp==0.8.5`).
+#   * pyglet: the viewer renders on the stable 2.x line; a 3.0.dev build appeared
+#     upstream and would otherwise be picked. (rerun-sdk is intentionally left
+#     alone: stable 0.33 needs psutil>=7, which isaacsim 6.0 pins to 5.9.8.)
+#   * tinyobjloader: isaacsim-core requires this exact RC transitively; making it
+#     explicit lets the narrow prerelease policy admit it during a fresh lock.
+#   * mujoco-usd-converter: the NVIDIA index contains only incompatible 0.1.0
+#     prereleases, so this exact requirement is routed to PyPI below.
+constraint-dependencies = [
+    "pin-pink==3.1.0",
```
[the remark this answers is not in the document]
rafaelcathomen: Confirmed and accepted as an intentional trust boundary. The NVIDIA index supplies 61 package names in this stack, and uv source selection neither cascades through transitive dependencies nor supports a namespace wildcard, so making it explicit would require a fragile manual map. Under strict first-index, compatible collisions select the trusted NVIDIA index by design and incompatible collisions fail loudly; the sole known incompatible collision is explicitly routed to PyPI. The committed lock pins current sources and hashes, and future relocks require provenance review.

isaaclab_arena/variations/camera_extrinsics_variation.py
```diff
@@ -142,6 +142,10 @@ def __call__(
             # method claims to return w,x,y,z, but it actually returns x,y,z,w.
             # I am testing this in the test test_isaaclab_bug_get_local_poses.py.
             t_parent_C, q_parent_C_xyzw = view.get_local_poses()
+            if not isinstance(t_parent_C, torch.Tensor):
+                t_parent_C = t_parent_C.torch
+            if not isinstance(q_parent_C_xyzw, torch.Tensor):
+                q_parent_C_xyzw = q_parent_C_xyzw.torch
```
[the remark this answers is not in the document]
rafaelcathomen: The two supported API shapes are a torch Tensor and the public-wheel Warp ProxyArray, whose torch property is used here. An unexpected third type already fails immediately with a type-specific AttributeError. A separate guard would improve only a hypothetical diagnostic and is not required for the pinned compatibility contract.

pyproject.toml
```diff
@@ -1,3 +1,175 @@
...
+]
+namespaces = false
+
+# ----------------------------------------------------------------------------
+# Existing tool configuration (migrated unchanged from the previous pyproject)
+# ----------------------------------------------------------------------------
+
 [tool.isort]
 
 py_version = 310
```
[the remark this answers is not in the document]
rafaelcathomen: This is pre-existing tool configuration rather than runtime metadata. isort, Pyright, and pyupgrade still target the older floor; changing only isort here would leave the tooling configuration internally inconsistent. A coordinated tooling-floor update is better handled separately.

pyproject.toml
```diff
@@ -1,3 +1,175 @@
...
+readme = "README.md"
+requires-python = ">=3.12,<3.13"
+license = { text = "Apache-2.0" }
+dependencies = [
+    "typing_extensions",
+    "onnxruntime",
+    "vuer[all]",
+    "lightwheel-sdk",
+    "pytest",
+    "pydantic>=2.0",
```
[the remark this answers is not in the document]
rafaelcathomen: Keeping this unchanged is intentional for this migration. pytest was already a runtime dependency in setup.py, and the documented bare uv sync flow promises that uv run pytest works without extra flags. Moving it only to the dev extra would make that workflow rely on a transitive dependency or require a broader dependency-group redesign, which belongs in a separate packaging change.

isaaclab_arena/variations/camera_extrinsics_variation.py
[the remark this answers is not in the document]
rafaelcathomen: There is one production conversion site. The other occurrences are regression tests that intentionally inspect the raw upstream return type; sharing the production helper with those tests would couple the oracle to the implementation and weaken coverage, so the local conversions are intentional.

pyproject.toml
```diff
@@ -1,3 +1,164 @@
+[build-system]
+requires = ["setuptools>=61"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "isaaclab_arena"
+version = "1.0.0"
+description = "Isaac Lab - Arena. An Isaac Lab extension for robotic policy evaluation."
+readme = "README.md"
+requires-python = ">=3.12,<3.13"
```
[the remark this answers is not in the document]
rafaelcathomen: Confirmed: /isaac-sim/python.sh in the current CI image is Python 3.12.12. Exact-head run 28574693615 installed the PEP 621 checkout through that interpreter and all Docker-backed jobs passed, so the narrowed range is compatible with the supported Docker path.

.github/workflows/ci.yml
```diff
@@ -95,6 +95,54 @@ jobs:
...
+              "isaaclab_arena_environments/eval_jobs_configs/zero_action_jobs_config.json",
+              "isaaclab_arena_g1/g1_env/config/loco_manip_g1_joints_order_43dof.yaml",
+              "isaaclab_arena_gr00t/policy/config/droid_manip_gr00t_closedloop_config.yaml",
+          )
+          for runtime_resource in runtime_resources:
+              assert (repo_root / runtime_resource).is_file(), runtime_resource
+
+          version = metadata.version("isaaclab_arena")
+          print(f"editable package check passed: {version}")
+          PY
```
alexmillane, line 163: Am I corrent that this test only checks that the package is installable, but not that the install package works.

Should we also run the tests?

I would also move this testing out of the `pre-commit` job and into its own job.

docs/pages/quickstart/installation.rst
```diff
@@ -1,17 +1,87 @@
...
 Hardware requirements for Isaac Lab Arena are shared with Isaac Sim, and are detailed in
 `Isaac Sim Requirements <https://docs.isaacsim.omniverse.nvidia.com/6.0.0/installation/requirements.html>`_.
 
 
+Native uv developer setup
+-------------------------
+
+Isaac Lab Arena can be installed natively with `uv <https://docs.astral.sh/uv/>`_
+against the public Isaac Lab and Isaac Sim wheels -- no Docker, no submodule
+initialization, and no ``ISAACLAB_PATH`` required.
```
alexmillane: ~~no Docker, no submodule initialization, and no ``ISAACLAB_PATH`` required.~~

docs/pages/quickstart/installation.rst
```diff
@@ -1,17 +1,87 @@
...
+.. code-block:: bash
+
+    uv run python isaaclab_arena/evaluation/policy_runner.py \
+      --headless --policy_type zero_action --num_steps 20 \
+      --output_base_dir "$PWD/output" cube_goal_pose
+
+.. note::
+   The policy and evaluation runners retain the Docker-compatible
+   ``/eval/output`` default. Pass ``--output_base_dir`` with a writable path
+   when running them natively.
```
alexmillane: We should change the default output path such that it works for both docker and native installs.

docs/pages/quickstart/installation.rst
```diff
@@ -1,17 +1,87 @@
...
+
+.. note::
+   The third, subprocess-based phase (``-m with_subprocess``) runs the full
+   evaluation pipeline on the bundled example environments. Those tests expect
+   the Docker eval mounts (e.g. ``/eval``) and exercise contact-sensor success
+   detection that currently depends on the Docker Isaac Lab build, so run that
+   phase from the Docker workflow below.
```
alexmillane: I don't think we have any changes to the contact sensors in our version of lab.

If contact sensors don't work in the uv-installed version of Arena... That's a bit of a bad sign. Many things wont work without that.

I think this needs more investigation.

isaaclab_arena/tests/test_object_set.py
```diff
@@ -17,6 +17,15 @@
 OBJECT_SET_BOTTLES_PRIM_PATH = "/World/envs/env_.*/ObjectSet_Bottles"
 
 
+def _get_current_stage():
+    """Return Isaac Lab's ``get_current_stage`` across Isaac Sim versions."""
+    try:
+        from isaaclab.sim.utils.stage import get_current_stage
+    except ImportError:
+        from isaacsim.core.utils.stage import get_current_stage
+    return get_current_stage()
```
alexmillane, line 31: Am I correct that this is a compatibility layer that accounts for different versions of isaac sim in the docker and in the uv install?

I'm thinking we should try to align those two versions, rather than dealing with two versions on the arena side.

isaaclab_arena/variations/camera_extrinsics_variation.py
alexmillane, line 151: What's going on here?

I would guess that isaac lab be changed the return types of some of their functions between the version we have in the docker and the version we get in uv.

I think we should attempt to align these versions, rather than supporting both in Arena.

pyproject.toml
```diff
@@ -1,3 +1,175 @@
...
+[[tool.uv.index]]
+name = "pytorch-cu130"
+url = "https://download.pytorch.org/whl/cu130"
+explicit = true
+
+[tool.uv.sources]
+# torch / torchvision / torchaudio are pulled transitively by the isaaclab
+# wheel; we only steer them to the right CUDA index per platform. x86_64/AMD64
+# Linux + win32 -> cu128; aarch64/arm64 Linux -> cu130.
+mujoco-usd-converter = { index = "pypi" }
```
alexmillane, line 131: Right now Arena only support linux x86. Could we remove this section in that case?

README.md
```diff
@@ -66,12 +66,40 @@ Arena solves this by making environment variation a first-class concept. Swap an
...
+uv run pytest -q isaaclab_arena/tests/test_camera_observation.py::test_camera_observation
+```
+
+Launch a short native rollout with a writable output directory:
+
+```bash
+uv run python isaaclab_arena/evaluation/policy_runner.py \
+  --headless --policy_type zero_action --num_steps 20 \
+  --output_base_dir "$PWD/output" cube_goal_pose
+```
```
alexmillane: ~~Launch a short native rollout with a writable output directory:~~

-> Launch a short rollout

Let's change the default output directory such that it works for both docker and uv.

README.md
```diff
@@ -66,12 +66,40 @@ Arena solves this by making environment variation a first-class concept. Swap an
...
-Isaac Lab-Arena currently supports **installation from source inside a Docker container**.
+**Native developer setup with uv:**
+
+```bash
+git clone git@github.com:isaac-sim/IsaacLab-Arena.git
+cd IsaacLab-Arena
+uv sync
+```
+
+`uv sync` creates a Python 3.12 virtual environment in `.venv/`, installs Isaac Lab-Arena, and pulls `isaaclab[isaacsim,all]==3.0.0b2` with the matching Isaac Sim 6.0, PyTorch, and Newton wheels. No submodule init and no `ISAACLAB_PATH` are required. Accept the Isaac Sim EULA non-interactively and run the non-camera and camera smoke tests:
```
alexmillane: ~~No submodule init and no `ISAACLAB_PATH` are required.~~

Let's try to say what we do, rather than what we don't do.

docs/pages/quickstart/installation.rst
```diff
@@ -1,17 +1,87 @@
...
+   the native smoke tests and in-process test coverage work without it.
+   ``numba``-backed *upstream* Isaac Lab functionality is outside the validated
+   native-``uv`` scope -- use the Docker workflow for that.
+
+With ``isaaclab_arena`` installed you're ready to build your first environment;
+see :doc:`first_arena_env`.
+
+.. note::
+   The ``uv`` path is the native developer setup. The Docker workflow below is
+   still supported and is what CI and large-scale evaluation use today.
```
alexmillane: Remove.

docs/pages/quickstart/installation.rst
```diff
@@ -1,17 +1,87 @@
...
+    uv run pytest -q isaaclab_arena/tests/test_achieve_cube_goal_pose.py::test_achieve_cube_goal_pose_initial_state
+    uv run pytest -q isaaclab_arena/tests/test_camera_observation.py::test_camera_observation
+
+Launch a short zero-action rollout:
+
+.. code-block:: bash
+
+    uv run python isaaclab_arena/evaluation/policy_runner.py \
+      --headless --policy_type zero_action --num_steps 20 \
+      --output_base_dir "$PWD/output" cube_goal_pose
```
alexmillane: Let's align the validation between the docker and uv paths.

I like the validation you've added there. Maybe let's reflect that below in the docker workflow, and maybe add an optional step here for running the full test suite.

alexmillane (commented):
Thanks for doing this!!!

Looks like it's almost there.

I have a few minor comments. The major comment is that: There are a few places where we have shims to deal with the different versions of Isaac Lab in the docker and in the uv install. Ideally we wouldn't have that. Perhaps we should upgrade the version of isaaclab that we use in the docker to align with the public version? I think some of our tests don't pass for this reason (i.e. the subprocess requiring tests). Is that correct? I feel we need all the tests to pass. I'm interested in your opinion on how we can sort this out.

README.md
rafaelcathomen: Fixed — rephrased positively.

docs/pages/quickstart/installation.rst
rafaelcathomen: Fixed — now says what it does (the lockfile pins the complete environment).

README.md
rafaelcathomen: Done — flag dropped from the example; the default now works everywhere (see below).

docs/pages/quickstart/installation.rst
rafaelcathomen: Done — the runners now default to `./outputs` (gitignored, works natively and in Docker); `--output_base_dir` overrides, e.g. to the `/eval` mount.

docs/pages/quickstart/installation.rst
rafaelcathomen: Removed.

pyproject.toml
rafaelcathomen: Done — resolution is now locked to Linux x86_64 via `tool.uv.environments`; the per-platform markers and the cu130 index are gone (lockfile shrank by ~1900 lines).

docs/pages/quickstart/installation.rst
rafaelcathomen: Done — the native section now has the same optional test-phase verification as the Docker section (in-process phases; the subprocess phase stays Docker-side for now).

.github/workflows/ci.yml
rafaelcathomen: Correct — it validates the lock and editable packaging only. Moved into its own `uv_package` job. Running the sim suite needs the GPU runner plus the full Isaac Sim wheel install; that still runs in the Docker-backed `test` job — a native-uv test job is a good follow-up.

isaaclab_arena/tests/test_object_set.py
rafaelcathomen: Correct — the public wheel exposes it at `isaaclab.sim.utils.stage`, the Docker build only at `isaacsim.core.utils.stage`. Documented in the docstring and marked for removal on version convergence — agreed on aligning, discussing in the main thread.

isaaclab_arena/variations/camera_extrinsics_variation.py
rafaelcathomen: The public wheel returns warp ProxyArrays from `get_local_poses()` (`.torch` converts); the Docker build returns torch Tensors. Added a NOTE marking it for removal once the versions converge.

docs/pages/quickstart/installation.rst
rafaelcathomen: Agreed this needs a root cause before shipping — investigating whether it's the Isaac Lab build difference or the tests' Docker mount assumptions; will report on the main thread.

rafaelcathomen:
**Root cause of the native `with_subprocess` failures** — following up on the installation.rst and shim threads above.

The primary cause was the hardcoded Docker interpreter path (`submodules/IsaacLab/_isaac_sim/python.sh`) in the test constants: natively, every subprocess test failed at spawn before running any sim code. Fixed in 98492587d with a `sys.executable` fallback; the phase now runs natively.

With that fix the native result is **9 passed / 11 skipped** (482f3abc0 adds scoped skipif markers, active only outside Docker; the Docker run is unchanged and runs the full set). The 11 skips are environmental, in two buckets:

- **8 × `requires_docker_assets`** — the tests load `maple_table.usda` / `srl_robolab_assets`, which are not yet promoted to the public Nucleus (404 natively). Not fixable by version alignment; needs asset promotion.
- **3 × `requires_docker_isaaclab`** — genuine API differences between the Docker Isaac Lab build and the public 3.0.0b2 wheel: `InteractiveScene` rejecting `ArticulationCfg` (RL training env) and the contact-view `filter_count` API (mimic data generation). These are exactly the tests version convergence fixes; the markers carry a remove-on-convergence note.

**Contact sensors: exonerated.** They are not the cause of any native failure — contact-sensor usage in general (e.g. the pick-and-place success/termination path) works natively. The earlier installation.rst warning reflected the pre-fix failure mix and has been removed; the only contact-related seam is the mimic data-gen `filter_count` API difference above.

**On the dual-version shims:** agreed that aligning the Docker and uv Isaac Lab versions is the right end state; the shims are transitional and each is marked for removal on convergence. Note that alignment alone would clear only the 3 `requires_docker_isaaclab` skips — the 8 asset skips need Nucleus promotion regardless. Proposing to track version convergence as a follow-up rather than a blocker here, since the Docker path is unchanged by this PR.

rafaelcathomen:
**Pre-merge evidence at exact head 2598c941f** — both remaining verification gates are green:

1. **Full CI dispatch run** [28804317761](https://github.com/isaac-sim/IsaacLab-Arena/actions/runs/28804317761): all jobs pass — uv package, pre-commit, docs, GR00T closed-loop E2E, tests, and subprocess tests (the two test jobs run Docker-backed, so the Docker path is exercised end to end).

2. **No-cache Docker image rebuild + smoke**: rebuilt `isaaclab_arena:latest` with `--no-cache --pull` from a pristine checkout of this head. In the resulting image: `isaaclab==4.5.24` installed editable from `submodules/IsaacLab` (not the public wheel), `isaacsim` provided by the base image (not pip), `isaaclab_arena==1.0.0` editable, and `import isaaclab_arena, isaaclab` succeeds through `/isaac-sim/python.sh` — i.e. the setup.py → PEP 621 migration leaves the Docker install path unchanged.

Remaining gate is review sign-off; the four open threads are answered above and the root-cause summary is in the previous comment.

rafaelcathomen:
Rebased onto latest `main` (head is now a6ee8a8ba) to resolve conflicts from the recent merges (#864, #862/#863, #843 et al.). Two conflict resolutions, both unions: kept the `NV_API_KEY` skipif on top of the rewritten live-endpoint test in `test_environment_generation_agent.py`, and merged the import lists in `test_all_environments.py`. No packaging changes were needed — `main` didn't touch `setup.py`, so `pyproject.toml`/`uv.lock` carry over unchanged.

Post-rebase native verification: fresh `uv sync` from the lock is clean, and `pytest --collect-only isaaclab_arena/tests` collects 637 tests with no errors. CI is rerunning on the new head.

One observation for a follow-up (not this PR): two test modules outside the documented native scope now fail *collection* natively because `main`-side changes import Docker-only deps at module level — `isaaclab_arena_gr00t/policy/gr00t_remote_closedloop_policy.py:19` imports `gr00t` (submodule-only), and `test_review_gui.py` now pulls `streamlit` (dev extra) via `editor_panel`. Both fine in Docker; a `pytest.importorskip` guard would restore native collection of those dirs.

PR #851, opened 2026-07-02
Add typed Hydra environment configuration MVP

cvolkcvolk:
## Summary
Add typed Hydra experiment configuration

## Detailed description
- Move reusable Arena environment and experiment Cfg types into core and resolve concrete environment schemas through EnvironmentRegistry.
- Keep the YAML-driven Hydra runner as an example frontend with a required suite path and sequential experiment dispatch through eval-runner Jobs.
- Preserve the legacy argparse path while making the experiment-to-Job compatibility boundary explicit.
- Cover composition, registry resolution, invalid input, Job mapping, and simulator subprocess execution.


isaaclab_arena_environments/example_environment_base.py
cvolkcvolk, line 1: We should move this out of "example_environment_base" as it would essentially be our base environment contract (requiring a `.build()`)

PR #858, opened 2026-07-03
Add typed configurations to Arena environments

cvolkcvolk:
## Summary
Decouple environment construction from argparse

## Detailed description
- Move every first-party environment's construction inputs from untyped `argparse.Namespace` reads into environment-specific dataclass configs.
- Make `build(cfg)` the typed, frontend-independent construction boundary so Hydra and future dispatchers can compose environments without emulating CLI state.
- Keep `get_env(args_cli)` as a temporary compatibility adapter, preserving existing command-line workflows and behavior.

Follow up: Make `ExampleEnvironmentBase()` a core base class


isaaclab_arena_environments/example_environment_base.py
```diff
@@ -7,13 +7,17 @@
...
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Generic, TypeVar
+
+from isaaclab_arena.environments.arena_environment_cfg import ArenaEnvironmentCfg
 
 if TYPE_CHECKING:
     from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
 
+EnvironmentCfgT = TypeVar("EnvironmentCfgT", bound=ArenaEnvironmentCfg)
```
cvolkcvolk: Do we need this type config? Could we rename it to ArenaEnvironmentCfgT or something similar more descriptive?

isaaclab_arena_environments/pick_and_place_maple_table_environment.py
```diff
@@ -33,8 +68,8 @@ def get_env(self, args_cli: argparse.Namespace) -> IsaacLabArenaEnvironment:
 
         # Step 1: Retrieve assets from the registry
         background = self.asset_registry.get_asset_by_name("maple_table_robolab")()
-        pick_up_object = self.asset_registry.get_asset_by_name(args_cli.pick_up_object)()
-        destination_location = self.asset_registry.get_asset_by_name(args_cli.destination_location)()
+        pick_up_object = self.asset_registry.get_asset_by_name(cfg.pick_up_object_asset_name)()
```
cvolkcvolk: Let's continue using the previous naming of the variables. Don't rename in this PR just yet. Eg stick with pick_up_object, destination_location, additional_table_objects, embodiment etc in this file

isaaclab_arena_environments/pick_and_place_maple_table_environment.py
```diff
@@ -92,6 +127,10 @@ def _set_viewer_cfg(env_cfg):
             task=task,
             env_cfg_callback=_set_viewer_cfg,
         )
+        if cfg.teleoperation_device_name is not None:
```
cvolkcvolk: This addition is not needed if it wasn't in the original environment. Let's only add this block if we actually need it.

isaaclab_arena_environments/pick_and_place_maple_table_environment.py
```diff
@@ -6,21 +6,56 @@
...
+@dataclass
+class PickAndPlaceMapleTableEnvironmentCfg(ArenaEnvironmentCfg):
+    """Configure the Maple-table pick-and-place environment."""
+
+    name: str = "pick_and_place_maple_table"
+    embodiment_asset_name: str = "droid_abs_joint_pos"
+    teleoperation_device_name: str | None = None
+    high_dynamic_range_image_name: str | None = None
+    light_intensity: float = 500.0
+    pick_up_object_asset_name: str = "rubiks_cube_hot3d_robolab"
```
cvolkcvolk: See other comment about the naming of those variables.

cvolkcvolk: Done

isaaclab_arena_environments/pick_and_place_maple_table_environment.py
cvolkcvolk: Done

isaaclab_arena_environments/example_environment_base.py
cvolkcvolk: Done

cvolkcvolk: Renamed the TypeVar to `ArenaEnvironmentCfgT` so its scope and purpose are explicit.

isaaclab_arena_environments/pick_and_place_maple_table_environment.py
cvolkcvolk: Restored the existing environment argument names throughout this file (`pick_up_object`, `destination_location`, `additional_table_objects`, `embodiment`, and related fields).

isaaclab_arena_environments/pick_and_place_maple_table_environment.py
cvolkcvolk: Removed the teleoperation-device assignment so this refactor preserves the original environment behavior.

isaaclab_arena_environments/pick_and_place_maple_table_environment.py
cvolkcvolk: Updated the dataclass fields to retain the existing argument names in this PR.

isaaclab_arena_environments/pick_and_place_maple_table_environment.py
```diff
@@ -6,21 +6,54 @@
...
     from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
 
 
+@dataclass
+class PickAndPlaceMapleTableEnvironmentCfg(ArenaEnvironmentCfg):
+    """Configure the Maple-table pick-and-place environment."""
+
+    enable_cameras: bool = False
+    embodiment: str = "droid_abs_joint_pos"
+    teleop_device: str | None = None
```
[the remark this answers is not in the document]
cvolkcvolk: Updated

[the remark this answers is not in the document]
cvolkcvolk: updated

isaaclab_arena_environments/cube_goal_pose_environment.py
alexmillane: Could we generate the flags from a cfg object automatically? Maybe in a function in the base class?

Then we can ensure the (loose) contract here that the flags match the config struct? Wdyt?

isaaclab_arena_environments/example_environment_base.py
```diff
@@ -7,13 +7,19 @@
...
 if TYPE_CHECKING:
     from isaaclab_arena.environments.isaaclab_arena_environment import IsaacLabArenaEnvironment
 
+ArenaEnvironmentCfgT = TypeVar("ArenaEnvironmentCfgT", bound=ArenaEnvironmentCfg)
+
 
-class ExampleEnvironmentBase(ABC):
+# TODO(cvolk, 2026-07-03): Co-locate ArenaEnvironmentCfg and this base in the core
+# arena_environment_factory module as ArenaEnvironmentFactoryBase.
+class ExampleEnvironmentBase(ABC, Generic[ArenaEnvironmentCfgT]):
```
alexmillane, line 84: cool!

isaaclab_arena_environments/example_environment_base.py
```diff
@@ -26,8 +32,14 @@ def __init__(self):
 
     @abstractmethod
     def get_env(self, args_cli: argparse.Namespace) -> IsaacLabArenaEnvironment:
+        # TODO(cvolk, 2026-07-03): Deprecate this legacy argparse entry point; build(cfg) will
+        # become the primary environment construction API.
         pass
 
+    def build(self, cfg: ArenaEnvironmentCfgT) -> IsaacLabArenaEnvironment:
+        """Build an Arena environment from its typed configuration."""
+        raise NotImplementedError(f"{type(self).__name__} does not support typed environment configuration")
```
alexmillane, line 118: Out of interest, why not enforce this as above? Potentially we have some legacy environments hanging around?

alexmillane (approved):
LGTM. Configuring this way makes a lot more sense.

My only substantial comment is whether we could generate the argparse from the config class difinition? That would eliminate one (loose) contract for during the transition.

isaaclab_arena_environments/example_environment_base.py
cvolkcvolk: Introduced and made abstract/ mandatory in #860. Should land today.

isaaclab_arena_environments/cube_goal_pose_environment.py
cvolkcvolk: Good suggestion. Implementated that the base class now generates the flags from the config dataclass and uses the same fields to translate the arguments back into a config (with some few intentional exceptions for arguments owned by shared runners). Hope to get rid of them within the next days though

PR #862, opened 2026-07-03
Add typed configurations to Arena policies

cvolkcvolk:
## Summary
Decouple policy construction from argparse

## Detailed description
- Add a generic `PolicyCfg` contract and limit `PolicyBase` to typed runtime configs.
- Rename first-party policy configurations consistently with the `Cfg` convention.
- Preserve the existing CLI through the shared dataclass-to-argparse compatibility adapter.
- Continue the broader migration from argparse namespaces to strongly typed configurations.

### DIFF
--- a/isaaclab_arena/assets/register.py
+++ b/isaaclab_arena/assets/register.py
@@ -62,9 +62,9 @@ def _register_policy(cls: type["PolicyBase"], cfg_type: type["PolicyCfg"] | None
 def register_policy(policy_type: type["PolicyBase"] | None = None, *, cfg_type: type["PolicyCfg"] | None = None):
     """Register a policy and its typed configuration."""
     if cfg_type is None:
-        # TODO(cvolk, 2026-07-03): After policy_runner and eval_runner stop using policy-owned
-        # argparse adapters, remove bare @register_policy support and this warning branch. Make
-        # cfg_type required and keyword-only, and remove the optional policy_type argument.
+        # TODO(cvolk, 2026-07-03): After the policy_runner and eval_runner compatibility
+        # fallbacks for untyped downstream policies are removed, require cfg_type and delete
+        # the optional policy_type argument and this warning branch.
         assert policy_type is not None, "Typed policy registration requires cfg_type"
         warnings.warn(
             "Bare @register_policy is deprecated; use @register_policy(cfg_type=PolicyCfgType)",
--- a/isaaclab_arena/assets/register.py
+++ b/isaaclab_arena/assets/register.py
@@ -3,6 +3,9 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import warnings
+from typing import TYPE_CHECKING
+
 from isaaclab_arena.assets.registries import (
     AssetRegistry,
     DeviceRegistry,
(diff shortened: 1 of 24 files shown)


isaaclab_arena/assets/register.py
```diff
@@ -44,15 +50,37 @@ def register_retargeter(cls):
...
+            stacklevel=2,
+        )
+        return _register_policy(policy_type, None)
+
+    assert policy_type is None, "Pass the policy config with the cfg_type keyword"
+
+    def decorator(cls):
+        return _register_policy(cls, cfg_type)
+
+    return decorator
```
alexmillane: Nit: This is a bit hard to follow what's going on. Could we:

- Get rid of the first return? I think that is we fall through `cfg_type` is None, so the second return still works no?
- If we can do that, we can delete `_register_policy` by just putting it inline in `decorator`

I think that will improve the readability here.

isaaclab_arena/cli/dataclass_cli.py
```diff
@@ -0,0 +1,114 @@
...
+    elif get_origin(cli_value_type) is Literal:
+        choices = get_args(cli_value_type)
+        choice_types = {type(choice) for choice in choices}
+        assert len(choice_types) == 1, f"{config_field.name} Literal choices must have one value type"
+        argument_options.update(type=choice_types.pop(), choices=choices)
+    elif cli_value_type is bool:
+        argument_options["action"] = "store_true" if default is False else argparse.BooleanOptionalAction
+    else:
+        argument_options["type"] = cli_value_type
+    return argument_options
```
alexmillane: Mega nit: But I feel that internal functions should go at the end of the file. That way when someone opens the file they see the public functions, i.e. the functions that they should be looking at, first.

isaaclab_arena/evaluation/policy_runner_cli.py
```diff
@@ -4,6 +4,82 @@
...
+        _add_policy_cfg_arguments(parser, policy_cfg_type)
+        return parser
+
+    # TODO(cvolk, 2026-07-03): Delete this fallback when downstream policies register PolicyCfg.
+    warnings.warn(
+        f"Policy {policy_type.__name__} uses deprecated policy-owned argparse flags; register a PolicyCfg",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    return policy_type.add_args_to_parser(parser)
```
alexmillane: Do we still have policies in the repo that don't use the Cfg method? To me is looks as if they all use the Cfg?

alexmillane (approved):
LGTM!

One question I have is, in a few places you state that there are compatibility branches to policies that don't use config classes. To me it looks as if everything has been switched over. What still remains?

isaaclab_arena/assets/register.py
cvolkcvolk: Actually given that all Policies are accepting the new decorator `@register_policy() ` I think we can drop this lecacy branch in here altogether already

isaaclab_arena/evaluation/policy_runner_cli.py
cvolkcvolk: Good point. That makes me wonder wether we need the legacy support in the decorator at all. I think we can already drop it

isaaclab_arena/cli/dataclass_cli.py
cvolkcvolk: Done

PR #863, opened 2026-07-03
Use typed config for ArenaEnvBuilder

cvolkcvolk:
## Summary
Type ArenaEnvBuilder configuration

## Detailed description
- Decouple `ArenaEnvBuilder` from argparse so any frontend can construct it from typed data.
- Add `ArenaEnvBuilderCfg` and update existing callers, documentation, and tests.
- Keep the current CLI working through a dated compatibility adapter that can be removed with the argparse route.

PR #867, opened 2026-07-06
Add typed Arena experiment runner

cvolkcvolk:
## Summary
Run evaluations from typed experiment configs. Add `ArenaExperimentCfg` and `RolloutCfg` so an experiment describes its environment, policy, and rollout without making execution depend on an argparse namespace.

## Detailed description
- Keep current JSON files working by converting each legacy evaluation config into concrete ArenaExperimentCfg instances with `experiment_cfgs_from_legacy_eval_config()`.
- Register each environment config with its factory so `build_and_run_experiment()` can run an `ArenaExperimentCfg` directly, without a temporary callback wrapper.
- Keep graph-based environments on the existing argparse path until they have typed configs.

Part of a broader move away from passing `argparse.Namespace` objects through Arena’s runtime. Environments, policies, builders, and evaluations will consume source-independent typed configs, allowing CLI, YAML/Hydra, and Python callers to share the same execution path.

Depends on #863


isaaclab_arena/evaluation/experiment_runner.py
```diff
@@ -0,0 +1,208 @@
...
+
+    from isaaclab_arena.environments.arena_env_builder import ArenaEnvBuilder
+    from isaaclab_arena.environments.arena_environment_factory import ArenaEnvironmentFactory
+    from isaaclab_arena.metrics.metric_data import MetricsDataCollection
+    from isaaclab_arena.policy.policy_base import PolicyBase, PolicyCfg
+
+ArenaBuilderFactory = Callable[[ArenaExperimentCfg], "ArenaEnvBuilder"]
+
+
+def build_registered_arena_builder(
```
cvolkcvolk: This class seems to be only used in a test

isaaclab_arena/evaluation/experiment_runner.py
```diff
@@ -0,0 +1,208 @@
...
+    return ArenaEnvBuilder(
+        arena_environment,
+        cfg.environment_builder,
+        hydra_overrides=hydra_overrides_from_nested_dict(cfg.variations),
+    )
+
+
+def run_experiment(
+    cfg: ArenaExperimentCfg,
+    *,
```
cvolkcvolk: Do we need a wildcard here or could we just pass named arguments to improve readability?

isaaclab_arena/evaluation/experiment_execution.py
```diff
@@ -0,0 +1,208 @@
...
+        cfg.environment_builder,
+        hydra_overrides=hydra_overrides_from_nested_dict(cfg.variations),
+    )
+
+
+def run_experiment(
+    cfg: ArenaExperimentCfg,
+    *,
+    output_dir: str | Path,
+    arena_builder_factory: ArenaBuilderFactory,
```
cvolkcvolk: run_experiment() sounds like an experiment with its environment and configs was already constructed before that stage. Could we rename or factor the "experiment building" step out of actually running an experiment?

isaaclab_arena/evaluation/experiment_runner.py
```diff
@@ -0,0 +1,208 @@
...
+        return [None] * num_rebuilds
+    assert num_episodes >= num_rebuilds, (
+        f"Experiment '{experiment_name}': num_episodes ({num_episodes}) must be >= num_rebuilds "
+        f"({num_rebuilds}) so each rebuild runs at least one episode"
+    )
+    base, remainder = divmod(num_episodes, num_rebuilds)
+    return [base + int(rebuild_index < remainder) for rebuild_index in range(num_rebuilds)]
+
+
+def _aggregate_metrics(metrics_per_rebuild: list[MetricsDataCollection]) -> MetricsDataCollection | None:
```
cvolkcvolk: Do we need a function for this one liner?

isaaclab_arena/evaluation/experiment_runner.py
```diff
@@ -0,0 +1,208 @@
...
+    """Aggregate available rebuild metrics, or return ``None`` when a task defines none."""
+    return aggregate_metrics(metrics_per_rebuild) if metrics_per_rebuild else None
+
+
+def _close_policy(policy: PolicyBase | None) -> None:
```
cvolkcvolk: Could we move out these helper that close and free resources into a seperate file?

isaaclab_arena/utils/hydra_overrides.py
```diff
@@ -29,3 +31,21 @@ def assert_hydra_overrides(args: list[str], parser: argparse.ArgumentParser) ->
             bad.append(arg)
     if bad:
         parser.error(f"Unrecognized arguments: {' '.join(bad)}")
+
+
+def hydra_overrides_from_nested_dict(values: dict[str, Any]) -> list[str]:
```
cvolkcvolk: Why do we need this function? I think we have not introduced any hydra machinery yet

isaaclab_arena/assets/register.py
```diff
@@ -109,9 +107,6 @@ def agent_ready(cls):
     return cls
 
 
-# TODO(cvolk, 2026-07-06): Remove this resolver and the PolicyRegistry config-type
```
cvolkcvolk, line 112: We only removed the comment, but not the "resolver and the metadata". Intentional?

isaaclab_arena/assets/register.py
```diff
@@ -129,3 +125,24 @@ def _policy_cfg_type_from_policy(policy_type: type["PolicyBase"]) -> type["Polic
         cfg_type, PolicyCfg
     ), f"{policy_type.__name__} must use a concrete PolicyCfg subclass"
     return cfg_type
+
+
+def _environment_cfg_type_from_factory(
+    factory_type: type["ArenaEnvironmentFactory"],
+) -> type["ArenaEnvironmentCfg"]:
+    """Return the concrete config declared by ``ArenaEnvironmentFactory[Cfg]``."""
+    from isaaclab_arena.environments.arena_environment_factory import ArenaEnvironmentCfg, ArenaEnvironmentFactory
```
cvolkcvolk: Let's move all imports at the top of the files

isaaclab_arena/assets/register.py
```diff
@@ -129,3 +125,24 @@ def _policy_cfg_type_from_policy(policy_type: type["PolicyBase"]) -> type["Polic
         cfg_type, PolicyCfg
     ), f"{policy_type.__name__} must use a concrete PolicyCfg subclass"
     return cfg_type
+
+
+def _environment_cfg_type_from_factory(
```
cvolkcvolk, line 130: Let us make this function implementation more descriptive

isaaclab_arena/assets/registries.py
```diff
@@ -257,10 +265,41 @@ def get_random_hdr_by_tag(self, tag: str) -> type["HDRImage"]:
...
+    def get_environment_cfg_type(
+        self,
+        factory_type: type["ArenaEnvironmentFactory"],
+    ) -> type["ArenaEnvironmentCfg"]:
+        """Get the concrete config consumed by a registered environment factory."""
+        ensure_assets_registered()
+        assert factory_type in self._cfg_types, f"Environment {factory_type.__name__} must register a config"
+        return self._cfg_types[factory_type]
+
+    def get_factory_type_for_cfg(self, cfg: "ArenaEnvironmentCfg") -> type["ArenaEnvironmentFactory"]:
```
cvolkcvolk, line 305: Is this implementation needed once we move away from the legacy argparse namespace to fully typed config dicts? If so, lets add a note to remove.

isaaclab_arena/assets/registries.py
```diff
@@ -257,10 +265,41 @@ def get_random_hdr_by_tag(self, tag: str) -> type["HDRImage"]:
...
+    ) -> None:
+        """Register an environment factory and the config it consumes."""
+        assert (
+            cfg_type not in self._factory_types_by_cfg_type
+        ), f"Environment config {cfg_type.__name__} already registered"
+        self.register(factory_type, factory_type.name)
+        self._cfg_types[factory_type] = cfg_type
+        self._factory_types_by_cfg_type[cfg_type] = factory_type
+
+    def get_environment_cfg_type(
```
cvolkcvolk, line 292: Is this implementation needed once we move away from the legacy argparse namespace to fully typed config dicts? If so, lets add a note to remove.

isaaclab_arena/evaluation/experiment_execution.py
```diff
@@ -0,0 +1,157 @@
...
+) -> gym.Env:
+    """Compile and instantiate an experiment's environment."""
+    arena_builder = _build_arena_builder(cfg)
+    _, env_cfg, env_kwargs = arena_builder.build_registered()
+    if env_cfg.recorders is not None:
+        env_cfg.recorders.dataset_filename = f"dataset_{cfg.name}"
+    return arena_builder.make_registered(env_cfg, env_kwargs, render_mode=render_mode)
+
+
+def _build_arena_builder(cfg: ArenaExperimentCfg) -> ArenaEnvBuilder:
```
cvolkcvolk: Wdyt to make the transition clear, could we rename all current adapter functions to make them more descriptive to append eihter _from_cfg or _from_legacy (or _from_namespace)?

isaaclab_arena/evaluation/run_execution.py
```diff
@@ -0,0 +1,157 @@
...
+    arena_builder = _build_arena_builder(cfg)
+    _, env_cfg, env_kwargs = arena_builder.build_registered()
+    if env_cfg.recorders is not None:
+        env_cfg.recorders.dataset_filename = f"dataset_{cfg.name}"
+    return arena_builder.make_registered(env_cfg, env_kwargs, render_mode=render_mode)
+
+
+def _build_arena_builder(cfg: ArenaExperimentCfg) -> ArenaEnvBuilder:
+    """Build an Arena environment builder from the experiment's concrete config."""
+    from isaaclab_arena.environments.arena_env_builder import ArenaEnvBuilder
```
cvolkcvolk, line 112: Could we move imports to the top of the file?

isaaclab_arena/variations/variations_hydra.py
```diff
@@ -19,6 +20,23 @@
     from isaaclab_arena.variations.variation_base import VariationBase
 
 
+def overrides_from_dict(values: dict[str, Any]) -> list[str]:
```
cvolkcvolk, line 23: Moved out of job_manager.py

isaaclab_arena/assets/registries.py
```diff
@@ -257,10 +265,50 @@ def get_random_hdr_by_tag(self, tag: str) -> type["HDRImage"]:
 
 
 class EnvironmentRegistry(Registry):
-    """Registry for example environment classes."""
+    """Registry for Arena environment factories and their concrete configs."""
```
alexmillane: Registry for Arena environment factories and their ~~concrete~~ configs

isaaclab_arena/assets/registries.py
```diff
@@ -257,10 +265,50 @@ def get_random_hdr_by_tag(self, tag: str) -> type["HDRImage"]:
...
+        assert (
+            factory_type in self._cfg_types_by_factory_type
+        ), f"Environment {factory_type.__name__} must register a config"
+        return self._cfg_types_by_factory_type[factory_type]
+
+    def get_factory_type_for_cfg(self, cfg: "ArenaEnvironmentCfg") -> type["ArenaEnvironmentFactory"]:
+        """Resolve the registered factory used to execute a typed environment config.
+
+        This reverse lookup is part of the typed runtime path and remains after the
+        legacy argparse and JSON adapters are removed.
```
alexmillane: Maybe convert to a note like the other notes about the temporary stuff so we can go through and remove them.

Perhaps it would be good to have a key in all the notes on this topic so we can grep the code-base in a week when we're cleaning all this up.

isaaclab_arena/evaluation/arena_run.py
```diff
@@ -0,0 +1,93 @@
...
+    """Variation values applied when the environment is compiled."""
+
+    def __post_init__(self) -> None:
+        assert self.name, "experiment name must not be empty"
+        assert self.num_rebuilds > 0, "num_rebuilds must be greater than zero"
+        if self.rollout.num_episodes is not None:
+            assert self.rollout.num_episodes >= self.num_rebuilds, (
+                f"Experiment '{self.name}': num_episodes ({self.rollout.num_episodes}) must be >= num_rebuilds "
+                f"({self.num_rebuilds}) so each rebuild runs at least one episode"
+            )
```
alexmillane, line 72: This describes a single environment in the json file.

To me the json describes an experiment. The experiment has multiple environments over which it runs, however it is a single experiment. I would tend to call the json file an experiment config (or better yet just an experiment file). The question then is what to call a single environment. What about a `Trial` or a `Run`? Wdyt? An experiment would then be a composition of several `Trial`s or `Run`s?

isaaclab_arena/evaluation/arena_run.py
```diff
@@ -0,0 +1,93 @@
...
+
+    num_episodes: int | None = None
+    """Number of completed episodes, or ``None`` for a step-driven rollout."""
+
+    def __post_init__(self) -> None:
+        assert not (
+            self.num_steps is not None and self.num_episodes is not None
+        ), "num_steps and num_episodes are mutually exclusive"
+        assert self.num_steps is None or self.num_steps > 0, "num_steps must be greater than zero"
+        assert self.num_episodes is None or self.num_episodes > 0, "num_episodes must be greater than zero"
```
alexmillane, line 37: This describes the duration only right?

What about calling this a `RunDurationCfg`?

See comment below.

alexmillane (approved):
LGTM.

Only comment is on what we call the experiments. Let's discuss that.

isaaclab_arena/evaluation/arena_run.py
cvolkcvolk: Agree, renamed the current `ArenaExperimentCfg` to `ArenaRunCfg`. An experiment is now composed of multiple runs (eg a full json file). 
“Trial” sounds more like one individual attempt. Lets go with "Run"

isaaclab_arena/evaluation/arena_run.py
cvolkcvolk: Good point. I would however actaully prefer `RolloutLimitCfg` over `RunDurationCfg` because the values are step/episode limits rather than elapsed duration (for now)

isaaclab_arena/assets/registries.py
cvolkcvolk: Done

isaaclab_arena/assets/registries.py
cvolkcvolk: Done

PR #874, opened 2026-07-08
Narrow config registry loading boundaries

cvolkcvolk:
## Summary
Separate registry loading by ownership

## Detailed description
- Registry lookups previously used one REGISTRIES list to call ensure_assets_registered(). Asking PolicyRegistry for config types therefore also imported unrelated assets, tasks, embodiments, and pxr before SimulationApp started.
- Replace that central list with a loading hook owned by each registry. Asset, device, retargeter, HDR, relation, and task registries keep the existing shared asset-registration behavior.
- PolicyRegistry imports only the built-in policies, so zero_action, replay, and rsl_rl still work without manual imports. EnvironmentRegistry does no lazy loading because extension packages explicitly register their environment factories.
- Typed Experiment composition can now discover policy and environment config types before simulation starts. Regression tests cover both narrowed lookup paths.
- Other components, such as tasks and embodiments, still cannot be discovered before SimulationApp because their shared loader imports pxr. A TODO records that follow-up.

### DIFF
--- a/isaaclab_arena/assets/registries.py
+++ b/isaaclab_arena/assets/registries.py
@@ -275,8 +275,8 @@ class HDRImageRegistry(_GloballyLoadedRegistry):
 class EnvironmentRegistry(Registry):
     """Store factories and configs registered by environment extension packages.
 
-    This registry has no lazy loader. First-party callers populate it through
-    ensure_environments_registered() before lookup.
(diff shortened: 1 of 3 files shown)


cvolkcvolk:
Closed. We now defer discovery and composition until after startup

PR #875, opened 2026-07-08
Add predicate library with object settling check

peterd-NV:
## Summary
Adds a small predicate library with spatial predicates and a special object settling predicate.

Spatial predicates:

- Moves existing termination functions (objects on destination) to the new spatial predicate library
- Adds additional spatial predicates for object moved and object lifted

Object settling:

- Adds an object settling predicate that returns True when objects settle and fall below a threshold velocity after env startup.
- Add feature to log settled object positions for other predicates to optionally use (e.g. object_lifted can use it to check if the object is lifted above it's initial height)
- Adds object settling reset to ProgressTracker to handle resetting object initial position logs during reset. When object settling predicate is unused, this event is a passthrough. No effect on existing envs or tasks. 

Task updates:

- Updates imports on existing tasks (sorting_task, assembly_task) to use the object_on_destination function from the new predicate library.
- Adds progress tracking to pick and place task using new predicates (settled, lifted, on_destination).

Other changes:

- Adds "progress" key to recorder tests. Progress tracking is now enabled in pick place task so it contributes an additional key to the recorder.


isaaclab_arena/tasks/predicates/object_settling.py
```diff
@@ -0,0 +1,99 @@
...
+
+def objects_settled(
+    env,
+    object_names: list[str],
+    velocity_threshold: float = 1e-3,
+    env_id: int | None = None,
+) -> torch.Tensor:
+    """True per env when every object in the env is at rest, records each object's position on first settle."""
+    for object_name in object_names:
+        assert object_name in get_env(env).scene.keys(), f"Object '{object_name}' not found in scene"
```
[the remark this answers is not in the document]
peterd-NV: Removed assert entirely. Downstream already raises an informative error immediately when non-existent scene object is looked up. This check wasn't doing anything extra and just adding bloat.

isaaclab_arena/tasks/predicates/spatial.py
```diff
@@ -0,0 +1,157 @@
...
+    velocity_threshold: float = 0.5,
+) -> torch.Tensor:
+    """ Checks if an object is in contact with it's detination location via a contact sensor.
+
+    Returns True when the object is in contact with destination above a force threshold 
+    and below a velocity threshold.
+    """
+
+    object: RigidObject = env.unwrapped.scene[object_cfg.name]
+    sensor: ContactSensor = env.unwrapped.scene[contact_sensor_cfg.name]
```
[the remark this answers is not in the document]
peterd-NV: done

isaaclab_arena/tasks/predicates/spatial.py
```diff
@@ -0,0 +1,157 @@
...
+    for object_cfg, contact_sensor_cfg in zip(object_cfg_list, contact_sensor_cfg_list):
+        single_condition = object_on_destination(
+            env=env,
+            object_cfg=object_cfg,
+            contact_sensor_cfg=contact_sensor_cfg,
+            force_threshold=force_threshold,
+            velocity_threshold=velocity_threshold,
+        )
+        condition_met = torch.logical_and(condition_met, single_condition)
+    return condition_met
```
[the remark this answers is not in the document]
peterd-NV: done

isaaclab_arena/tasks/predicates/object_settling.py
```diff
@@ -0,0 +1,99 @@
...
+
+from isaaclab_arena.tasks.predicates.predicate_utils import get_env, get_root_lin_vel_w, get_root_pos_w, select
+
+_SETTLED_OBJ_POS_ATTR = "_settled_object_positions"
+
+
+def objects_settled(
+    env,
+    object_names: list[str],
+    velocity_threshold: float = 1e-3,
```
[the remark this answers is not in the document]
peterd-NV: added

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -253,6 +254,7 @@ def compose_manager_cfg(self) -> tuple[IsaacLabArenaManagerBasedRLEnvCfg, dict[s
             placement_event_cfg,
             variations_event_cfg,
             progress_tracking_events_cfg,
+            make_object_settling_reset_event_cfg(),
```
cvolkcvolk: Mh I feel like this smells a bit that we would need to add the implementation detail of one progress milestone on the arena_env_builder level. It should not know about one particular predicate implementatio

isaaclab_arena/tasks/predicates/spatial.py
```diff
@@ -0,0 +1,165 @@
...
+    object_z = get_root_pos_w(env, object_name)[:, 2]
+    if use_settled_state:
+        settled_pos, has_settled = get_object_settled_state(env, object_name)
+        result = has_settled & (object_z > (settled_pos[:, 2] + distance))
+    else:
+        result = object_z > (surface_height + distance)
+    return select(result, env_id)
+
+
+def object_moved(
```
cvolkcvolk: Suggestion renaming to `object_moving`

isaaclab_arena/tasks/predicates/spatial.py
```diff
@@ -0,0 +1,165 @@
...
+    env: ManagerBasedRLEnv,
+    object_name: str,
+    velocity_threshold: float = 1e-2,
+    env_id: int | None = None,
+) -> torch.Tensor:
+    """Checks if an object has moved.
```
cvolkcvolk:
```suggestion
    """Checks if an object is moving.
```

isaaclab_arena/tasks/predicates/spatial.py
```diff
@@ -0,0 +1,165 @@
...
+from isaaclab.assets import RigidObject
+from isaaclab.envs import ManagerBasedRLEnv
+from isaaclab.managers import SceneEntityCfg
+from isaaclab.sensors.contact_sensor.contact_sensor import ContactSensor
+
+from isaaclab_arena.tasks.predicates.object_settling import get_object_settled_state
+from isaaclab_arena.tasks.predicates.predicate_utils import get_env, get_root_lin_vel_w, get_root_pos_w, select
+
+
+def object_lifted(
```
cvolkcvolk: Maybe we could give this a more descriptive name like `object_is_above_reference_height`?

isaaclab_arena/tasks/predicates/object_settling.py
```diff
@@ -0,0 +1,117 @@
...
+    """Record ``name``'s position for envs that just settled and weren't recorded yet."""
+
+    entry = _get_entry(env, name)
+    new = settled & ~entry["settled"]
+    if bool(new.any()):
+        entry["pos"] = torch.where(new.unsqueeze(-1), get_root_pos_w(env, name), entry["pos"])
+        entry["settled"] = entry["settled"] | new
+
+
+def reset_settled(env, env_ids=None) -> None:
```
cvolkcvolk: Could the recorded rest poses be owned by a resettable progress component instead of being stored as a private attribute on the environment?

I think if we avoid exposing `reset_settled()` as a predicate-specific global reset event we could remove the need of passing `make_object_settling_reset_event_cfg()` on the `arena_env_builder.py` level. 

The cached poses are currently used by progress tracking, and ProgressTracker already has an episode-reset lifecycle

ProgressTracker.reset() could maybe own the lifecycle like:

```
class ProgressTracker:
     def reset(self, env_ids) -> None:
      """Reset progress and objective-owned episode state."""
         for runner in self.runners:
             runner.reset(env_ids)

       for component in self.resettable_components:
           component.reset_environments(env_ids)
```

isaaclab_arena/tasks/predicates/object_settling.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def get_object_settled_state(env, name: str) -> tuple[torch.Tensor, torch.Tensor]:
+    """Return the settled state of an object.
+
+    The settled state is a tuple of two tensors, the settled pos and the settled mask.
+    The settled pos (num_envs, 3) is meaningful when the settled mask (num_envs,) is True.
+    """
+
+    entry = _get_entry(env, name)
```
cvolkcvolk: `_get_entry` appears to be a lookup, but it also creates the store, registers an object, allocates tensors, and mutates the environment. 
Could this be avoided if  we provide this state an explicit owner, such as `InitialObjectRestPoseRecorder`, initialized with the relevant object names, `num_envs`, and device. 
It could expose `record_once()`, `get_record()`, and `reset()` methods. The read path should not create state.

isaaclab_arena/environments/arena_env_builder.py
cvolkcvolk: Maybe there is a way around. See `reset_settled()`

isaaclab_arena/tasks/predicates/object_settling.py
```diff
@@ -0,0 +1,117 @@
...
+class ObjectSettlingResetEventCfg:
+    """Clears recorded settle data as envs reset."""
+
+    reset_settled: EventTermCfg = EventTermCfg(func=reset_settled, mode="reset")
+
+
+def make_object_settling_reset_event_cfg() -> ObjectSettlingResetEventCfg:
+    """Reset-event cfg that clears recorded settle data on env reset."""
+
+    return ObjectSettlingResetEventCfg()
```
cvolkcvolk: Nit, could we just inline this when needed?

isaaclab_arena/tasks/predicates/object_settling.py
peterd-NV: done, add explicit owner class for storing/resetting the state

isaaclab_arena/tasks/predicates/object_settling.py
peterd-NV: removed this explicit event cfg after moving the reset ownership to progress tracker

isaaclab_arena/tasks/predicates/object_settling.py
peterd-NV: done, moved object settling reset func to progress tracker and eliminated explicit reset event in arena env builder.

isaaclab_arena/tasks/predicates/spatial.py
peterd-NV: done

isaaclab_arena/tasks/predicates/spatial.py
peterd-NV: done

isaaclab_arena/tasks/predicates/spatial.py
peterd-NV: done

isaaclab_arena/environments/arena_env_builder.py
peterd-NV: removed from env builder

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -427,6 +428,7 @@ def progress_tracking_reset_func(env, env_ids, progress_objectives: list[Progres
     elif torch.is_tensor(env_ids):
         env_ids = env_ids.tolist()
     progress_tracker.reset(env_ids)
+    reset_rest_pose_recorder(env, env_ids)
```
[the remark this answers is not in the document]
peterd-NV: The coupling of the obj settling predicate and progress tracking is intentionally done here as obj settling is a feature used only for progress tracking and a "special case" predicate. Having the progress tracker take ownership of reset here eliminates the need for a dedicated event term in the arena env builder and is overall a better organization for the feature.

isaaclab_arena/tasks/predicates/object_settling.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def get_rest_pose_recorder(env) -> ObjectInitialRestPoseRecorder:
+    """Return the env's ``ObjectInitialRestPoseRecorder``, lazily creating and caching it on first use."""
+
+    env = get_env(env)
+    recorder = getattr(env, _INITIAL_REST_POSE_RECORDER_ATTR, None)
+    if recorder is None:
+        recorder = ObjectInitialRestPoseRecorder(env.num_envs, env.device)
+        setattr(env, _INITIAL_REST_POSE_RECORDER_ATTR, recorder)
```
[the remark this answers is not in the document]
peterd-NV: removed usage of get/set attr.

isaaclab_arena/tasks/predicates/object_settling.py
```diff
@@ -0,0 +1,129 @@
...
+
+from isaaclab_arena.tasks.predicates.predicate_utils import (
+    get_env,
+    get_root_ang_vel_w,
+    get_root_lin_vel_w,
+    get_root_pos_w,
+    select,
+)
+
+_recorders: WeakKeyDictionary[object, ObjectInitialRestPoseRecorder] = WeakKeyDictionary()
```
cvolkcvolk: Suggestion to rename to maybe _rest_pose_recorders_by_env?

alexmillane: Having static module-level objects like this is not great. The issue is that this guy is effectively a global variable, which causes all the usual issues with globals. For example, is this guy reset between rebuilds of new envs? Probably not.

I would recommend placing this guy within `IsaacLabArenaManagerBasedRLEnv`. That's pretty much how we capture state that needs to be exchanged through the program. For example the `VariationRecorder` sits in the `env` is written to by a bunch of events, and later read by the writer.

alexmillane (approved):
LGTM

isaaclab_arena/tasks/predicates/object_settling.py
peterd-NV: makes sense, I was trying to use weak key dict to handle the reset between new envs but it's better to do it this way. Moved and renamed recorder into `IsaacLabArenaManagerBasedRLEnv`

PR #873, opened 2026-07-08
Add typed YAML experiment composition

cvolkcvolk:
## Summary
Add typed YAML composition for Arena Experiments as the next step in replacing argparse namespaces and legacy JSON evaluation jobs with strongly typed configuration.

A follow-up connects this loader to `eval_runner.py` before the JSON path is deprecated and removed.

  ## Detailed description
  - Build on #871’s runtime model: an Arena Experiment is an ordered list of typed Runs
  - Transition experiment definitions to YAML because it fits the configuration hierarchy directly. Ports `getting_started_experiment.yaml`.
  - Use Hydra as the composition layer, rather than merely as a YAML parser. Hydra selects and constructs the concrete typed environment and policy configs, rejects unsupported.
  - Load each YAML Run into an ArenaRunCfg, composing its concrete environment and policy configurations through an encapsulated Hydra integration.
  - Keep eval_runner and legacy JSON handling unchanged in this PR

PR #876, opened 2026-07-09
Run typed YAML experiments from eval_runner.py

cvolkcvolk:
## Summary
Connect typed YAML Experiments to eval_runner while preserving legacy JSON evaluation.

## Detailed description
- Build on #873 by accepting typed YAML arena experiments through the `--experiment_config` option and forwarding Hydra overrides.
- Compose typed Runs only after AppLauncher starts, avoiding pre-start runtime registry and pxr imports without changing registry boundaries.
- Preserve legacy JSON behavior and the `--eval_jobs_config` alias until legacy json files are ported.

### DIFF
--- a/isaaclab_arena/evaluation/arena_experiment_config_loader.py
+++ b/isaaclab_arena/evaluation/arena_experiment_config_loader.py
@@ -10,13 +10,14 @@ from __future__ import annotations
 import json
 from dataclasses import replace
 from pathlib import Path
-from typing import TYPE_CHECKING
 
+from isaaclab_arena.assets.registries import EnvironmentRegistry, PolicyRegistry
+from isaaclab_arena.environments.arena_environment_factory import ArenaEnvironmentCfg
 from isaaclab_arena.evaluation.arena_experiment import ArenaExperiment
-
-if TYPE_CHECKING:
-    from isaaclab_arena.environments.arena_environment_factory import ArenaEnvironmentCfg
-    from isaaclab_arena.policy.policy_base import PolicyCfg
+from isaaclab_arena.evaluation.legacy_eval_config import run_cfgs_from_legacy_eval_config
+from isaaclab_arena.hydra.arena_experiment import load_arena_experiment_from_yaml
+from isaaclab_arena.policy.policy_base import PolicyCfg
+from isaaclab_arena_environments.cli import ensure_environments_registered
 
 
 def load_arena_experiment_from_config_file(
@@ -45,30 +46,31 @@ def load_arena_experiment_from_config_file(
     assert suffix in {".json", ".yaml", ".yml"}, f"Experiment config must use .json, .yaml, or .yml, got '{path}'"
 
     if suffix == ".json":
-        from isaaclab_arena.evaluation.legacy_eval_config import run_cfgs_from_legacy_eval_config
-
         assert not overrides, "Experiment overrides are supported only for typed YAML Experiments"
         with path.open(encoding="utf-8") as experiment_config_file:
             legacy_experiment_config = json.load(experiment_config_file)
-        experiment = run_cfgs_from_legacy_eval_config(legacy_experiment_config, device=device)
-    else:
-        from isaaclab_arena.hydra.arena_experiment import load_arena_experiment_from_yaml
+        return run_cfgs_from_legacy_eval_config(legacy_experiment_config, device=device)
 
-        experiment = load_arena_experiment_from_yaml(
-            path,
-            environment_cfg_types=_registered_environment_cfg_types(),
-            policy_cfg_types=_registered_policy_cfg_types(),
-            overrides=overrides,
+    yaml_experiment = load_arena_experiment_from_yaml(
+        path,
+        environment_cfg_types=_registered_environment_cfg_types(),
+        policy_cfg_types=_registered_policy_cfg_types(),
+        overrides=overrides,
+    )
+
+    experiment_with_process_device: ArenaExperiment = []
+    for run_config in yaml_experiment:
+        environment_builder_with_process_device = replace(run_config.environment_builder, device=device)
+        run_config_with_process_device = replace(
+            run_config,
+            environment_builder=environment_builder_with_process_device,
         )
-
-    return [replace(run, environment_builder=replace(run.environment_builder, device=device)) for run in experiment]
+        experiment_with_process_device.append(run_config_with_process_device)
+    return experiment_with_process_device
 
 
 def _registered_environment_cfg_types() -> dict[str, type[ArenaEnvironmentCfg]]:
     """Return registered environment selector names and their config types."""
-    from isaaclab_arena.assets.registries import EnvironmentRegistry
-    from isaaclab_arena_environments.cli import ensure_environments_registered
-
     ensure_environments_registered()
     registry = EnvironmentRegistry()
     return {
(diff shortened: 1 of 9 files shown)


isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -104,81 +111,123 @@ def _run_chunk(chunk_label: str, chunk_jobs: list[dict]) -> int:
...
-
-    # Load the experiment before starting simulation so process-wide requirements
-    # can be configured before Isaac Sim starts.
     add_eval_runner_arguments(args_parser)
     args_cli, _ = args_parser.parse_known_args()
     assert not args_cli.distributed, "Distributed evaluation is not supported yet"
+    return args_cli
+
+
+def _validate_experiment_config_path(experiment_config: str | Path) -> Path:
```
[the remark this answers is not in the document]
cvolkcvolk: Updated

isaaclab_arena/evaluation/arena_experiment_config_loader.py
```diff
@@ -0,0 +1,98 @@
...
+    experiment_config_path: str | Path,
+    *,
+    device: str,
+    overrides: list[str] | None = None,
+) -> ArenaExperiment:
+    """Load a JSON or YAML Arena Experiment and apply its process device.
+
+    Call this only after SimulationApp starts because resolving the complete
+    configuration registries loads Isaac Sim modules.
+
```
alexmillane: I feel we can drop this in the docstring. This comment applies to probably 75% of our functions.

isaaclab_arena/evaluation/arena_experiment_config_loader.py
```diff
@@ -0,0 +1,98 @@
...
+    return experiment_with_process_device
+
+
+def _registered_environment_cfg_types() -> dict[str, type[ArenaEnvironmentCfg]]:
+    """Return registered environment selector names and their config types."""
+    ensure_environments_registered()
+    registry = EnvironmentRegistry()
+    return {
+        name: registry.get_environment_cfg_type(registry.get_component_by_name(name))
+        for name in registry.get_all_keys()
```
alexmillane: nit: spit over a couple of lines.

isaaclab_arena/evaluation/arena_experiment_config_loader.py
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+
+def _registered_policy_cfg_types() -> dict[str, type[PolicyCfg]]:
+    """Return registered policy selector names and their config types."""
+    registry = PolicyRegistry()
+    return {
+        name: registry.get_policy_cfg_type(registry.get_component_by_name(name)) for name in registry.get_all_keys()
```
alexmillane: see comment above.

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -47,19 +51,28 @@ def legacy_json_experiment_requires_cameras(experiment_config: dict) -> bool:
     )
 
 
-def _run_chunk(chunk_label: str, chunk_jobs: list[dict]) -> int:
-    """Run ``chunk_jobs`` in a fresh ``eval_runner`` subprocess and return its exit code."""
+def _assert_camera_support_enabled(experiment: ArenaExperiment, enable_cameras: bool) -> None:
+    """Check that AppLauncher enabled camera support requested by typed Runs."""
+    camera_run_names = [run_cfg.name for run_cfg in experiment if getattr(run_cfg.environment, "enable_cameras", False)]
```
alexmillane, line 46: We have to do get_attr here because we basically leave it up to the user to optionally define `enable_cameras` in their child of `ArenaEnvironmentCfg`.

Wdyt about moving flags like `enable_cameras` into the parent to guaruntee that they exist for all environments. We could probably move some others like `teleop_device`, `mimic`, potentially `embodiment`.

Many of these are in all environments, but we basically are relying on the user to add them.

isaaclab_arena/evaluation/eval_runner.py
```diff
@@ -69,80 +82,110 @@ def _run_chunk(chunk_label: str, chunk_jobs: list[dict]) -> int:
...
+
+    Args:
+        experiment_config_path: Validated path to an Experiment configuration.
+        experiment_overrides: Hydra overrides supplied for a typed YAML Experiment.
+
+    Returns:
+        The legacy JSON mapping, or None when the Experiment uses YAML.
+    """
+    if experiment_config_path.suffix.lower() != ".json":
+        return None
```
alexmillane: I wonder if we should minimize the eval runner to just the run loop by moving everything out into another file.

Then the eval_runner will be minimal thing that just contains starting the sim, loading the experiment cfg (through functions) and running it. It could be beautiful. WDyt?

isaaclab_arena/evaluation/eval_runner_cli.py
```diff
@@ -5,41 +5,57 @@
...
+    parser.add_argument(
+        "--experiment_override",
+        action="append",
+        default=[],
+        metavar="KEY=VALUE",
+        help=(
+            "Hydra override for a typed YAML Experiment. Repeat for multiple overrides. "
+            "Also pass --enable_cameras when an override enables environment cameras."
+        ),
     )
```
alexmillane: So we move the hydra overrides related to the experiment in through a flag containing hydra tokens?

This is a bit uglier than the variations where the hydra tokens are picked up off the args in their native form (i.e. without needing `--experiment_override`).

Potentially that would be hard here? Because it would require asasembling all sources of hydra options into one place before the parse...

What's your thinking here. More than happy to merge this as is, just wondering what the thinking is.

isaaclab_arena/tests/test_eval_runner.py
```diff
@@ -46,6 +54,36 @@ def run_eval_runner(jobs_config_path: str, headless: bool = HEADLESS):
...
+""",
+        encoding="utf-8",
+    )
+
+    run_eval_runner(
+        str(experiment_config_path),
+        config_option="--experiment_config",
+        extra_args=[
+            "--experiment_override",
+            "runs.yaml_baseline.rollout_limit.num_steps=2",
```
alexmillane: We don't check we get num_steps actually being 2, or that the variable is overriden? Could we add a check for that given the slightly complex form of the experiment override flags?

alexmillane (approved):
LGTM!!!

isaaclab_arena/evaluation/arena_experiment_config_loader.py
cvolkcvolk: deleted

isaaclab_arena/evaluation/arena_experiment_config_loader.py
cvolkcvolk: Done

isaaclab_arena/evaluation/arena_experiment_config_loader.py
cvolkcvolk: Done

isaaclab_arena/evaluation/eval_runner.py
cvolkcvolk: Good point. I think there is yet another thing wrong about needing environment.enable_cameras=true and --enable_cameras. One for the env and one for the SimApp. I added a TODO to adress this and removing the getattr() in a follow up as it would go beyond the scope of this PR

isaaclab_arena/evaluation/eval_runner_cli.py
cvolkcvolk: The original idea was that overrides are clearly yaml based and easy to reject. And it minimized changes while adding the Hydra overrides... But I think doing this internally is worth it, keeping this layer as is

isaaclab_arena/evaluation/eval_runner.py
cvolkcvolk: Moved the legacy functions into their own file. We are getting there.

isaaclab_arena/tests/test_eval_runner.py
cvolkcvolk: Yea I replaced the  `--experiment_override` from above now with native Hydra override tokens, so the CLI is now simpler. And updated the subprocess test to assert here

PR #880, opened 2026-07-09
Relation solver places robot

qianl-nv:
## Summary
Let the relation solver place robot embodiments

## Detailed description
- Introduce a shared `PlacementEntity` protocol so the relation solver, object
  placer, and overlap checks operate on both scene objects and embodiments;
  `EmbodimentBase` and `Object` now implement it.
- Add USD-derived embodiment placement footprints (`embodiment_placement.py`)
  and wire Droid so its pose can be solved from spatial relations (e.g. `on`
  floor, `next_to` counter) instead of a hardcoded initial pose.
- Expand the relation solver / object placer to write solved poses back to the
  correct Isaac Lab scene entity (`robot` for embodiments), and expand graph
  conversion to attach embodiment relations.
- Defer the pxr-importing `usd_prim_tree` import in the pass-2 resolver so
  importing the generation agent no longer pulls in pxr before SimulationApp.
- Add `DummyEmbodiment`/`DummyObject` test doubles plus tests for embodiment
  bbox, graph conversion, and solver placement; add example graph YAMLs
  (`kitchen_droid_placement_only`, `kitchen_embodiment_placement`,
  `kitchen_pick_and_place`).


qianl-nv:
Split into https://github.com/isaac-sim/IsaacLab-Arena/pull/915 and https://github.com/isaac-sim/IsaacLab-Arena/pull/937

PR #871, opened 2026-07-07
Remove legacy evaluation job model

cvolkcvolk:
## Summary
Use typed Runs for experiment orchestration. Replace the CLI-oriented `Job` and `JobManager` layer with the typed Run configuration and result model introduced in #867.

## Detailed description
- Treat each JSON document as an ordered Experiment of Runs and execute them directly through build_and_run.
- Preserve the legacy JSON schema, CLI, chunking, camera preflight, reporting, and continue-on-error behavior.

PR #884, opened 2026-07-09
Address evaluation runner review feedback

cvolkcvolk:
## Summary
Address late review feedback from #871

## Detailed description
- Move Run table construction alongside the typed Run model while leaving printing to eval_runner.
- Make legacy camera detection a boolean helper that does not mutate the argparse namespace.
- Use Path for experiment output directories and flush failure diagnostics before a process exits.
- Follow up separately because #871 merged before these approved review changes could be pushed.


alexmillane (approved):
Thanks!

PR #885, opened 2026-07-09
Add light color temperature variation

alexmillane:
## Summary
Adds a light color variation.

## Detailed description
- Light color plugs into an existing Lab parameter for all lights.


isaaclab_arena/assets/object_library.py
```diff
@@ -389,6 +389,15 @@ def set_color(self, color: tuple[float, float, float]) -> None:
         self.spawner_cfg.color = tuple(color)
         self.object_cfg = self._init_object_cfg()
 
+    def set_color_temperature(self, color_temperature: float) -> None:
+        """Enable and set the light's white-point color temperature in Kelvin, in [1000, 10000]."""
```
cvolkcvolk: Is there a reason to constrain it to 1000, 10000?

cvolkcvolk (approved):
lgtm

isaaclab_arena/assets/object_library.py
alexmillane: That is taken from Isaac Lab. I'm not sure of the motivation for that constraint, but I've just copied it here.

PR #888, opened 2026-07-09
Add FaceTo in Relation Solver

zhx06:
## Summary
Add FaceTo object orientation relation

## Detailed description
- Enables objects to face another object after placement.
- Computes world Z-yaw, revalidates rotated bounds, and handles invalid directions.

### DIFF
--- a/isaaclab_arena/relations/object_placer.py
+++ b/isaaclab_arena/relations/object_placer.py
@@ -60,7 +60,7 @@ class PlacementCandidate:
     """Per-check validation results for this candidate's layout."""
 
     orientations: dict[ObjectBase, float] = field(default_factory=dict)
-    """Absolute world Z-yaw per oriented object. Empty when unrotated."""
+    """Placement-computed absolute world Z-yaws. Omitted objects retain their marker orientation."""
 
     @property
     def is_valid(self) -> bool:
@@ -125,12 +125,13 @@ class ObjectPlacer:
         )
         results_per_env = [env_results[0] for env_results in ranked_results_per_env]
 
-        for env_idx, result in enumerate(results_per_env):
-            if not result.success:
-                print(
-                    f"  env {env_idx}: no valid layout; using lowest-loss fallback "
-                    f"(failed: {result.validation_results.get_failed_validation_check_names})"
-                )
+        if self.params.verbose:
+            for env_idx, result in enumerate(results_per_env):
+                if not result.success:
+                    print(
+                        f"  env {env_idx}: no valid layout; using lowest-loss fallback "
+                        f"(failed: {result.validation_results.get_failed_validation_check_names})"
+                    )
 
         if self.params.apply_positions_to_objects:
             positions_per_env = [r.positions for r in results_per_env]
@@ -178,31 +179,9 @@ class ObjectPlacer:
                 f"Object '{obj.name}' has no relations. All objects passed to place() must have "
                 "at least one relation (e.g., On(), NextTo(), or IsAnchor())."
             )
-            face_to_relations = [relation for relation in obj.get_relations() if isinstance(relation, FaceTo)]
-            assert len(face_to_relations) <= 1, f"Object '{obj.name}' has more than one FaceTo relation."
-            if face_to_relations:
-                assert not obj.is_anchor, f"Anchor object '{obj.name}' cannot have a FaceTo relation."
-                assert face_to_relations[0].parent is not obj, f"Object '{obj.name}' cannot face itself."
-                assert (
-                    face_to_relations[0].parent in object_set
-                ), f"FaceTo parent '{face_to_relations[0].parent.name}' must participate in placement."
-                assert (
-                    self._get_rotate_around_solution(obj) is None
-                ), f"Object '{obj.name}' cannot combine FaceTo with RotateAroundSolution."
-                subject_randomization = self._get_random_around_solution(obj)
-                if subject_randomization is not None:
-                    assert (
-                        subject_randomization.x_half_m == 0.0
-                        and subject_randomization.y_half_m == 0.0
-                        and subject_randomization.roll_half_rad == 0.0
-                        and subject_randomization.pitch_half_rad == 0.0
-                        and subject_randomization.yaw_half_rad == 0.0
-                    ), f"Object '{obj.name}' cannot randomize XY or rotation while using FaceTo."
-                parent_randomization = self._get_random_around_solution(face_to_relations[0].parent)
-                if parent_randomization is not None:
-                    assert (
-                        parent_randomization.x_half_m == 0.0 and parent_randomization.y_half_m == 0.0
-                    ), f"FaceTo parent '{face_to_relations[0].parent.name}' cannot randomize XY."
+            face_to = self._get_face_to(obj)
+            if face_to is not None:
+                self._validate_face_to_configuration(obj, face_to, object_set)
 
         anchor_objects = get_anchor_objects(objects)
         assert len(anchor_objects) > 0, (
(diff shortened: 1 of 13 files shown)


isaaclab_arena/relations/object_placer.py
```diff
@@ -951,13 +1009,49 @@ def _get_random_around_solution(self, obj: ObjectBase) -> RandomAroundSolution |
...
+        face_to_count = sum(isinstance(relation, FaceTo) for relation in obj.get_relations())
+        assert face_to_count == 1, f"Object '{obj.name}' has more than one FaceTo relation."
+        assert not obj.is_anchor, f"Anchor object '{obj.name}' cannot have a FaceTo relation."
+        assert face_to.parent is not obj, f"Object '{obj.name}' cannot face itself."
+        assert face_to.parent in object_set, f"FaceTo parent '{face_to.parent.name}' must participate in placement."
+        assert (
+            self._get_rotate_around_solution(obj) is None
+        ), f"Object '{obj.name}' cannot combine FaceTo with RotateAroundSolution."
+
+        subject_randomization = self._get_random_around_solution(obj)
```
qianl-nv: I feel face_to and randomizing rotate is not strictly conflicting. most times user might want to specify face_to with some angle tolerance, so we can either add some tolerance param to FaceTo or allow it to work with random_around_solution. wdyt?

isaaclab_arena/relations/object_placer.py
```diff
@@ -173,11 +173,15 @@ def _prepare_placement(
         objects: list[ObjectBase],
     ) -> tuple[set[ObjectBase], torch.Generator | None]:
         """Validate placement inputs and allocate an RNG seeded per candidate later."""
+        object_set = set(objects)
```
qianl-nv, line 179: not sure if it belongs to this MR, but I'd find this file getting very long with some methods that can be consolidated

for example _get_random_around_solution, _get_rotate_around_solution, _get_face_to can consolidate into one

T = TypeVar("T", bound=RelationBase)
@staticmethod
def _get_relation(obj: ObjectBase, relation_type: type[T]) -> T | None:
    return next((r for r in obj.get_relations() if isinstance(r, relation_type)), None)

or

@staticmethod
def _get_relations(obj: ObjectBase, relation_type: type[T]) -> list[T]:
    return [r for r in obj.get_relations() if isinstance(r, relation_type)]
assert len(self._get_relations(obj, FaceTo)) == 1

isaaclab_arena/relations/object_placer.py
```diff
@@ -951,13 +1009,49 @@ def _get_random_around_solution(self, obj: ObjectBase) -> RandomAroundSolution |
                 return rel
         return None
 
+    def _validate_face_to_configuration(
```
qianl-nv: can we define a more generate .validate_configuration method to each RelationBase, can default to true

qianl-nv (commented):
This is an acceptable solution.

I have some suggestion for cleaning up this isaaclab_arena/relations/object_placer.py file which has grown quite large, your judgement if it belongs to this MR or its better to be in a separate one.

isaaclab_arena/relations/object_placer.py
zhx06: Done, we consolidate 3 into one` _get_relation(obj, relation_type)`

isaaclab_arena/relations/object_placer.py
zhx06: Done. RelationBase now provides a default `validate_configuration()`

isaaclab_arena/relations/object_placer.py
zhx06: Agreed. `FaceTo` sets the center heading; `RandomAroundSolution` adds angular variation. XY randomization remains unsupported.

isaaclab_arena/relations/object_placer.py
```diff
@@ -173,11 +177,14 @@ def _prepare_placement(
...
     ) -> tuple[set[ObjectBase], torch.Generator | None]:
         """Validate placement inputs and allocate an RNG seeded per candidate later."""
+        object_set = set(objects)
         for obj in objects:
             assert obj.get_relations(), (
                 f"Object '{obj.name}' has no relations. All objects passed to place() must have "
                 "at least one relation (e.g., On(), NextTo(), or IsAnchor())."
             )
+            for relation in obj.get_relations():
+                relation.validate_configuration(obj, object_set)
```
alexmillane: I feel that we need a more descriptive name for `validate_configuration`. Not clear from the call site, what is being validated.

isaaclab_arena/relations/object_placer.py
alexmillane: I also agree with @qianl-nv that this file is getting very long. There's also lots of branching.

Let's get `v0.3` done, but I think we need to come back and consider how we can refactor this file to simplify at some point in the future.

isaaclab_arena/relations/object_placer.py
```diff
@@ -235,12 +242,19 @@ def _place_ranked(
...
+        )
 
         all_positions = self._solver.solve(
             objects, initial_positions, env_bboxes=candidate_bboxes, orientations=orientations_per_candidate
         )
+        self._apply_face_to_orientations(all_positions, orientations_per_candidate)
+        # FaceTo yaw is only known after solving, so rebuild from unrotated boxes before validation.
+        candidate_bboxes = self._rotate_candidate_bboxes(
+            objects, unrotated_candidate_bboxes, orientations_per_candidate
+        )
```
alexmillane, line 256: This is fine for this MR, however the way we treat rotations in general is becoming quite complicated due to lots of options.

I think for v0.4 we should come back and clean up how we deal we rotations in general. I'm thinking that if we just take object yaw into the solver, i.e. optimize for object yaw too, we can clean up a lot of this branching.

isaaclab_arena/relations/object_placer.py
```diff
@@ -911,26 +973,26 @@ def _apply_poses(
...
 
-            rotate_marker = self._get_rotate_around_solution(obj)
+            rotate_marker = self._get_relation(obj, RotateAroundSolution)
             base_rotation = rotate_marker.get_rotation_xyzw() if rotate_marker else (0.0, 0.0, 0.0, 1.0)
             marker_yaw = yaw_from_quat_xyzw(base_rotation)
 
-            def _sampled_yaw_delta(env_idx: int) -> float:
-                """Total yaw minus marker yaw = solver-sampled delta."""
+            def _yaw_delta(env_idx: int) -> float:
+                """Return the yaw to compose with the base rotation."""
```
alexmillane: in the docstring, by "base rotation", do we mean the marker yaw?

isaaclab_arena/relations/relations.py
```diff
@@ -73,6 +74,56 @@ def __init__(self, parent: ObjectBase, relation_loss_weight: float = 1.0):
...
+            (relation for relation in subject.get_relations() if isinstance(relation, RandomAroundSolution)), None
+        )
+        if subject_randomization is not None:
+            assert (
+                subject_randomization.x_half_m == 0.0 and subject_randomization.y_half_m == 0.0
+            ), f"Object '{subject.name}' cannot randomize XY while using FaceTo."
+
+        parent_randomization = next(
+            (relation for relation in self.parent.get_relations() if isinstance(relation, RandomAroundSolution)), None
+        )
```
alexmillane, line 120: Thank you for being so thorough and thinking about `RandomAroundSolution`.

I feel like this relation was added during the v0.2 Developer Preview to account for the fact that we didn't have some features that we have now added. The same is true of the `RotateAroundSolution` marker.

Supporting this stuff is leading to a lot of branching and complexity.

I'm thinking that after `v0.3` we should go back through and consider what we still need. And what we could unify by moving to support rotations in the solver.

isaaclab_arena/utils/pose.py
```diff
@@ -58,6 +58,34 @@ def wrap_angle_to_pi(angle_rad: float) -> float:
     return (angle_rad + math.pi) % (2.0 * math.pi) - math.pi
 
 
+MINIMUM_FACING_DIRECTION_XY_M = 1e-6
+"""XY distances at or below this are too short to define a facing direction."""
+
+
+def yaw_toward_positions(
```
alexmillane: Let's move this function and the yaw related functions below to a new file called:

`isaaclab_arena/utils/yaw.py`

alexmillane (approved):
A few nits and comments for the future, but it looks good. Feel free to adress and merge!

isaaclab_arena/relations/object_placer.py
zhx06: Done, Renamed to `validate_placement_configuration()`

isaaclab_arena/relations/object_placer.py
zhx06: Agree. Will revisit in v0.4

isaaclab_arena/relations/object_placer.py
zhx06: Yes, it is the marker rotation. Renamed `base_rotation` to `marker_rotation` and clarified the docstring.

isaaclab_arena/utils/pose.py
zhx06: Done. The yaw utilities have been moved to `isaaclab_arena/utils/yaw.py`

PR #889, opened 2026-07-10
Add cross-pool OSMO evaluation workflows for the DreamZero policy

TuurStuyck:
## Summary
Fully-remote DreamZero policy evaluation on OSMO from a single command: server and policy runner in separate compute pools, connected by an in-task tunnel and torn down together.

## Detailed description
- **Why**: the DreamZero server needs an H100 (CUDA-OOMs on an L40's 48 GiB during WAN VAE encode) while the policy runner's Isaac Sim rendering needs an RTX-capable GPU (the RTX renderer crashes with Vulkan `ERROR_DEVICE_LOST` on dgx-h100 nodes). No OSMO pool offers both platforms and a workflow's gang-scheduled group is single-pool, so the pair runs as two workflows connected by a supervised in-task `osmo workflow port-forward` tunnel (the OSMO CLI is baked into the runner image via `COPY --from=nvcr.io/nvidia/osmo/client:6.3.1` — image builders need pull access to that registry path).
- **What**: `--policy dreamzero` submits both workflows in one command — the server workflow first, then the runner auto-wired to its captured workflow ID — and the runner cancels the server workflow on exit (success or failure), so both finish together with no manual cleanup (best-effort: a hard-killed runner leaves the server until its exec timeout). `DreamZeroRemotePolicy` moves fully onto the typed-config path and gains paced connect/reconnect waits so server startup or tunnel drops don't kill an eval. Outputs (evaluation report, per-camera/viewport videos, episode results) upload to swift exactly like the other policy workflows. The standalone server YAML (`isaaclab_arena_dreamzero/docker/dreamzero_inference_server.yaml`) is updated for the local-development flow (server on OSMO, local port-forwarded policy runner).
- **Verified end-to-end on OSMO**: repeated 5-episode `pick_and_place_maple_table` (droid embodiment) evals through the tunnel (success_rate 0.6–1.0 across runs); the final validation pair (`dreamzero-final-server-1` + `dreamzero-final-1`) completed and tore down automatically, with artifacts at `swift://pdx.s8k.io/AUTH_team-isaac/isaaclab_arena/workflows/dreamzero-final-1`.
- **Impact**: DreamZero evals no longer require a local simulator with port-forwarding; pi0/gr00t/zero_action workflows are unchanged apart from the optional `--policy_runner_image` override.
- **Prerequisites (one-time account setup)**: the standard `omni_svc` GENERIC credential and a DATA credential for `swift://pdx.s8k.io/AUTH_team-isaac` (as for the other policies), plus a GENERIC credential `osmo-token` holding a base64 OSMO `login.yaml` (field `login_yaml_b64`) that authenticates the in-task tunnel and teardown.

Example:
```bash
python osmo/submit_evaluation_workflow.py \
    --policy dreamzero \
    --arena_env pick_and_place_maple_table \
    --arena_env_args '--embodiment droid_abs_joint_pos' \
    --policy_runner_args '--num_episodes 5 --enable_cameras --record_camera_video \
        --language_instruction "Pick up the cube and place it in the bowl"'
# both workflows finish on their own; results land at
# https://pdx.s8k.io/v1/AUTH_team-isaac/isaaclab_arena/workflows/<runner-workflow-id>
```


docker/Dockerfile.isaaclab_arena
```diff
@@ -140,6 +141,11 @@ RUN echo "alias pytest='/isaac-sim/python.sh -m pytest'" >> /etc/bash.bashrc
 #    configuration from the Run and Debug panel.
 RUN echo "alias debugpy='python -Xfrozen_modules=off -m debugpy --listen localhost:5678 --wait-for-client'" >> /etc/bash.bashrc
 
+# OSMO CLI, used by OSMO tasks that tunnel to a policy server running in another
+# OSMO workflow (see osmo/tasks/dreamzero_policy_runner_task.py).
```
alexmillane: I think we actually want this installed regardless of the Dream Zero requirement.

Suggestion to update this comment to `# OSMO CLI`

osmo/tasks/dreamzero_policy_runner_task.py
```diff
@@ -0,0 +1,130 @@
...
+    def __init__(
+        self,
+        task_cfg: DreamZeroPolicyRunnerTaskCfg,
+        server_workflow_id: str,
+        lead: bool | None = None,
+    ) -> None:
+        super().__init__(task_cfg=task_cfg, lead=lead)
+        # The ID is interpolated into the bash entry script; reject anything that
+        # word-splits or expands there.
+        assert re.fullmatch(r"[A-Za-z0-9._-]+", server_workflow_id), f"Invalid OSMO workflow ID: {server_workflow_id!r}"
```
alexmillane: Not sure I understand the comment here.

Suggestion to move this to a `utils` file. Perhaps

```
assert is_valid_workflow_id(server_workflow_id)
```

osmo/workflows/workflow.py
```diff
@@ -165,11 +172,21 @@ def _submit_rendered_workflow(self, rendered: str) -> int:
         print(f"  {' '.join(cmd)}\n")
 
         try:
-            result = subprocess.run(cmd)
+            # Capture stdout only, to parse the workflow ID; stderr stays inherited so
+            # interactive output (login prompts, progress) remains visible live.
+            result = subprocess.run(cmd, stdout=subprocess.PIPE, text=True)
+            print(result.stdout, end="")
+            self.submitted_workflow_id = self._parse_submitted_workflow_id(result.stdout)
             return result.returncode
```
alexmillane: Suggestion to upgrade the type of the return on this function to something that can hold both the return code and submitted_workflow_id (plus anything else we need in the future)

```
return WorkflowSubmissionResult(
   returncode=result.returncode,
   workflow_id=self._parse_submitted_workflow_id(result.stdout)
)
```

This will avoid storing things as members for later access.

osmo/workflows/dreamzero_split_workflows.py
```diff
@@ -0,0 +1,165 @@
...
+        """Return a workflow config carrying the server's resource values and workflow name."""
+        return replace(
+            self.workflow_cfg,
+            workflow_name=f"{self.workflow_cfg.workflow_name}-server",
+            pool=self.task_cfg.server_pool,
+            platform=self.task_cfg.server_platform,
+            cpus=self.task_cfg.server_cpus,
+            memory=self.task_cfg.server_memory,
+            storage=self.task_cfg.server_storage,
+        )
```
alexmillane: I guess we're doing this because this `Workflow` effectively has two `workflow_cfg`s but this is dis-allowed by the `Workflow` base class?

What do you think about introducing some `CompositeWorkflow` class that naturally supports this. And supports the other stuff going on in this file.

We may have some issues with repeated flags on the CLI that we'll need to sort out with a prefix or something.

alexmillane (approved):
Looks great.

One substantial comment.

Merge once you've addressed it!

PR #891, opened 2026-07-10
Remove light from purple_crate object usd

qianl-nv:
## Summary
Switch purple_crate to staging server without light.

## Detailed description
- Production server version has dome light as part of the object usd, so the bottles_crate.yaml RoboLab env loads without a global DomeLight.
- A new version of the purple crate with its dome light removed is saved to the staging server. Update object library to use the staging server version.
- Now the bottles_crate.yaml loads with correct dome light.

Before:
image

After: 
image


alexmillane (approved):
Thanks for doing that!

PR #892, opened 2026-07-10
Use typed configuration for local Experiments

cvolkcvolk:
## Summary
Use typed configuration for local Experiment execution

## Detailed description
- Depends on #876 and should merge after it; this draft is stacked on the #876 branch that introduces typed YAML execution through eval_runner.py.
- Prepares eval_runner.py to become the single entry point for both local execution and OSMO submission by replacing the Namespace flow with EvalRunnerCfg and typed AppLauncher values; explicit routing remains a later PR.
- Keeps local execution in eval_runner.py until the non-local path exists, avoiding a premature execution-module abstraction.
- Preserves native Hydra overrides, legacy JSON and chunk behavior, and existing camera semantics; camera preflight and OSMO behavior are intentionally excluded.


isaaclab_arena/evaluation/local_evaluation.py
```diff
@@ -0,0 +1,124 @@
...
+def _load_local_experiment(cfg: EvalRunnerCfg, experiment_config_path: Path) -> ArenaExperiment:
+    """Compose an Experiment after SimulationApp resolves the process device."""
+    return load_arena_experiment_from_config_file(
+        experiment_config_path,
+        device=cfg.device,
+        overrides=cfg.experiment_overrides,
+    )
+
+
+def run_local_experiment(cfg: EvalRunnerCfg) -> int:
```
cvolkcvolk: Is this the prior content of def main() in eval_runner.py?  If yes could we keep it in eval_runner.py for now until we add the non local execution path?

isaaclab_arena/evaluation/local_evaluation.py
```diff
@@ -0,0 +1,124 @@
...
+def _assert_camera_support_enabled(experiment: ArenaExperiment, enable_cameras: bool) -> None:
+    """Check that AppLauncher enabled camera support requested by typed Runs."""
+    camera_run_names = [run_cfg.name for run_cfg in experiment if getattr(run_cfg.environment, "enable_cameras", False)]
+    assert not camera_run_names or enable_cameras, (
+        f"Runs {camera_run_names} enable environment cameras. Pass --enable_cameras so AppLauncher enables "
+        "camera support before the typed Experiment is composed."
+    )
+
+
+def _load_local_experiment(cfg: EvalRunnerCfg, experiment_config_path: Path) -> ArenaExperiment:
```
cvolkcvolk: Why does this need its own function?

isaaclab_arena/utils/isaaclab_utils/simulation_app.py
```diff
@@ -23,12 +24,19 @@ def get_isaac_sim_version() -> str:
...
 
 
-def get_app_launcher(args: argparse.Namespace) -> AppLauncher:
+def get_app_launcher(args: argparse.Namespace | dict[str, Any]) -> AppLauncher:
     """Get an app launcher."""
     import time
 
     t0 = time.monotonic()
-    app_launcher = AppLauncher(args)
+    # AppLauncher consumes and mutates dictionaries while resolving settings,
```
cvolkcvolk: This looks like a workaround. It this really required or is it covering up something we should investigate?

cvolkcvolk: I investigated this in Isaac Lab's `AppLauncher`. It intentionally mutates the supplied mapping: a Namespace is converted to its `__dict__`, launcher-only keys are popped, and an implicit XR device is written back for downstream consumers. I removed the copy and manual propagation; the typed dictionary is now passed directly, restoring the pre-PR semantics. The test now verifies in-place consumption and resolved-device propagation.

isaaclab_arena/evaluation/local_evaluation.py
cvolkcvolk: Agreed, this helper did not add a useful boundary. I removed `_load_local_experiment()` and inlined Experiment composition in both `SimulationAppContext` branches.

isaaclab_arena/evaluation/local_evaluation.py
cvolkcvolk: Yes. This was the prior local `main()` body. I moved it back into `eval_runner.py` and deleted `local_evaluation.py`; `run_local_experiment()` remains in the same file as the small future routing seam.

isaaclab_arena/evaluation/eval_runner_cli.py
```diff
@@ -3,91 +3,179 @@
...
+    arguments = list(sys.argv[1:] if cli_args is None else cli_args)
+    parser = argparse.ArgumentParser(
+        description="Run an Isaac Lab-Arena Experiment locally.",
+        allow_abbrev=False,
+    )
     parser.add_argument(
+        "--experiment-config",
         "--experiment_config",
+        "--eval-jobs-config",
         "--eval_jobs_config",
```
cvolkcvolk, line 83: Let's stick with only the underscore version

isaaclab_arena/evaluation/eval_runner_cli.py
```diff
@@ -3,91 +3,179 @@
...
+        type=Path,
         default=_DEFAULT_EXPERIMENT_CONFIG_PATH,
-        help=(
-            "Path to a typed YAML Experiment or legacy JSON evaluation config. "
-            "For YAML, append Hydra KEY=VALUE overrides to the command."
-        ),
+        help="Path to a typed YAML Experiment or deprecated JSON evaluation config.",
     )
     parser.add_argument(
+        "--record-viewport-video",
```
cvolkcvolk, line 90: Lets keep the underscore version, here and for all other flags

PR #896, opened 2026-07-13
Rename eval_runner.py to experiment_runner.py

cvolkcvolk:
## Summary
Align runner naming with Experiments. Renames `eval_runner.py` to `experiment_runner.py` because its unit of work is one `ArenaExperiment`: an ordered `list[ArenaRunCfg]`.

## Detailed description
- Uses Experiment Runner for the already-namespaced component while retaining the Arena prefix on public domain types.
- Updates Python APIs, helpers, tests, and documentation without retaining a compatibility wrapper or changing execution behavior.

### DIFF
--- a/docs/pages/concepts/policy/concept_evaluation_types.rst
+++ b/docs/pages/concepts/policy/concept_evaluation_types.rst
@@ -2,7 +2,7 @@ Evaluation Types
 =================
 
 Isaac Lab Arena supports two main ways to run policy evaluation: a single-job
-**policy runner** (single or multi-GPU) and an **Arena Experiment Runner** for
+**policy runner** (single or multi-GPU) and an **Experiment Runner** for
 multiple jobs in one process. This section summarizes when to use each and how
 they work. Each section below links to the relevant concept docs:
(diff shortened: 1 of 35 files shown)


docs/pages/concepts/policy/concept_evaluation_types.rst
```diff
@@ -2,7 +2,7 @@ Evaluation Types
 =================
 
 Isaac Lab Arena supports two main ways to run policy evaluation: a single-job
-**policy runner** (single or multi-GPU) and a **sequential batch eval runner** for
+**policy runner** (single or multi-GPU) and an **Arena Experiment Runner** for
```
xyao-nv: Do we have to call it **arena** experiment runner?  Everything is Arena xxx in this repo.
Can it just be experiment runner?

cvolkcvolk: Yeah I think you are right it stutters the namespace a bit. Chaning to `experiment_runner.py`

alexmillane (approved):
Makes a ton of sense. LGTM

PR #898, opened 2026-07-13
Represent Experiment Runs as a named mapping

cvolkcvolk:
## Summary
Represent Arena Experiments as typed, named Run mappings that Hydra can consume directly.

## Detailed description
Before this change, an Experiment declared its Runs as a YAML sequence, with each Run carrying a separate `name` field:

```yaml
runs:
  - name: first_run
    environment:
      type: pick_and_place_maple_table
    policy:
      type: zero_action

  - name: second_run
    environment:
      type: pick_and_place_maple_table
    policy:
      type: zero_action
    environment_builder:
      num_envs: 64
```

That representation is readable, but its user-facing name is only a value inside a sequence item. 

This PR makes the name the mapping key:

```yaml
runs:
  first_run:
    environment:
      type: pick_and_place_maple_table
    policy:
      type: zero_action

  second_run:
    environment:
      type: pick_and_place_maple_table
    policy:
      type: zero_action
    environment_builder:
      num_envs: 64
```

The key now appears before the Run value, remains the single source of Run identity, and creates a real Hydra path. A user can override the parallel Run without knowing its position:

```text
runs.second_run.environment_builder.num_envs=8
```
instead of
```text
runs.1.environment_builder.num_envs=8
```

The named mapping is represented explicitly through a new Cfg:

```python
@dataclass
class ArenaExperimentCfg:
    """Configure the named Runs that form one Arena Experiment."""

    runs: dict[str, ArenaRunCfg]
```

`ArenaExperimentCfg` is the single Experiment representation. The former `ArenaExperiment = list[ArenaRunCfg]` Definition is removed

Typed YAML composition now resolves `policy.type` as each Run is built. Core policies may continue using a registered name such as `zero_action`; an extension policy may provide a dotted class path such as `package.policy.Policy`. Importing that class registers its typed policy config without making the core loader import extension packages up front.

### DIFF
--- a/isaaclab_arena/evaluation/arena_experiment.py
+++ b/isaaclab_arena/evaluation/arena_experiment.py
@@ -3,17 +3,16 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""Represent declarative and executable Arena Experiments."""
+"""Represent Arena Experiment configurations."""
 
 from dataclasses import dataclass
-from typing import TypeAlias
 
 from isaaclab_arena.evaluation.arena_run import ArenaRunCfg
 
 
 @dataclass
-class ArenaExperimentDefinitionCfg:
-    """Declare the named Runs that form one Arena Experiment."""
+class ArenaExperimentCfg:
+    """Configure the named Runs that form one Arena Experiment."""
 
     runs: dict[str, ArenaRunCfg]
     """Runs keyed by the names used for overrides, execution, and results."""
@@ -27,7 +26,3 @@ class ArenaExperimentDefinitionCfg:
                 "Run name is defined by its Experiment Definition mapping key and cannot be overridden: "
                 f"expected '{run_name}', got '{run_cfg.name}'"
             )
-
-
-ArenaExperiment: TypeAlias = list[ArenaRunCfg]
-"""An Experiment expressed as its ordered list of Runs."""
--- a/isaaclab_arena/evaluation/arena_experiment.py
+++ b/isaaclab_arena/evaluation/arena_experiment.py
@@ -3,11 +3,31 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""Represent an Arena Experiment as an ordered list of Runs."""
+"""Represent declarative and executable Arena Experiments."""
 
+from dataclasses import dataclass
 from typing import TypeAlias
 
 from isaaclab_arena.evaluation.arena_run import ArenaRunCfg
+
+
+@dataclass
+class ArenaExperimentDefinitionCfg:
+    """Declare the named Runs that form one Arena Experiment."""
+
+    runs: dict[str, ArenaRunCfg]
+    """Runs keyed by the names used for overrides, execution, and results."""
+
+    def __post_init__(self) -> None:
+        assert isinstance(self.runs, dict) and self.runs, "Experiment Definition must contain at least one Run"
+        for run_name, run_cfg in self.runs.items():
+            assert isinstance(run_name, str) and run_name, "Experiment Definition Run names must be non-empty strings"
+            assert isinstance(run_cfg, ArenaRunCfg), f"Experiment Definition Run '{run_name}' must be an ArenaRunCfg"
+            assert run_cfg.name == run_name, (
+                "Run name is defined by its Experiment Definition mapping key and cannot be overridden: "
+                f"expected '{run_name}', got '{run_cfg.name}'"
+            )
+
 
 ArenaExperiment: TypeAlias = list[ArenaRunCfg]
 """An Experiment expressed as its ordered list of Runs."""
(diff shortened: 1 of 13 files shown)


isaaclab_arena/evaluation/arena_experiment.py
```diff
@@ -3,11 +3,26 @@
...
+
+    def __post_init__(self) -> None:
+        assert isinstance(self.runs, dict) and self.runs, "Experiment Definition must contain at least one Run"
+        for run_name, run_cfg in self.runs.items():
+            assert isinstance(run_name, str) and run_name, "Experiment Definition Run names must be non-empty strings"
+            assert isinstance(run_cfg, ArenaRunCfg), f"Experiment Definition Run '{run_name}' must be an ArenaRunCfg"
+            assert run_cfg.name == run_name, (
+                "Run name is defined by its Experiment Definition mapping key and cannot be overridden: "
+                f"expected '{run_name}', got '{run_cfg.name}'"
+            )
```
alexmillane, line 28: Cool!

isaaclab_arena/evaluation/arena_experiment_config_loader.py
```diff
@@ -46,35 +48,36 @@ def load_arena_experiment_from_config_file(
         overrides: Hydra overrides applied to typed YAML Experiments.
 
     Returns:
-        The ordered typed Runs that make up the Experiment.
+        The composed Experiment configuration with named typed Runs.
```
alexmillane: Mega nit: I personally don't like the way that hydra calls things "composed".

Basically this is just returning the _loaded_ `ArenaExperimentCfg`.

I guess there's a reason that hydra calls this thing "composed", but the uninformed user, it's a bit of a weird term IMO.

isaaclab_arena/evaluation/experiment_runner.py
```diff
@@ -41,9 +41,13 @@ def list_variations(experiment: ArenaExperiment) -> None:
...
 # values and Hydra overrides before startup. Enable AppLauncher when any Run enables
 # cameras, then remove this getattr and the requirement to also pass --enable_cameras.
-def _assert_camera_support_enabled(experiment: ArenaExperiment, enable_cameras: bool) -> None:
+def _assert_camera_support_enabled(experiment_cfg: ArenaExperimentCfg, enable_cameras: bool) -> None:
     """Check that AppLauncher enabled camera support requested by typed Runs."""
-    camera_run_names = [run_cfg.name for run_cfg in experiment if getattr(run_cfg.environment, "enable_cameras", False)]
+    camera_run_names = [
+        run_cfg.name
+        for run_cfg in experiment_cfg.runs.values()
+        if getattr(run_cfg.environment, "enable_cameras", False)
```
alexmillane, line 49: at some point we discussed adding `enable_cameras` to some default list of parameters that all environments get. Which would obviate the `getattr` here.

Not really important now, but we could do that in the future.

isaaclab_arena_environments/experiment_configs/getting_started_experiment.yaml
```diff
@@ -5,52 +5,52 @@
...
+      embodiment: droid_rel_joint_pos
+      pick_up_object: rubiks_cube_hot3d_robolab
+      destination_location: bowl_ycb_robolab
+    policy:
+      type: zero_action
+    environment_builder:
+      num_envs: 64
+      env_spacing: 2.5
+    rollout_limit:
+      num_steps: 100
```
alexmillane, line 56: nice!

alexmillane (approved):
Great improvement!

isaaclab_arena/hydra/experiment_composition.py
```diff
@@ -74,19 +75,22 @@ def load_arena_experiment_from_yaml(
...
             }
             hydra_experiment_config_name = f"{hydra_config_namespace}_root"
-            config_store.store(name=hydra_experiment_config_name, node={"runs": arena_runs_by_name})
+            config_store.store(
+                name=hydra_experiment_config_name,
+                node=ArenaExperimentCfg(runs=arena_runs_by_name),
+            )
             composed_experiment = compose(config_name=hydra_experiment_config_name, overrides=overrides or [])
-            experiment = [OmegaConf.to_object(composed_experiment.runs[run_name]) for run_name in run_values_by_name]
+            experiment_cfg = OmegaConf.to_object(composed_experiment)
```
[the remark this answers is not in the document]
cvolkcvolk: Experiment overrides currently modify fields on Runs declared in the YAML. They do not create new typed Runs. Expected behaviour and clarified in the docstring

isaaclab_arena/evaluation/arena_experiment_config_loader.py
cvolkcvolk: Yea me neither. Claude likes to call all functions touching hydra `compose_*`. Will keep the internal variables aligned with the hydra naming convention but try to rename user facing variables/funcions. Updated the docstring here.

isaaclab_arena/evaluation/experiment_runner.py
cvolkcvolk: Agree. Its already captured in the above TODO() to do that in a follow up.

PR #907, opened 2026-07-14
Let run-time variations apply build-time effects

alexmillane:
## Summary
Allow any variation (not just build-time ones) to perform build-time mutations.

## Detail
- To support some run-time variations, sometimes you need to set something up in the `EnvironmentCfg` at build time.
- This makes this possible.
- Rename the `BuildTimeVariationBase` abstract `apply()` to `apply_build_time_effects()`.

## Notes
- This is added to support forthcoming support for camera intrinsic variations.


isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -145,18 +144,19 @@ def _compose_variations_event_cfg(self) -> Any | None:
...
-        """Sample and apply every enabled :class:`BuildTimeVariationBase`.
+        """Apply every enabled variation's build-time effects before ``scene_cfg`` is materialised.
 
-        These mutate asset configs in place (e.g. a dome light's spawner
-        texture), so this must run before ``scene_cfg`` is materialised.
+        Calls ``apply_build_time_effects`` on each enabled variation regardless of type: a
+        :class:`BuildTimeVariationBase` realises its whole effect there (e.g. a dome light's
+        spawner texture), while a run-time variation uses it for build-time preconditions (e.g.
+        forcing its camera untiled). These mutate asset configs in place, so this must run before
+        the scene is materialised.
```
alexmillane: Restore the details section of the docstring to what it was.

isaaclab_arena/variations/variation_base.py
```diff
@@ -91,6 +91,15 @@ def add_sample_listener(self, listener: Callable[[Any, torch.Tensor | None], Non
         if self._sampler is not None:
             self._sampler.add_listener(listener)
 
+    def apply_build_time_effects(self) -> None:
+        """Sample and mutate the bound asset config(s) in place before the scene is materialised.
+
+        Called once per env build for every enabled variation, regardless of type. Default:
+        no-op. A :class:`BuildTimeVariationBase` realises its whole effect here; a run-time
+        variation overrides this only when it needs a build-time precondition (e.g. forcing a
+        camera untiled so per-env run-time edits take effect).
```
alexmillane: remove: regardless of type

remove :class: markings.

remove "(e.g. forcing a...)

isaaclab_arena/variations/variation_base.py
```diff
@@ -129,11 +138,11 @@ class BuildTimeVariationBase(VariationBase):
 
     Use for properties that can't change in-flight: HDR maps, USD swaps,
     spawner params baked into a config. Subclasses hold references to the
-    asset(s) they mutate.
+    asset(s) they mutate and realise the effect in :meth:`apply_build_time_effects`.
```
alexmillane: remove :meth: marker.

alexmillane (commented):
self review 2

isaaclab_arena/variations/variation_base.py
```diff
@@ -91,6 +91,14 @@ def add_sample_listener(self, listener: Callable[[Any, torch.Tensor | None], Non
         if self._sampler is not None:
             self._sampler.add_listener(listener)
 
+    def apply_build_time_effects(self) -> None:
+        """Sample and mutate the bound asset config(s) in place before the scene is materialised.
```
alexmillane: Update: Sample and mutate the bound assets in place in the ArenaEnvBuilder before the environment cfg is composed.

alexmillane (commented):
self review 3

isaaclab_arena/variations/variation_base.py
```diff
@@ -91,6 +91,14 @@ def add_sample_listener(self, listener: Callable[[Any, torch.Tensor | None], Non
         if self._sampler is not None:
             self._sampler.add_listener(listener)
 
+    def apply_build_time_effects(self) -> None:
```
cvolkcvolk: Is "effect" a bit vaque here? I think we are doing build-time _realization_ and _preparation_ for later run-time realization here. So this function serves two different purposes: a `BuildTimeVariationBase` realizes its sampled variation, whereas a run-time variation only prepares configuration required by its later event.


Only a suggestion, could we make the distinction explicit while leaving the builder with one call with something like

```
  class VariationBase(ABC):
      def prepare_for_build(self) -> None:
          """Configure prerequisites required before environment construction."""
          pass

      def realize_at_build_time(self) -> None:
          """Realize this variation during environment construction."""
          pass

      def configure_at_build_time(self) -> None:
          """Run this variation's build-time preparation and realization."""
          self.prepare_for_build()
          self.realize_at_build_time()
```


`BuildTimeVariationBase` makes `self.realize_at_build_time()` mandatory:

```
  class BuildTimeVariationBase(VariationBase):
      @abstractmethod
      def realize_at_build_time(self) -> None:
          """Sample and apply this variation to its target configuration."""
          ...
```

A run-time variation overrides `self.prepare_for_build()` when necessary:

```
  class CameraIntrinsicsRunTimeVariation(RunTimeVariationBase):
      def prepare_for_build(self) -> None:
          """Configure the camera for per-environment run-time updates."""
          if self._camera_rig is not None:
              self._camera_rig.set_use_tiled_camera(False)
```

and the arena_env_builder.py retains one call:

`variation.configure_at_build_time()`

cvolkcvolk (approved):
Great!

isaaclab_arena/variations/variation_base.py
alexmillane: Good idea! Done!

PR #908, opened 2026-07-14
uv source install

alexmillane:
## Summary
Add a native (non-Docker) uv-based install path for Arena.

## Detailed description
- Given a source checkout, one can now install Arena via `uv sync`
- packaging moves from `setup.py` to a full `pyproject.toml` with a committed `uv.lock`.
- Arena code to work with the public Isaac Lab wheel.
- Adds a `uv_test` CI job that runs the test suite natively in a clean `nvidia/cuda:12.8-ubuntu24.04`.
- Updates the installation docs and README test commands to cover the native uv workflow alongside Docker.
- Based on #846 but rebased on the Isaac Lab 3.0 Beta 2.0 upgrade.

PR #910, opened 2026-07-14
Fix graph camera setup for DROID scenes

lgulich:
## Summary
Fix graph camera setup for DROID graph specs

## Detailed description
- Generated graph YAMLs could not request the reviewed DROID exterior RGB-D camera profile, so camera setup depended on non-graph Maple environment wiring.
- Adds registered camera profiles, routes the DROID exterior RGB-D profile through agentic generation and graph-spec builds, and fails closed when a profile is selected without cameras enabled.
- Shares CAP Maple/DROID staging and local asset routing with graph-YAML builds so generated DROID scenes can run directly without mutating shared registered asset classes.
- Validated with `pre-commit run --all-files` and focused graph/camera/asset tests: 53 passed, 3 skipped.

### DIFF
--- a/isaaclab_arena/agentic_environment_generation/environment_generation_agent.py
+++ b/isaaclab_arena/agentic_environment_generation/environment_generation_agent.py
@@ -131,8 +131,17 @@ class AssetCatalogue:
 
     def to_catalog_string(self) -> str:
         """Format this catalogue as the user-message vocabulary block."""
+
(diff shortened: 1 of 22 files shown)


lgulich:
Closing this stacked CAP-specific change. The replacement keeps CAP/GaP composition in Isaac-cap and will propose only independently useful Arena framework changes on top of #908.

PR #911, opened 2026-07-14
Add configurable stand_height to Droid embodiment

xyao-nv:
## Summary
Add configurable stand_height to Droid embodiment.

## Detailed description
- The Droid base stand had a hardcoded size and no way to resize it independently of the robot; users needed to adjust stand height/footprint without altering the robot articulation.
- Only stand_height is being configured thru inputs params of Embodiment class. Surfaced it in one robolab env.
- It also changes the base pose z values automatically.

3.0
image
1.7
image


peterd-NV (approved):

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -61,9 +64,11 @@ def __init__(
         initial_joint_pose: list[float] | None = None,
         concatenate_observation_terms: bool = False,
         arm_mode: ArmMode | None = None,
+        stand_scale: tuple[float, float, float] = _DEFAULT_STAND_SCALE,
```
alexmillane: What's the goal of the scale? I'm thinking that the goal is likely to be able to adjust the height? Am I correct there, the PR description doesn't mention the intention here.

I'm wondering if the goal is to be able to adjust the robot height, if we should expose that in another way. I.e. add a constructor variable `stand_height`. Right now the user has to know that the stand scale can be used to affect the height. They have to know that the stand stretches around the robot base (rather than for example scaling around the stand's center which would cause collisions).

alexmillane (approved):
Looks good.

I have one comment about renaming the scale to make the purpose clear to a user.

isaaclab_arena/embodiments/droid/droid.py
xyao-nv: yeah it's for adjusting the height. taking your suggestion. thx

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -60,9 +98,16 @@ def __init__(
         initial_joint_pose: list[float] | None = None,
         concatenate_observation_terms: bool = False,
         arm_mode: ArmMode | None = None,
+        stand_height: float = _DEFAULT_STAND_HEIGHT_SCALE,
```
qianl-nv: stand_height sounds like a metric value by name but is actually a scale.
I'd prefer to update the implementation to actually treat it as an absolute height in meter. the default 1.7 scale is a bit arbitrary anyway

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -42,6 +44,42 @@
...
+
+# The stand's top is pinned at the robot base and it extends downward to the floor, so scaling its
+# height moves the floor contact, not the mount. We lift the robot base by the amount the stand
+# grows (native height x scale delta), keeping the stand's bottom on the floor while the robot rides
+# higher. The native height is read from the stand USD; this measured value is the fallback used when
+# the asset can't be opened (e.g. no asset resolver / network outside a running SimulationApp).
+_FALLBACK_STAND_UNIT_HEIGHT_M: float = 0.795
+
+
+@functools.cache
```
qianl-nv: move this private helper to the back of the file. 
nicer to see class DroidEmbodimentBase at the top...

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -42,6 +44,42 @@
 from isaaclab_arena.utils.pose import Pose
 from isaaclab_arena.variations.camera_extrinsics_variation import CameraExtrinsicsVariation
 
+# The base stand's x/y footprint is fixed; only its height (the z-axis scale factor) is
+# configurable, so the stand can be made taller or shorter without changing its footprint.
+_STAND_FOOTPRINT_SCALE_XY: tuple[float, float] = (1.2, 1.2)
+_DEFAULT_STAND_HEIGHT_SCALE: float = 1.7
+
+# The stand's top is pinned at the robot base and it extends downward to the floor, so scaling its
```
qianl-nv: these docstring should move to line 107.

qianl-nv (approved):
Thanks for the change, a few nits.

isaaclab_arena/embodiments/droid/droid.py
xyao-nv: done

isaaclab_arena/embodiments/droid/droid.py
xyao-nv: change to meter

isaaclab_arena/embodiments/droid/droid.py
xyao-nv: done

PR #914, opened 2026-07-15
Add an optional build-time, sim-free gate to keep only the object layouts a robot can actually reach, using cuRobo IK (Part 2/3)

xyao-nv:
## Summary
Today the solver only checks that objects fit geometrically, before placement event. This MR adds a second, optional check: a layout is stored only when it passes geometry and the robot can reach a top-down grasp at *every movable object*.  The placement loop keeps re-sampling until every env has enough reachable layouts.

## Detailed description
- IK_reachability is registered to the validator registry and loaded if curobo deps exist when requested by enable_checks/required_checks by the env.yaml
- A new SimApp-free IK solver that can be used to reconstruct each object's world pose from a layout, build one collision box per object, and batch-check top-down grasp IK feasibility. This validator only runs on geometry-valid candidates so the expensive IK check never wastes time on layouts that already failed.
 - Consolidated the cuRobo IK helpers. Added supportive utils funcs.
 
## How to
Run in curobo image (-c).  Include ik_reachability in enabled_checks
```
python isaaclab_arena/evaluation/policy_runner.py       --viz kit       --policy_type zero_action       --enable_cameras       --num_steps 100  --env_graph_spec_yaml isaaclab_arena_environments/robolab/tasks/butter_above_rasin.yaml
```
Logs will show how many layouts are rejected after validation upon each refill

> 
[placement] Validated 50 candidate layout(s); passed per check: on_relation=48/50, next_to=50/50, not_next_to=50/50, face_to=50/50, no_overlap=41/50, ik_reachable=1/40
image
image

(Green dots: reachable voxels, red bbox: objects unreachable by eef)

## Note
- Different than #837 where it's an isolated, post-validation, sim-live, env-coupled process. Here it refactors previous Curobo machinery and uses Curobo in a sim-free manner.
- Instead of using live USD mesh for all its compute, it approximates the world by using cuboid with object's AABB. It's a simpler machinery compared with a throwaway env to avoid conflict from calling env.reset() terms.

## TODO
1. Apply reachability only among task-centric objects.
2. Disable hand joint to check collision between world <> arms (Curobo.success)


isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -95,6 +97,17 @@ def _solve_relations(self) -> None:
             )
             if self.cfg.resolve_on_reset is not None:
                 placer_params.resolve_on_reset = self.cfg.resolve_on_reset
+
+        # Gate pooled placement on reachability on demand.
+        if self.cfg.validate_reachability and placer_params.reachability_validator is None:
+            embodiment = self.arena_env.embodiment
+            assert embodiment is not None, "--validate_reachability requires an environment with a robot embodiment."
+            # Load the extension providing the gate, then pull its validator into the pool's solve loop.
+            extensions = ["isaaclab_arena_curobo"]
```
xyao-nv: really?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -95,6 +97,17 @@ def _solve_relations(self) -> None:
...
+            import_extensions(extensions)
+            placer_params.reachability_validator = resolve_reachability_validator(
```
xyao-nv: I cannot think of other ways to pass this callable as optional to solver call. Open for suggestions.

isaaclab_arena/environments/arena_env_builder_cfg.py
```diff
@@ -27,5 +28,13 @@ class ArenaEnvBuilderCfg:
     device: str = "cuda:0"
     language_instruction: str | None = None
 
+    validate_reachability: bool = False
```
xyao-nv: Because curobo does not support pip, so leaving reachability validation as optional

isaaclab_arena/environments/arena_env_builder.py
xyao-nv: removed

isaaclab_arena_curobo/curobo_ik_utils.py
```diff
@@ -0,0 +1,293 @@
...
+    quat_wxyz = tuple(float(v) for v in math_utils.convert_quat(quat_t, to="wxyz").tolist())
+    return AABBCollisionCuboid(
+        name=obj.name,
+        center_xyz=tuple(float(v) for v in center_world.tolist()),
+        dims_xyz=dims,
+        quat_wxyz=quat_wxyz,
+    )
+
+
+def check_ik_feasibility_batch_goal_poses(
```
xyao-nv: can be unified

isaaclab_arena_curobo/curobo_frame_utils.py
```diff
@@ -0,0 +1,148 @@
...
+    ik_result = ik_solver.solve_batch(goal_pose, seed_config=ik_seed)
+
+    num_poses = positions.shape[0]
+    pos_err = ik_result.position_error.view(num_poses, -1)
+    rot_err = ik_result.rotation_error.view(num_poses, -1)
+    best_idx = pos_err.argmin(dim=1, keepdim=True)
+    best_pos_err = pos_err.gather(1, best_idx).squeeze(1)
+    best_rot_err = rot_err.gather(1, best_idx).squeeze(1)
+    # TODO(xinjieyao): add a check for curobo.success for collision-free joints
+    feasible = (best_pos_err < position_threshold) & (best_rot_err < rotation_threshold)
```
zhx06: **This is the crux of the geometry-vs-IK question.** `feasible` is gated purely on position/rotation residual and never consults `ik_result.success`. But `success` is the flag that folds in cuRobo's collision-aware check against the `WorldConfig` cuboids we build per layout. As written, the collision world assembled in `world_config_from_cuboids` / `update_world` is loaded into the solver but its verdict is discarded — a joint config that reaches the pose *through* an obstacle is still reported reachable. So the gate validates "an IK pose exists" but not "a collision-free IK pose exists," which is the geometric guarantee the caller presumably wants.

The `TODO(xinjieyao)` on the line above acknowledges this, but I'd treat it as blocking rather than a follow-up, since without it the whole AABB-cuboid machinery is inert. Suggest:
```python
success = ik_result.success.view(num_poses, -1)
feasible = success.any(dim=1)  # or combine with the residual thresholds below
```

isaaclab_arena_curobo/curobo_frame_utils.py
```diff
@@ -0,0 +1,148 @@
...
+        ik_seed = pose_ctx._to_curobo_device(seed_config)
+        while ik_seed.dim() < 3:
+            ik_seed = ik_seed.unsqueeze(0)
+
+    ik_result = ik_solver.solve_batch(goal_pose, seed_config=ik_seed)
+
+    num_poses = positions.shape[0]
+    pos_err = ik_result.position_error.view(num_poses, -1)
+    rot_err = ik_result.rotation_error.view(num_poses, -1)
+    best_idx = pos_err.argmin(dim=1, keepdim=True)
```
zhx06: Subtle false-negative in the best-seed selection: you pick the seed with minimum *position* error (`pos_err.argmin`), then test that one seed's rotation error against `rotation_threshold`. A different seed could satisfy *both* thresholds while the min-position seed fails rotation, and you'd wrongly report infeasible. Feasibility should be "does any seed meet both bars":
```python
ok = (pos_err < position_threshold) & (rot_err < rotation_threshold)
feasible = ok.any(dim=1)
```
Once collision `success` is added (comment below), the per-pose reduction should likewise be over *successful* seeds only, otherwise `best_pos_err`/`best_rot_err` can report an in-collision solution's residuals.

isaaclab_arena_curobo/ik_reachability_validator.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+    base_pos, base_quat_xyzw = embodiment.get_base_pose()
+
+    def validator(result: PlacementResult) -> bool:
+        objects = list(result.positions.keys())
+        anchors = set(get_anchor_objects(objects))
+        base_rotations = get_base_rotation_per_object(objects)
+
+        world_poses = {obj: get_object_world_pose_from_layout(result, obj, base_rotations) for obj in objects}
+        cuboids = [get_aabb_collision_cuboid_for_object(obj, *world_poses[obj]) for obj in objects]
```
zhx06: The collision world is built from **every** object (`for obj in objects`), including the movable object we're about to grasp. That's fine today because collision is ignored, but the moment `ik_result.success` is enforced (see `solve_ik_feasibility`), a top-down grasp at `object_center + 0.02` will collide with the target's *own* AABB cuboid and every movable object will be reported unreachable. Typically the grasp target (or at least its cuboid) should be excluded from the world for its own IK query — e.g. build a per-object world that drops `obj`, or disable that obstacle for its solve. Worth resolving together with the collision-success change so the two land consistent.

isaaclab_arena_curobo/ik_reachability_validator.py
```diff
@@ -0,0 +1,126 @@
...
+        ik_pos_threshold: Max IK position error (m) for a grasp to count as reachable.
+        ik_rot_threshold: Max IK rotation error (rad) for a grasp to count as reachable.
+        device: Explicit CUDA device (e.g. ``"cuda:0"``); defaults to the current device.
+        stamp_results: Record the verdict as a required ``PlacementCheck.IK_REACHABLE`` check on each
+            layout, so it shows up in the audit report and gates that layout's ``.success``.
+    """
+    solver = StandaloneIKReachability(
+        embodiment_curobo_cfg(embodiment),
+        device=device,
+        position_threshold=ik_pos_threshold,
```
zhx06: `ik_pos_threshold`/`ik_rot_threshold` do double duty here: they're passed as cuRobo's *internal convergence* thresholds (constructor, lines 78-79) **and** as the *acceptance* thresholds in `check_ik_feasibility` (lines 110-111). Those are conceptually different knobs — convergence controls when the optimizer declares a seed solved; acceptance controls how tight a residual you'll count as reachable. Coupling them means you can't run a tighter solver with a looser gate (or vice-versa), and it silently overrides `StandaloneIKReachability`'s own defaults (0.005/0.05) with the looser 0.01/0.1. Consider separate params, or at least a comment noting the intentional coupling.

isaaclab_arena_curobo/curobo_ik_utils.py
```diff
@@ -0,0 +1,265 @@
...
+    """
+
+    dev = resolve_cuda_device(device)
+    robot_pos = torch.tensor(robot_base_pos_w, dtype=torch.float32, device=dev)
+    robot_quat = torch.tensor(robot_base_quat_w_xyzw, dtype=torch.float32, device=dev)
+
+    curobo_cuboids = []
+    for c in cuboids:
+        pos_w = torch.tensor(c.center_xyz, dtype=torch.float32, device=dev)
+        # StandaloneCuboid holds a wxyz world quat; the transform math works in xyzw.
```
zhx06: Comment rot: `StandaloneCuboid` isn't the type here — this is `AABBCollisionCuboid` (`c.quat_wxyz`). Rename in the comment to avoid confusion for the next reader.

zhx06 (commented):
Focused pass on the reachability-solver interface — i.e. how the IK gate validates the geometry-valid layouts. The plumbing (registry indirection, geometry-first short-circuit, reject-&-refill) is clean and keeps core vendor-agnostic. My main concern is that the current IK check is *pose-residual only* and does not yet enforce the collision world it goes to some trouble to build, so "IK reachable" doesn't actually validate the geometric/collision feasibility a grasp needs. Inline notes below.

isaaclab_arena/relations/pooled_object_placer.py
```diff
@@ -203,7 +219,8 @@ def _store_env_matched_results(
         fallback_envs = []
         for cur_env in range(self._num_envs):
             env_results = ranked_results_per_env[cur_env][:layouts_per_env]
-            valid_results = [r for r in env_results if r.success]
+            # if reachability validator is not set, results are only checked through geometry validation
+            valid_results = [r for r in env_results if r.success and self._passes_reachability_validator(r)]
```
zhx06: The expensive validator runs more than it needs to, in two ways:

1. **Already-full envs are re-validated every batch.** `_solve_and_store` loops up to `max_placement_attempts` batches, sizing each batch by the *least-full* env (`max_missing = target - min(available)`). But this line computes `valid_results` — and thus runs the IK validator — for *every* env, including ones where `missing <= 0` (the `else` branch on line 231 just bumps a counter and stores nothing). So an env that filled in batch 0 keeps paying a full round of cuRobo IK solves on every later batch until the slowest env catches up.

2. **More candidates validated than stored.** Even for a not-yet-full env you validate all `layouts_per_env` candidates, then keep only `valid_results[:missing]`.

Suggest gating on `missing` before validating and stopping once `missing` are accepted, e.g.:
```python
missing = target_num_layouts_per_env - self._env_pools[cur_env].available
if missing <= 0:
    continue  # env already satisfied — don't run the validator
valid_results = []
for r in env_results:
    if r.success and self._passes_reachability_validator(r):
        valid_results.append(r)
        if len(valid_results) >= missing:
            break
```
This keeps the `total_solved`/`total_valid` log honest for the envs that matter and can cut IK calls substantially when envs fill at different rates.

isaaclab_arena/relations/pooled_object_placer.py
```diff
@@ -203,7 +219,8 @@ def _store_env_matched_results(
         fallback_envs = []
         for cur_env in range(self._num_envs):
             env_results = ranked_results_per_env[cur_env][:layouts_per_env]
```
zhx06, line 204: When a `reachability_validator` is set, produce more geometry candidates than needed so the IK gate has fallbacks in one batch — otherwise unreachable top-`missing` candidates force a full geometry+cuRobo re-solve every refill. Concretely: pass `results_per_env = layouts_per_env * overfetch` (e.g. 4×, gated on `self._reachability_validator`) into `place_ranked_per_env`, drop this `[:layouts_per_env]` slice so the tail is reachable, and stop validating at `missing` (the `:223` early-stop). Geometry sampling is cheap next to a cuRobo solve, so this trades a bit of sampling for far fewer expensive batches.

isaaclab_arena/relations/object_placer_params.py
```diff
@@ -49,3 +57,12 @@ class ObjectPlacerParams:
 
     allow_best_loss_fallbacks: bool = True
     """Whether pooled placement may use best-loss layouts when no valid layout is found."""
+
+    reachability_validator: Callable[[PlacementResult], bool] | None = None
```
zhx06: The type advertises a pure predicate — `Callable[[PlacementResult], bool]` — but the cuRobo validator supplied by the extension **mutates** the argument: `make_standalone_ik_layout_validator` stamps a required `PlacementCheck.IK_REACHABLE` onto `result.validation_results` (flipping `result.success` when it rejects). A caller reading only this signature won't expect the passed result to change, and the stamped `.success=False` then conflates "geometry failed" with "IK rejected" for any downstream consumer that inspects the layout. Worth (a) documenting the side-effect here, and (b) confirming the `.success` conflation is intended — an audit reader can no longer tell a geometry failure from a reachability rejection.

zhx06 (commented):
Follow-up focused on the relation-solver side of the interface — where the gate actually plugs into `PooledObjectPlacer`. The predicate hook itself is a clean seam, but the placement of the call inside the refill loop has a real efficiency cost, and the `Callable[..., bool]` contract hides a mutation. Notes inline.

isaaclab_arena/relations/pooled_object_placer.py
zhx06: It seems `_passes_reachability_validator` is time-consuming. Can we check `missing` first to avoid extra computing?

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -59,6 +59,22 @@ def set_joint_initial_pos(self, joint_pos: Mapping[str, float]) -> None:
...
+        if self.initial_pose is not None:
+            return tuple(self.initial_pose.position_xyz), tuple(self.initial_pose.rotation_xyzw)
+        # otherwise use the robot's init_state
+        init_state = getattr(getattr(self.scene_config, "robot", None), "init_state", None)
+        if (
+            init_state is not None
+            and getattr(init_state, "pos", None) is not None
+            and getattr(init_state, "rot", None) is not None
+        ):
+            return tuple(float(v) for v in init_state.pos), tuple(float(v) for v in init_state.rot)
```
zhx06: **Quaternion-convention bug.** The return contract is `rotation_xyzw`, and the `initial_pose` branch honors it. But Isaac Lab's `init_state.rot` is **wxyz** (identity `(1,0,0,0)`), so returning it verbatim mislabels wxyz as xyzw — identity becomes `(1,0,0,0)` read as xyzw, i.e. a 180° rotation about X. The default fallback correctly uses xyzw identity `(0,0,0,1)`, which makes the mismatch concrete. `get_base_pose()` feeds `base_quat_xyzw` straight into the IK validator, so a non-identity robot base orientation gets corrupted. Convert:
```python
from isaaclab.utils.math import convert_quat
rot_xyzw = convert_quat(torch.tensor(init_state.rot), to="xyzw")
return tuple(float(v) for v in init_state.pos), tuple(float(v) for v in rot_xyzw)
```

isaaclab_arena/relations/reachability_validator_registry.py
```diff
@@ -0,0 +1,53 @@
...
+    Args:
+        embodiment: Embodiment the validator is built for.
+        module_names: Enabled extension modules to search.
+        kwargs: Extra tuning forwarded verbatim to the entry point (e.g. IK thresholds).
+    """
+    for module_name in module_names or []:
+        factory = getattr(importlib.import_module(module_name), _REACHABILITY_VALIDATOR_FACTORY_ATTR, None)
+        if factory is not None:
+            return factory(embodiment, **kwargs)
+    raise AssertionError(
```
zhx06: Use `RuntimeError`, not `AssertionError`, for "no extension provided a factory" — this is a real runtime/config error, and `assert`/`AssertionError` gets stripped under `python -O`, turning the friendly message into a bare `TypeError` later when `None` is called.

isaaclab_arena/relations/reachability_validator_registry.py
```diff
@@ -0,0 +1,53 @@
...
+    Imports each enabled extension (cached if already imported) and looks for a
+    ``make_reachability_validator(embodiment, **kwargs)`` entry point, calling the first one found.
+
+    Args:
+        embodiment: Embodiment the validator is built for.
+        module_names: Enabled extension modules to search.
+        kwargs: Extra tuning forwarded verbatim to the entry point (e.g. IK thresholds).
+    """
+    for module_name in module_names or []:
+        factory = getattr(importlib.import_module(module_name), _REACHABILITY_VALIDATOR_FACTORY_ATTR, None)
```
zhx06: `import_module` raises a raw `ImportError` when the extension's deps aren't installed (e.g. cuRobo missing), which is exactly the case the CLI help promises to handle ("make sure dev environment has curobo deps installed"). Wrap it so a missing extension is skipped or re-raised with that guidance instead of a bare `ImportError`:
```python
try:
    module = importlib.import_module(module_name)
except ImportError as e:
    raise RuntimeError(f"{module_name} enabled for --validate_reachability but import failed: {e}. Install its deps (e.g. cuRobo).") from e
```

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -95,6 +96,16 @@ def _solve_relations(self) -> None:
...
                 placer_params.resolve_on_reset = self.cfg.resolve_on_reset
+
+        # Gate pooled placement on reachability on demand.
+        if self.cfg.validate_reachability and placer_params.reachability_validator is None:
+            embodiment = self.arena_env.embodiment
+            assert embodiment is not None, "--validate_reachability requires an environment with a robot embodiment."
+            # Pull the gate exported by an enabled extension into the pool's solve loop. resolve_*
+            # imports the extension by name, so no cuRobo import appears in core.
+            placer_params.reachability_validator = resolve_reachability_validator(
+                embodiment, ["isaaclab_arena_curobo"], **self.cfg.reachability_validator_kwargs
```
zhx06: The extension list is hardcoded `["isaaclab_arena_curobo"]` here, which quietly contradicts the registry's stated design ("the module names come from config ... no cuRobo import appears in core"). The string is fine, but consider sourcing it from a config field (e.g. `self.cfg.reachability_validator_modules`) so core carries no vendor name and a different backend can be wired without editing the builder.

zhx06 (commented):
A few more fixable items on the core relation-solver side (the geometry path that feeds the gate), independent of cuRobo.

isaaclab_arena/environments/arena_env_builder.py
zhx06: Is it possible to make it an input to  `solve_and_apply_relation_placement`, so that we can also remove it from  `ObjectPlacerParams`? The reachability check is quite detached from the solver, so I don't have a very good suggestion at this moment

isaaclab_arena_curobo/ik_reachability_validator.py
xyao-nv: That will be addressed in following MR by disabling the hand link as a standard approach.

isaaclab_arena/embodiments/embodiment_base.py
xyao-nv: are you/your agent sure Lab's init_state.rot is wxyz?

isaaclab_arena_curobo/ik_reachability_validator.py
xyao-nv: I don't think it's an issue.

isaaclab_arena_curobo/curobo_ik_utils.py
xyao-nv: done

isaaclab_arena_curobo/curobo_frame_utils.py
xyao-nv: - Current (argmin): it first ranks seeds by position error alone, picks that one seed, then checks both position and rotation on it. So the rotation test is applied to whatever seed happened to win on position.
  - Yours (.any): each seed must clear both thresholds itself; a pose is feasible if any seed does.

Actually internally for Curobo, right now.
It optimizes num_seeds (12) in parallel, but it returns  only the top return_seeds = 1 solution per pose — has already reduced to its own single best seed.

I can adopt yours.

isaaclab_arena_curobo/curobo_frame_utils.py
xyao-nv: it is a followup

isaaclab_arena/relations/pooled_object_placer.py
xyao-nv: Well it shall be another improvement MR and you can add it according to your agent suggestion :)

isaaclab_arena_curobo/ik_reachability_validator.py
```diff
@@ -0,0 +1,126 @@
...
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Build-time cuRobo IK-reachability gate for pooled placement, standalone from SimApp.
+
+The pool's solve loop calls it on each geometry-valid candidate; a candidate is stored only when the robot can reach a
+top-down grasp at every movable object, so the loop keeps solving (reject-&-refill) until every env has enough reachable layouts.
+
+Requires a CUDA GPU.
```
alexmillane: I think our whole framework requires a CUDA GPU right? Suggestion to remove this?

isaaclab_arena/cli/isaaclab_arena_cli.py
```diff
@@ -111,6 +111,16 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
...
+    arena_group.add_argument(
+        "--validate_reachability",
+        action="store_true",
+        default=False,
+        help=(
+            "Gate build-time pooled placement on reachability, storing only layouts whose objects the"
+            " robot can reach (cuRobo top-down-grasp IK). isaaclab_arena_curobo is automatically imported"
+            " when this flag is set. Please make sure dev environment has curobo deps installed."
+        ),
+    )
```
alexmillane: Is there some way that we can enable this on a per-environment level so a user needn't worry about whether or not curobo validation is required? This shifts the responsibility from the person running the environment to the person creating the environment, which makes more sense to me.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -59,6 +59,22 @@ def set_joint_initial_pos(self, joint_pos: Mapping[str, float]) -> None:
             raise RuntimeError("scene_config must be populated with a `robot` before calling `set_joint_initial_pos`.")
         self.scene_config.robot.init_state.joint_pos.update(joint_pos)
 
+    def get_base_pose(self) -> tuple[tuple[float, float, float], tuple[float, float, float, float]]:
```
alexmillane: elsewhere we use our `Pose` object to communicate poses. Any reason to do something different here?

If you need the pose as tuples, suggestion to add a conversion to our pose object.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -95,6 +96,16 @@ def _solve_relations(self) -> None:
...
+            )
```
alexmillane: I feel like this validator (actually, all validators) should only be visible inside the solver. They should be activated in the config, and then the solver should sort out how they're exectude. Having one particular validator appearing on the top-level in env-builder seems a bit off.

Is there some reason that we need this here? One reason is that currently this is controlled by a flag. See above for my comment on that.

isaaclab_arena/environments/arena_env_builder_cfg.py
```diff
@@ -27,5 +28,13 @@ class ArenaEnvBuilderCfg:
...
+    """Gate build-time pooled placement on reachability, storing only layouts whose objects the robot can
+    reach (cuRobo top-down-grasp IK, supplied by the cuRobo extension).."""
+
+    reachability_validator_kwargs: dict[str, Any] = field(default_factory=dict)
+    """Keyword tuning forwarded to the reachability-validator factory (e.g. IK thresholds). Ignored when
+    ``validate_reachability`` is ``False`` or a callable was set directly on the placer params."""
```
alexmillane: Are we in control of the validator, or is this a curobo thing? If this is us, could we pass the kwargs as a dataclass. A dataclass communicates valid args to the outside word. With a dict, a user has no idea what options are available. They just put some stuff in, maybe it works, maybe it crashes.

isaaclab_arena/relations/object_placer_params.py
alexmillane: Do we need to have the actual callable in the params struct?

I would have imagined that instead we have in the param struct the flag that activates the check, and maybe some params, but then the placer is in charge of actually calling the checker.

isaaclab_arena/relations/pooled_object_placer.py
```diff
@@ -185,6 +190,17 @@ def _solve_env_ranked_layouts(self, num_layouts: int) -> tuple[list[list[Placeme
 
         return ranked_results_per_env, layouts_per_env
 
+    def _passes_reachability_validator(self, result: PlacementResult) -> bool:
+        """Whether *result* is accepted by the optional build-time validator (True when none is set).
+
+        Called only on geometry-valid candidates (the caller short-circuits on ``result.success``), so
+        an expensive validator such as the cuRobo IK-reachability check never runs on a layout that
+        already failed geometry.
```
alexmillane: This docstring details what the caller does before calling this function.

Suggestion to keep the docstrings only detail what the function they're documenting does.

isaaclab_arena/relations/pooled_object_placer.py
alexmillane: Seems to me like we're doing validation in a function that is labeled as for storing results.

I would suggest moving the validation to a function about validation.

As far as I recall, we do other types of validation on the results. Am I correct there? Could we do the curobo validation in the same place as the other validation?

isaaclab_arena_curobo/ik_reachability_validator.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def make_standalone_ik_layout_validator(
+    embodiment: EmbodimentBase,
+    grasp_z_offset: float = 0.02,
+    ik_pos_threshold: float = 0.01,
+    ik_rot_threshold: float = 0.1,
+    device: str | torch.device | None = None,
+    stamp_results: bool = True,
+) -> Callable[[PlacementResult], bool]:
```
cvolkcvolk: I think the cuRobo validator should remain read-only with respect to Arena-owned state such as PlacementResult. Could it simply return whether the layout is reachable and let for example PooledObjectPlacer record the result?

isaaclab_arena/relations/reachability_validator_registry.py
```diff
@@ -0,0 +1,53 @@
...
+        if factory is not None:
+            return factory(embodiment, **kwargs)
+    raise AssertionError(
+        f"No enabled extension exports {_REACHABILITY_VALIDATOR_FACTORY_ATTR}(); "
+        "enable one that provides the build-time reachability gate (e.g. 'isaaclab_arena_curobo')."
+    )
```
alexmillane: I guess that the module name is statically know right? like this function wont work unless `mudule_names=[isaaclab_arena_curobo]`

isaaclab_arena_curobo/curobo_frame_utils.py
```diff
@@ -0,0 +1,160 @@
...
+        R_grasp = R_down
+
+    q_grasp_xyzw = math_utils.quat_from_matrix(R_grasp.unsqueeze(0))[0]
+    pose = Pose(position_xyz=tuple(t_R_O.tolist()), rotation_xyzw=tuple(q_grasp_xyzw.tolist()))
+    return pose.to_transform_matrix(device)
+
+
+def solve_ik_feasibility(
+    pose_ctx,
+    ik_solver,
```
alexmillane: Could we type annotate these?

isaaclab_arena_curobo/curobo_frame_utils.py
```diff
@@ -0,0 +1,160 @@
...
+    seed_config: torch.Tensor | None = None,
+    position_threshold: float = 0.01,
+    rotation_threshold: float = 0.1,
+    require_collision_free: bool = False,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """Batched IK feasibility of all ``target_poses`` against a cuRobo IK solver. Shared between standalone and env-coupled paths.
+    Runs a single ``solve_batch`` for one layout.
+
+    Args:
+        pose_ctx: Object providing ``_to_curobo_device``, ``_make_pose`` and ``logger``. A ``CuroboPlanner`` or a ``StandaloneIKReachability``.
```
alexmillane: I'm not sure I understand the naming here. Sounds like it's a pose?

What are the differences between the options `CuroboPlanner` and `StandaloneIKReachability`?

Are these objects under our control? If so perhaps we could add a class specifying the requirements on their interface?

isaaclab_arena_curobo/curobo_frame_utils.py
```diff
@@ -0,0 +1,160 @@
...
+        ik_solver: cuRobo ``IKSolver`` exposing ``solve_batch``.
```
alexmillane: I'm a little surprised that we seem to pass in two curobo solver-related objects.

`pose_ctx: CuroboPlanner`

and 

`ik_solver: IKSolver`

What's the difference here? Why do we need so planner/solvers to solve the feasibility problem?

isaaclab_arena_curobo/curobo_ik_utils.py
```diff
@@ -0,0 +1,281 @@
...
+    """A collision obstacle described by an axis-aligned bounding box in the world frame.
+
+    ``dims_xyz`` are full extents (edge lengths), matching cuRobo's ``Cuboid.dims``. ``quat_wxyz`` is
+    the box orientation in the world frame; for an axis-aligned box built from a layout it is identity.
+    """
+
+    name: str
+    center_xyz: tuple[float, float, float]
+    dims_xyz: tuple[float, float, float]
+    quat_wxyz: tuple[float, float, float, float] = (1.0, 0.0, 0.0, 0.0)
```
alexmillane: elsewhere in the framework we use xyzw. I suppose that this is because curobo uses wxyz?

I would tend to stick to the isaaclab convention within our code, and then convert at the last moment.

alexmillane: Could we combine the center and orientation into one of our pose objects?

isaaclab_arena_curobo/ik_reachability_validator.py
```diff
@@ -0,0 +1,126 @@
...
+    from isaaclab_arena.assets.object_base import ObjectBase
+    from isaaclab_arena.embodiments.embodiment_base import EmbodimentBase
+    from isaaclab_arena.relations.placement_result import PlacementResult
+
+
+def get_object_world_pose_from_layout(
+    result: PlacementResult,
+    obj: ObjectBase,
+    base_rotations: dict,
+) -> tuple[tuple[float, float, float], tuple[float, ...]]:
```
alexmillane: Maybe this is a limitation of the sovler, but I'm wondering why we're passing around `base_rotations` separately.

Could we not combine the rotations into the `PlacementResult` and treat the solution as one complete thing. Not something that we need to add some random rotations to?

alexmillane: Ah maybe this is because were trying to stay away from kit?

isaaclab_arena_curobo/curobo_ik_utils.py
```diff
@@ -0,0 +1,281 @@
...
+    t_W_O = torch.tensor(object_pos_w, dtype=torch.float32, device=dev)
+    q_W_O_xyzw = torch.tensor(object_quat_w_xyzw, dtype=torch.float32, device=dev)
+    t_W_R = torch.tensor(robot_base_pos_w, dtype=torch.float32, device=dev)
+    q_W_R_xyzw = torch.tensor(robot_base_quat_w_xyzw, dtype=torch.float32, device=dev)
+
+    t_R_O, q_R_O_xyzw = world_pose_to_robot_frame(t_W_O, q_W_O_xyzw, t_W_R, q_W_R_xyzw)
+    return top_down_grasp_matrix(t_R_O, q_R_O_xyzw, grasp_z_offset, align_yaw_to_object)
+
+
+class StandaloneIKReachability:
```
alexmillane: How does this class relate to the non-standalone version? Presumably, they do very similar things? Do they currently follow completely different code paths?

Is there some way we could unify them. For example the kit-requiring version can hold one of these?

alexmillane (commented):
Thanks for putting this together. Great that it's working!

I've left a few comments. I think we need to investigate the class heirachy and see how we can build this into the solver such that's not passed in from the top level.

isaaclab_arena_curobo/curobo_ik_utils.py
xyao-nv: Yes curobo uses wxyz. Sure

isaaclab_arena_curobo/ik_reachability_validator.py
xyao-nv: hmm, isn't the solver still runnable on CPU? But this is too verbal for a validator, removed.

isaaclab_arena/embodiments/embodiment_base.py
xyao-nv: sure, updated.

isaaclab_arena_curobo/curobo_ik_utils.py
xyao-nv: done

isaaclab_arena_curobo/curobo_frame_utils.py
xyao-nv: added

isaaclab_arena_curobo/curobo_frame_utils.py
xyao-nv: renamed it to ik_context.

Both are under ctrl by us. Its one for simfree and one for env-coupled. Added an explicit interface.

isaaclab_arena_curobo/curobo_frame_utils.py
xyao-nv: actually ik solver shall be fetchable thur the ctx. See updates.

isaaclab_arena/environments/arena_env_builder.py
xyao-nv: I added enable_checks/required_checks in placer params. And once set and curobo deps loadable, the validator will be turned on.

isaaclab_arena/relations/object_placer_params.py
xyao-nv: outdated

isaaclab_arena/cli/isaaclab_arena_cli.py
xyao-nv: yes now it's enabled thru required_checks and enabled_checks.

And see butter_raisin_box.yaml env for example.

isaaclab_arena/environments/arena_env_builder_cfg.py
xyao-nv: Yes introduced a reachabilityConfig class that capsulates a few hyper-params for IK.

isaaclab_arena_curobo/curobo_ik_utils.py
xyao-nv: Sim-free vs env coupled. Main diff is where env obs (live vs. cuboids) come from and whether it pulls modules more than IK solver, like motion gen related funcs.

I tried to unify most of the IK compute frame utils funcs. Unfortunately it has to be another class as i don't have ctrl over CuroboPlanner class.

isaaclab_arena/relations/object_placer_params.py
```diff
@@ -3,11 +3,34 @@
...
 
-__all__ = ["CollisionMode", "ObjectPlacerParams"]
+if TYPE_CHECKING:
+    from isaaclab_arena.relations.placement_validators import PlacementValidator
+
+__all__ = ["CollisionMode", "ObjectPlacerParams", "ReachabilityConfig"]
+
+
+@dataclass
+class ReachabilityConfig:
```
xyao-nv, line 18: Note on outdated comments: IK solver related params can be configured by users thru this dataclass.

isaaclab_arena/relations/object_placer_params.py
```diff
@@ -57,3 +80,9 @@ class ObjectPlacerParams:
     required_checks: set[str] | None = None
     """Check names that must pass for a layout to count as valid (gates rejection/refill in the pool).
     None requires every enabled check; otherwise should be a subset of enabled_checks."""
+
+    reachability_config: ReachabilityConfig = field(default_factory=ReachabilityConfig)
+    """Tuning for the optional ``ik_reachable`` build-time check. See ReachabilityConfig for more details."""
+
+    extra_validators: list[PlacementValidator] | None = None
```
xyao-nv: Note: register on the fly because Curobo dep is not installed in the base image/uv env.

isaaclab_arena_environments/robolab/scenes/butter_raisin_box.yaml
```diff
@@ -18,6 +18,9 @@ objects:
 - id: raisin_box_hope_robolab
   registry_name: raisin_box_hope_robolab
   params: {}
+placement_validators:
+  enabled_checks: [no_overlap, on_relation, ik_reachable]
```
xyao-nv, line 22: Note on outdated comments: Can be propagated to other envs once this is merged.

isaaclab_arena_curobo/ik_reachability_validator.py
```diff
@@ -0,0 +1,211 @@
...
+    Args:
+        embodiment: Embodiment who has a registered CuroboEmbodimentCfg.
+        grasp_z_offset: Height (m) above each object's root for the top-down grasp pose.
+        ik_pos_threshold: Max IK position error (m) for a grasp to count as reachable.
+        ik_rot_threshold: Max IK rotation error (rad) for a grasp to count as reachable.
+        device: Explicit CUDA device (e.g. ``"cuda:0"``); defaults to the current device.
+        stamp_results: Record the verdict as a required ``PlacementCheck.IK_REACHABLE`` check on each
+            layout, so it shows up in the audit report and gates that layout's ``.success``.
+    """
+    solver = SimFreeIKSolver(
```
xyao-nv: Note on outdated comments: The IK solver between env-coupled (from CuroboPlanner) and sim-free one used in build time is different as it requires world obs either come from env or cuboids. CuroboPlanner also introduced motion_gen related modules.
And CuroboPlanner is set in Lab and I don't we ctrl over it.

isaaclab_arena_curobo/ik_reachability_validator.py
xyao-nv: The reason base_rotations is separate is the current PlacementResult orientation model: result.orientations[obj] is only a scalar absolute Z-yaw — that's all the placer solves. The object's roll/pitch lives in a separate full quaternion, the RotateAroundSolution marker (base_rotations). In fact objects with marker roll/pitch are deliberately skipped in orientations, so today orientations alone can't represent their full pose. The final world orientation is base_quat with its yaw overridden by the solved total_yaw — hence combine the two.

I can consume the one from PlacementResult once the refactor is done post v0.3.

isaaclab_arena/relations/pooled_object_placer.py
xyao-nv: outdated but addressed

isaaclab_arena/relations/reachability_validator_registry.py
xyao-nv: outdated

isaaclab_arena_curobo/ik_reachability_validator.py
xyao-nv: it's how other validators are writing the results and having Placer to handle aggregating the results seems a bit too much to me.

isaaclab_arena/environments/arena_env_builder.py
xyao-nv: Your proposal has been implemented.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -59,6 +59,25 @@ def set_joint_initial_pos(self, joint_pos: Mapping[str, float]) -> None:
             raise RuntimeError("scene_config must be populated with a `robot` before calling `set_joint_initial_pos`.")
         self.scene_config.robot.init_state.joint_pos.update(joint_pos)
 
+    def get_base_pose(self) -> Pose:
```
qianl-nv: not for this MR, but we need to keep in mind @zhx06 's MR https://github.com/isaac-sim/IsaacLab-Arena/pull/915 to enable robot placement by solver, in which case curobo needs to get the solved pose of the robot.

isaaclab_arena/relations/object_placer_params.py
```diff
@@ -3,11 +3,34 @@
...
 from dataclasses import dataclass, field
+from typing import TYPE_CHECKING
 
 from isaaclab_arena.relations.relation_solver_params import CollisionMode, RelationSolverParams
 
-__all__ = ["CollisionMode", "ObjectPlacerParams"]
+if TYPE_CHECKING:
+    from isaaclab_arena.relations.placement_validators import PlacementValidator
+
+__all__ = ["CollisionMode", "ObjectPlacerParams", "ReachabilityConfig"]
```
qianl-nv: is this re-export required? suggest to remove if possible, we don't use this anywhere else in arena

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -59,6 +59,25 @@ def set_joint_initial_pos(self, joint_pos: Mapping[str, float]) -> None:
...
+            init_state is not None
+            and getattr(init_state, "pos", None) is not None
+            and getattr(init_state, "rot", None) is not None
+        ):
+            return Pose(
+                position_xyz=tuple(float(v) for v in init_state.pos),
+                rotation_xyzw=tuple(float(v) for v in init_state.rot),
+            )
+        # default to origin and identity
+        return Pose.identity()
```
alexmillane: Suggestion to return `None` rather than identity if the embodiment doesn't have a pose set.

We should leave it up to the caller to decide what is appropriate in the case of no pose being set.

Alternately, if the true default is identity. As in isaaclab initializes at identity if there's None here, there may be an argument to initialize `self.initial_pose=Pose.identity()` to capture that.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -59,6 +59,25 @@ def set_joint_initial_pos(self, joint_pos: Mapping[str, float]) -> None:
             raise RuntimeError("scene_config must be populated with a `robot` before calling `set_joint_initial_pos`.")
         self.scene_config.robot.init_state.joint_pos.update(joint_pos)
 
+    def get_base_pose(self) -> Pose:
+        """Env-local robot base pose, preferring ``initial_pose``, then the robot's init_state, then identity."""
+        # initial_pose if set
+        if self.initial_pose is not None:
+            return self.initial_pose
+        # otherwise use the robot's init_state
+        init_state = getattr(getattr(self.scene_config, "robot", None), "init_state", None)
```
alexmillane: Let's assert `asset hasattr(self.scene_config, "robot")`.

the access unconditionally.

All valid embodiments should define this no?

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -59,6 +59,25 @@ def set_joint_initial_pos(self, joint_pos: Mapping[str, float]) -> None:
...
+        if (
+            init_state is not None
+            and getattr(init_state, "pos", None) is not None
+            and getattr(init_state, "rot", None) is not None
+        ):
+            return Pose(
+                position_xyz=tuple(float(v) for v in init_state.pos),
+                rotation_xyzw=tuple(float(v) for v in init_state.rot),
+            )
```
alexmillane: Are their valid cases where this would occur? That the initial pose is None, but we have a non-None pose on the robot.

I would think that that's maybe an issue? It's certainly confusing?

I would actually add this as an invariant to our class to be enforced on construction, and checked whenever someone sets the pose?

isaaclab_arena/embodiments/embodiment_base.py
alexmillane: What's the "base" in base pose? I would suggest to change this to initial pose, which is what we call it above.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -96,6 +97,9 @@ def _solve_relations(self) -> None:
             placer_params.placement_seed = self.cfg.placement_seed
         if self.cfg.resolve_on_reset is not None:
             placer_params.resolve_on_reset = self.cfg.resolve_on_reset
+
+        # No-ops unless the embodiment has a registered cuRobo config and the solver deps are importable.
+        configure_reachability_validator(placer_params, self.arena_env.embodiment)
```
alexmillane: I'm wondering why we configure the reachability validator in the arena env builder? Could we push this deeper to where the other validators live? Or is there some reason we need this here?

isaaclab_arena/relations/object_placer.py
```diff
@@ -592,12 +592,40 @@ def _validate_candidates(
...
-        }
         # required_checks=None means "every enabled check is required"; an empty set means no checks.
         required = self.params.required_checks
+        num_candidates = len(positions)
+        # Per-check count of layouts evaluated by that check: all candidates for non-gated checks, only the
+        # geometry-passing subset for gated ones (they skip candidates already rejected on cheaper checks).
+        evaluated_by_check: dict[str, int] = {}
+        verdicts_by_check: dict[str, list[bool]] = {}
+        for validator in self._validators:
+            if not validator.gated:
```
alexmillane: What does gated mean? Does this mean that the check is disabled? Suggestion to use `validator.enabled`. Or maybe it means something different?

isaaclab_arena/relations/object_placer.py
```diff
@@ -592,12 +592,40 @@ def _validate_candidates(
...
+                    [positions[i] for i in kept],
+                    [orientations[i] for i in kept],
+                    [bboxes[i] for i in kept],
+                    collision_objects,
+                )
+                verdicts = [False] * num_candidates
+                for sub_idx, cand_idx in enumerate(kept):
+                    verdicts[cand_idx] = sub_verdicts[sub_idx]
+                verdicts_by_check[validator.check] = verdicts
+                evaluated_by_check[validator.check] = len(kept)
```
alexmillane: This is difficult to read code because we have nested for sequential and nested for loops.

Can we break these out into a function. I would also recommend adding some comments here to say what is going on. Could we perhaps combine these into a single loop. and add some comments. it's really not clear what's happening.

isaaclab_arena/relations/object_placer_params.py
```diff
@@ -57,3 +80,9 @@ class ObjectPlacerParams:
     required_checks: set[str] | None = None
     """Check names that must pass for a layout to count as valid (gates rejection/refill in the pool).
     None requires every enabled check; otherwise should be a subset of enabled_checks."""
+
+    reachability_config: ReachabilityConfig = field(default_factory=ReachabilityConfig)
+    """Tuning for the optional ``ik_reachable`` build-time check. See ReachabilityConfig for more details."""
+
+    extra_validators: list[PlacementValidator] = field(default_factory=list)
+    """Pre-built validators to run alongside the registered ones, in case they need to be injected by integration code."""
```
alexmillane: " case they need to be injected by integration code"? What does that mean?

isaaclab_arena/relations/placement_validators.py
```diff
@@ -43,9 +43,18 @@ class PlacementValidator(ABC):
     """The check name this validator reports; its registry key and result key. Built-ins use a
     PlacementCheck constant; external validators may use any unique string."""
 
+    gated: ClassVar[bool] = False
+    """If True, run this validator only on candidates that already pass every required non-gated check,
+    so an expensive check (e.g. IK reachability) never runs on a layout rejected on cheaper geometry."""
```
alexmillane: Let's choose a variable name that can be straightforwardly interpretted.

Suggestion `run_after_inexpensive_checks` or something?

isaaclab_arena/relations/placement_validators.py
```diff
@@ -73,28 +82,46 @@ def get_build_time_checks() -> tuple[str, ...]:
...
+        )
+        registered_checks = tuple(check for check in registered if check in enabled_checks)
+        extra_validators = [v for v in params.extra_validators if v.check in enabled_checks]
+    else:
+        # Default: run every registered check that opts in thru enabled_by_default, plus every params.extra_validators.
+        registered_checks = tuple(
+            check for check in registered if registry.get_validator_by_name(check).enabled_by_default(params)
         )
-        checks = tuple(check for check in checks if check in enabled_checks)
+        extra_validators = list(params.extra_validators)
```
alexmillane: This is a bit confusing. Am I correct that either:
- the user requests enabled_checks in which case we use those plus the extras, or
- the user doesn't write enabled_checks, in which case we use the default enabled checks plus the extras.

The main flexibility this affords us is to turn off some of the default validators (by setting enabled_checks).

Is this required at this stage? Could we just leave the defaults always on and just add the extras if requested. 

That will make things a bit easier.

isaaclab_arena/relations/placement_validators.py
```diff
@@ -43,9 +43,18 @@ class PlacementValidator(ABC):
...
     PlacementCheck constant; external validators may use any unique string."""
 
+    gated: ClassVar[bool] = False
+    """If True, run this validator only on candidates that already pass every required non-gated check,
+    so an expensive check (e.g. IK reachability) never runs on a layout rejected on cheaper geometry."""
+
     def __init__(self, params: ObjectPlacerParams) -> None:
         self._params = params
 
+    @classmethod
```
qianl-nv, line 54: this is never overwritten by any validator classes and the params argument is unused. 
do we need to add it in this MR?

isaaclab_arena_curobo/curobo_ik_utils.py
```diff
@@ -0,0 +1,255 @@
...
+    t_W_O = torch.tensor(object_pos_w, dtype=torch.float32, device=dev)
+    q_W_O_xyzw = torch.tensor(object_quat_w_xyzw, dtype=torch.float32, device=dev)
+    t_W_R = torch.tensor(robot_base_pos_w, dtype=torch.float32, device=dev)
+    q_W_R_xyzw = torch.tensor(robot_base_quat_w_xyzw, dtype=torch.float32, device=dev)
+
+    t_R_O, q_R_O_xyzw = world_pose_to_robot_frame(t_W_O, q_W_O_xyzw, t_W_R, q_W_R_xyzw)
+    return top_down_grasp_matrix(t_R_O, q_R_O_xyzw, grasp_z_offset, align_yaw_to_object)
+
+
+class SimFreeIKSolver:
```
qianl-nv: How about we just call this CuroboIKSolver? The counter part CuroboPlanner doesn't mention Sim in the name, so this naming SimFree just brings more questions.
The doc string below already explains this is sim-free

isaaclab_arena_curobo/curobo_ik_utils.py
```diff
@@ -0,0 +1,255 @@
...
+        moving obstacle poses, so this handles both moved objects and a changed object set per layout,
+        as long as the obstacle count stays within the collision cache.
+        """
+        world_cfg = world_config_from_cuboids(cuboids, robot_base_pos_w, robot_base_quat_w_xyzw, self.device)
+        self.ik_solver.update_world(world_cfg)
+        if torch.cuda.is_available():
+            torch.cuda.synchronize()
+
+
+def check_ik_feasibility(
```
qianl-nv: this is a very thin wrapper around solve_ik_feasibility. can we just merge the two and let user call solve_ik_feasibility instead?

qianl-nv (commented):
Small nits. Will check again when current comments are all addressed. Thanks!

isaaclab_arena_curobo/curobo_frame_utils.py
```diff
@@ -0,0 +1,189 @@
...
+class IKSolverContext(Protocol):
+    """The host that owns a cuRobo IK solver plus the device/pose plumbing to drive it."""
+
+    logger: logging.Logger
+
+    def _to_curobo_device(self, tensor: torch.Tensor) -> torch.Tensor:
+        """Move a tensor onto the cuRobo device/dtype."""
+
+    def _make_pose(self, position: torch.Tensor, quaternion: torch.Tensor, *, quat_is_xyzw: bool = True) -> CuroboPose:
+        """Build a cuRobo ``Pose`` on the cuRobo device."""
```
alexmillane: Is the idea here to require some interface of different solvers?

That makes sense, but the interface here seems pretty slim. It doesn't really enforce anything about being a `IKSolver` even though it's called `IKSolverContext`.

I feel like either we enforce more, like that things conforming to this protocol can be used as an IKSolver. Another option is just removing this class. As it is, it doesn't appear to do much.

isaaclab_arena_curobo/curobo_frame_utils.py
```diff
@@ -0,0 +1,189 @@
...
+def resolve_ik_solver(ik_solver_context: IKSolverContext) -> IKSolver:
+    """Get the cuRobo ``IKSolver`` from a host, whichever way it exposes it.
+
+    ``SimFreeIKSolver`` holds it as ``ik_solver``; the upstream ``CuroboPlanner`` (not ours to change)
+    holds it as ``motion_gen.ik_solver``. Centralizing the lookup lets the solve take just the host.
+    """
+    ik_solver = getattr(ik_solver_context, "ik_solver", None)
+    if ik_solver is None:
+        ik_solver = ik_solver_context.motion_gen.ik_solver
+    return ik_solver
```
alexmillane: This seems like a strange function.

We take a Protocol (i.e. something that defines an interface), but then touch a member that is not necessarily defined by that protocol.

I'm wondering if the type annotation should be changed to something like `SimFreeIKSolver | CuroboPlanner`

isaaclab_arena_curobo/curobo_frame_utils.py
```diff
@@ -0,0 +1,189 @@
...
+
+def solve_ik_feasibility(
+    ik_solver_context: IKSolverContext,
+    target_poses: torch.Tensor,
+    seed_config: torch.Tensor | None = None,
+    position_threshold: float = 0.01,
+    rotation_threshold: float = 0.1,
+    require_collision_free: bool = False,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """Batched IK feasibility of all ``target_poses`` against a cuRobo IK solver. Shared between sim-free and env-coupled paths.
```
alexmillane: Some of the functions in this file don't seem to fit to the definition of "frame utils" (some of them are indeed to do with rigid transforms/frames.

Should we break this file up into stuff to do with frames and those which are not?

alexmillane: This seems like kind of a main entrypoint "solve_ik_feasibility" but is hidden in the "frame_utils" file.

I'm wondering if we should work on the structure of functions, files, and directories in this module a bit.

isaaclab_arena_curobo/curobo_ik_utils.py
```diff
@@ -0,0 +1,255 @@
...
+class AABBCollisionCuboid:
+    """A collision obstacle described by an axis-aligned bounding box in the world frame.
+
+    ``dims_xyz`` are full extents (edge lengths), matching cuRobo's ``Cuboid.dims``. ``pose`` is the
+    box pose (center + orientation) in the world frame.
+    """
+
+    name: str
+    dims_xyz: tuple[float, float, float]
+    pose: Pose = field(default_factory=Pose.identity)
```
alexmillane: Suggestion to add frame annotations here.

Perhaps: `pose_W_O` i.e. the pose that transforms a point in the object frame to the world frame.

isaaclab_arena_curobo/curobo_ik_utils.py
alexmillane: Suggestion to move this class to its own file.

alexmillane (commented):
Review round 2.

Looking good. I have quite a few comments and questions.

Happy to have another look when you're ready.

isaaclab_arena/embodiments/embodiment_base.py
xyao-nv: thx will update once it's merged. Added TODO.

isaaclab_arena/embodiments/embodiment_base.py
xyao-nv: yes, robot is required in scenen cfg. Added.

isaaclab_arena/embodiments/embodiment_base.py
xyao-nv: Since Lab initializes initial-state to be identify, so it's never the case of None for us to catch. Removed it.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -47,6 +47,7 @@
 from isaaclab_arena.variations import variations_hydra, variations_printing
 from isaaclab_arena.variations.variation_base import RunTimeVariationBase, VariationBase
 from isaaclab_arena.variations.variation_recorder import VariationRecorder
+from isaaclab_arena_curobo import configure_reachability_validator
```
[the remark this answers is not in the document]
xyao-nv: it's imported but if deps do not exist, it will skip it. I think it's fine.

isaaclab_arena/relations/placement_validators.py
xyao-nv: updated

isaaclab_arena/embodiments/embodiment_base.py
xyao-nv: I coded it as in droid.py `initial_state.pos != initial_pose.position_xyz` due to a change I recetly added in https://github.com/isaac-sim/IsaacLab-Arena/pull/911 about `stand_height_m`.
But now I regretted this change, and I made changes to droid.py to make these two fields same value in that MR https://github.com/isaac-sim/IsaacLab-Arena/pull/945.

isaaclab_arena/embodiments/embodiment_base.py
xyao-nv: See droid.py changes to ensure `initial_state.pos == initial_pose.position_xyz`. With that, Curobo ik solver just need robot base's initial pose.

isaaclab_arena/relations/object_placer.py
xyao-nv: Now it's called run_after_inexpensive_checks.

isaaclab_arena/embodiments/embodiment_base.py
xyao-nv: Switched to `get_initial_pose` and included in docstring fallback to initial_state.pos if not set.

isaaclab_arena/relations/object_placer.py
xyao-nv: sorry about that -- added comments, renamed vars, broke into small funcs.

isaaclab_arena_curobo/curobo_ik_utils.py
xyao-nv: gotcha, added.

isaaclab_arena_curobo/curobo_ik_utils.py
xyao-nv: added, done

xyao-nv: done

isaaclab_arena_curobo/curobo_frame_utils.py
xyao-nv: Moved it out, and cleaned up the file structure

isaaclab_arena/relations/object_placer_params.py
xyao-nv: done

isaaclab_arena/relations/placement_validators.py
xyao-nv: removed

isaaclab_arena/relations/placement_validators.py
xyao-nv: I'm sorry it confused you. The major factors causing this level of complexity are Curobo is not in base image by default, and some embodiment may not come with Curobo cfg. I was thinking given the actual realization is gated by these 2 pre-reqs, then having it in `extra_validators` to register at run-time could be a way.

Meaning, if pre-reqs are met, bring it to `extra_validator.` And it won't block the solver even if user want reachability in required/enabled checks but its pre-reqs are not met.

Now given how much it confuses, I simplify (hopefully) it to below.
 
`ik_reachable` is now a first-class registered check. It will self-delist at build time, if the embodiment has a registered cuRobo config and the solver deps are importable. And all extra-validator logics are removed.

isaaclab_arena/relations/object_placer_params.py
xyao-nv: meaning if curobo reqs are met (integration code), then inject (add to extra_validators).

But this is removed now, reachability will auto self-delist if pre-reqs are not met.

isaaclab_arena/relations/object_placer.py
```diff
@@ -592,12 +592,40 @@ def _validate_candidates(
...
+        if verdicts_by_check:
+            summary = ", ".join(
+                f"{check}={sum(verdicts)}/{evaluated_by_check[check]}" for check, verdicts in verdicts_by_check.items()
+            )
+            print(f"[placement] Validated {num_candidates} candidate layout(s); passed per check: {summary}")
```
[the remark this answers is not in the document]
xyao-nv: can go in another MR

isaaclab_arena/environments/arena_env_builder.py
xyao-nv: It was a hack around checking if reachability pre-reqs are actually met. See below thread on what it was, why it was a bad idea, and what I have updated to.

isaaclab_arena_environments/robolab/scenes/butter_raisin_box.yaml
```diff
@@ -18,6 +18,9 @@ objects:
...
+  required_checks: [no_overlap, on_relation]
```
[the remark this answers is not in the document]
xyao-nv: no gating as base image does not have curobo

isaaclab_arena_curobo/curobo_ik_utils.py
xyao-nv: done

isaaclab_arena_curobo/curobo_frame_utils.py
xyao-nv: yeah it's inappropriate to call it as a protocol under this case.

Just rename the type annotation to reflect the change of member.

isaaclab_arena_curobo/curobo_frame_utils.py
xyao-nv: It's indeed not that much and removing this class makes sense to me. The extra plumbs between sim-free IKSolver, vs env-coupled CuroboPlanner.motion_gen.ik_solver is not a lot honestly.

Just use type_annotation to distinguish these two instead.

isaaclab_arena_curobo/ik_reachability_validator.py
```diff
@@ -0,0 +1,188 @@
...
+                position_threshold=ik_pos_threshold,
+                rotation_threshold=ik_rot_threshold,
+            )
+            reachable = bool(feasible.all().item())
+
+        if stamp_results:
+            result.validation_results.add_validation_check(PlacementCheck.IK_REACHABLE, reachable, required=True)
+        return reachable
+
+    return validator
```
alexmillane: What do you think about making this a class, with a `__call__` operator such that it conforms to the validator interface. I think that that would sort of make sense. We would encapsulate the surrounding function, which serves to capture some parameters/state, in the class constructor, which is a more natural way to do things.

alexmillane: Ah I see there is a class below. See comments there.

isaaclab_arena_curobo/ik_reachability_validator.py
```diff
@@ -0,0 +1,188 @@
...
+    def __init__(self, params: ObjectPlacerParams) -> None:
+        super().__init__(params)
+        config = params.reachability_config
+        self._validator_fn = make_ik_reachability_validator(
+            config.embodiment,
+            grasp_z_offset=config.grasp_z_offset_m,
+            ik_pos_threshold=config.ik_position_threshold_m,
+            ik_rot_threshold=config.ik_rotation_threshold_rad,
+            stamp_results=False,
+        )
```
alexmillane: I'm wondering why we have this odd construction where the validator class splits out the validation to a function, created by `make_ik_reachability_validator`. It would seem more natural to me to just put that stuff in this class and skip the complexity of creating a function object in another function and returning it.

Perhaps there's a good reason for this though? Like we need to access the function above in a context where we can't use this object?

alexmillane (approved):
Thanks for cleaning this all up. it's looking much better. clean!

I had a couple of questions around the structure of the validator, but you're clear to merge.

qianl-nv (approved):
Thanks for multi rounds of updates. LGTM!

isaaclab_arena_curobo/ik_reachability_validator.py
xyao-nv: It's because the ik solver build needs to be done once, and the actual func to run IK is called per candidate layout.

xyao-nv: I could fold the solver build into init method and drop the factory.

xyao-nv: Updated to a simpler caller.

isaaclab_arena_curobo/curobo_frame_utils.py
```diff
@@ -0,0 +1,189 @@
...
+    num_poses = positions.shape[0]
+    pos_err = ik_result.position_error.view(num_poses, -1)
+    rot_err = ik_result.rotation_error.view(num_poses, -1)
+
+    ok = (pos_err < position_threshold) & (rot_err < rotation_threshold)
+    # TODO(xinjieyao, 2026-07-15): Support collision-free IK by disabling the hand-link spheres during collision checking
+    assert require_collision_free is False, "Collision-free IK is not supported yet. Needs extra machinery."
+    if require_collision_free:
+        # cuRobo folds collision-free-ness into ``success`` (success = converged AND feasible).
+        ok = ok & ik_result.success.view(num_poses, -1).bool()
```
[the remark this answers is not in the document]
xyao-nv: it's will be relaxed soon

PR #915, opened 2026-07-15
Introduce PlaceableAsset

zhx06:
## Summary
Unify object and robot relation placement

## Detailed description
- Enables robot embodiments to use the existing relation solver.
- Adds `PlaceableAsset` and generalizes the placement pipeline from objects to placement assets.
- Preserves existing object placement while adding single-root, bounding-box-based robot placement.


isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -37,23 +41,25 @@ def _get_passive_collision_objects(
...
     placer_params: ObjectPlacerParams | None = None,
     collision_objects: list[CollisionObject] | None = None,
     scene_assets: Iterable[Asset | RigidObjectSet] | None = None,
+    scene_entity_names: Mapping[str, str] | None = None,
 ) -> EventTermCfg | None:
     """Solve relation placement and apply the result to object reset/static state.
 
     Args:
-        objects: Objects with spatial predicates that should be relation-solved.
+        objects: Entities with spatial predicates that should be relation-solved.
```
xyao-nv: nit, rename objects to entities?

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -50,14 +56,43 @@ def __init__(
...
+    def get_world_bounding_box(self) -> AxisAlignedBoundingBox:
+        """Return bounds transformed by the configured root pose."""
+        bounding_box = self.get_bounding_box()
+        if self.initial_pose is None:
+            return bounding_box
+        quarters = quaternion_to_90_deg_z_quarters(self.initial_pose.rotation_xyzw)
+        return bounding_box.rotated_90_around_z(quarters).translated(self.initial_pose.position_xyz)
+
+    def get_collision_mesh(self) -> trimesh.Trimesh | None:
+        """Return no mesh because embodiment placement uses bounds."""
```
xyao-nv: add a TODO as mesh shall also depend on initial joint pos

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -85,7 +84,12 @@ def _solve_relations(self) -> None:
         * **False** — applies one layout per environment so per-object reset
           events restore the same layout every time.
         """
-        objects_with_relations = self.arena_env.scene.get_objects_with_relations()
+        placement_entities = list(self.arena_env.scene.get_objects_with_relations())
+        scene_entity_names = {entity.name: entity.name for entity in placement_entities}
+        embodiment = self.arena_env.embodiment
+        if embodiment is not None and embodiment.get_relations():
```
xyao-nv, line 91: If we are modeling single-arm manipulator on a stand as a 3D bbox, does it mean it will put the robot far away from the table that may be even hard to reach? E.g. the stand footprint of droid is pretty large.
How's nextTo(Robot) being defined here? Using a boxy AABB edge?
On(Robot)? a footprint on the top face of the robot?

Shall we have some restrictions/definitions of embodiment-related relations?

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -50,14 +56,43 @@ def __init__(
...
-    def set_initial_pose(self, pose: Pose) -> None:
+    def get_bounding_box(self) -> AxisAlignedBoundingBox:
+        """Return root-relative bounds for the USD-authored articulation pose."""
+        from isaaclab_arena.utils.usd_helpers import compute_local_bounding_box_from_usd
+
+        assert self.scene_config is not None, "scene_config must be populated before placement"
+        spawn = self.scene_config.robot.spawn
+        assert spawn.usd_path is not None, "scene_config.robot must use a USD spawn for placement"
+        scale = tuple(spawn.scale or (1.0, 1.0, 1.0))
+        # TODO: Compute bounds at configured initial joint positions when joint_pos is not None.
```
xyao-nv: please include your name in the TODO

isaaclab_arena/tests/dummy_embodiment.py
cvolkcvolk, line 1: This class seems to be used only by tests. Suggestion to move into the test directory

isaaclab_arena/tests/dummy_object.py
cvolkcvolk, line 1: Suggestion to move to tests/

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -50,14 +56,43 @@ def __init__(
...
+            return bounding_box
+        quarters = quaternion_to_90_deg_z_quarters(self.initial_pose.rotation_xyzw)
+        return bounding_box.rotated_90_around_z(quarters).translated(self.initial_pose.position_xyz)
+
+    def get_collision_mesh(self) -> trimesh.Trimesh | None:
+        """Return no mesh because embodiment placement uses bounds."""
+
+    def set_initial_pose(self, pose: Pose | PoseRange | PosePerEnv) -> None:
+        """Set the embodiment root pose."""
+        assert isinstance(pose, Pose), "Embodiments require one root Pose"
```
cvolkcvolk: `pose: Pose | PoseRange | PosePerEnv` seems to contradict "assert isinstance(pose, Pose)"?

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -50,14 +56,43 @@ def __init__(
...
+
+    def get_world_bounding_box(self) -> AxisAlignedBoundingBox:
+        """Return bounds transformed by the configured root pose."""
+        bounding_box = self.get_bounding_box()
+        if self.initial_pose is None:
+            return bounding_box
+        quarters = quaternion_to_90_deg_z_quarters(self.initial_pose.rotation_xyzw)
+        return bounding_box.rotated_90_around_z(quarters).translated(self.initial_pose.position_xyz)
+
+    def get_collision_mesh(self) -> trimesh.Trimesh | None:
```
cvolkcvolk, line 76: Could we make this a default implementation on `PlacementEntity`, then remove the redundant no-op implementations from `EmbodimentBase` and `ObjectBase`?

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -116,10 +151,11 @@ def get_camera_cfg(self) -> Any:
         return self.camera_config.get_cfg()
 
     def _update_scene_cfg_with_robot_initial_pose(self, scene_config: Any, pose: Pose) -> Any:
-        if scene_config is None or not hasattr(scene_config, "robot"):
-            raise RuntimeError("scene_config must be populated with a `robot` before calling `set_robot_initial_pose`.")
-        scene_config.robot.init_state.pos = pose.position_xyz
-        scene_config.robot.init_state.rot = pose.rotation_xyzw
+        assert scene_config is not None, "scene_config.robot must be populated before setting the root pose"
```
cvolkcvolk: The assert message does not match the actual assert

cvolkcvolk:
```suggestion
        assert scene_config is not None, "scene_config must be populated before setting the root pose"
```

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -116,10 +151,11 @@ def get_camera_cfg(self) -> Any:
...
+        robot = scene_config.robot
+        assert robot is not None, "scene_config.robot must be populated before setting the root pose"
```
cvolkcvolk, line 223: Nit, can directly do  `assert scene_config.robot is not None, "scene_config.robot must be populated before setting the root pose"`,as the declared variable is not directly used

isaaclab_arena/relations/placement_entity.py
```diff
@@ -0,0 +1,83 @@
...
+from isaaclab_arena.relations.relations import IsAnchor, Relation, RelationBase, UnaryRelation
+from isaaclab_arena.utils.pose import Pose, PosePerEnv, PoseRange
+
+if TYPE_CHECKING:
+    import trimesh
+
+    from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox
+
+
+class PlacementEntity(Asset, ABC):
```
cvolkcvolk: Cool!

isaaclab_arena/assets/object_base.py
```diff
@@ -115,6 +87,17 @@ def set_initial_pose(self, pose: Pose | PoseRange | PosePerEnv) -> None:
             self.object_cfg.init_state.rot = initial_pose.rotation_xyzw
         self.event_cfg = self._init_event_cfg()
 
+    def set_placement_initial_pose(self, pose: Pose) -> None:
+        """Set a solved root pose without configuring an independent reset."""
+        self.initial_pose = pose
+        if self.object_cfg is not None:
+            self.object_cfg.init_state.pos = pose.position_xyz
+            self.object_cfg.init_state.rot = pose.rotation_xyzw
```
alexmillane: Why's this required?

Are there places where we want to set a pose that doesn't reset?

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -50,14 +56,43 @@ def __init__(
         self.xr: Any | None = None
         self.termination_cfg: Any | None = None
 
-    def set_initial_pose(self, pose: Pose) -> None:
+    def get_bounding_box(self) -> AxisAlignedBoundingBox:
+        """Return root-relative bounds for the USD-authored articulation pose."""
+        from isaaclab_arena.utils.usd_helpers import compute_local_bounding_box_from_usd
```
alexmillane, line 65: Is this import required to be local?

isaaclab_arena/embodiments/embodiment_base.py
alexmillane: Should this print a warning or something? Is this triggered when a user requests that the embodiment is placed via mesh? If this is called on user request, I feel that we should error out. If returning `None` is just our way of indicating that a bounding box should be used, then it's fine as is.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -85,7 +84,12 @@ def _solve_relations(self) -> None:
         * **False** — applies one layout per environment so per-object reset
           events restore the same layout every time.
         """
-        objects_with_relations = self.arena_env.scene.get_objects_with_relations()
+        placement_entities = list(self.arena_env.scene.get_objects_with_relations())
```
alexmillane: nit: out of interest why is the proceeding `list` required here. The function returns a `list`. I thought maybe the point of the `list` is to do a copy, but the function actually builds the returned list from scratch. I.e. you don't get a reference to some internal list you don't wanna modify.

alexmillane: One issue with this MR is that we've introduced another term `entities`. We now have `assets`, `objects` and `entities`.

I feel that we should maybe only have two. In this case both objects and embodiments inherit from the `Asset` class. So perhaps we should call them assets?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -96,10 +100,11 @@ def _solve_relations(self) -> None:
             if self.cfg.resolve_on_reset is not None:
                 placer_params.resolve_on_reset = self.cfg.resolve_on_reset
         self._placement_event_cfg = solve_and_apply_relation_placement(
-            objects_with_relations,
+            placement_entities,
             num_envs=self.cfg.num_envs,
             placer_params=placer_params,
             scene_assets=self.arena_env.scene.assets.values(),
+            scene_entity_names=scene_entity_names,
```
alexmillane: Here we expand the number of arguments passed to this function in order to pass the names separately. However the names are already contained within `placement_entities`. In fact we extract them from that list right here.

Is there a reason to pass them separately?

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -63,6 +69,11 @@ def solve_and_apply_relation_placement(
     if not objects:
         print("No objects with relations found in scene. Skipping relation solving.")
         return None
+    entity_names = {obj.name for obj in objects}
+    assert len(entity_names) == len(objects), "Placement entity names must be unique"
+    if scene_entity_names is None:
+        scene_entity_names = {obj.name: obj.name for obj in objects}
+    assert set(scene_entity_names) == entity_names, "scene_entity_names must contain every placement entity name"
```
alexmillane: What is going on here?

We construct our own list of names, then check it against the (optionally) passed list of names and check that they're equal? Why bother passing the list of names at all if we extract it inside and expect to get exactly the same thing passed in?

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -139,91 +150,108 @@ def _apply_relation_placement_result(
...
+            num_envs=num_envs,
+        )
+        return None
+    return _apply_static_spawn_pose(
         objects=objects,
         placement_pool=placement_pool,
         anchor_objects_set=anchor_objects_set,
         num_envs=num_envs,
+        scene_entity_names=scene_entity_names,
     )
```
alexmillane: What's the difference in behaviour between these two branches?

Like, how does the generated scene differ? In the second case, do we, for example, lose different poses between the environments? That would be a big change in behaviour... Probably something that we don't want to do.

Could you explain what's going on here. I would add such an explanation as a (lengthy) comment. However, let's agree here on what's going on and what should happen first.

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -139,91 +150,108 @@ def _apply_relation_placement_result(
...
-        object_cfg.init_state.rot = rot
+    _set_placement_initial_poses(objects, anchor_objects_set, layout)
 
     return EventTermCfg(
         func=solve_and_place_objects,
         mode="reset",
         params={
             "objects": objects,
             "placement_pool": placement_pool,
+            "scene_entity_names": scene_entity_names,
```
alexmillane: why is this now required?

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -139,91 +150,108 @@ def _apply_relation_placement_result(
...
+    _set_placement_initial_poses(objects, anchor_objects_set, layouts[0])
+    return EventTermCfg(
+        func=place_entities_from_layouts,
+        mode="reset",
+        params={
+            "objects": objects,
+            "layouts": layouts,
+            "scene_entity_names": scene_entity_names,
+        },
+    )
```
alexmillane: Ok this is related to my comment above. Seems like we're introducing new behaviour that is degraded from what was there in some circumstances.

It seems like now, we may end up with the same poses per environment, rather than the varied poses we had before. That is a big change in behaviour no?

Is that coming because we can't set the robot pose per-env?

isaaclab_arena/relations/placement_entity.py
alexmillane: As I commented above, I'm not sure we should introduce the term `Entity`, because then we'll have three overlapping things `Assets` `Objects` and `Entities`.

This class is always intended to be combined with something deriving from `Asset` no? In that case this class is a "mixin" to add functionality for placement no? What do you think about calling this class `Placeable`. That indicates that this class adds place-ability to another object?

alexmillane (commented):
High level review #1

isaaclab_arena/relations/placement_entity.py
cvolkcvolk: Agree with the naming. `Entitiy` is quite vague as well. Placeable would be a good fit, however already taken by one of our affordances `IsaacLab-Arena/isaaclab_arena/affordances/placeable.py`
Arguable `Placeable` would be a better fit for this use case here as the current `Placeable` affordance is just checking for orientation via `place_upright()`

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Yes. This is about USD and this is to avoid using simulation

isaaclab_arena/environments/arena_env_builder.py
zhx06: Addressed.

isaaclab_arena/relations/placement_entity.py
zhx06: I have renamed the class to be  `PlacementAsset`.

isaaclab_arena/environments/relation_solver_interface.py
zhx06: The first layout is used only to initialize the single-pose scene configuration during construction. The poses will be applied separately for each environment.

isaaclab_arena/environments/relation_solver_interface.py
zhx06: This is to deal with an edge case. It is now removed.

isaaclab_arena/environments/arena_env_builder.py
zhx06: We call it `PlacementAsset`.

isaaclab_arena/environments/arena_env_builder.py
zhx06: `scene_entity_names` has been removed entirely. It was designed for an edge case, which is not necessary.

isaaclab_arena/environments/relation_solver_interface.py
zhx06: Addressed. It is removed.

isaaclab_arena/embodiments/embodiment_base.py
zhx06: `None` means the asset has no collision mesh, so placement falls back to bounding-box method.

isaaclab_arena/assets/object_base.py
zhx06: It is designed for robots that does not have  `object_cfg`

isaaclab_arena/environments/relation_solver_interface.py
zhx06: Both branches preserve per-environment variation. The reason is that objects and embodiments currently use different reset mechanisms.

The first branch handles object-only scenes, where each object’s reset event stores its per-environment pose.

The second handles scenes containing an embodiment. Robot configs store only one construction pose, so a reset event writes each environment’s saved layout at reset.

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Addressed.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -140,6 +165,10 @@ def get_termination_cfg(self) -> Any:
     def get_embodiment_name_in_scene(self) -> str:
         return "robot"
 
+    def get_scene_name(self) -> str:
+        """Return the embodiment's Isaac Lab scene key."""
+        return self.get_embodiment_name_in_scene()
```
qianl-nv: remove get_embodiment_name_in_scene and just use get_scene_name everywhere

isaaclab_arena/tests/dummy_object.py
qianl-nv: +1 on this suggestion

isaaclab_arena/tests/test_relation_solver_embodiment.py
```diff
@@ -0,0 +1,72 @@
...
+        initial_pose=Pose(
+            position_xyz=(3.0, 4.0, 0.0),
+            rotation_xyzw=(0.0, 0.0, 2**-0.5, 2**-0.5),
+        ),
+    )
+
+    world_bbox = asset.get_world_bounding_box()
+
+    assert torch.allclose(world_bbox.min_point, torch.tensor([[2.0, 4.0, 0.0]]))
+    assert torch.allclose(world_bbox.max_point, torch.tensor([[3.0, 6.0, 1.0]]))
```
qianl-nv, line 74: can we add an integration test using a test yaml data? For examples this one https://github.com/isaac-sim/IsaacLab-Arena/pull/937/changes#diff-71de32477f1759846c5ba9d4ac203e1bbcbb4461b4e22496099fec57dfeb5604

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Addressed.

isaaclab_arena/tests/dummy_object.py
zhx06: Addressed.

isaaclab_arena/tests/test_relation_solver_embodiment.py
zhx06: Test added using `isaaclab_arena/tests/test_data/pick_and_place_maple_table_env_graph.yaml `

isaaclab_arena/environments/relation_solver_interface.py
qianl-nv: if we need to handle object vs embodiment differently at relation_solver_interface.py level, it feels like we are missing some abstraction layers at PlacementAsset/EmbodimentBase class so we can unify the interface with ObjectBase. The root cause appears to be embodiment's rejecting PosePerEnv. Is this something we can extend the EmbodimentBase to support, so we provide one unified the per-env & reset pose behavior for object and embodiment?

zhx06: Addressed. The new commit unify these 2 classes. One remaining thing is how to deal with the stand of the driod.

isaaclab_arena/assets/asset.py
```diff
@@ -47,3 +47,7 @@ def get_variation(self, name: str) -> VariationBase:
     def get_variations(self) -> list[VariationBase]:
         """Return every variation attached to this asset, enabled or not."""
         return list(self.variations.values())
+
+    def get_scene_name(self) -> str:
```
alexmillane: `get_scene_key`

isaaclab_arena/assets/object_base.py
```diff
@@ -40,7 +32,7 @@
 ]
 
 
-class ObjectBase(Asset, ABC):
+class ObjectBase(PlacementAsset, ABC):
```
alexmillane: `PlaceableAsset`

isaaclab_arena/assets/object_base.py
```diff
@@ -111,12 +76,25 @@ def set_initial_pose(self, pose: Pose | PoseRange | PosePerEnv) -> None:
...
         initial_pose = self._get_initial_pose_as_pose()
         if initial_pose is not None and self.object_cfg is not None:
             self.object_cfg.init_state.pos = initial_pose.position_xyz
             self.object_cfg.init_state.rot = initial_pose.rotation_xyzw
-        self.event_cfg = self._init_event_cfg()
+
+    def set_spawn_pose(self, pose: Pose) -> None:
+        """Set the scene-construction pose without rebuilding the reset event."""
+        assert self.object_cfg is not None, "object_cfg must be initialized before setting the spawn pose"
+        self._set_pose_state(pose)
```
alexmillane: Can we return this to a single function with a flag for `create_reset_event=True`

Right now, these three names: `set_initial_pose`, `_set_pose_state`, and `set_spawn_pose` are all basically synonyms. One cannot read the names and know what the difference is.

If we have a single function called `set_initial_pose` with  `create_reset_event=True`, it's super clear to the reader what is going on.

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -98,9 +98,16 @@ def __init__(
...
+    def set_initial_pose(self, pose: Pose | PosePerEnv) -> None:
+        """Store the requested base pose(s), lifted by the stand-height offset to match the spawned base."""
+        if isinstance(pose, PosePerEnv):
+            super().set_initial_pose(PosePerEnv(poses=[p.translate(self._robot_base_offset) for p in pose.poses]))
+        else:
+            super().set_initial_pose(pose.translate(self._robot_base_offset))
+
+    def set_spawn_pose(self, pose: Pose) -> None:
+        """Set the scene-construction base pose, lifted by the stand-height offset like ``set_initial_pose``."""
+        super().set_spawn_pose(pose.translate(self._robot_base_offset))
```
alexmillane: Am I correct that this function just calls the underlying function, except that it first translates the pose by the `self._robot_base_offset`?

I think it probably makes sense we eventually put the stand into the USD with the robot to skip this backflipping here.

For the time being, let's go with this but:

```
def _translate_pose(pose: Pose | PosePerEnv):
        if isinstance(pose, PosePerEnv):
            translated_pose = PosePerEnv(poses=[p.translate(self._robot_base_offset) for p in pose.poses])
        else:
            translated_pose = pose.translate(self._robot_base_offset)
        return translated_pose

def set_initial_pose(self, pose: Pose | PosePerEnv)
    pose = _translate_pose(pose)
    super().

```

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -49,17 +56,103 @@ def __init__(
...
+            assert robot is not None, "scene_config.robot must be populated before placement"
+            spawn = robot.spawn
+            assert spawn.usd_path is not None, "scene_config.robot must use a USD spawn for placement"
+            scale = tuple(spawn.scale or (1.0, 1.0, 1.0))
+            stage = Usd.Stage.Open(spawn.usd_path)
+            assert stage is not None, f"could not open robot USD: {spawn.usd_path}"
+            default_prim = stage.GetDefaultPrim() or stage.GetPseudoRoot()
+            # The default prim scopes extraction to the robot, excluding sibling scene props
+            # (ground planes, stray objects) baked into some flattened articulation USDs.
+            self._collision_mesh = extract_trimesh_from_prim(stage, default_prim.GetPath().pathString, scale)
```
alexmillane: What do you think about folding these into a (small) function next to `extract_trimesh_from_prim`. Something like `extract_trimesh_from_usd_path`

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -49,17 +56,103 @@ def __init__(
...
+        assert isinstance(initial_pose, Pose), "Embodiments support a fixed Pose or PosePerEnv only"
+        return EventTermCfg(
+            func=reset_placement_asset_pose,
+            mode="reset",
+            params={"scene_writes": self.layout_pose_to_scene_writes(initial_pose)},
+        )
+
+    def has_pose_reset_event(self) -> bool:
+        """Return whether the embodiment owns a root-pose reset event."""
+        return self.pose_event_cfg is not None
```
alexmillane: Is all of this an exact duplicate of the functions in `ObjectBase`?

Could we fold this into another mixin class? That attaches setting pose behaviour to objects?

Can we factor out this common behaviour somehow?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -83,10 +82,14 @@ def _solve_relations(self) -> None:
 
         * **True** (default) — registers a reset event that draws a fresh layout
           from the pool for each resetting environment.
-        * **False** — applies one layout per environment so per-object reset
-          events restore the same layout every time.
+        * **False** — assigns one fixed layout per environment. Object-only
+          scenes use per-object reset events; scenes with an embodiment use one
+          coordinated reset event.
```
alexmillane: "Object-only scenes use per-object reset events; scenes with an embodiment use one coordinated reset event."

Is there any way we can unify this behaviour?

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -59,10 +59,14 @@ def solve_and_apply_relation_placement(
         Reset event config to attach to the environment when placement should be
         resolved on reset. Returns ``None`` when no reset event is needed.
     """
-    objects = list(objects)
-    if not objects:
-        print("No objects with relations found in scene. Skipping relation solving.")
+    assets = list(assets)
```
alexmillane: why do we need this? The according to the type annotation we already pass in a `list`?

Is this to get a copy?

isaaclab_arena/relations/placement_asset.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        bounding_box = self.get_bounding_box()
+        initial_pose = self.get_initial_pose()
+        if not isinstance(initial_pose, Pose):
+            return bounding_box
+        quarters = quaternion_to_90_deg_z_quarters(initial_pose.rotation_xyzw)
+        return bounding_box.rotated_90_around_z(quarters).translated(initial_pose.position_xyz)
+
+    def get_collision_mesh(self) -> trimesh.Trimesh | None:
+        """Return this asset's collision mesh, or ``None`` if it has none."""
```
alexmillane: Is this an @abstractmethod?

isaaclab_arena/relations/placement_asset.py
```diff
@@ -0,0 +1,99 @@
...
+        """Set the configured root pose.
+
+        Accepts a fixed ``Pose``, a ``PoseRange`` (randomized on reset), or a ``PosePerEnv``
+        (a distinct pose per environment); the interpretation is subclass-specific.
+        """
+        self.initial_pose = pose
+
+    def set_spawn_pose(self, pose: Pose) -> None:
+        """Set the root pose used when constructing the scene."""
+        self.set_initial_pose(pose)
```
alexmillane: These functions appear to me to be duplicated (overridden) in the classes that this class is mixed into.

Is there a reason for that? It would be amazing if we could centralize the treatment of poses here.

Also there appears to me to be a bit of duplication between `ObjectBase` and `EmbodimentBase`. Another reason to centralize the implementation here.

Is there anything preventing us from doing that?

alexmillane (commented):
Looks pretty much ready to go. A few small comments and questions about pose-setting.

isaaclab_arena/assets/asset.py
zhx06: Addressed.

isaaclab_arena/assets/object_base.py
zhx06: Addressed.

isaaclab_arena/relations/placement_asset.py
zhx06: Yeah, there was some duplicated code here. In the newest commit, all pose-related functions share the same interface in `PlaceableAsset`. `ObjectBase`/`EmbodimentBase` only override the two different methods according to the actual needs.

isaaclab_arena/relations/placement_asset.py
zhx06: No, it is not. `None` here means "No collision mesh,  fall back to the AABB." It is supposed to stay concrete.

isaaclab_arena/environments/relation_solver_interface.py
zhx06: Addressed. This is removed.

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Addressed. It is all in `PlaceableAsset `.

isaaclab_arena/environments/arena_env_builder.py
zhx06: Yes, their behaviors are unified now. Each non-anchor asset stores its per-env pose and owns its own reset event.

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Addressed. We now have `extract_trimesh_from_usd_path` in `usd_helpers.py`

isaaclab_arena/assets/object_base.py
zhx06: Addressed. We now have `set_initial_pose(pose, create_reset_event=True)`. `_set_pose_state`, and `set_spawn_pose` are gone and `set_initial_pose` is the public interface.

isaaclab_arena/embodiments/droid/droid.py
zhx06: Yes, that's correct. We now have `_translate_pose` and `set_initial_pose` for droid in `droid.py`, and add a  TODO for adding stand into the USD.

alexmillane (approved):

isaaclab_arena/environments/relation_solver_interface.py
zhx06: Addressed.

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Addressed.

zhx06: Addressed.

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Addressed

isaaclab_arena/assets/object_base.py
```diff
@@ -57,66 +49,16 @@ def __init__(
...
-
-    def set_initial_pose(self, pose: Pose | PoseRange | PosePerEnv) -> None:
-        """Set / override the initial pose and rebuild derived configs.
-
-        Args:
-            pose: A fixed ``Pose``, a ``PoseRange`` (randomised on reset),
-                or a ``PosePerEnv`` (distinct pose per environment).
-        """
+    def _materialize_pose_state(self, pose: Pose | PoseRange | PosePerEnv) -> None:
+        """Store the pose and write its construction values into the materialized object config."""
```
alexmillane: Suggestion to call this `_set_initial_pose`. I guess we've introduced this because it is called in `PlaceableAsset` and we need diferent implementations per sub-class?

We don't really use the word "materialize" elsewhere, and we just call "pose-state" pose elsewhere. So suggestion to just make it `_set_initial_pose`

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -98,9 +98,27 @@ def __init__(
...
+        """Lift a base pose (or per-env poses) by the stand-height offset to match the spawned base."""
+        if isinstance(pose, PosePerEnv):
+            return PosePerEnv(poses=[p.translate(self._robot_base_offset) for p in pose.poses])
+        return pose.translate(self._robot_base_offset)
+
+    def set_initial_pose(self, pose: Pose | PosePerEnv, create_reset_event: bool = True) -> None:
+        """Store the requested base pose(s), lifted by the stand-height offset to match the spawned base.
+
+        Both the reset path and the construction path (``create_reset_event=False``) route through here,
+        so the stand lift is applied consistently in either case.
```
alexmillane: remove long comment.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -49,17 +56,76 @@ def __init__(
...
+            spawn = robot.spawn
+            assert spawn.usd_path is not None, "scene_config.robot must use a USD spawn for placement"
+            scale = tuple(spawn.scale or (1.0, 1.0, 1.0))
+            self._collision_mesh = extract_trimesh_from_usd_path(spawn.usd_path, scale)
+        return self._collision_mesh
+
+    def _materialize_pose_state(self, pose: Pose | PoseRange | PosePerEnv) -> None:
+        """Store the configured pose; the construction pose is materialized in ``get_scene_cfg``."""
+        assert isinstance(pose, (Pose, PosePerEnv)), "Embodiments support a fixed Pose or PosePerEnv only"
         self.initial_pose = pose
```
alexmillane, line 94: See comment above. Suggestion to rename -> `_set_initial_pose`

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -49,17 +56,76 @@ def __init__(
...
+                func=reset_placement_asset_pose_per_env,
+                mode="reset",
+                params={"write_pose_list": [self.layout_pose_to_scene_writes(pose) for pose in initial_pose.poses]},
+            )
+        assert isinstance(initial_pose, Pose), "Embodiments support a fixed Pose or PosePerEnv only"
+        return EventTermCfg(
+            func=reset_placement_asset_pose,
+            mode="reset",
+            params={"scene_writes": self.layout_pose_to_scene_writes(initial_pose)},
+        )
```
alexmillane, line 123: Is this shared with `ObjectBase`? Can we move it somewhere common? `PlaceableAsset`?

If it's not common I assume that it's pretty close to a duplicate, so could be worth adding a note.

_looking above I think that the difference may be the auxillary prims. In that case you definately want to leave a NOTE that we should clean this up and unify with ObjectBase, once we come up with a bit solution for the auxillary prims._

isaaclab_arena/environments/relation_solver_interface.py
```diff
@@ -127,111 +128,126 @@ def _should_include_background_mesh(
...
+    ``layout_pose_to_scene_writes``; any auxiliary prim it omits (e.g. Droid's static stand) would
+    stay at its env-0 spot. Fail loudly here rather than silently orphaning those prims.
+    """
+    for asset in assets:
+        if asset in anchor_assets:
+            continue
+        assert not asset.has_unplaced_auxiliary_prims(), (
+            f"Non-anchor asset '{asset.name}' owns auxiliary scene prims that a per-environment reset "
+            "does not reposition, so relation placement would orphan them (e.g. Droid's static stand). "
+            "Make it an anchor, or implement layout_pose_to_scene_writes for its auxiliary prims."
```
alexmillane, line 256: I'm thinking that @qianl-nv is introducing something to solve this limitation?

Should we add a NOTE here to delete this code once the auxillary prims correctly travel with their parents?

isaaclab_arena/relations/placement_asset.py
```diff
@@ -0,0 +1,148 @@
...
+                pose on every environment reset. Pass False to set only the scene-construction
+                pose (e.g. when relation solving seeds a fixed per-environment layout).
+        """
+        self._materialize_pose_state(pose)
+        if create_reset_event:
+            self._pose_event_cfg = self._build_reset_event()
+
+    def _materialize_pose_state(self, pose: Pose | PoseRange | PosePerEnv) -> None:
+        """Store the configured pose; subclasses also materialize any derived construction config."""
+        self.initial_pose = pose
```
alexmillane, line 80: See comments above.

isaaclab_arena/relations/placement_asset.py
```diff
@@ -0,0 +1,148 @@
...
+            return None
+        if isinstance(pose, PosePerEnv):
+            return pose.poses[0]
+        if isinstance(pose, PoseRange):
+            return pose.get_midpoint()
+        return pose
+
+    def _get_initial_pose_as_pose(self) -> Pose | None:
+        """Return the resolved root pose (relation- or reference-derived) as a single ``Pose``."""
+        return self._collapse_pose_to_single(self.get_initial_pose())
```
alexmillane, line 105: Is there some reason to have two functions here? (maybe there is)?

alexmillane (approved):
A few remaining items. After fixing feel free to merge.

isaaclab_arena/assets/object_base.py
zhx06: Addressed. It is now called `_set_initial_pose`

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Addressed.

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Note has been added.

isaaclab_arena/environments/relation_solver_interface.py
zhx06: Note added.

isaaclab_arena/relations/placement_asset.py
zhx06: Addressed.

isaaclab_arena/relations/placement_asset.py
zhx06: The collapse logic all lives in `_collapse_pose_to_single`. `_get_initial_pose_as_pose` is a thin wrapper so `EmbodimentBase` can override the input pose (self.initial_pose instead of get_initial_pose()) without duplicating the collapse.

isaaclab_arena/environments/arena_env_builder.py
zhx06: Yeah, this is the limitation of the current method so we observe a huge gap between droid and the kitchen top in the example. We will have follow-up MRs for addressing this.

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Addressed. The assert now allows `Pose` and `PosePerEnv`

isaaclab_arena/embodiments/embodiment_base.py
zhx06: There is a refactor so it is now used.

PR #916, opened 2026-07-15
Add object mass variation

TuurStuyck:
## Summary
Add runtime object mass variation

## Detailed description
- Adds a rigid-object runtime variation for per-env sampled mass.
- Registers mass on rigid ObjectBase assets for Hydra configuration and recording.
- Covers event composition, Hydra overrides, recorder output, partial reset, and invalid mass handling.

mass_sensitivity_combined

### DIFF
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -54,6 +54,10 @@ class ObjectBase(Asset, ABC):
             prim_path = "{ENV_REGEX_NS}/" + self.name
         self.prim_path = prim_path
         self.object_type = object_type
+        if self.object_type == ObjectType.RIGID:
+            from isaaclab_arena.variations.object_mass_variation import ObjectMassVariation  # noqa: PLC0415
+
+            self.add_variation(ObjectMassVariation(self.name))
         self.initial_pose: Pose | PoseRange | PosePerEnv | None = None
         self.initial_velocity: Velocity | None = None
         self.object_cfg = None
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -32,6 +32,7 @@ from isaaclab_arena.terms.events import set_object_pose, set_object_pose_per_env
 from isaaclab_arena.utils.bounding_box import AxisAlignedBoundingBox
 from isaaclab_arena.utils.pose import Pose, PosePerEnv, PoseRange
 from isaaclab_arena.utils.velocity import Velocity
+from isaaclab_arena.variations.object_mass_variation import ObjectMassVariation
 
 __all__ = [
     "ObjectBase",
@@ -55,8 +56,6 @@ class ObjectBase(Asset, ABC):
         self.prim_path = prim_path
         self.object_type = object_type
         if self.object_type == ObjectType.RIGID:
-            from isaaclab_arena.variations.object_mass_variation import ObjectMassVariation  # noqa: PLC0415
-
             self.add_variation(ObjectMassVariation(self.name))
         self.initial_pose: Pose | PoseRange | PosePerEnv | None = None
         self.initial_velocity: Velocity | None = None
--- a/isaaclab_arena/variations/object_mass_variation.py
+++ b/isaaclab_arena/variations/object_mass_variation.py
@@ -26,6 +26,10 @@ from isaaclab_arena.variations.variation_base import RunTimeVariationBase, Varia
 if TYPE_CHECKING:
     from isaaclab.envs import ManagerBasedEnv
 
+# Smallest physically meaningful mass [kg]. Sampled masses below this would inject invalid
+# (near-zero / NaN-prone) values into the physics engine, so they are rejected rather than clamped.
+_MIN_PHYSICAL_MASS_KG = 1e-6
+
 
 @configclass
 class ObjectMassVariationCfg(VariationBaseCfg):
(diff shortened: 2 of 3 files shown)


isaaclab_arena/assets/object_base.py
```diff
@@ -54,6 +54,10 @@ def __init__(
             prim_path = "{ENV_REGEX_NS}/" + self.name
         self.prim_path = prim_path
         self.object_type = object_type
+        if self.object_type == ObjectType.RIGID:
+            from isaaclab_arena.variations.object_mass_variation import ObjectMassVariation  # noqa: PLC0415
```
alexmillane: Am I correct that we need a local import here due to this file being used in contexts where sim isn't started?

isaaclab_arena/variations/object_mass_variation.py
```diff
@@ -0,0 +1,197 @@
...
+    """Configuration for ObjectMassVariation."""
+
+    sampler_cfg: UniformSamplerCfg = field(default_factory=lambda: UniformSamplerCfg(low=[0.05], high=[2.0]))
+    """Uniform distribution over absolute object mass [kg]."""
+
+    recompute_inertia: bool = True
+    """Whether to scale inertia tensors by the sampled-mass/default-mass ratio."""
+
+    min_mass: float = 1e-6
+    """Minimum allowed sampled mass [kg]. Values below this fail instead of being clamped."""
```
alexmillane: Not sure I understand the purpose of this guy. Do we need this as a configurable parameter? Seems like if the user is requesting a minimum mass explicityly via the sampler, we should let them have it unless it's less than some very small value?

isaaclab_arena/variations/object_mass_variation.py
```diff
@@ -0,0 +1,197 @@
...
+            f"got {tuple(sampler.shape_per_sample)}."
+        )
+        assert (
+            min_mass >= 1e-6
+        ), f"apply_object_mass_from_sampler requires min_mass >= 1e-6 to avoid invalid physics masses; got {min_mass}."
+        selected_bodies = self._selected_body_count()
+        assert selected_bodies == 1, (
+            "apply_object_mass_from_sampler currently supports exactly one selected body; "
+            f"'{self.asset_cfg.name}' selected {selected_bodies} bodies."
+        )
```
alexmillane: nit: `selected_bodies` -> `selected_bodies_count`

isaaclab_arena/variations/object_mass_variation.py
```diff
@@ -0,0 +1,197 @@
...
+        assert self._default_inertia is not None
+
+        if env_ids is None:
+            env_ids = torch.arange(env.scene.num_envs, device=self.asset.device, dtype=torch.int32)
+        else:
+            env_ids = torch.as_tensor(env_ids, device=self.asset.device, dtype=torch.int32).reshape(-1)
+        body_ids = self._body_ids_tensor()
+
+        sample = sampler.sample(num_samples=len(env_ids), env_ids=env_ids)
+        masses_to_apply = sample.to(device=self._default_mass.device, dtype=self._default_mass.dtype)
```
alexmillane, line 164: Hmmm. This is my bad, but ideally we'd pass the device to the sampler.

Sampling on the CPU and sending it to the GPU (and probably synchronizing) looks a bit weird to me.

Probably doesn't make too much difference but doing a CPU->GPU transfer unnecessarily looks a bit weird haha.

isaaclab_arena/variations/object_mass_variation.py
```diff
@@ -0,0 +1,197 @@
...
+                f"default masses={default_masses}."
+            )
+            ratios = masses_to_apply / default_masses
+            inertias = wp.to_torch(self.asset.data.body_inertia).clone()
+            inertias[env_ids[:, None], body_ids] = self._default_inertia[env_ids[:, None], body_ids] * ratios[..., None]
+            self.asset.set_inertias_index(
+                inertias=inertias[env_ids[:, None], body_ids],
+                body_ids=body_ids,
+                env_ids=env_ids,
+            )
```
alexmillane: Suggestion to add a small comment here about the approach taken. Basically that we scale the inertias by the same ratio that the mass changed by.

alexmillane (approved):
Looks great!!

TuurStuyck:
Thanks for the reviews! Addressed all the comments. Summary of what changed:

min_mass handling (@alexmillane, @arena-review-bot)
- Removed min_mass as a configurable cfg field. As suggested, the sampler range is now the user's knob; we only keep a fixed internal floor to reject non-physical masses.
- That floor is a single module constant _MIN_PHYSICAL_MASS_KG = 1e-6, replacing the previously duplicated 1e-6 literals.

Wasteful full-tensor clones (@greptile-apps, P2)
- Dropped the per-reset clones of the full body_mass / body_inertia tensors. 

snake_case class name (@greptile-apps, P2)
- Renamed apply_object_mass_from_sampler → ApplyObjectMassFromSampler (CapWords), updated all references.

Local import (@alexmillane)
- It wasn't needed. The variations layer has no import dependency back on assets, so no cycle exists. 

selected_bodies → selected_bodies_count (@alexmillane) — done.

Inertia-scaling comment (@alexmillane)
- Added a comment explaining we scale each inertia tensor by the same ratio the mass changed by (uniform-density, I ∝ m). Also added a test (test_object_mass_variation_scales_inertia) that asserts inertia / mass is constant across envs with distinct sampled masses, so the scaling is actually verified.

Safety asserts under python -O (@greptile-apps, P1)
- Kept these as assert to follow the repo convention.

Sampler CPU→GPU transfer (@alexmillane)
- Left an in-code comment at the transfer site documenting it so it isn't lost.

PR #926, opened 2026-07-17
Share one timestamped output directory across Experiment chunks

cvolkcvolk:
## Summary
Share timestamped output across chunks

## Detailed description
- Expose `execute_run(...)` as the focused building block for executing one `ArenaRunCfg` inside a caller-owned `SimulationApp`.
- Preserve the existing `--output_base_dir` behavior: every invocation creates one timestamped Experiment directory below the configured base.
- Choose that directory once in the parent and pass it privately to every legacy chunk worker.
- Keep a fresh `SimulationApp` per chunk while producing one directory and one combined `index.html` for the whole Experiment.
- Leave OSMO-specific timestamp discovery to the distributed workflow in #920.

## Output layout

Normal and chunked Experiments now use the same layout:

```text
<output_base_dir>/
└── <timestamp>/
    ├── index.html
    ├── run-a/
    │   └── episode_results_rebuild0.jsonl
    └── run-b/
        └── episode_results_rebuild0.jsonl
```

For example, `--output_base_dir ./other-directory` writes the Experiment below `./other-directory/<timestamp>/`. Chunk workers reuse that timestamped directory rather than creating one directory per chunk.


cvolkcvolk:
Closing for now. This PR addresses the separate legacy JSON `--chunk_size` output TODO, while #920 runs typed single-Run Experiments and no longer depends on it. We can revisit legacy chunk output sharing in a focused PR with a simpler parent/child path contract.

PR #928, opened 2026-07-18
Revert "Fix Franka assembly asset path (#919)"

qianl-nv:
This reverts commit fa28eb3ec4eaae73c61bab73bff7a9a51932470d.

- The Arena franka with stand USD has been fixed on staging server. Update the franka embodiment default to use it.
- Update the gear assembly tasks which uses a different cfg to also use the franka with stand.


alexmillane (commented):
LGTM!

cvolkcvolk (approved):
Thanks!

qianl-nv:
Merged only the franka embodiment path update in https://github.com/isaac-sim/IsaacLab-Arena/pull/925
keeping the assembly task config using the lab legacy path

PR #929, opened 2026-07-20
Distribute Arena Experiment Runs with exact outputs

cvolkcvolk:
## Summary
Run every Arena Run independently on OSMO and publish one Experiment report.

## Detailed description
- Create one independently scheduled OSMO group per Run. A Pi0 Run gets its own policy-server task in the same group.
- Keep the timestamped `--output_base_dir` behavior used by existing local commands. OSMO instead passes `--experiment_output_directory` so each Experiment Runner writes to the exact directory owned by its task.
- After every Run succeeds, the CPU-only `build-experiment-output` task collects the Run directories, generates `index.html`, and publishes one complete Experiment output. If any Run fails, the workflow fails and no combined report is published; retries and partial reports remain out of scope.

## Run it on OSMO
This submits the nine Runs in the OpenPI Experiment. The Experiment Runner uses the default `nvcr.io/nvstaging/isaac-amr/isaaclab_arena:latest` image.

```bash
/isaac-sim/python.sh -m osmo.submit_arena_experiment \
    --experiment_cfg isaaclab_arena_environments/experiment_configs/droid_pnp_srl_openpi_experiment.yaml \
    --policy_server pi0 \
    osmo.workflow_name=arena-parallel-test \
    osmo.pool=isaac-dev-l40-03 \
    osmo.platform=ovx-l40 \
    osmo.memory=120Gi \
    experiment_runner.image=nvcr.io/nvstaging/isaac-amr/isaaclab_arena:osmo_parallel

```


osmo/scripts/build_experiment_output.py
```diff
@@ -0,0 +1,130 @@
...
+def _parse_arguments() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--experiment-runner-output-directories-file",
+        required=True,
+        type=Path,
+        help="JSON mapping of each Run name to its Experiment Runner task output directory",
+    )
+    parser.add_argument(
+        "--experiment-output-directory",
```
xyao-nv, line 109: Nit, can you include a download script to retrieve the dir to local in another MR?

xyao-nv (approved):

osmo/scripts/build_experiment_output.py
cvolkcvolk: Yes good idea

osmo/tasks/base_task.py
```diff
@@ -28,14 +28,18 @@ def __init__(
         self,
         task_cfg: TaskCfg | None = None,
         lead: bool | None = None,
+        task_name: str | None = None,
+        resource: str | None = None,
     ) -> None:
         self.task_cfg = task_cfg
         self.lead = lead
+        self.task_name = task_name or self.get_task_name()
```
alexmillane: Why allow both methods of specifying the task name? I.e. via an argument to `super()`, or via overriding `get_task_name()`

osmo/tasks/collect_experiment_outputs_task.py
alexmillane, line 1: Am I correct this is for collecting all the outputs. Wdyt about `collect_experiment_outputs_task`

alexmillane (approved):
LGTM

osmo/tasks/base_task.py
cvolkcvolk: Mh  good point, I thinkg that was partly due to legacy code where `get_task_name()` provided defaults for static workflows, while `task_name=`  allowed to create multiple uniquely named instances.. I removed `get_task_name()` everywhere and made task_name a required constructor argument. I think that is more explicit

osmo/tasks/collect_experiment_outputs_task.py
cvolkcvolk: Updated

PR #931, opened 2026-07-21
Move RoboLab episode_length_s to composite task root

qianl-nv:
## Summary
Set RoboLab task episode limits on the composite root and wire conversion.

## Detailed description
- Add `episode_length_s` to `CompositeTaskSpec` and pass it into composite/atomic task constructors during graph conversion
- Move `episode_length_s` out of RoboLab subtask params into each task YAML root, using upstream RoboLab benchmark durations
- Update graph-spec tests and fixtures to match the new schema

### DIFF
--- a/isaaclab_arena/tests/test_data/robolab_task_overlay.yaml
+++ b/isaaclab_arena/tests/test_data/robolab_task_overlay.yaml
@@ -8,10 +8,10 @@ external_yaml: robolab_scene_base.yaml
 task:
   composition: atomic
   description: pick up the banana and place it in the bowl
+  episode_length_s: 50.0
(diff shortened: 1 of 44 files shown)


isaaclab_arena/tests/test_data/robolab_task_overlay.yaml
```diff
@@ -8,10 +8,10 @@ external_yaml: robolab_scene_base.yaml
 task:
   composition: atomic
   description: pick up the banana and place it in the bowl
+  episode_length_s: 50.0
```
qianl-nv: revert this from 50 back to 20

PR #934, opened 2026-07-21
Install OpenPI client in native uv environments

cvolkcvolk:
## Summary
Fix failing ci nightly on main. Install OpenPI client with native uv. 

Verification: Previously failed CI test passes locally.
```
uv run --no-sync python -m pytest -sv --durations=0 \
  isaaclab_arena/tests/test_osmo_experiment_workflow.py
```

## Detailed description
- Native uv CI could not collect the OSMO experiment tests because `openpi-client` was only installed by the Docker build.
- Add the client as a default uv dependency group, pinned to the same upstream revision used by the OpenPI server.
- Refresh the lockfile and guard the shared revision so client/server compatibility cannot drift.
- Leave the existing Docker installation path unchanged.

### DIFF
--- a/isaaclab_arena/tests/test_packaging.py
+++ b/isaaclab_arena/tests/test_packaging.py
@@ -12,7 +12,9 @@ bare ``--no-deps --editable`` install in the native uv job.
 """
 
 import importlib.util
+import tomllib
 from importlib import metadata
+from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


alexmillane (approved):
Thanks for fixing that.

I guess we need to move to conditionally testing the uv install pre-merge. This just shows how easy things are to break.

I'll work on that.

PR #939, opened 2026-07-22
Add radial bounds to PositionLimits

lgulich:
## Summary

- Extend the generic `PositionLimits` relation with optional world-XY disk/annulus bounds.
- Preserve legacy positional `relation_loss_weight` compatibility and add a deterministic exact-center lower-radius gradient path.
- Cover validation, batching, weighting, solver behavior, generic graph conversion, and public YAML/Python usage.

## Motivation

Rectangular bounds cannot express radial placement regions without approximating them with a box. This adds disk/annulus constraints while keeping existing axis limits composable.

## Validation

- `pre-commit` on changed files
- `compileall` / syntax checks
- YAML transport assertions and diff checks
- Independent task, whole-branch, and adversarial reviews

Focused Isaac Sim pytest remains pending: this checkout has no pinned local environment or running Arena Docker container. No bootstrap or image pull was performed.


lgulich:
Superseded by #940, which is rebased on current main and contains only the generic radial PositionLimits feature.

PR #945, opened 2026-07-22
Refactor: Fold stand-height lift into Droid initial_pose

xyao-nv:
## Summary
Make the Droid robot's base pose and the scene's robot init state always match, and move the pose-shift math into one shared helper.

## Detailed description
- Add translate_by_xyz_offset as a helper func
- With no explicit nitial_pose, lift the default robot and stand init states; with an explicit pose, lift that pose and write it straight into the scene robot init state.
- Lift an explicit base pose once, at set_initial_pose time, instead of re-applying the lift later. 

## Note on how embodiment pose is set
1. Users pass a pose into the constructor — DroidEmbodiment(initial_pose=somePose)
2. Users pass nothing — DroidEmbodiment() (the default, initial_pose=None)
3. Users call set_initial_pose(...) later

### DIFF
--- a/isaaclab_arena/utils/pose.py
+++ b/isaaclab_arena/utils/pose.py
@@ -83,13 +83,13 @@ def compose_poses(T_C_B: Pose, T_B_A: Pose) -> Pose:
 
 
 @overload
-def translate_by_xyz_offset(target: Pose, xyz_offset: tuple[float, float, float]) -> Pose: ...
+def translate_by_xyz_offset(target: Pose, xyz_offset: tuple[float, float, float]) -> Pose: ...  # noqa: E704
 
 
 @overload
 def translate_by_xyz_offset(
     target: tuple[float, float, float], xyz_offset: tuple[float, float, float]
-) -> tuple[float, float, float]: ...
+) -> tuple[float, float, float]: ...  # noqa: E704
 
 
 def translate_by_xyz_offset(
--- a/isaaclab_arena/utils/pose.py
+++ b/isaaclab_arena/utils/pose.py
@@ -5,7 +5,6 @@
 
 import torch
 from dataclasses import dataclass
-from typing import overload
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


isaaclab_arena/utils/pose.py
```diff
@@ -81,6 +81,23 @@ def compose_poses(T_C_B: Pose, T_B_A: Pose) -> Pose:
     return Pose(position_xyz=tuple(t_C_A.tolist()), rotation_xyzw=tuple(q_C_A.tolist()))
 
 
+def translate_by_xyz_offset(
+    target: tuple[float, float, float] | Pose, xyz_offset: tuple[float, float, float]
+) -> tuple[float, float, float] | Pose:
```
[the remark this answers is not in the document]
xyao-nv: good idea to seperate it out.

isaaclab_arena/utils/pose.py
```diff
@@ -81,6 +82,33 @@ def compose_poses(T_C_B: Pose, T_B_A: Pose) -> Pose:
...
+def translate_by_xyz_offset(target: Pose, xyz_offset: tuple[float, float, float]) -> Pose: ...  # noqa: E704
+
+
+@overload
+def translate_by_xyz_offset(
+    target: tuple[float, float, float], xyz_offset: tuple[float, float, float]
+) -> tuple[float, float, float]: ...  # noqa: E704
+
+
+def translate_by_xyz_offset(
```
qianl-nv: we have 4 @overload wrappers to do a simple "add a z-offset" operation, seems quite an overkill and hard to read. I don't think we should try to hide the tuple vs Pose representation difference just by adding unnecessary layers of abstraction. 

how about 
- define a standalone translate_by_xyz_offset just for tuple
- then a Pose class method and use it for the Pose cases.

```
class Pose:
    def translate(self, xyz_offset: tuple[float, float, float]) -> "Pose":
        """Return this pose shifted by ``xyz_offset`` (rotation unchanged)."""
        return Pose(position_xyz=translate_by_xyz_offset(self.position_xyz, xyz_offset), rotation_xyzw=self.rotation_xyzw)
```

on droid.py we store 

```
self._robot_base_offset = (0, 0, stand_height_m - _DEFAULT_STAND_HEIGHT_M)
...
self.scene_config.robot.init_state.pos = translate_by_xyz_offset(self.scene_config.robot.init_state.pos, self._robot_base_offset)
...
    def set_initial_pose(self, pose: Pose) -> None:
        super().set_initial_pose(pose.translate(self._robot_base_offset))
```

qianl-nv (approved):
Thanks for cleaning this up. Good change, just one nit for readability, feel free to merge after fixing.

isaaclab_arena/utils/pose.py
xyao-nv: good suggestion, updated.

PR #947, opened 2026-07-24
Reachability only validates tasks-relevant objects (Part 2.5/3)

xyao-nv:
## Summary
Reachability used to validate over all non-anchor objects in the candidate layout (https://github.com/isaac-sim/IsaacLab-Arena/pull/914). In this MR, it checks reachability against task-related objects. For PnP tasks, it is auto-configured as pickup objects and place destination objects.

## Detailed description
- Introduced `task_constraints` in the env graph spec to capture the reachability constraint.
- Reachability of an `Object` is an `RelationBase` tied to the object, which affects the geometry of layout (only at the validation step, not solving step). Defined as `RequiresReachability`.
- `PickAndPlaceTask` specified what objects need to be reached thru a init data member `reachability_target_objects`. This data is also added to the `TaskBase.`
- Tested on one Robolab env

## TODO
1. More testing over other Robolab envs
2. Enable joint collision checking

Screenshot from 2026-07-23 21-45-10
(banana and bowl are checked against reachability.)


isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
```diff
@@ -38,6 +39,7 @@ def build_arena_env_from_graph_spec(graph_spec: ArenaEnvGraphSpec, enable_camera
     assets_by_node_id = instantiate_assets_from_spec(graph_spec, AssetRegistry(), enable_cameras=enable_cameras)
     _ensure_scene_lighting(graph_spec, assets_by_node_id)
     _attach_spatial_relations_to_assets(graph_spec.relations, assets_by_node_id)
+    _attach_reachability_relations_to_assets(graph_spec.get_reachability_target_object_ids(), assets_by_node_id)
```
xyao-nv: Open for discussions if any alternative to tie the reachability to object. It does not make sense to me to add it as a property to the object class. So I treat it as a Relation instead.

isaaclab_arena/environment_spec/arena_env_graph_spec.py
```diff
@@ -66,9 +72,47 @@ def validate(self) -> Self:
...
+        object_ids = {*(obj.id for obj in self.objects), *(ref.id for ref in (self.object_references or []))}
+        # dict.fromkeys de-duplicates while preserving order: explicit 'reachable' constraints first, then auto-targets.
+        return tuple(
+            dict.fromkeys((
+                *(c.subject for c in (self.task_constraints or []) if c.type is TaskConstraintType.REACHABLE),
+                *self._select_task_declared_reachable_subjects(object_ids),
+            ))
+        )
+
+    def _select_task_declared_reachable_subjects(self, object_ids: set[str]) -> list[str]:
```
xyao-nv: Open for discussion if this shall be handled by the agent. I think it's reasonable to be configured by the task interface. Apart from a navigation task, most manipulation tasks require object-robot interaction.

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -52,6 +52,13 @@ def _factory(arm_mode):
     The factory receives ``arm_mode`` from the env builder and returns a constructed cfg.
     """
 
+    reachability_target_objects = ("pick_up_object", "destination_location")
+    """The robot must reach the object it picks up and the it places onto.
+
+    An object reference or the background scene (a static location, not a placed object) is skipped
```
xyao-nv: I'm debating the tradeoff to support object reference for reachability validation. Here is my thinking process. Open for discussion.
Object reference is never a placed entity thru the solver (e.g. shelf inside the fridge). But its parent object (the fridge) could be inside the background scene (e.g. kitchen) or a foreground object (e.g. the microwave on table). It does not make sense to reject layout if it's a background for unreachable as none of them will be returned. 
Can be in another MR if enable the object reference if it's a foreground placeable object is necessary.

isaaclab_arena_environments/robolab/tasks/banana_on_plate.yaml
```diff
@@ -5,6 +5,11 @@
 
 env_name: banana_on_plate
 external_yaml: ../scenes/bagel_plate_banana_bowl.yaml
+task_constraints:
+- type: reachable
+  subject: banana
+- type: reachable
+  subject: plate
```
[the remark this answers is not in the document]
qianl-nv: I agree, the explicit task_constraints in yaml/env seems unnecessary
the task interface already defines which objects are reachability_target_objects, there's no need to expose this again through env/yaml. I'd suggest remove this and the task_constraints TaskConstraintSpec  ect entirely.

a user can design to switch on/off reachability for an env through validator params?

isaaclab_arena/relations/relations.py
```diff
@@ -275,6 +275,25 @@ def is_unary() -> bool:
         return True
 
 
+@register_object_relation
+class RequiresReachability(RelationBase):
```
qianl-nv, line 279: parent class should be UnaryRelation

isaaclab_arena/environment_spec/arena_env_graph_spec.py
```diff
@@ -46,11 +48,15 @@ class ArenaEnvGraphSpec(BaseModel):
         description="Per-env placement validators; none runs all build-time checks.",
     )
     task: CompositeTaskSpec = Field(description="Root task the robot performs to manipulate the objects.")
+    task_constraints: list[TaskConstraintSpec] | None = Field(
+        default=None,
+        description="Optional per-object task constraints.",
+    )
```
alexmillane: Point for discussion. Given that the reachability is described as a relation (which is probably correct IMO) is there a good reason to introduce another field here to hold it? Could we not have the reachability relation with the other relations.

With the splitting of the graph `yaml`s into scenes and tasks, relations can be added in a task-specific manner. They live inside the task portion of the yaml file(s) and are added to the scene if that task is used.

Wdyt?

@qianl-nv for viz

isaaclab_arena/environment_spec/arena_env_graph_types.py
```diff
@@ -147,6 +147,22 @@ def _validate_composition_task_count(self) -> CompositeTaskSpec:
...
+
+
+class TaskConstraintSpec(BaseModel):
+    """One task constraint: a requirement of ``type`` on the object named by ``subject``."""
+
+    type: TaskConstraintType = Field(description="Constraint kind; must match TaskConstraintType exactly.")
+    subject: str = Field(
+        min_length=1,
+        description="Graph node id of the object the constraint applies to. Must match an object id exactly.",
+    )
```
alexmillane: See comment above. Any reason not to just treat the reachable constraint as another relation?

Probably later, we'll actually want it to be a binary relation, rather than a unary one. The reachable constrain _relates_ the object and the robot.

isaaclab_arena/relations/relations.py
```diff
@@ -275,6 +275,25 @@ def is_unary() -> bool:
...
+    Usage:
+        banana.add_relation(RequiresReachability())  # IK-check reachability of the banana
+    """
+
+    name = "requires_reachability"
+
+    @staticmethod
+    def is_unary() -> bool:
+        """Return whether the relation constrains a single object."""
+        return True
```
alexmillane, line 293: We should possibly make this a binary constraint between the robot and the object. Right now the robot is sort of implied.

I wonder if this constraint could also have an affect on the optimization by adding a reach radius constraint to the optimization between the robot (which will soon be placeable) and the object(s).

alexmillane: Perhaps this should be the subject of a future MR though.

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -52,6 +52,13 @@ def _factory(arm_mode):
     The factory receives ``arm_mode`` from the env builder and returns a constructed cfg.
     """
 
+    reachability_target_objects = ("pick_up_object", "destination_location")
```
alexmillane: I'm wondering why these need to be added here?

Aren't we relying on the user to specify the reachability relation manually? I'm wondering why they're here? (maybe there's a good reason, I am just unsure).

isaaclab_arena/environment_spec/arena_env_graph_spec.py
```diff
@@ -66,9 +72,53 @@ def validate(self) -> Self:
...
+        """Object ids the robot shall be able to reach."""
+        object_ids = self._placed_object_ids()
+        # dict.fromkeys de-duplicates repeated object ids.
+        # It also preserves order esp for composite tasks for consistency, or if the user adds explicit 'reachable' constraints.
+        return tuple(
+            dict.fromkeys((
+                *(c.subject for c in (self.task_constraints or []) if c.type is TaskConstraintType.REACHABLE),
+                *self._select_task_declared_reachable_subjects(object_ids),
+            ))
+        )
```
alexmillane: nit: large compound comprehensions like this (i find) are basically impossible for a human to read. Suggestion to break things like this up over multiple lines and statements.

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
alexmillane: I'm thinking that attaching reachability constraints to the environment in the graph layer is the wrong place to do it. For starters, it means that the python way of specifying environments does get these.

A good comparison might be the non-collision constraints which are added in the env-builder / object placer.

I'm thinking that we could add to `PickAndPlaceTask` a function that adds reachability constraints to it's input objects? That's not perfect either but I feel it's better than having it in the graph layer.

alexmillane (commented):
Look pretty good. I left some points for discussion.

isaaclab_arena/environment_spec/arena_env_graph_spec.py
qianl-nv: I would actually go one step further and don't expose task_constraints at the env/yaml level.
I feel adding reachable relation to objects should be handled internally in the Task class, and user should not need to add the task-based constraints explicitly.

isaaclab_arena/relations/relations.py
qianl-nv: I would lean towards this being a "unary" constraint, since one of the end of the relation is always the robot (it doesn't make sense to be any other asset), and there's only one embodiment in an arena env.
so it's either define as unary, or as binary + asset that the reference is always THE embodiment. In that case going for unary seems simpler.

isaaclab_arena/tasks/pick_and_place_task.py
qianl-nv: Ha I was suggesting the exact opposite in my comment above lol... to not let user specify the reachability relation and also add them in the PickAndPlaceTask automatically.

isaaclab_arena/assets/object_base.py
```diff
@@ -190,6 +190,11 @@ def is_anchor(self) -> bool:
         """True if this object has an IsAnchor relation."""
         return any(isinstance(r, IsAnchor) for r in self.relations)
 
+    @property
```
qianl-nv: It might be a bad idea to add relation-type-specific API to the object_base class. Imaging we would have other task constraints and each time adding a new API to object_base sounds a bad idea.

either we make this a generic def has_relation(self, relation_type: Relation) in object_base
or make this a helper/util in the relations module instead.

qianl-nv (commented):
Thanks for the change, it's a good place to start refining.

Do we have any use cases where we want to add reachable relation to non-task-related object? If not it feels better to have the Task class automatically add the relation to its pick/target objects instead of always having the user specific the relation outside of the task.

The only main concern for that approach is backward compatibility: some of our existing envs (esp those hand crafted ones) might not even satisfy these relations in the first place. For those we might just explicitly switch the reachability checks off through PlacementParams instead.

wdyt? maybe I'm missing other scenarios where the user might want to have manual control over the reachability relations.

isaaclab_arena/relations/relations.py
xyao-nv: I don't agree, UnaryRelation has been using  for solver-optimized position constraints. If inheriting that, it will crash the solver as no loss strategy is implemented (yet) for this reachability.

We either rename the following func to be explicit on those are used in solver optimization in another MR, or change this to `UnaryRelation` once reachability loss is added.

If you insist suggesting UnaryRelation to use here, then a zero-loss strategy needed to be added as a no-op loss for non-geometric relations. For simplicity, I prefer subclassing `RelationBase.`

In object_base.py,
```
    def get_spatial_relations(self) -> list[RelationBase]:
        """Get only spatial relations (On, NextTo, AtPosition, etc.), excluding markers like IsAnchor."""
        return [r for r in self.relations if isinstance(r, (Relation, UnaryRelation))]
```

xyao-nv: Subclassing RelationBase is similiar to how `is_anchor` is used.

isaaclab_arena/relations/relations.py
xyao-nv: It makes sense to me, once radius constraint is added and tested, this can be changed to subclassing `BinaryRelation` class. Added a TODO.

Once robot-object is in, it could be binary relation.

xyao-nv: On a side note, there are other better candidates for a surrogate loss than the radius constraint for reachability. Lio's MR is just 2D, and does not reflect robot's reachability map.

isaaclab_arena/tasks/pick_and_place_task.py
xyao-nv: I think configuring them thru Task interface is a better design. It's a requirement from the task, and task interface can easily derive the target to be checked.

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -52,6 +52,13 @@ def _factory(arm_mode):
...
+    """The robot must reach the object it picks up and the it places onto.
```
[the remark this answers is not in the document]
xyao-nv: done

isaaclab_arena/environment_spec/arena_env_graph_types.py
xyao-nv: Removed task constraint.

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -52,6 +52,13 @@ def _factory(arm_mode):
...
+
+    An object reference or the background scene (a static location, not a placed object) is skipped
+    for reachability checks during layout validation.
+    """
```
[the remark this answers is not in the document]
xyao-nv: done

isaaclab_arena/tasks/task_base.py
```diff
@@ -19,6 +19,9 @@ class TaskBase(ABC):
 
     DEFAULT_EPISODE_LENGTH_S: float = 20.0
 
+    reachability_target_objects: tuple[str, ...] = ()
```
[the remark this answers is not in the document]
xyao-nv: Param is too board to me.

isaaclab_arena_environments/robolab/tasks/banana_on_plate.yaml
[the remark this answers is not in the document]
xyao-nv: Yes it was just for showcase. But I agree earlier to remove the task_constraints entirely.

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
xyao-nv: Yes, it was added.
And now it's only configured in the task.

isaaclab_arena/assets/object_base.py
xyao-nv: valid, updated it to has_relation.

isaaclab_arena/tasks/task_base.py
```diff
@@ -70,6 +75,14 @@ def get_task_description(self) -> str | None:
     def get_progress_objectives(self) -> list[ProgressObjective]:
         return []
 
+    def apply_reachability_constraints(self) -> None:
+        """Apply RequiresReachability relations to each placeable object named in reachability_target_objects."""
+        for param_name in self.reachability_target_objects:
+            target = getattr(self, param_name)
+            # Only placed objects are IK-checked; an object reference or background location is skipped.
+            if isinstance(target, Object) and not target.has_relation(RequiresReachability):
+                target.add_relation(RequiresReachability())
```
alexmillane, line 88: Suggestion to move this function down to the child class.

We're effectively reaching into child class members in the base class. This, in general, is not something that you should do. I would view `getattr` as a sign that something has gone wrong in your design, unless you're explicitly doing meta-programming. Because it means you're not sure about the interface of the object you're dealing with.

In this case, I would just move this to the child, who knows about what objects should have reachability constraints.

Alternately you could do

```
# task_base.py
def apply_reachability_constraints(self, object: list[ObjectBase]) -> None:
    pass

def _apply_reachability_constraints(self, object: list[ObjectBase]) -> None:
    for target in objects:
        target.add_relation(RequiresReachability())

# pick_and_place.py
def apply_reachability_constraints(self, object) -> None:
    super._apply_reachability_constraints([self.pick_up_object, self.destination_object])
```

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -52,6 +52,13 @@ def _factory(arm_mode):
     The factory receives ``arm_mode`` from the env builder and returns a constructed cfg.
     """
 
+    reachability_target_objects = ("pick_up_object", "destination_location")
+    """The robot must reach the object it picks up and the location it places onto.
+
+    An object reference or the background scene (a static location, not a placed object) is skipped
+    for reachability checks during layout validation.
+    """
```
alexmillane: Remove these. See comment below.

alexmillane (approved):
Looks much cleaner.

I have one comment about the implementation of one function. Feel free to merge after addressing.

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
```diff
@@ -50,10 +50,9 @@ def build_arena_env_from_graph_spec(graph_spec: ArenaEnvGraphSpec, enable_camera
 
 def build_checks_for_placer_params(graph_spec: ArenaEnvGraphSpec) -> ObjectPlacerParams:
     """Build placement params defining what checks to run during layout validation for this env."""
-    placement = graph_spec.placement_validators
```
qianl-nv, line 53: seems a unrelated stray rename-only change, suggest to revert if not needed

isaaclab_arena/relations/relations.py
qianl-nv: You convinced me here 👍

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -86,6 +86,9 @@ def _solve_relations(self) -> None:
         * **False** — applies one layout per environment so per-object reset
           events restore the same layout every time.
         """
+        # Reachability constraints are defined in the task, so apply them before placement.
+        if self.arena_env.task is not None:
+            self.arena_env.task.apply_reachability_constraints()
```
qianl-nv, line 91: Suggest to rename apply_reachability_constraints() to apply_task_constraints() or simply apply_constraints(). we might have other types of task-based constraints in the future that can all be bundled into one external API.

qianl-nv (approved):
Approved with a renaming suggestion and a small nit. Thanks for a very well organized review round!

isaaclab_arena/tasks/task_base.py
xyao-nv: Good suggestion done!

isaaclab_arena/tasks/pick_and_place_task.py
xyao-nv: done

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
xyao-nv: it's better naming.

isaaclab_arena/environments/arena_env_builder.py
xyao-nv: I don't agree. It's an explicit function call at this point as there is only 1 task-based constraint. Can be renamed once more are added.

isaaclab_arena/relations/placement_asset.py
```diff
@@ -50,11 +50,20 @@ def get_spatial_relations(self) -> list[RelationBase]:
...
+        return any(isinstance(relation, relation_type) for relation in self.relations)
+
     @property
     def is_anchor(self) -> bool:
         """Return whether the asset is fixed during relation solving."""
         return any(isinstance(relation, IsAnchor) for relation in self.relations)
 
+    @property
+    def requires_reachability(self) -> bool:
+        """Return whether the asset carries a RequiresReachability marker (from a 'reachable' task constraint)."""
```
xyao-nv: nah

PR #954, opened 2026-07-27
Support Embodiment Placement for Droid/Franka

qianl-nv:
## Summary
Local on-stand USD compose for Franka and Droid to support placement with stand/camera

## Detailed description
- Stand and Droid external cameras were detaching on relation reset when spawn and reset layouts differed; parenting them under `panda_link0` lets a single `write_root_pose_to_sim` move the full assembly in Fabric/Kit.
- Franka and Droid now compose cached robot+stand USDs locally (`robot_on_stand_utils`, embodiment-specific mount helpers) instead of relying on Nucleus combined assets or a separate Droid stand scene entity.
- Remove has_unplaced_auxiliary_prims API since the only use case Droid has been updated to a single USD.
- Added `test_robot_on_stand_utils.py` and merge `test_droid_stand_height.py` into it.

Before:
image

After:
image


isaaclab_arena/embodiments/droid/droid_stand_usd.py
```diff
@@ -0,0 +1,78 @@
...
+        stand_translate_xyz: Stand child translate in the robot-base frame.
+        stand_footprint_scale_xy: Stand x/y scale.
+        layout: Prim-name layout under the default root.
+
+    Returns:
+        Local path to the composed on-stand USD.
+    """
+    unit_height = measure_mesh_prim_height_m(stand_usd_path, stand_mesh_prim_path)
+    scale_z = stand_height_m / unit_height
+    stand_resolved = retrieve_file_path(stand_usd_path)
```
qianl-nv: can remove this and use stand_path in the mount_stand_fn?

isaaclab_arena/embodiments/droid/droid_stand_usd.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def mount_stand_fn(stage: Usd.Stage, stand_layout: OnStandPrimLayout, stand_path: str) -> None:
```
qianl-nv: it seems the only droid vs franka difference is in the mount_stand_fn
and it's due to droid needed to additionally handle stand_height_m before composing the stand usd

if droid and franka needs different handling for the stand  due to different origin of the stand, then can the composed usd has the an additional parent prim to the stand, so it's default pose is always zero (corresponding the stand top of mesh at 0), with the stand_height_m based scale; its child (the actual stand USD reference) has a fixed z shift based on default height to align the stand top of mesh to zero

can we unify both? then droid_stand_usd.py and franka_stand_usd.py can be removed

isaaclab_arena/embodiments/robot_on_stand_utils.py
```diff
@@ -0,0 +1,248 @@
...
+from dataclasses import dataclass
+
+from isaaclab.utils.assets import retrieve_file_path
+from pxr import Gf, Usd, UsdGeom
+
+from isaaclab_arena.assets.asset_cache import get_arena_asset_cache_dir
+from isaaclab_arena.utils.usd_helpers import compute_local_bounding_box_from_prim, open_stage
+
+# Prim layout shared by Franka and Droid on-stand assemblies (default prim ``/panda``).
+ON_STAND_ROOT_PRIM_NAME = "panda"
```
qianl-nv: this should probably be called panda_on_stand_utils.py, since it hard codes the panda robot USD prim structure

isaaclab_arena/tests/test_on_stand_root_link_follow.py
```diff
@@ -0,0 +1,159 @@
+# Copyright (c) 2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
qianl-nv: rename this to test_robot_on_stand_utils.py. merge test_droid_stand_height.py into this

qianl-nv (commented):
self review 1

isaaclab_arena/embodiments/franka/franka.py
```diff
@@ -449,3 +452,15 @@ def get_object_poses(self, env_ids: Sequence[int] | None = None):
         object_pose_matrix = get_rigid_and_articulated_object_poses(state, env_ids)
 
         return object_pose_matrix
+
+
+def _franka_robot_cfg_on_stand(robot_cfg: ArticulationCfg) -> ArticulationCfg:
+    """Copy ``robot_cfg`` onto ``{ENV_REGEX_NS}/Robot`` with the composed on-stand USD."""
+    cfg = robot_cfg.replace(prim_path="{ENV_REGEX_NS}/Robot")
```
qianl-nv, line 461: this line is not in original code. safe to remove?

isaaclab_arena/tests/test_embodiment_collision_mesh.py
```diff
@@ -20,8 +20,9 @@ def _test_embodiment_provides_robot_collision_mesh(simulation_app) -> bool:
         assert mesh is not None, "embodiment must expose a collision mesh; None forces the loose bbox fallback"
         assert len(mesh.vertices) > 0
 
-        # The default prim scopes extraction to the arm. The Droid USD also bakes in a 50 m ground
```
qianl-nv, line 23: revert doc change

isaaclab_arena/tests/test_robot_on_stand_utils.py
```diff
@@ -0,0 +1,285 @@
...
+import pytest
+import warp as wp
+
+from isaaclab_arena.tests.utils.subprocess import run_simulation_app_function
+
+_HEIGHT_ATOL = 1e-3
+_ROOT_WRITE_ATOL = 5e-3
+_ROOT_WRITE_DELTA_X = 1.0
+_CUSTOM_DROID_STAND_HEIGHT_M = 2.0
+_HEADLESS = True
```
qianl-nv: do we need this?

isaaclab_arena/tests/test_robot_on_stand_utils.py
```diff
@@ -0,0 +1,285 @@
...
+from pathlib import Path
+
+import pytest
+import warp as wp
+
+from isaaclab_arena.tests.utils.subprocess import run_simulation_app_function
+
+_HEIGHT_ATOL = 1e-3
+_ROOT_WRITE_ATOL = 5e-3
+_ROOT_WRITE_DELTA_X = 1.0
```
qianl-nv, line 22: what's this constant for? check it

isaaclab_arena/tests/test_robot_on_stand_utils.py
```diff
@@ -0,0 +1,285 @@
...
+            payload_child_name="Stand",
+        )
+    except Exception as e:
+        print(f"Error: {e}")
+        traceback.print_exc()
+        return False
+    return True
+
+
+def _test_franka_stand_under_link0_follows_root_write(simulation_app) -> bool:
```
qianl-nv: this can be removed?

isaaclab_arena/tests/test_robot_on_stand_utils.py
```diff
@@ -0,0 +1,285 @@
...
+        default_usd = _assert_compose_on_stand_usd(
+            _DROID_ROBOT_USD_PATH,
+            _DROID_STAND_MOUNT,
+            stand_height_m=_DEFAULT_STAND_HEIGHT_M,
+            output_basename="droid_franka_robotiq_on_stand",
+            payload_child_name="franka_table",
+            check_orient_180z=True,
+        )
+
+        custom_usd = _assert_compose_on_stand_usd(
```
qianl-nv, line 137: remove this 180 rot usd

isaaclab_arena/tests/test_robot_on_stand_utils.py
```diff
@@ -0,0 +1,285 @@
...
+            _DROID_STAND_MOUNT,
+            stand_height_m=_CUSTOM_DROID_STAND_HEIGHT_M,
+            output_basename="droid_franka_robotiq_on_stand",
+            payload_child_name="franka_table",
+        )
+        assert custom_usd != default_usd
+        assert abs(_stand_top_z(Usd.Stage.Open(custom_usd)) - _stand_top_z(Usd.Stage.Open(default_usd))) < _HEIGHT_ATOL
+
+        default_emb = DroidAbsoluteJointPositionEmbodiment()
+        custom_emb = DroidAbsoluteJointPositionEmbodiment(stand_height_m=_CUSTOM_DROID_STAND_HEIGHT_M)
```
qianl-nv, line 153: need to check the actual stand height matches config

qianl-nv (commented):
self review 2

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -70,24 +70,14 @@ def __init__(
...
-            self.scene_config.robot.init_state.pos = translate_by_xyz_offset(
-                self.scene_config.robot.init_state.pos, self._robot_base_offset
-            )
-            self.scene_config.stand.init_state.pos = translate_by_xyz_offset(
-                self.scene_config.stand.init_state.pos, self._robot_base_offset
-            )
-        else:
-            # Explicit base pose: lift it via set_initial_pose; get_scene_cfg writes the scene config later.
-            self.set_initial_pose(self.initial_pose)
+        self.scene_config.robot.spawn.usd_path = compose_on_stand_usd(
```
zhx06, line 86: Are we assuming that there will not be object parts? My [MR](https://github.com/isaac-sim/IsaacLab-Arena/pull/946) addresses the same issue but in different perspective. Should I rework my MR after yours got merged into main?

isaaclab_arena/embodiments/robot_on_stand_utils.py
```diff
@@ -0,0 +1,151 @@
...
+
+    try:
+        stage = Usd.Stage.CreateNew(str(tmp_path))
+        root = stage.DefinePrim(ROBOT_ROOT_PRIM_PATH, "Xform")
+        robot_resolved = retrieve_file_path(robot_usd_path)
+        root.GetReferences().AddReference(robot_resolved, ROBOT_ROOT_PRIM_PATH)
+        stage.SetDefaultPrim(root)
+
+        _mount_stand_normalized(stage, mount, stand_height_m)
+        stage.GetRootLayer().Save()
```
zhx06: Will `Save()` returns non-zero value but not get catched?

isaaclab_arena/embodiments/robot_on_stand_utils.py
```diff
@@ -0,0 +1,151 @@
...
+from pathlib import Path
+
+from isaaclab.utils.assets import retrieve_file_path
+from pxr import Gf, Usd, UsdGeom
+
+from isaaclab_arena.assets.asset_cache import get_arena_asset_cache_dir
+
+ROBOT_ROOT_PRIM_PATH = "/panda"
+# The PhysX articulation root link modified by reset ``write_root_pose_to_sim``.
+ROBOT_BASE_PRIM_PATH = f"{ROBOT_ROOT_PRIM_PATH}/panda_link0"
```
zhx06: We introduced some path here, and also in `droid.py`. Is there a way to organize the paths so the paths have one source of truth and are easy to update later?

xyao-nv: The default is very specific to franka-arm-based robot, and I would suggest reading it from robot.py, instead of a hyper-param in this file.

isaaclab_arena/embodiments/robot_on_stand_utils.py
```diff
@@ -0,0 +1,151 @@
...
+
+    return str(out_path)
+
+
+def _mount_stand_normalized(
+    stage: Usd.Stage,
+    mount: StandMountSpec,
+    stand_height_m: float,
+) -> None:
+    """Parent a stand payload under ``panda_link0``, scale, and align to the robot base.
```
xyao-nv: ``panda_link0`` is franka-arm specific. You can change it to a common name referring to the foundation mounted to floor, wall, or mobile platform, ``base_link``.

isaaclab_arena/embodiments/robot_on_stand_utils.py
```diff
@@ -0,0 +1,151 @@
...
+ROBOT_BASE_PRIM_PATH = f"{ROBOT_ROOT_PRIM_PATH}/panda_link0"
+STAND_PRIM_PATH = f"{ROBOT_BASE_PRIM_PATH}/stand_instanceable"
+_ROBOT_ON_STAND_USD_CACHE_DIR = "robot_on_stand"
+
+_HEIGHT_ATOL = 1e-3
+_ALIGN_ATOL = 5e-2
+
+
+@dataclass(frozen=True)
+class StandMountSpec:
```
xyao-nv: Maybe beyond this MR, but can start with moving this data class out of this utils.py.

A `stand/` can be added, so we can have ``metal_pillar_stand.py``, ``operating_table_stand.py``. And fix-based manipulator can choose different stand to mount to their base link.

xyao-nv (approved):

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -70,24 +70,14 @@ def __init__(
         stand_height_m: float = _DEFAULT_STAND_HEIGHT_M,
```
xyao-nv: Can you update docstring explaining the impact to the bottom location of the stand, if user is to set it manually?

qianl-nv: done

isaaclab_arena/embodiments/droid/droid.py
qianl-nv: yes we assume the embodiment gets only a single USD with articulation root.
Your MR won't be needed as is with this change. It would be more useful to extend it to handle the multi-parts by articulation links, not by separate scene entities.

isaaclab_arena/embodiments/franka/franka.py
qianl-nv: this is actually inline on main. should not drop

isaaclab_arena/embodiments/robot_on_stand_utils.py
qianl-nv: good point, moved all constants into the robot-specific .py. see Xinjie's comment below as well

qianl-nv: great suggestion. updated

isaaclab_arena/embodiments/robot_on_stand_utils.py
qianl-nv: good point, long term we should probably have a separate “stand_library.py" with all available stands USD. Let's do it later when we have a use case to swap stands.

isaaclab_arena/embodiments/robot_on_stand_utils.py
```diff
@@ -0,0 +1,151 @@
...
+        output_basename: Stable basename for the composed USD (embodiment-specific).
+
+    Returns:
+        Local path to the composed on-stand USD.
+    """
+    assert stand_height_m > 0.0, f"stand_height_m must be positive, got {stand_height_m}"
+
+    cache_root = get_arena_asset_cache_dir().parent / "usd" / _ROBOT_ON_STAND_USD_CACHE_DIR
+    cache_root.mkdir(parents=True, exist_ok=True)
+    out_path = cache_root / f"{output_basename}_{stand_height_m:.3f}.usd"
```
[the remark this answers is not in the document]
qianl-nv: 3 decimal = 1mm accurate, this is definitely enough precison for our use case.

isaaclab_arena/embodiments/robot_on_stand_utils.py
qianl-nv: good catch! add an assert to catch the non zero return value. comi

isaaclab_arena/embodiments/robot_on_stand_utils.py
qianl-nv: Updated to base_link, good suggestion!

isaaclab_arena/tests/test_embodiment_collision_mesh.py
qianl-nv: done

isaaclab_arena/tests/test_robot_on_stand_utils.py
qianl-nv: done

isaaclab_arena/tests/test_robot_on_stand_utils.py
qianl-nv: keep it to cache the 180 droid stand rotatino regression

isaaclab_arena/tests/test_robot_on_stand_utils.py
qianl-nv: already there in _assert_compose_on_stand_usd

isaaclab_arena/tests/test_robot_on_stand_utils.py
qianl-nv: this the value the test resets on the robot stand so it should be large enough. it's not a closeness threshold

isaaclab_arena/tests/test_robot_on_stand_utils.py
qianl-nv: yes. default is already headless

qianl-nv (commented):
Addressed all review comments

zhx06 (approved):
LGTM!

PR #960, opened 2026-07-27
Bump Isaac Lab submodule and enable imitation learning script interop

peterd-NV:
## Summary
Bump Isaac Lab submodule and enable Arena interop with Isaac Lab imitation learning scripts:

1. Updates Isaac Lab to latest release branch commit to pickup script changes required for Arena imitation learning interop
2. Update Arena docs to use Isaac Lab scripts for record, annotate, data generation
3. Add demo recorder config to Arena env to enable passing Arena's custom recorder configs to Lab's IL scripts
4. Update tests to use Lab IL scripts
5. Remove test that triggered when upstream Lab fixed the legacy franka path patch and replaced with a TODO comment. UV from wheel tests still use old artifacts with broken paths so the path patch is still needed even if from-source builds are fixed.
6. Update pin-pink version in pyproject.toml to match with Lab submodule's version.
7. Update uv.lock with new package versions to match with Lab's submodule's versions
8. Rebuild and push new CI base image for Isaac Lab. Tagged image with `lab_beta_2_af1bab4dc17`. Can be replaced with latest tag post merge in separate PR.


qianl-nv (commented):
Clean change, thanks!
One question: shall we remove the Arena scripts (isaaclab_arena/scripts/imitation_learning/replay_demos.py, ect) now that we have the interop?

qianl-nv (approved):

PR #968, opened 2026-07-28
Joint Dependent Mesh/BBox for Embodiment

zhx06:
## Summary
Joint-dependent mesh/bbox for embodiments

## Detailed description
- Compute embodiment bounding boxes at configured initial joint positions
- Use one combined posed link-box collision mesh for Droid and Franka
- Preserve existing list-based joint-pose APIs
- Enable Droid mesh collision in the Lightwheel kitchen; other embodiment configurations retain bbox collision by default

## Tests
- Joint-dependent bounding boxes and collision meshes match configured Franka and Droid joint poses.
- Cached bounding boxes and meshes return mutation-safe copies.
- Completed 20 episodes for both Droid bbox and mesh collision modes with `placement_seed=17`.
- Across five placement solves, `on_relation`, `not_next_to`, `face_to`, and `no_overlap` validation counts matched between modes. `next_to` remained comparable.

Command to run:
```bash
/isaac-sim/python.sh isaaclab_arena/evaluation/policy_runner.py \
  --viz kit \
  --policy_type zero_action \
  --num_episodes 20 \
  --placement_seed 17 \
  --disable_fabric \
  --device cpu \
  --env_graph_spec_yaml isaaclab_arena_environments/kitchen_bench/droid_pick_and_place_lightwheel_kitchen.yaml
```

With Droid mesh collision and solver logging enabled, five solves averaged:
```text
[RelationSolver] solve: 4434.8 ms average | batch=50 | objects=3 optimizable + 3 anchors | no-overlap pairs=8 | iters=600 (7.39 ms/iter)
```

Versus the previous Droid bbox configuration:
```text
[RelationSolver] solve: 4140.6 ms average | batch=50 | objects=3 optimizable + 3 anchors | no-overlap pairs=12 | iters=600 (6.90 ms/iter)
```

This is an end-to-end configuration comparison: Droid mesh collision was approximately 7.1% slower, while using fewer no-overlap pairs.


isaaclab_arena/assets/asset_cache.py
```diff
@@ -13,3 +13,11 @@ def get_arena_asset_cache_dir() -> pathlib.Path:
     if not asset_cache_dir.exists():
         asset_cache_dir.mkdir(parents=True, exist_ok=True)
     return asset_cache_dir
+
+
+def get_arena_usd_cache_dir() -> pathlib.Path:
+    """Return the cache root for USDs Arena generates, such as composed and baked-geometry assets.
+
+    The directory is not created, so callers that only compute a path do not leave one behind.
```
qianl-nv: is there any risk of having an empty directory? if not let's just create the dir here following get_arena_asset_cache_dir

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -56,37 +81,46 @@ def __init__(
...
         spawn = robot.spawn
         assert spawn.usd_path is not None, "scene_config.robot must use a USD spawn for placement"
-        scale = tuple(spawn.scale or (1.0, 1.0, 1.0))
-        # TODO(zihaox): Account for configured initial joint positions in bounds and collision meshes.
-        return compute_local_bounding_box_from_usd(spawn.usd_path, scale)
+        scale_x, scale_y, scale_z = spawn.scale or (1.0, 1.0, 1.0)
+        return PlacementGeometrySource(
+            usd_path=spawn.usd_path,
+            scale=(scale_x, scale_y, scale_z),
+            joint_pos=dict(robot.init_state.joint_pos or {}),
```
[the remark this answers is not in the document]
qianl-nv: This is better handle in droid/franka. set_initial_joint_pose currently only updates the reset event's default_pose. 
can expand set_initial_joint_pose to also all self.set_joint_initial_pos(...) with a name→value map matching set_default_joint_pose order.

isaaclab_arena/embodiments/robot_on_stand_utils.py
```diff
@@ -80,7 +80,7 @@ def compose_on_stand_usd(
     """
     assert stand_height_m > 0.0, f"stand_height_m must be positive, got {stand_height_m}"
 
-    cache_root = get_arena_asset_cache_dir().parent / "usd" / _ROBOT_ON_STAND_USD_CACHE_DIR
+    cache_root = get_arena_usd_cache_dir() / _ROBOT_ON_STAND_USD_CACHE_DIR
```
qianl-nv: thanks for the clean up!

qianl-nv (commented):
I have some general questions on why we need the "local/Omniverse cache for pre-computed mesh" part. 
- collecting the mesh for droid is only taking about 1s, it's hardly the bottlenet in the overall pipeline atm. we don't think we need to go done for the perf there using cache. Finding ways the speed up the mesh mode for Background (where we absolutely need it) is more important imo.
- for embodiment, what's blocking is actually the joint-angle-based bounding box collection. unless we are confident of shipping v0.3 with both background and embodiment using mesh mode (so far it has always take forever for solver), we need a working version of background in mesh mode + embodiment in bbox mode.

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -113,8 +111,15 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
         prim_path = _DROID_ROBOT_PRIM.stand_prim_path if self.placement_bbox_stand_only else None
         return super().get_bounding_box(prim_path=prim_path)
 
-    def set_initial_joint_pose(self, initial_joint_pose: list[float]) -> None:
-        self.event_config.init_franka_arm_pose.params["default_pose"] = initial_joint_pose
+    def set_initial_joint_pose(self, initial_joint_pose: Mapping[str, float]) -> None:
```
qianl-nv: can we keep this initial_joint_pose: list[float] type unchanged, and pre-define the list of joint name constants in matching order? I'd avoid unnecessary modification to existing API interface if possible

isaaclab_arena/embodiments/franka/franka.py
```diff
@@ -89,8 +86,16 @@ def __init__(
         self.observation_config.policy.concatenate_terms = self.concatenate_observation_terms
         self.add_camera_variations(self.camera_config)
 
-    def set_initial_joint_pose(self, initial_joint_pose: list[float]) -> None:
-        self.event_config.init_franka_arm_pose.params["default_pose"] = initial_joint_pose
+    def set_initial_joint_pose(self, initial_joint_pose: Mapping[str, float]) -> None:
```
qianl-nv: ditto

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -78,7 +79,6 @@ def __init__(
         self,
         enable_cameras: bool = False,
         initial_pose: Pose | None = None,
-        initial_joint_pose: list[float] | None = None,
```
qianl-nv, line 81: why remove this from init?

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -56,42 +71,57 @@ def __init__(
...
+        )
 
     def get_bounding_box(self, prim_path: str | None = None) -> AxisAlignedBoundingBox:
-        """Return root-relative bounds computed from the articulation's USD geometry.
+        """Return root-relative bounds of the articulation posed at its configured joint positions.
+
+        Shared and cached across callers, so treat the result as read-only.
```
qianl-nv: since _compute_local_bounding_box_from_usd_at_joint_pos results can stored in lru_caches and modification to it persist, it's best to make a copy of it in the compute_local_bounding_box_from_usd_at_joint_pos wrapper before returning, so we don't rely on the caller to not accidentally modify it.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -56,42 +71,57 @@ def __init__(
...
-            robot = self.scene_config.robot
-            assert robot is not None, "scene_config.robot must be populated before placement"
-            spawn = robot.spawn
-            assert spawn.usd_path is not None, "scene_config.robot must use a USD spawn for placement"
-            scale = tuple(spawn.scale or (1.0, 1.0, 1.0))
-            self._collision_mesh = extract_trimesh_from_usd_path(spawn.usd_path, scale)
-        return self._collision_mesh
+        """Return one mesh containing all posed link boxes."""
+        # Import locally because USD/pxr is available only after simulation initialization.
+        from isaaclab_arena.utils.usd_helpers import extract_trimesh_from_usd_at_joint_pos
```
qianl-nv: ditto about cache copy

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -56,42 +71,57 @@ def __init__(
...
+
+        source = self.get_placement_geometry_source()
+        return extract_trimesh_from_usd_at_joint_pos(source.usd_path, source.joint_pos, source.scale)
+
+    def get_collision_meshes(self) -> tuple[trimesh.Trimesh, ...]:
```
qianl-nv: I suggest we don't return a new API get_collision_meshes. we just overwrite the get_collision_mesh in droid/franka. also can we union the extract_link_bbox_meshes_from_usd_at_joint_pos before returning so it acts as one mesh?

qianl-nv (commented):
For droid/franka let's just overwrite get_collision_mesh instead of creating a new get_collsion_meshes. The return should be one mesh, this should remove a lot of downstream changes. 

Please rebase the MR onto qianl/feature/mesh-optimization
Update the commit message with solver/validation logging before/after this change for the droid kitchen example

Will do another pass after above modification.

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -6,6 +6,7 @@
```
qianl-nv, line 7: please also update the droid pnp kitchen yaml to switch Droid collsion mode to mesh so this feature in used .
default for other envs/yamls should still be bbox mode

zhx06: The lines for comments are different after rebase, but I will still discuss the changes here. 
Addressed. Now `collision_mode: mesh` for `droid_pick_and_place_lightwheel_kitchen.yaml`. Others are in bbox mode.

isaaclab_arena/embodiments/droid/droid.py
zhx06: Sure, it still takes `initial_joint_pose: list[float]`

isaaclab_arena/embodiments/franka/franka.py
zhx06: Same as above

isaaclab_arena/embodiments/droid/droid.py
zhx06: It is restored.

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Done. The public wrapper now constructs a new bounding box, avoiding the chance of modifying it directly.

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Addressed. It now returns `mesh.copy()`

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Done. `get_collision_meshes` is removed. Droid and Franka now override the existing `get_collision_mesh`

isaaclab_arena/assets/asset_cache.py
zhx06: This function was removed.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -23,6 +25,20 @@
     import trimesh
 
 
+@dataclass(frozen=True)
+class PlacementGeometrySource:
```
qianl-nv: ArticulationGeometrySpec or EmbodimentGeometrySpec would be more accurate.

isaaclab_arena/utils/usd_articulation.py
```diff
@@ -0,0 +1,263 @@
+# Copyright (c) 2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Forward kinematics over a USD articulation's physics joints.
+
+Authored USD transforms only describe the one joint configuration an asset happens to be saved in,
```
qianl-nv: Instead of explaining "how" the transform are computed, you need to explain why we need to implement this forward kinematics here (instead of re-using lab/curobo ect)

isaaclab_arena/utils/usd_helpers.py
```diff
@@ -437,3 +466,240 @@ def extract_trimesh_from_usd_path(
     assert stage is not None, f"could not open USD: {usd_path}"
     default_prim = stage.GetDefaultPrim() or stage.GetPseudoRoot()
     return extract_trimesh_from_prim(stage, default_prim.GetPath().pathString, scale)
+
```
qianl-nv, line 469: let's add a code block separation here for better grouping as this file gets long

```
# -----------------------------------------------------------------------------
# Joint-posed articulation geometry helpers
# -----------------------------------------------------------------------------
```

isaaclab_arena/utils/usd_helpers.py
```diff
@@ -437,3 +466,240 @@ def extract_trimesh_from_usd_path(
...
+
+def extract_link_bbox_meshes_from_usd_at_joint_pos(
```
qianl-nv: this wrapper is not used anymore, remove

isaaclab_arena/utils/usd_helpers.py
```diff
@@ -437,3 +466,240 @@ def extract_trimesh_from_usd_path(
...
+        Watertight link-box meshes in the scaled default-prim frame.
+    """
+    meshes = _extract_link_bbox_meshes_from_usd_at_joint_pos(usd_path, tuple(sorted(joint_pos.items())), tuple(scale))
+    return tuple(mesh.copy() for mesh in meshes)
+
+
+# NOTE(zihaox, 2026-07-28): Cache here rather than on the asset. Isaac Lab reaches assets through
+# EventTermCfg params, and configclass's validation walk tracks no visited set, so a trimesh held by
+# an asset sends it recursing through trimesh's internal back-references until the stack overflows.
+@functools.lru_cache(maxsize=_POSED_GEOMETRY_CACHE_SIZE)
```
qianl-nv, line 489: since this is now only called from _extract_trimesh_from_usd_at_joint_pos which is also lru cached, we can remove the cache here.

isaaclab_arena/utils/usd_helpers.py
```diff
@@ -437,3 +466,240 @@ def extract_trimesh_from_usd_path(
...
+    usd_path: str,
+    joint_pos_items: tuple[tuple[str, float], ...],
+    scale: tuple[float, float, float],
+) -> trimesh.Trimesh:
+    """Return a cached mesh containing all of an articulation's link boxes."""
+    meshes = _extract_link_bbox_meshes_from_usd_at_joint_pos(usd_path, joint_pos_items, scale)
+    return trimesh.util.concatenate(meshes)
+
+
+def extract_trimesh_from_usd_at_joint_pos(
```
qianl-nv: let's move the wrapper before _extract_trimesh_from_usd_at_joint_pos for better readibility

isaaclab_arena/utils/usd_articulation.py
```diff
@@ -0,0 +1,263 @@
...
+
+    local_0: np.ndarray
+    """Row-vector 4x4 mapping the joint frame into body0."""
+
+    local_1: np.ndarray
+    """Row-vector 4x4 mapping the joint frame into body1."""
+
+
+def _joint_edges(
+    stage: Usd.Stage,
```
qianl-nv: stage not used

isaaclab_arena/utils/usd_articulation.py
```diff
@@ -0,0 +1,263 @@
...
+    assert not unknown_joints, (
+        f"Joint positions name joints absent under {root_prim_path}: {sorted(unknown_joints)}."
+        f" Available joints: {sorted(joint_prims)}."
+    )
+
+    edges = _joint_edges(stage, root_prim, joint_prims, joint_pos)
+    if not edges:
+        return {}
+
+    rest_transforms = {
```
qianl-nv: can reuse one UsdGeom.XformCache for body transforms to avoid repeatingly construct it in the for loop
can also inline and remove the one-use _local_to_world helper

```
    xform_cache = UsdGeom.XformCache(Usd.TimeCode.Default())
    rest_transforms: dict[str, np.ndarray] = {}
    for path in {path for edge in edges for path in (edge.parent, edge.child) if path}:
        prim = stage.GetPrimAtPath(path)
        assert prim, f"Joint references a missing prim: {path}"
        rest_transforms[path] = np.array(xform_cache.GetLocalToWorldTransform(prim), dtype=np.float64)
```

isaaclab_arena/utils/usd_articulation.py
```diff
@@ -0,0 +1,263 @@
...
+    joint_prims: Mapping[str, Usd.Prim],
+    joint_pos: Mapping[str, float],
+) -> list[_JointEdge]:
+    """Build the articulation's parent-to-child edges, including fixed joints that carry no motion."""
+    root_path = root_prim.GetPath().pathString
+    movable_paths = {prim.GetPath().pathString for prim in joint_prims.values()}
+    values_by_path = {joint_prims[name].GetPath().pathString: value for name, value in joint_pos.items()}
+
+    edges: list[_JointEdge] = []
+    for prim in Usd.PrimRange(root_prim):
```
qianl-nv, line 159: store root_sdf_path = Sdf.Path(root_path)
outside the for loop

qianl-nv (commented):
Some more nits. 
Plz do one more pass in cleaning / simplying the util functions

isaaclab_arena/embodiments/embodiment_base.py
zhx06: Addressed. Renamed to `ArticulationGeometrySpec `

isaaclab_arena/utils/usd_articulation.py
zhx06: Addressed.

isaaclab_arena/utils/usd_helpers.py
zhx06: Addressed. This wrapper has been removed.

isaaclab_arena/utils/usd_helpers.py
zhx06: Addressed. Section separator added.

isaaclab_arena/utils/usd_helpers.py
zhx06: Addressed. The redundant inner cache was removed.

isaaclab_arena/utils/usd_helpers.py
zhx06: Addressed. Moved the public `extract_trimesh_from_usd_at_joint_pos` wrapper before the private cached implementation.

isaaclab_arena/utils/usd_articulation.py
zhx06: This is removed.

isaaclab_arena/utils/usd_articulation.py
zhx06: Addressed. The changes have been applied as suggested.

isaaclab_arena/utils/usd_articulation.py
zhx06: Addressed. he root path is now stored once as `root_sdf_path` and stays out of the for loop.

isaaclab_arena/utils/usd_helpers.py
```diff
@@ -437,3 +466,218 @@ def extract_trimesh_from_usd_path(
...
+# EventTermCfg params, and configclass's validation walk tracks no visited set, so a trimesh held by
+# an asset sends it recursing through trimesh's internal back-references until the stack overflows.
+@functools.lru_cache(maxsize=_POSED_GEOMETRY_CACHE_SIZE)
+def _extract_trimesh_from_usd_at_joint_pos(
+    usd_path: str,
+    joint_pos_items: tuple[tuple[str, float], ...],
+    scale: tuple[float, float, float],
+) -> trimesh.Trimesh:
+    """Return a cached mesh containing all of an articulation's link boxes."""
+    meshes = _extract_link_bbox_meshes_from_usd_at_joint_pos(usd_path, joint_pos_items, scale)
```
qianl-nv: super nits: now that this becomes a one-line wrapper to _extract_link_bbox_meshes_from_usd_at_joint_pos and _extract_link_bbox_meshes_from_usd_at_joint_pos isn't use anywhere else, I'd recommend removing _extract_link_bbox_meshes_from_usd_at_joint_pos and have it inline in _extract_trimesh_from_usd_at_joint_pos.

qianl-nv (approved):
Thanks a lot for addressing all the comments. Looks a lot cleaner now. 
Approved!
one super small nit, see if you can fix it when rebasing to main.

isaaclab_arena/utils/usd_helpers.py
zhx06: Addressed. Changed into inlined method

PR #969, opened 2026-07-29
Fix CI uv test failure from #941

qianl-nv:
## Summary
Fix Streamlit uv lock version and move failing test out from CI

## Detailed description
- `uv.lock` still listed Streamlit `>=1.30` after the pin moved to `>=1.56`, so `uv lock --check` failed
- #941 adds test_gui_review to CI which is failing in uv. Revert the change. A follow up MR will fix the failure and re-enable the test.


alexmillane (approved):
LGTM.

Thanks for fixing this.

I guess this issue crept in because the uv tests are currently not required to merge? I'll fix that.

qianl-nv:
@alexmillane thanks for reviewing!
Yes it creep in since the uv test does not block merging, and I turned on the Enable auto-merge overnight before all the CI results were in.

PR #972, opened 2026-07-29
Enable lightwheel kitchen mesh placement

qianl-nv:
## Summary
Set lightwheel kitchen placement to mesh and fix crashes from placement pool

## Detailed description
- Pass mesh collision mode through PlaceableAsset kwargs; set it to mesh for the droid pick and place yaml to avoid foreground objects on counter top intersecting with existing background scene objects. Keep the default as bbox for other existing environments.
- For the droid pick and place yaml with kitchen in mesh mode, env bringup crashes with two issues: EventTermCfg deepcopying placement pool with Warp ctypes caches ("ctypes objects containing pointers cannot be pickled"); configclass validation on the placement pool recurse into cyclic asset graphs throwing RecurssionError. Fix both by introducing a PlacementPoolHandle the wraps the pool to deepcopy and limits validation depth.

Known issue: With Fabric enabled, live Warp caches on the pool can hide the Droid stand_instanceable in Kit viz while physics stays correct; use --disable_fabric --device cpu for viewport runs (4 env /  2000-step benchmark: 87.0 vs 86.5 ms/step rollout, no meaningful difference). Alternative to fix this problem requires releasing the warp cache before passing into EventTermCfg. This slows down the first refill solve significantly (x2).
 
 - With background bbox mode (collision with backgroudn object)
image

- With background mesh mode + fix + fabric (stand visual disappear)
image

- With background mesh mode + fix + disable fabric
image

### DIFF
--- a/isaaclab_arena/relations/placement_asset.py
+++ b/isaaclab_arena/relations/placement_asset.py
@@ -28,15 +28,19 @@ class PlaceableAsset(Asset, ABC):
     """Asset whose root pose can be constrained by spatial relations."""
 
     def __init__(self, name: str, tags: list[str] | None = None, **kwargs) -> None:
+        collision_mode = kwargs.pop("collision_mode", None)
+        repair_collision_mesh_non_watertight = kwargs.pop("repair_collision_mesh_non_watertight", True)
         super().__init__(name=name, tags=tags, **kwargs)
         self.initial_pose: Pose | PoseRange | PosePerEnv | None = None
         self._pose_event_cfg: EventTermCfg | None = None
         """Reset event restoring this asset's root pose; ``None`` until a pose with a reset event is set."""
         self.relations: list[RelationBase] = []
         # None delegates collision-mode selection to the solver.
-        self.collision_mode: CollisionMode | None = None
+        if collision_mode is not None:
+            collision_mode = CollisionMode(collision_mode)
+        self.collision_mode: CollisionMode | None = collision_mode
         # Whether to replace a non-watertight collision mesh with its convex hull.
-        self.repair_collision_mesh_non_watertight = True
+        self.repair_collision_mesh_non_watertight = repair_collision_mesh_non_watertight
 
     def add_relation(self, relation: RelationBase) -> None:
         """Attach a relation to the asset."""
--- a/isaaclab_arena/relations/placement_asset.py
+++ b/isaaclab_arena/relations/placement_asset.py
@@ -27,9 +27,14 @@ if TYPE_CHECKING:
 class PlaceableAsset(Asset, ABC):
     """Asset whose root pose can be constrained by spatial relations."""
 
-    def __init__(self, name: str, tags: list[str] | None = None, **kwargs) -> None:
-        collision_mode = kwargs.pop("collision_mode", None)
-        repair_collision_mesh_non_watertight = kwargs.pop("repair_collision_mesh_non_watertight", True)
+    def __init__(
+        self,
+        name: str,
+        tags: list[str] | None = None,
+        collision_mode: CollisionMode | str | None = None,
+        repair_collision_mesh_non_watertight: bool = True,
+        **kwargs,
+    ) -> None:
         super().__init__(name=name, tags=tags, **kwargs)
         self.initial_pose: Pose | PoseRange | PosePerEnv | None = None
         self._pose_event_cfg: EventTermCfg | None = None
@@ -38,7 +43,7 @@ class PlaceableAsset(Asset, ABC):
         # None delegates collision-mode selection to the solver.
         if collision_mode is not None:
             collision_mode = CollisionMode(collision_mode)
-        self.collision_mode: CollisionMode | None = collision_mode
+        self.collision_mode = collision_mode
         # Whether to replace a non-watertight collision mesh with its convex hull.
         self.repair_collision_mesh_non_watertight = repair_collision_mesh_non_watertight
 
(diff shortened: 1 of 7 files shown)


isaaclab_arena/relations/placement_events.py
```diff
@@ -26,6 +26,38 @@
 PLACEMENT_RESET_EVENT_NAME = "placement_reset"
 
 
+class PlacementPoolHandle:
+    """Opaque EventTermCfg param holding a runtime placement pool.
+
+    Isaac Lab ``configclass._validate`` recursively walks any object with ``__dict__`` and has no
```
cvolkcvolk: Suggestion to rewrite this docstring in _human_ :) 

It exists because Isaac Lab deep-copies every termination parameter and would then hold a potential outdated copy?

isaaclab_arena/relations/placement_events.py
```diff
@@ -113,8 +145,7 @@ def write_layout_to_sim(
 def solve_and_place_objects(
     env: ManagerBasedEnv,
     env_ids: torch.Tensor | None,
-    assets: list[PlaceableAsset],
-    placement_pool: PooledObjectPlacer,
+    placement_pool: PooledObjectPlacer | PlacementPoolHandle,
```
cvolkcvolk: If we require a `PlacementPoolHandle` for it to savely copy, shall we make it the only accepted type hint?

isaaclab_arena/relations/placement_events.py
```diff
@@ -26,6 +26,38 @@
...
+    def __init__(self, pool: PooledObjectPlacer) -> None:
+        self.pool = pool
+
+    def __deepcopy__(self, memo: dict[int, object]) -> PlacementPoolHandle:
+        """Share the live pool across ``copy.deepcopy`` of EventTermCfg params."""
+        memo[id(self)] = self
+        return self
+
+
+def resolve_placement_pool(value: PooledObjectPlacer | PlacementPoolHandle | None) -> PooledObjectPlacer | None:
```
cvolkcvolk: Same comments below, If we require a PlacementPoolHandle for it to savely copy, shall we make it the only accepted type hint?

isaaclab_arena/relations/placement_events.py
```diff
@@ -26,6 +26,38 @@
...
+    """Return the underlying pool, unwrapping a handle when present."""
+    if value is None:
+        return None
+    if isinstance(value, PlacementPoolHandle):
+        return value.pool
+    return value
```
cvolkcvolk: Something looks off with this function... `get_placement_pool()` below should return None only when the event is absent and if the event exists, it should require a handle and return `handle.pool` and we could remove this function `removes resolve_placement_pool()`?

isaaclab_arena/relations/placement_asset.py
```diff
@@ -28,15 +28,19 @@ class PlaceableAsset(Asset, ABC):
     """Asset whose root pose can be constrained by spatial relations."""
 
     def __init__(self, name: str, tags: list[str] | None = None, **kwargs) -> None:
+        collision_mode = kwargs.pop("collision_mode", None)
```
zhx06: `kwargs.pop()` makes these supported options difficult to discover by users and is rarely used in our repo. Can we declare it explicitly? For example, `collision_mode: CollisionMode | str | None = None` for `__init__`?

isaaclab_arena/relations/placement_events.py
```diff
@@ -26,6 +26,38 @@
 PLACEMENT_RESET_EVENT_NAME = "placement_reset"
 
 
+class PlacementPoolHandle:
```
zhx06, line 29: `PlacementPoolHandle` doesn’t make its purpose very clear. Could we use a more descriptive name, such as `PlacementResetState`?

qianl-nv: Handle is a commonly used program object to holds reference to other resources. It's used for passing references through APIs that cannot accept the underlying object directly, and avoid accidental copying/serialization. This is exactly the purpose we introduce this wrapper here. 
I think the name is fitting for what it is & its purpose. Prefer to keep.

isaaclab_arena/relations/placement_events.py
qianl-nv: It's not for potential outdated copy. 
on main, if we switch any asset to mesh mode instead of bbox, a warp GPU cache (wp.Mesh for BVHs) is created, which causes deepcopy to crash (ctype no pickled). so we need to either release that cache before sim starts (which requires recreating it on the first placement layout refill with extra runtime), or just avoid deepcopying all together.
After that fix there's also another RecurssionError issue with config validation.

qianl-nv: I've updated the docstring to make that more clear.

isaaclab_arena/relations/placement_events.py
qianl-nv: good point, switched all to use PlacementPoolHandle

isaaclab_arena/relations/placement_events.py
qianl-nv: +1. updated to PlacementPoolHandle only

isaaclab_arena/relations/placement_events.py
qianl-nv: Good point. the above two comments enforcing PlacementPoolHandle as input type already reoslves this issue. resolve_placement_pool() is removed and one small assert is added to get_placement_pool() directly.

isaaclab_arena/relations/placement_asset.py
qianl-nv: Great suggestion! move both params to init explicitly.

zhx06 (approved):
LGTM!

cvolkcvolk (approved):
lgtm

PR #973, opened 2026-07-29
Add interactive environment runner

cvolkcvolk:
## Summary
Add an interactive Arena environment runner, launch a registered Arena environment or environment graph spec
[banana_drag_drop_bowl.webm](https://github.com/user-attachments/assets/674b2848-8730-46ed-abbf-07020874f01e)

Enables to
- Apply forces via SHIFT+Left Mouse to objects
- Stop, reposition assets & Start Sim again
- Aligns origins assets in the SimUI

## Usage

Registered environment:

```bash
python isaaclab_arena/scripts/environment_runner.py \
  pick_and_place_maple_table \
  --embodiment droid_rel_joint_pos \
  --pick_up_object banana_ycb_robolab \
  --destination_location bowl_ycb_robolab
```

Environment graph spec:

```bash
python isaaclab_arena/scripts/environment_runner.py \
  --env_graph_spec_yaml isaaclab_arena_environments/robolab/tasks/banana_in_bowl.yaml
```


isaaclab_arena/evaluation/environment_runner.py
```diff
@@ -0,0 +1,220 @@
...
+from isaaclab_arena.cli.isaaclab_arena_cli import get_isaaclab_arena_cli_parser
+from isaaclab_arena.utils.hydra_overrides import assert_hydra_overrides
+from isaaclab_arena.utils.isaaclab_utils.simulation_app import SimulationAppContext
+from isaaclab_arena_environments.cli import get_arena_builder_from_cli, get_isaaclab_arena_environments_cli_parser
+
+if TYPE_CHECKING:
+    import gymnasium as gym
+
+
+class RealTimeRateLimiter:
```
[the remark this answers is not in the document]
cvolkcvolk: Updated

isaaclab_arena/evaluation/environment_runner.py
```diff
@@ -0,0 +1,197 @@
...
+    assert args_cli.num_envs == 1, "environment_runner supports exactly one environment"
+    assert not args_cli.distributed, "environment_runner does not support distributed execution"
+    assert args_cli.presets != "newton", "environment_runner mouse interaction currently requires PhysX"
+    assert not args_cli.list_variations, "environment_runner does not support --list_variations"
+    assert args_cli.device == "cpu", "environment_runner mouse interaction requires CPU PhysX; use --device cpu"
+
+
+def assert_gui_environment() -> None:
+    """Check that environment variables do not force Isaac Sim into headless mode."""
+    assert os.environ.get("HEADLESS", "0") == "0", "environment_runner requires the Kit GUI; unset HEADLESS"
```
cvolkcvolk: I think we can remove this function... This script is never run HEADLESS anyway

isaaclab_arena/evaluation/environment_runner.py
```diff
@@ -0,0 +1,197 @@
...
+    for term_name, term_cfg in vars(terminations_cfg).items():
+        if term_cfg is not None and getattr(term_cfg, "time_out", False):
+            setattr(terminations_cfg, term_name, None)
+            if term_name not in disabled_term_names:
+                disabled_term_names.append(term_name)
+
+    return disabled_term_names
+
+
+def get_idle_actions(env: gym.Env, env_cfg) -> torch.Tensor:
```
cvolkcvolk: Could we just use the zero_action policy runner?

isaaclab_arena/evaluation/environment_runner.py
```diff
@@ -0,0 +1,197 @@
...
+    termination_manager = env.unwrapped.termination_manager
+    fired_terms: dict[str, list[int]] = {}
+    for term_name in termination_manager.active_terms:
+        term_env_ids = termination_manager.get_term(term_name).nonzero(as_tuple=False).flatten().tolist()
+        if term_env_ids:
+            fired_terms[term_name] = term_env_ids
+    return fired_terms
+
+
+def log_fired_termination_terms(env: gym.Env, terminated: torch.Tensor, truncated: torch.Tensor) -> None:
```
cvolkcvolk: We can delete this logging function

isaaclab_arena/evaluation/environment_runner.py
```diff
@@ -0,0 +1,197 @@
...
+def get_idle_actions(env: gym.Env, env_cfg) -> torch.Tensor:
+    """Return the configured idle action or a zero action for the environment."""
+    device = torch.device(env.unwrapped.device)
+    configured_idle_action = getattr(env_cfg, "idle_action", None)
+    if configured_idle_action is not None:
+        return configured_idle_action.repeat(env.unwrapped.num_envs, 1).to(device)
+    return torch.zeros(env.action_space.shape, device=device)
+
+
+def get_fired_termination_terms(env: gym.Env) -> dict[str, list[int]]:
```
cvolkcvolk: If this function is only needed for logging, lets delete it

isaaclab_arena/evaluation/environment_runner.py
```diff
@@ -0,0 +1,197 @@
...
+    app_args, _ = args_parser.parse_known_args()
+    use_kit_visualizer_by_default(app_args)
+    assert_interactive_runner_args(app_args)
+    assert_gui_environment()
+    print("[environment_runner] Using CPU physics for interactive viewport manipulation.", flush=True)
+
+    with SimulationAppContext(app_args) as simulation_app:
+        args_parser = get_isaaclab_arena_environments_cli_parser(args_parser)
+        args_cli, hydra_overrides = args_parser.parse_known_args()
+        use_kit_visualizer_by_default(args_cli)
```
cvolkcvolk: It seems we are duplicating some args_cli again down here. Could we keep them before SimulationAppContext?

cvolkcvolk: Why do we call use_kit_visualizer_by_default and  assert_interactive_runner_args here again?

isaaclab_arena/evaluation/environment_runner.py
cvolkcvolk: Removed in 131086592. The existing `--headless` and Kit visualizer argument validation remains sufficient for this GUI-only runner.

isaaclab_arena/evaluation/environment_runner.py
cvolkcvolk: Not safely for this workflow. `zero_action` always emits literal zeros, which is not an idle command for absolute-position controllers; that is why the Franka moved away from its reset pose during our test. `policy_runner` also requires a finite step/episode limit and creates evaluation outputs. This runner instead uses `env_cfg.idle_action` when available, runs until Kit closes, and applies the CPU PhysX, scene-query, mouse-interaction, timeout, and recorder settings needed for manual inspection. Adding all of that as a special mode to the runner we intend to deprecate would conflate the two workflows.

isaaclab_arena/evaluation/environment_runner.py
cvolkcvolk: Removed in e0275a67a together with the logging function that was its only caller.

isaaclab_arena/evaluation/environment_runner.py
cvolkcvolk: Removed in e0275a67a. The loop now relies on normal automatic reset behavior, which is directly visible during interactive inspection.

isaaclab_arena/evaluation/environment_runner.py
cvolkcvolk: Environment-specific parsing must remain after `SimulationAppContext` because registering those subcommands imports Isaac Sim environment modules. Updated in 9dc2ec0c9 to reuse the original validated `args_cli` namespace, so the app settings remain established before startup and only environment arguments are added afterward.

cvolkcvolk: Removed the duplicate calls in 9dc2ec0c9. The second parse now uses `namespace=args_cli`, preserving the CPU and Kit selections without applying defaults or validation again.

cvolkcvolk: Good point. Simplified in 89aa483f6: the complete environment parser is now built and parsed once before SimulationAppContext, with CPU and Kit expressed as parser defaults. This removes the preliminary parse and Namespace reuse; first-party and graph parser construction does not require a running Kit app.

isaaclab_arena/evaluation/environment_runner.py
```diff
@@ -0,0 +1,220 @@
...
+    for term_name, term_cfg in vars(terminations_cfg).items():
+        if term_cfg is not None and getattr(term_cfg, "time_out", False):
+            setattr(terminations_cfg, term_name, None)
+            if term_name not in disabled_term_names:
+                disabled_term_names.append(term_name)
+
+    return disabled_term_names
+
+
+def get_idle_actions(env: gym.Env, env_cfg) -> torch.Tensor:
```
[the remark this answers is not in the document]
cvolkcvolk: Simplified in 26c3535ff. environment_runner now creates zero actions directly in its loop, so it no longer duplicates the configurable idle-action selection from replay_demos.py.

isaaclab_arena/evaluation/environment_runner.py
cvolkcvolk: Follow-up correction: you were right that zero_action is usable with a zero-neutral controller such as droid_rel_joint_pos. The earlier response overstated the distinction; environment_runner now also uses literal zeros directly. Its remaining purpose is the interactive PhysX mouse setup, real-time pacing, app-lifetime loop, and lack of evaluation outputs.

xyao-nv (approved):

PR #974, opened 2026-07-29
Fix collision Loss Scaling for Large-Object Placement

zhx06:
## Summary
Normalize collision loss for large objects

## Detailed description
- Large robots could overwhelm relation losses and fail placement after repeated attempts.
- Normalize bbox and mesh penetration by object size.
- Align mesh dispatch between optimization and validation.
- Exclude relation anchors from aggregated background meshes.
- Improves `NextTo` placement reliability without increasing retries.

### DIFF
--- a/isaaclab_arena/assets/object_reference.py
+++ b/isaaclab_arena/assets/object_reference.py
@@ -95,6 +95,11 @@ class ObjectReference(ObjectBase):
         quarters = quaternion_to_90_deg_z_quarters(parent_pose.rotation_xyzw)
         return box.rotated_90_around_z(quarters).translated(world_position)
 
+    def get_prim_path_in_parent_usd(self) -> str:
+        """Return the referenced prim's absolute path in its parent USD stage."""
+        with open_stage(self.parent_asset.usd_path) as parent_stage:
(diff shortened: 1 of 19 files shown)


qianl-nv (commented):
This is a relative large change to the relation solver loss compute and carry a lot of risk this late in the stage. Let's take a more systematic approach to understand the perf gain of each optimization and if they introduces any regressions

I see several changes in this MR that should be separately measured
1. AABB loss: Volume overlap (slope=10000) → shortest-axis penetration / subject size (slope=10)
this carries the highest risk for regressing existing scenes. Let's benchmark this change alone, over all the existing robolab scenes (in isaaclab_arena_environments/robolab), and check how (1) layout success rate and interation count changes (2) physics stability rate on the layouts using isaaclab_arena/scripts/run_placement_pool_validation.py 

2. Mean sphere penetration → max normalized penetration per pair
3. Excludes anchor ObjectReference subtrees from aggregated background meshes
4. Droid stand collison_mode=BBOX

For 2-4, let's use the scenes in isaaclab_arena_environment/kitchen_bench with mesh mode enable for the kitchen, and check the impact on build/solver time, layout success rate, interations ect. Let's only merge the changes that has significant impact on perf.

It also seems 4 makes 2 unnecessary at least for kitchen_bench scenes?

PR #981, opened 2026-07-30
Enable lightwheel kitchen mesh placement

qianl-nv:
## Summary
Mesh kitchen placement; bind pool on env runtime

## Detailed description
- Lightwheel kitchen bench uses mesh collision mode through PlaceableAsset kwargs; set it to mesh for the droid pick and place yaml; keep the default as bbox for other existing environments.
- Pass the live placement pool through `env_kwargs` instead of `EventTermCfg` params so mesh placement Warp caches do not hit configclass deepcopy/pickle ("ctypes objects containing pointers cannot be pickled")
- Reset `placement_reset` event uses empty params and reads the  pool from `IsaacLabArenaManagerBasedRLEnv.placement_pool`

Note: With Fabric enabled, Warp caches on the pool can hide the Droid `stand_instanceable` in Kit viz while physics stays correct; use `--disable_fabric` for viewport runs (4 env / 2000-step benchmark: 87.0 vs 86.5 ms/step rollout, no meaningful difference)
 
 
 - With background bbox mode (collision with backgroudn object)
image

- With background mesh mode + fix + fabric (stand visual disappear)
image

- With background mesh mode + fix + disable fabric
image

### DIFF
--- a/isaaclab_arena/assets/background_library.py
+++ b/isaaclab_arena/assets/background_library.py
@@ -173,7 +173,12 @@ class LightwheelKitchenBackground(LibraryBackground):
     initial_pose = Pose.identity()
     object_min_z = -0.2
 
-    def __init__(self, layout_id: int = 1, style_id: int = 1):
+    def __init__(
+        self,
+        layout_id: int = 1,
+        style_id: int = 1,
+        **kwargs,
+    ):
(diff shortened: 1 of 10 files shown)


qianl-nv:
Won't go with this approach. willl use https://github.com/isaac-sim/IsaacLab-Arena/pull/972 instead.

PR #988, opened 2026-07-31
Update kitchen p&p yaml configs to avoid collision

qianl-nv:
- Add paper roll objet reference to avoid collision with foreground objects on spawn
- Increase robot to counter distance


alexmillane (approved):
Nice!

PR #995, opened 2026-07-31
Fix test fail associated with bad merge.

alexmillane:
## Summary
FIx an import error on test startup

## Detailed description
- Causing tests to fail on `main`
- Came from a bad merge between #986 and #973


cvolkcvolk (approved):
Thanks!

PR #997, opened 2026-07-31
Robustify parallel osmo experiment executions

cvolkcvolk:
Keep OSMO collection running after individual experiment failures and show both successful results and failed-run statuses in the combined report.

## Architecture

    runner 0 ----\
    runner 1 -----+--> collector --> aggregated index.html
       ... -------+
    runner N ----/

Previously, every runner was a hard dependency of the collector. If only one runner exited with an error, OSMO marked the collector as `FAILED_UPSTREAM` and never executed it leading to brittle experiment execution on osmo (Known limitation of osmo, not supporting _optional_ task inputs).

Now If Arena exits with an error, the wrapper records the failure and then exits **successfully from the perspective of OSMO**: Each runner invokes the Arena experiment normally. When that process finishes, the runner writes `experiment_runner_result.json` with either `completed` or `failed` plus the original process exit code.

The collector then makes the real decision:

  - Completed run: copy its results and videos.
  - Failed run: preserve its failure status and exit code, but discard partial episode artifacts.
  - Write all (includinding the failed) experiments into the report to visualize unsuccesful runs


osmo/tasks/experiment_runner_task.py
```diff
@@ -70,7 +73,8 @@ def _get_files_to_create(self) -> list[dict[str, Any]]:
         ]
 
     def _get_run_script(self) -> str:
-        command = [
+        """Record the runner result while keeping application failures available to the collector."""
```
alexmillane: Suggestion to change this docstring. It's too specific to this change.

osmo/tasks/experiment_runner_task.py
```diff
@@ -80,5 +84,28 @@ def _get_run_script(self) -> str:
...
+        write_experiment_runner_result_command = (
+            'printf \'{"execution_status":"%s","process_exit_code":%d}\\n\' '
+            '"$experiment_runner_execution_status" "$experiment_runner_process_exit_code" '
+            f"> {experiment_runner_result_path}"
+        )
+        return "\n".join([
+            "set -euo pipefail",
+            "",
+            "# Treat an Arena process failure as a result for the collector, not as an OSMO task failure.",
+            "# Bash allows a command used as an if condition to return nonzero even when set -e is enabled.",
```
alexmillane: Out of interest is there a reason the `set -e` and then use this trick?

If we don't want to fail should we not just not `set -e`?

osmo/scripts/build_experiment_output.py
```diff
@@ -54,25 +81,55 @@ def load_experiment_runner_output_directories_by_run_name(
...
+            experiment_runner_output_directory,
+            run_name,
+        )
+        run_execution_reports.append(
+            RunExecutionReport(
+                run_name=run_name,
+                status=execution_status,
+                process_exit_code=process_exit_code,
+            )
         )
```
alexmillane: Here we call this function, return 2 values which are immediately packed into an object. Is there a reason to not just pack the object inside the function and skip this juggling of values into objects at the caller level?

alexmillane (approved):
LGTM. Couple of tiny nits.

osmo/tasks/experiment_runner_task.py
cvolkcvolk: Updated

osmo/tasks/experiment_runner_task.py
cvolkcvolk: Welp, Codex really thought its required. Turns out it added a test that printf does not fail silently. I think its fair to assume that this won't happen regularly. Removed it for clarity

osmo/scripts/build_experiment_output.py
cvolkcvolk: Updated

PR #1007, opened 2026-08-03
Use a distinct seed per environment rebuild

alexmillane:
## Summary
Distinct env-builder seed per rebuild

## Detailed description
- **Reason:** Every rebuild constructed the environment with the same seed, so build-time variations (e.g. directional-light direction) resolved to the same value on each rebuild, defeating the purpose of rebuilding.
- **Change:** Added `_cfg_for_rebuild`, which offsets `environment_builder.seed` by the rebuild index. Rebuild 0 keeps the configured seed so single-rebuild runs are unchanged, and the original config is never mutated (a fresh copy via `dataclasses.replace`).
- **Impact:** Multi-rebuild runs now see genuinely different build-time variation draws across rebuilds; single-rebuild behaviour is untouched. Covered by a new unit test plus an updated budget-split test.

### DIFF
--- a/isaaclab_arena/evaluation/run_execution.py
+++ b/isaaclab_arena/evaluation/run_execution.py
@@ -103,12 +103,13 @@ def build_and_run(
                 video_base_dir=output_dir,
                 camera_name_prefix=f"robot-cam-rebuild{rebuild_index}",
             )
-            env = _build_environment_from_cfg(cfg, rebuild_video_cfg.render_mode)
+            rebuild_cfg = _cfg_for_rebuild(cfg, rebuild_index)
+            env = _build_environment_from_cfg(rebuild_cfg, rebuild_video_cfg.render_mode)
             results_path = os.path.join(output_dir, f"episode_results_rebuild{rebuild_index}.jsonl")
             env.unwrapped.episode_recorder.set_job_name(cfg.name)
             env.unwrapped.episode_recorder.set_output_path(results_path)
 
-            policy = _build_policy_from_cfg(cfg)
+            policy = _build_policy_from_cfg(rebuild_cfg)
             num_steps, num_episodes = _resolve_rollout_limit(
                 cfg,
                 policy,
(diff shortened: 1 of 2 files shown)


isaaclab_arena/tests/test_run_execution.py
```diff
@@ -82,11 +82,27 @@ def record_rollout(env, policy, num_steps, num_episodes):
         output_dir=tmp_path,
     )
 
+    base_seed = run.environment_builder.seed
     assert result.run_name == "test_run"
     assert result.status is RunStatus.COMPLETED
     assert rollout_limits == [(None, 3), (None, 2)]
-    assert received_run_cfgs == [run, run]
+    # Rebuild 0 builds the run unchanged; each later rebuild offsets the seed so its fresh
+    # construction differs (e.g. a build-time light-direction variation resolves to a new value).
```
alexmillane: Shorten to. Runs are the same except for their seeds.

isaaclab_arena/tests/test_run_execution.py
```diff
@@ -82,11 +82,27 @@ def record_rollout(env, policy, num_steps, num_episodes):
...
+    assert received_run_cfgs[0] is run
+    assert [cfg.environment_builder.seed for cfg in received_run_cfgs] == [base_seed, base_seed + 1]
     assert run.rollout_limit == RolloutLimitCfg(num_episodes=5)
+    assert run.environment_builder.seed == base_seed
```
alexmillane: This doesn't actually test that the second run is the same _except for the seed_. Consider copying run, altering the seed and testing:

```
assert received_run_cfgs == [run_seed_1, run_seed_2]
```

alexmillane (commented):
Self review.

isaaclab_arena/evaluation/run_execution.py
```diff
@@ -128,6 +130,18 @@ def build_and_run(
     )
 
 
+def _seed_cfg_for_rebuild(cfg: ArenaRunCfg, rebuild_index: int) -> ArenaRunCfg:
+    """Offset the environment-builder seed for a rebuild so each fresh construction differs.
+
+    Without this every rebuild constructs with the same seed, so build-time variations
+    (e.g. the directional-light direction) resolve to the same value on each rebuild.
+    Rebuild 0 keeps the configured seed (``seed + 0``) so single-rebuild runs are unchanged.
```
alexmillane: Remove the details here.

alexmillane (commented):
self review 2

cvolkcvolk (approved):
lgtm

PR #1008, opened 2026-08-03
Mrossi/feature/trajectory extraction

rossimattia:
## Summary
Record per-step trajectories in eval runs
## Detailed description
- Exported datasets carried only the metric terms, so the end-effector and
  gripper trajectories needed for downstream training and analysis could not be
  reconstructed from an evaluation run.
- Adds `--record_trajectories` to `experiment_runner`, which extends each Run's
  recorder config with scene states, pre- and post-step actions, episode
  identity, end-effector poses and velocities, and gripper state, and gives
  each rebuild its own dataset file. Camera observations are not included.
- End-effector linear and angular velocity are taken from the parent link
  (`v = v_link + ω × r`) in the same env-aligned frame as scene root pose /
  velocity, and are checked in tests against finite differences of position.
- Fixes the Droid `ee_frame`, which targeted `panda_link0` and carried Franka
  offsets, so recorded end-effector and fingertip poses tracked where the robot
  is mounted rather than where its hand is.
- Off by default, and existing metric terms are preserved, because metrics are
  derived by reading them back out of this same dataset.
- Adds tests for the merged recorder config, per-rebuild dataset naming, and an
  end-to-end rollout that verifies the terms reach the exported HDF5.


isaaclab_arena/evaluation/run_execution.py
```diff
@@ -131,12 +142,27 @@ def build_and_run(
...
+    """
     arena_builder = build_arena_builder_from_run_cfg(cfg)
     _, env_cfg, env_kwargs = arena_builder.build_registered()
     if env_cfg.recorders is not None:
-        env_cfg.recorders.dataset_filename = f"dataset_{cfg.name}"
+        if record_trajectories:
+            env_cfg.recorders = with_trajectory_recorder_terms(env_cfg.recorders)
+            env_cfg.recorders.dataset_export_dir_path = str(output_dir)
+        # Each rebuild recreates the dataset file, so the index keeps earlier rebuilds intact.
+        env_cfg.recorders.dataset_filename = f"dataset_{cfg.name}_rebuild{rebuild_index}"
```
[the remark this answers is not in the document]
rossimattia: Renaming was intended.

isaaclab_arena/utils/isaaclab_utils/recorders.py
```diff
@@ -50,3 +69,186 @@ class ArenaEnvRecorderManagerCfg(ActionStateRecorderManagerCfg):
...
+            if not isinstance(term, BinaryJointAction):
+                continue
+            # BinaryJointAction exposes no public accessor for its joints or its commands.
+            asset: Articulation = term._asset  # noqa: SLF001
+            joint_ids = term._joint_ids  # noqa: SLF001
+            open_command = term._open_command  # noqa: SLF001
+            close_command = term._close_command  # noqa: SLF001
+
+            travel = close_command - open_command
+            has_travel = travel != 0
```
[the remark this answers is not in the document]
rossimattia: Change applied.

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -277,30 +277,44 @@ class DroidSceneCfg:
...
     )
 
-    # The end-effector frame marker
+    # The end-effector frame marker.
+    #
+    # Every offset below is a constant measured from the flattened robot asset, which zeroes each gripper
+    # link's origin on the gripper base instead of on the geometry the link is named after. The links
+    # still translate with the joint, so a constant offset reaches the same point at any aperture.
+    # Re-exporting the robot unflattened would move those origins onto their real locations, and all
+    # three offsets would need re-measuring.
```
alexmillane: Remove comment. 

The comment below covers the measurement process.

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -277,30 +277,44 @@ class DroidSceneCfg:
...
+                prim_path="{ENV_REGEX_NS}/Robot/Gripper/Robotiq_2F_85/base_link",
                 name="end_effector",
                 offset=OffsetCfg(
-                    pos=[0.0, 0.0, 0.1034],
+                    pos=(0.131, 0.0, 0.0),
                 ),
             ),
+            # Each finger is offset from its link to the centre of its pad's gripping face, measured
+            # from the pad meshes. The two frames therefore meet on the centreline at full closure,
+            # and their separation is the gripper's aperture in metres.
```
alexmillane: This comment describes that the closure point is on the centerline, which is somewhat obvious. Should we describe (briefly) how you measured the critical measurement which is the ~13cm?

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -277,30 +277,44 @@ class DroidSceneCfg:
...
     ee_frame: FrameTransformerCfg = FrameTransformerCfg(
         prim_path="{ENV_REGEX_NS}/Robot/panda_link0",
         debug_vis=False,
         target_frames=[
             FrameTransformerCfg.FrameCfg(
-                prim_path="{ENV_REGEX_NS}/Robot/panda_link0",
+                # The Robotiq gripper base link, offset along +x (its approach axis) to the grasp point
+                # midway between the finger pads, so recorded poses can be compared against object poses.
+                # This sits ahead of the body ``ee_pos``/``ee_quat`` report. Being rigid to the base, it
+                # marks the grasp point at the open aperture; the pads travel ~14 mm further out as the
```
alexmillane: What is "open aperture"?

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -277,30 +277,44 @@ class DroidSceneCfg:
...
+                # linkage closes, so use the fingertip midpoint where that difference matters.
```
alexmillane: Suggestion to add a breif commend about how we measured this.

isaaclab_arena/embodiments/droid/droid.py
```diff
@@ -277,30 +277,44 @@ class DroidSceneCfg:
...
             ),
+            # Each finger is offset from its link to the centre of its pad's gripping face, measured
+            # from the pad meshes. The two frames therefore meet on the centreline at full closure,
+            # and their separation is the gripper's aperture in metres.
             FrameTransformerCfg.FrameCfg(
                 prim_path="{ENV_REGEX_NS}/Robot/Gripper/Robotiq_2F_85/right_inner_finger",
                 name="tool_rightfinger",
                 offset=OffsetCfg(
-                    pos=(0.0, 0.0, 0.046),
+                    pos=(0.131, -0.0414, 0.0),
```
alexmillane, line 353: This doesn't look quite right to me. The x value matches the above, which is referenced to the gripper base_link, while this is referenced to the finger.

isaaclab_arena/evaluation/run_execution.py
alexmillane: This is being added in the wrong place. We should centralize all env_cfg composition in the `ArenaEnvBuilder`. There is a section in the function `compose_manager_cfg` where this is some.

Suggestion to move this code there.

isaaclab_arena/evaluation/run_execution.py
```diff
@@ -23,6 +23,7 @@
 from isaaclab_arena.evaluation.policy_runner import rollout_policy
 from isaaclab_arena.evaluation.resource_cleanup import close_run_resources
 from isaaclab_arena.metrics.aggregate_metrics import aggregate_metrics
+from isaaclab_arena.utils.isaaclab_utils.recorders import with_trajectory_recorder_terms
```
alexmillane: `with_trajectory_recorder_terms` is a bit of weird name. Generally methods should have verb names like `get_` or in this case `add_`

alexmillane (commented):
Partial review.

isaaclab_arena/embodiments/droid/droid.py
rossimattia: Fixed.

isaaclab_arena/embodiments/droid/droid.py
rossimattia: Bad naming: the meaning was "gripper aperture status is open". Comments have been improved.

isaaclab_arena/embodiments/droid/droid.py
rossimattia: Comments have been improved.

isaaclab_arena/embodiments/droid/droid.py
rossimattia: Done as part of the comment improvements.

isaaclab_arena/embodiments/droid/droid.py
rossimattia: Clarified in the updated comments. In the DROID "flattened" USD file, the left_inner_finger / right_inner_finger origins sit at base_link when the gripper is open (the geometric offsets are baked into the meshes vertices rather than the link frames). So the shared +x = 0.131 is in the same frame as the EE offset; the ±y values then place each frame on its pad. That’s also why we measured from the meshes/bounding boxes instead of reading the offsets from a finger-local origin.

isaaclab_arena/evaluation/run_execution.py
rossimattia: I tried to fix it. Please let me know if it looks better now.

isaaclab_arena/evaluation/run_execution.py
rossimattia: Changed the function name to ```add_trajectory_recorder_terms```.

isaaclab_arena/environments/arena_env_builder_cfg.py
```diff
@@ -26,6 +26,12 @@ class ArenaEnvBuilderCfg:
     presets: str | None = None
     device: str = "cuda:0"
     language_instruction: str | None = None
+    record_trajectories: bool = False
+    """Whether to extend the recorder manager with per-step trajectory terms."""
+    recorder_dataset_export_dir_path: str | None = None
+    """If set, overrides the recorder manager's dataset export directory."""
+    recorder_dataset_filename: str | None = None
+    """If set, overrides the recorder manager's dataset filename after the default timestamped name."""
```
alexmillane, line 34: Do we need both of these variables? Are there situations where we'd like to specify one and not the other. If not, could we reduce to one variable?

isaaclab_arena/evaluation/run_execution.py
```diff
@@ -140,13 +150,32 @@ def _seed_cfg_for_rebuild(cfg: ArenaRunCfg, rebuild_index: int) -> ArenaRunCfg:
...
+    # ArenaEnvBuilder.compose_manager_cfg sees them without mutating the caller's run config.
+    run_cfg = replace(
+        cfg,
+        environment_builder=replace(
+            cfg.environment_builder,
+            record_trajectories=record_trajectories,
+            recorder_dataset_export_dir_path=str(output_dir) if record_trajectories else None,
+            recorder_dataset_filename=f"dataset_{cfg.name}_rebuild{rebuild_index}",
+        ),
+    )
```
alexmillane, line 176: Suggestion to make this a deep copy of `cfg` and then just overwrite the relevant entries.

isaaclab_arena/utils/isaaclab_utils/recorders.py
alexmillane, line 1: suggestion to move to `isaaclab_arena/terms`

isaaclab_arena/utils/isaaclab_utils/recorders.py
```diff
@@ -50,3 +69,244 @@ class ArenaEnvRecorderManagerCfg(ActionStateRecorderManagerCfg):
...
+
+
+@configclass
+class EndEffectorPosesRecorderCfg(RecorderTermCfg):
+    """Configuration for the end-effector pose recorder term."""
+
+    class_type: type[RecorderTerm] = EndEffectorPosesRecorder
+
+    sensor_name: str = "ee_frame"
+    """Name of the scene's end-effector frame transformer; the term no-ops when it is absent."""
```
alexmillane, line 211: Suggestion `sensor_name` -> `frame_transformer_name`

isaaclab_arena/utils/isaaclab_utils/recorders.py
```diff
@@ -50,3 +69,244 @@ class ArenaEnvRecorderManagerCfg(ActionStateRecorderManagerCfg):
...
+                }
+        return frames
+
+    def record_post_reset(self, env_ids: Sequence[int] | None):
+        poses = self._end_effector_poses(env_ids)
+        return (None, None) if poses is None else ("initial_state/kinematics", poses)
+
+    def record_post_step(self):
+        poses = self._end_effector_poses(None)
+        return (None, None) if poses is None else ("states/kinematics", poses)
```
alexmillane, line 201: Am I correct that we want both these terms because if we were to just record post_step, we'd miss the first pose? Suggesting to put that (or some other reason) in a comment.

isaaclab_arena/utils/isaaclab_utils/recorders.py
```diff
@@ -50,3 +69,244 @@ class ArenaEnvRecorderManagerCfg(ActionStateRecorderManagerCfg):
...
+                    "angular_velocity": angular_velocity,
+                }
+        return frames
+
+    def record_post_reset(self, env_ids: Sequence[int] | None):
+        poses = self._end_effector_poses(env_ids)
+        return (None, None) if poses is None else ("initial_state/kinematics", poses)
+
+    def record_post_step(self):
+        poses = self._end_effector_poses(None)
```
alexmillane, line 200: nit: -> `self._end_effector_poses(env_ids=None)`

isaaclab_arena/utils/isaaclab_utils/recorders.py
```diff
@@ -50,3 +69,244 @@ class ArenaEnvRecorderManagerCfg(ActionStateRecorderManagerCfg):
...
+            travel = close_command - open_command
+            assert (travel != 0).all(), "Binary gripper open and close commands must be distinct"
+            joint_positions = wp.to_torch(asset.data.joint_pos)[:, joint_ids]
+            opening = (joint_positions - open_command) / travel
+            gripper_state = {"position": opening[env_ids] if env_ids is not None else opening}
+
+            if include_command:
+                # process_actions() assigns the command tensors verbatim, so this comparison is exact.
+                is_open = (term.processed_actions == open_command).all(dim=-1)
+                gripper_state["is_open"] = is_open[env_ids] if env_ids is not None else is_open
```
alexmillane, line 258: `is_open` seems a bit of a bit of a confusing name. `commanded_openness`?

isaaclab_arena/utils/isaaclab_utils/recorders.py
```diff
@@ -50,3 +69,244 @@ class ArenaEnvRecorderManagerCfg(ActionStateRecorderManagerCfg):
...
+    class_type: type[RecorderTerm] = GripperStateRecorder
+
+
+@configclass
+class TrajectoryRecorderTermsCfg:
+    """Recorder terms capturing per-step robot and object trajectories.
+
+    Camera observations are deliberately excluded: at Arena's default resolutions they outweigh the
+    state and action terms by three orders of magnitude. Use the imitation-learning scripts when
+    image data is wanted.
```
alexmillane, line 286: Remove this comment. It's totally fine that the "trajectory" recorder doesn't include images. It would be weird if it did actually.

isaaclab_arena/utils/isaaclab_utils/recorders.py
```diff
@@ -50,3 +69,244 @@ class ArenaEnvRecorderManagerCfg(ActionStateRecorderManagerCfg):
...
+    The metric terms already on ``recorder_cfg`` are preserved, because metrics are computed by
+    reading their terms back out of the same exported dataset.
+
+    Args:
+        recorder_cfg: The composed recorder manager config to extend.
+    """
+    return combine_configclass_instances(
+        "TrajectoryRecorderManagerCfg",
+        recorder_cfg,
+        TrajectoryRecorderTermsCfg(),
```
alexmillane, line 310: Here we use the default frame names unconditionally. Suggestion to move creation of the ee-recorder terms into the embodiments, such that we can create them with the correct frame names and robot names.

See comment above.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -300,7 +300,14 @@ def compose_manager_cfg(self) -> tuple[IsaacLabArenaManagerBasedRLEnvCfg, dict[s
             progress_tracking_recorder_cfg,
             bases=(RecorderManagerBaseCfg,),
         )
+        if self.cfg.record_trajectories:
+            recorder_manager_cfg = add_trajectory_recorder_terms(recorder_manager_cfg)
```
alexmillane, line 304: Suggestion to move creation of the trajectory recorder terms (which are related to the robot/embodiment) into `embodiment.get_recorder_term_cfg()`

alexmillane (commented):
Looking clean!

I have a bunch of small nits. My main comment is that the recorder terms should be moved into the emodiments.

We also have some bi-manual embodiments in the framework. Could you ensure that this works for them too?

isaaclab_arena/environments/arena_env_builder_cfg.py
rossimattia: Yes, they server different purposes.

isaaclab_arena/terms/recorders.py
rossimattia: Done.

isaaclab_arena/evaluation/run_execution.py
rossimattia: Done.

isaaclab_arena/utils/isaaclab_utils/recorders.py
rossimattia: Done.

isaaclab_arena/terms/recorders.py
```diff
@@ -50,3 +69,244 @@ class ArenaEnvRecorderManagerCfg(ActionStateRecorderManagerCfg):
...
+                }
+        return frames
+
+    def record_post_reset(self, env_ids: Sequence[int] | None):
+        poses = self._end_effector_poses(env_ids)
+        return (None, None) if poses is None else ("initial_state/kinematics", poses)
+
+    def record_post_step(self):
+        poses = self._end_effector_poses(None)
+        return (None, None) if poses is None else ("states/kinematics", poses)
```
rossimattia: Yes. Added clarifying comment.

isaaclab_arena/utils/isaaclab_utils/recorders.py
rossimattia: Done.

isaaclab_arena/utils/isaaclab_utils/recorders.py
rossimattia: Changed to `is_commanded_open` as it is a bool.

isaaclab_arena/utils/isaaclab_utils/recorders.py
rossimattia: Done.

isaaclab_arena/environments/arena_env_builder.py
rossimattia: Done.

isaaclab_arena/terms/recorders.py
```diff
@@ -50,3 +69,244 @@ class ArenaEnvRecorderManagerCfg(ActionStateRecorderManagerCfg):
...
+    The metric terms already on ``recorder_cfg`` are preserved, because metrics are computed by
+    reading their terms back out of the same exported dataset.
+
+    Args:
+        recorder_cfg: The composed recorder manager config to extend.
+    """
+    return combine_configclass_instances(
+        "TrajectoryRecorderManagerCfg",
+        recorder_cfg,
+        TrajectoryRecorderTermsCfg(),
```
rossimattia: Done.

rossimattia:
@alexmillane I've implemented all the requests.

> We also have some bi-manual embodiments in the framework. Could you ensure that this works for them too?

Only G1 does not support the current trajectory extraction, but it does not result in a crash, the recorders just do nothing. Supporting G1 would go beyond the scope of this PR. I'll consider making a future extension.

isaaclab_arena/embodiments/embodiment_base.py
```diff
@@ -248,8 +247,20 @@ def _update_scene_cfg_with_robot_initial_pose(self, scene_config: Any, pose: Pos
...
+            record_trajectories: Whether to also include the per-step trajectory recorder terms,
+                built with this embodiment's own frame transformer and scene key.
+        """
+        if not record_trajectories:
+            return None
+        from isaaclab_arena.terms.recorders import make_trajectory_recorder_terms_cfg
+
+        return make_trajectory_recorder_terms_cfg(
+            frame_transformer_name=self.get_ee_frame_transformer_name(), asset_name=self.get_scene_key()
+        )
```
alexmillane, line 263: From the naming, this seems specific to embodiments with a single end-effector.

We want the framework to support different, particularly bi-manual embodiments.

Could we update this such that `get_ee_frame_transformer_name() -> str` -> get_ee_frame_transformer_names() -> list[str]`

isaaclab_arena/evaluation/run_execution.py
```diff
@@ -140,13 +150,27 @@ def _seed_cfg_for_rebuild(cfg: ArenaRunCfg, rebuild_index: int) -> ArenaRunCfg:
...
+        record_trajectories: Whether to also record per-step robot and object trajectories.
+    """
+    # Stamp eval-only recorder options onto a copy of the run's builder cfg so composition in
+    # ArenaEnvBuilder.compose_manager_cfg sees them without mutating the caller's run config.
+    run_cfg = deepcopy(cfg)
+    run_cfg.environment_builder.record_trajectories = record_trajectories
+    run_cfg.environment_builder.recorder_dataset_export_dir_path = str(output_dir) if record_trajectories else None
+    run_cfg.environment_builder.recorder_dataset_filename = f"dataset_{cfg.name}_rebuild{rebuild_index}"
+    arena_builder = build_arena_builder_from_run_cfg(run_cfg)
+    return arena_builder.make_registered(render_mode=render_mode)
```
alexmillane, line 168: This looks a little strange to me.

Why do we need to override these variables in the `ArenaRunCfg`? Can't we just use the `ArenaRunCfg` variables directly?

@cvolkcvolk Could you have a look at this?

isaaclab_arena/terms/recorders.py
```diff
@@ -0,0 +1,307 @@
...
+
+    Args:
+        frame_transformer_name: Name of the scene's end-effector frame transformer sensor.
+        asset_name: Scene entity name of the articulation that owns the end-effector frames.
+    """
+    return TrajectoryRecorderTermsCfg(
+        record_end_effector_poses=EndEffectorPosesRecorderCfg(
+            frame_transformer_name=frame_transformer_name, asset_name=asset_name
+        )
+    )
```
alexmillane, line 329: See comment above. I feel that we should extend this to support multiple frames.

isaaclab_arena/tests/test_recorders.py
```diff
@@ -0,0 +1,30 @@
...
+    "record_end_effector_poses",
+    "record_gripper_state",
+)
+
+
+def test_trajectory_terms_cfg_has_all_terms():
+    terms_cfg = make_trajectory_recorder_terms_cfg()
+
+    for term_name in TRAJECTORY_TERM_NAMES:
+        assert getattr(terms_cfg, term_name, None) is not None, f"missing trajectory term {term_name}"
```
alexmillane, line 22: Suggestion to remove this test. It doesn't really test anything.

isaaclab_arena/tests/test_recorders.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_end_effector_poses_term_uses_the_given_names():
+    terms_cfg = make_trajectory_recorder_terms_cfg(frame_transformer_name="left_ee_frame", asset_name="left_arm")
+
+    assert terms_cfg.record_end_effector_poses.frame_transformer_name == "left_ee_frame"
+    assert terms_cfg.record_end_effector_poses.asset_name == "left_arm"
```
alexmillane: Suggestion to extend this test when we can pass multiple frames.

isaaclab_arena/tests/test_recorders.py
alexmillane: The test may make sense when we have multiple frame names, to ensure that all the terms get added for all the frames.

alexmillane (commented):
Looks great!

Thanks for addressing the comments from the last review.

I only have one substantial comment: we should extend support to embodiments with multiple end-effectors.

isaaclab_arena/embodiments/embodiment_base.py
rossimattia: Done, as it makes things clearer. However, multiple arms could be handled with just one single ee_frame, as this can be used to track multiple end-effectors.

isaaclab_arena/terms/recorders.py
rossimattia: Done, see comment above.

isaaclab_arena/evaluation/run_execution.py
rossimattia: `--record_trajectories` and the export directory are session/CLI settings I did not want to add in the job config, and the dataset filename depends on rebuild_index, which is only known at runtime.

isaaclab_arena/tests/test_recorders.py
rossimattia: Tests updated considering the extension to multiple end-effectors (see above).

isaaclab_arena/tests/test_recorders.py
rossimattia: Tests updated considering the extension to multiple end-effectors (see above).

alexmillane (approved):

isaaclab_arena/evaluation/run_execution.py
rossimattia: Fixed by @cvolkcvolk

PR #1039, opened 2026-08-06
Avoid Warp CUDA initialization before Fabric startup

qianl-nv:
## Summary
Avoid Warp CUDA initialization before Fabric startup

## Detailed description
- Initial GPU MESH placement could corrupt PhysX Fabric rendering, causing background geometry and robot parts to disappear.
- Run the initial pooled MESH solve on CPU, while retaining CUDA for runtime pool refills after Fabric initializes.
- Add explicit solver-device propagation and regression coverage.
- Confirmed correct rendering with Fabric GPU interop enabled.

## Benchmark

Replicator kitchen, synchronized `RelationSolver.solve` wall time, batch size 50 (num_envs=1).
isaaclab_arena_environments/kitchen_bench/replicator_kitchen_peninsula_mustard_bowl.yaml"
Before mode used CUDA for every solve; after mode used CPU initially and CUDA for refills.

- Initial: 3.107 s → 8.278 s
- Refill 1: 2.788 s → 3.017 s
- Refill 2: 2.743 s → 2.845 s
- Refill 3: 2.780 s → 2.843 s

The fix adds approximately 5.17 seconds at startup. Runtime refills remain close to the previous CUDA performance.

For batch size 200 (num_envs=4), CPU initial uses ~12s while GPU unchanges. Refills are similar to batch=50.

### DIFF
--- a/isaaclab_arena/relations/object_placer.py
+++ b/isaaclab_arena/relations/object_placer.py
@@ -87,6 +87,7 @@ class ObjectPlacer:
         objects: list[PlaceableAsset],
         num_envs: int = 1,
         collision_objects: list[CollisionObject] | None = None,
+        device: torch.device | str | None = None,
     ) -> list[PlacementResult]:
         """Place objects according to their spatial relations.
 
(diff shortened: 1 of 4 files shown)


qianl-nv:
Found a better solution, will not need this one

PR #1040, opened 2026-08-06
Share defaults across Experiment Runs

cvolkcvolk:
Let an Experiment YAML define a `shared` block once and use it as the base configuration for every Run. 
Common values such as variations, rollout limits, or environment settings no longer need to be repeated in each Run.

The overwrite precedence, from lowest to highest, is:

1. Defaults in the YAML `shared` block. _[lowest]_
2. `shared.<path>=<value>` command-line changes to that shared base.
3. Values written directly in an individual Run in the YAML.
4. Direct `runs.<name>.<path>=<value>` command-line overrides. _[highest]_

- Update YAML Experiments, `robolab_18_tasks_pi0_and_cosmos.yaml`, `robolab_20_tasks_pi0_and_cosmos.yaml` and `getting_started_experiment.yaml`

This also avoids creating a duplicate Experiment YAML just to apply the same change everywhere. 
So for example the `robolab_20_tasks_pi0_and_cosmos.yaml` would not need to be duplicated anymore just to enable variations. Just override the default: 

```
python isaaclab_arena/evaluation/experiment_runner.py \
    --experiment_config isaaclab_arena_environments/robolab/experiment_configs/robolab_20_tasks_pi0_and_cosmos.yaml \
    shared.variations.droid_abs_joint_pos.camera_extrinsics_wrist_camera.enabled=true
```


docs/pages/concepts/variations/variations.rst
```diff
@@ -141,6 +141,62 @@ Run it with:
      --viz kit \
      --experiment_config isaaclab_arena_environments/experiment_configs/droid_pnp_variations_experiment.yaml
 
+Sharing variation values across runs
```
cvolkcvolk: Lets delete this doc change. The shared values are not necessarly bound to variation.

isaaclab_arena/evaluation/experiment_runner_cli.py
```diff
@@ -25,7 +25,7 @@ def add_experiment_runner_arguments(parser: argparse.ArgumentParser) -> None:
         default=_DEFAULT_EXPERIMENT_CONFIG_PATH,
         help=(
             "Path to a typed YAML Experiment or legacy JSON evaluation config. "
-            "For YAML, append Hydra KEY=VALUE overrides for fields on declared Runs."
+            "For YAML, append Hydra KEY=VALUE overrides for shared values or fields on declared Runs."
```
cvolkcvolk: Could we be more specific here?

xyao-nv (approved):

isaaclab_arena/hydra/typed_experiment_loader.py
```diff
@@ -102,17 +107,49 @@ def _assert_run_name_is_hydra_compatible(run_name: str) -> None:
...
+    """
+    experiment_override_prefix = f"{experiment_config_prefix}." if experiment_config_prefix else ""
+    shared_override_prefix = f"{experiment_override_prefix}shared."
+    shared_default_overrides: list[str] = []
+    remaining_overrides: list[str] = []
+    for override in overrides:
+        if override.startswith(shared_override_prefix):
+            shared_default_overrides.append(override.removeprefix(experiment_override_prefix))
+        else:
+            remaining_overrides.append(override)
```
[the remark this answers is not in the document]
cvolkcvolk: Adressed

alexmillane (approved):
LGTM

PR #1050, opened 2026-08-06
Support external & internal inference endpoint for agentic env gen

xyao-nv:
## Summary
Support external & internal inference endpoint

## Detailed description
- Add both internal and public endpoints and mechanism to switch. Tape CI to use public endpoint with my personal key.
- Logged invalid spec to disk with a prefix "invalid_" and log on terminals how it's invalid against  the catalogue.
- Added doc to inform how to choose model/endpoint and Arena is not in charge of the model quality.
- CI still using internal endpoint. Support to enable current sampling workflows with the public endpoint will come a follow up PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,8 +105,9 @@ jobs:
     timeout-minutes: 80
     needs: [pre_commit]
     env:
-      # NV_API_KEY is consumed by the live-endpoint tests for agentic env gen.
-      NV_API_KEY: ${{ secrets.ARENA_NV_API_KEY }}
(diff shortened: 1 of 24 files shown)


qianl-nv (approved):
Given we are short on time I split your PR into two parts

- This current one keeps the dual endpoint switch, invalid spec logging and doc updates regarding env setup/model selection. CI still uses internal endpoint since public is failing at this point.
- All the remaining update to spec inference and prim path inference prompt to work with the public endpoint are moved to https://github.com/isaac-sim/IsaacLab-Arena/pull/1057

I'll merge part 1 first so we can iterate in parallel on part 2

PR #1059, opened 2026-08-07
Update kitchen bench yamls to make tasks more doable

qianl-nv:
## Summary
Update kitchen bench yamls to make tasks more doable

## Detailed description
- Removed unused paper roll from LW pick & place now that we use background mesh mode
- Adjust stand_height_m for open door tasks to make it easier
- Update task description for replicator to disambiguite the placement target red bowl from existing background bowls


zhx06 (approved):
LGTM. I also updated 3 gifs

PR #1062, opened 2026-08-07
Add OpenPI server port override

TuurStuyck:
## Summary
Add OpenPI server port override

## Detailed description
- Add a -p option and pass the selected port to serve_policy.py.
- Keep OpenPI's native bind/argument errors instead of adding wrapper preflight checks.
- Document matching non-default server ports with Arena remote_port.
- Merge this to main first, then cherry-pick the landed commit to release/0.3.0-prerelease.

### DIFF
--- a/isaaclab_arena_openpi/docker/run_openpi_server.sh
+++ b/isaaclab_arena_openpi/docker/run_openpi_server.sh
@@ -48,55 +48,6 @@ while getopts ":rp:v:h" opt; do
     esac
 done
 
-validate_port() {
-    local port_number
-
-    if [[ ! "$PORT" =~ ^[0-9]+$ ]]; then
-        echo "invalid -p port: $PORT (expected an integer from 1 to 65535)" >&2
-        exit 1
-    fi
-    port_number=$((10#$PORT))
-    if (( port_number < 1 || port_number > 65535 )); then
-        echo "invalid -p port: $PORT (expected an integer from 1 to 65535)" >&2
-        exit 1
-    fi
-    PORT="$port_number"
-}
-
-assert_port_available() {
-    if ! command -v python3 >/dev/null 2>&1; then
-        echo "python3 is required to check whether port ${PORT} is available" >&2
-        exit 1
-    fi
-
-    if ! python3 - "$PORT" <<'PY'
-import socket
-import sys
-
-port = int(sys.argv[1])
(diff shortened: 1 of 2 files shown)


TuurStuyck:
Fix for second issue reported here https://nvbugspro.nvidia.com/bug/6491232

isaaclab_arena_openpi/docker/run_openpi_server.sh
```diff
@@ -29,21 +31,72 @@ Usage:
...
         r) FORCE_REBUILD=true ;;
+        p) PORT="$OPTARG" ;;
         v) VARIANT="$OPTARG" ;;
         h) print_help; exit 0 ;;
         \?) echo "unknown option: -$OPTARG" >&2; print_help; exit 1 ;;
         :) echo "option -$OPTARG requires an argument" >&2; exit 1 ;;
     esac
 done
 
+validate_port() {
```
cvolkcvolk: Do we think we need this? 
If an existing port is blocked, as reported in [NVbug 6491232](https://nvbugspro.nvidia.com/bug/6491232#:~:text=OSError%3A%20%5BErrno%2098%5D%20error%20while%20attempting%20to%20bind%20on%20address%20(%270.0.0.0%27%2C%208000)%3A%20address%20already%20in%20use),
the OSError seems to be already similarly descriptive: 

`OSError: [Errno 98] error while attempting to bind on address ('0.0.0.0', 8000): address already in use`

cvolkcvolk: I would personally prefer to not add it though as we _should_ have the same checks then for all other servers as well

cvolkcvolk (approved):
Thanks for fixing! Lgtm
Enabling the port overwrite makes sense. I doubt however we need the additional code to check for a blocked port

TuurStuyck:
Updated in a follow-up commit: removed the proactive port validation/availability checks and kept only the `-p <port>` override, so OpenPI will surface the same native bind errors as the other server launchers.

I also checked the other policy-server paths: GR00T already exposes server port selection in the documented `run_gr00t_server.py --port ...` flow and via `SERVER_PORT` in the CI server script; DreamZero's standalone server workflow already templates `--port`; OSMO-managed servers use an internal fixed port in isolated tasks. I did not find another local policy-server wrapper that needs the same `-p` addition.

PR #1065, opened 2026-08-08
Shorten agentic env-spec CLI names

qianl-nv:
## Summary
Rename agentic env-spec CLI flag and runner

## Detailed description
- Long `--env_graph_spec_yaml` / `environment_generation_runner` names made docs and CLI usage harder to scan
- Renamed the flag to `--env_spec` and the script to `cli_runner.py`, updating runners, hydra/OSMO paths, tests, and workflow docs
- Breaking CLI change for anyone still passing `--env_graph_spec_yaml`

### DIFF
--- a/docs/pages/concepts/environment/environment_definition.rst
+++ b/docs/pages/concepts/environment/environment_definition.rst
@@ -1,5 +1,5 @@
-Environment Spec
-================
+Environment Definition
+======================
 
 An Arena environment is three parts — scene, embodiment, and task (see the
 :doc:`section overview <index>`). This page covers the two ways to define that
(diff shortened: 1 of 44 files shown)


docs/pages/concepts/environment/environment_definition.rst
```diff
@@ -331,7 +331,7 @@ How to spawn an environment
    * - .. code-block:: bash
 
           python isaaclab_arena/scripts/environment_runner.py \
-            --env_graph_spec_yaml \
+            --env_spec \
```
xyao-nv, line 334: Since yaml is the env spec now, shall we rename the this file too? It covers similarity and difference of py vs yaml for env description.

xyao-nv (approved):

docs/pages/concepts/environment/environment_definition.rst
qianl-nv: good point, renamed this to environent_definition.rst

PR #1084, opened 2026-08-11
Add native uv support for lightweight GR00T clients

xyao-nv:
## Summary
UV envs didn't load GR00T path correctly. It makes gr00t importable by expanding gr00t modulepath as a group in toml.

## Detailed description
- Adds a gr00t-client dependency group backed by the Isaac-GR00T submodule.
- Supplies minimal dependency metadata so uv can resolve only the client dependencies
    (msgpack, numpy, and pyzmq) without installing the full GR00T training stack.
- Updates uv.lock.


docker/Dockerfile.isaaclab_arena
```diff
@@ -64,9 +64,6 @@ RUN /isaac-sim/python.sh -m pip install -e \
 # Upgrade pip ahead of the editable install
 RUN /isaac-sim/python.sh -m pip install --upgrade pip
 
-# PyAV: video read/write backend used by isaaclab_arena_gr00t/lerobot/convert_hdf5_to_lerobot.py
```
xyao-nv, line 67: This is duplicate from previous line.

cvolkcvolk (approved):
Thanks for the update! Looks good to me. Could you upadte the PR title and body to the current implementation before merging? 
And just one small note on the CI: The the passed jobs do not currently select `--group gr00t-client`, so they have not tested the installation added by this PR.

PR #1085, opened 2026-08-11
[Doc] Add uv run & reachability validator instructions in agentic env gen

xyao-nv:
## Summary
Add uv run in pre-reqs in agentic env gen index, reachability validator instructions in kitchen PnP. Replace kitchen PnP gif to reachable layouts.


qianl-nv (approved):
Thanks, looks great!

PR #1092, opened 2026-08-13
Split native-uv validation into a self-contained CI workflow and add GR00T client coverage

xyao-nv:
## Summary
Split native-uv validation into a self-contained CI workflow and add GR00T client coverage.

## Detailed description
- Moves native-uv jobs from ci.yml into independently discovered uv-ci.yml.
- Runs both wheel and source variants when UV deps, submodules, or the **workflow itself changes**(new); also runs on main, nightly, and manually.
- Installs gr00t-client and runs policy client test on CI


.github/workflows/ci.yml
cvolkcvolk, line 13:
```suggestion
    # Nightly at 06:00 UTC
```

cvolkcvolk (approved):
Lgtm

PR #1097, opened 2026-08-17
Fix Agibot left-arm teleop defects and env_cfg_callback contract

xingyoujun:
## Summary

Fix three defects that make Agibot left-arm teleoperation unusable and crash documented `env_cfg_callback` usage.

- Add a robot-only reset event to `AgibotEmbodiment`: `event_config` was never set, so every joint reset to 0 and RMPFlow's lula model no longer described the actual robot. Robot-only rather than `mdp.reset_scene_to_default`, which would also reset every scene asset and trample task-level contracts such as `disable_reset_pose`.
- Fix the left arm's frame/body offset quaternions, written as (w,x,y,z) in (x,y,z,w) fields: read as a 180° flip, the arm ran away ~270 mm under a zero command after every reset (corrected: 9.6 mm) and could not be teleoperated.
- `ArenaEnvBuilder` accept an `env_cfg_callback` that must return a cfg. Update the doc and comments to state so.
- Harden `test_place_upright_mug_multi` to assert the env_ids/percentage plumbing at the moment of the write: it previously depended on marginal rotate-in-place physics that any change to the robot's reset behaviour flips.

### DIFF
--- a/isaaclab_arena/embodiments/agibot/agibot.py
+++ b/isaaclab_arena/embodiments/agibot/agibot.py
@@ -13,7 +13,6 @@ from isaaclab.envs.mdp.actions.rmpflow_actions_cfg import RMPFlowActionCfg
 from isaaclab.managers import EventTermCfg
 from isaaclab.managers import ObservationGroupCfg as ObsGroup
 from isaaclab.managers import ObservationTermCfg as ObsTerm
-from isaaclab.managers import SceneEntityCfg
 from isaaclab.markers.config import FRAME_MARKER_CFG
 from isaaclab.sensors.frame_transformer.frame_transformer_cfg import FrameTransformerCfg, OffsetCfg
 from isaaclab.utils.configclass import configclass
@@ -25,6 +24,7 @@ from isaaclab_arena.assets.register import register_asset
 from isaaclab_arena.embodiments.common.arm_mode import ArmMode
 from isaaclab_arena.embodiments.embodiment_base import EmbodimentBase
 from isaaclab_arena.embodiments.franka.franka import FrankaMimicEnv
+from isaaclab_arena.terms.events import reset_joint_position_and_velocity_to_defaults
 from isaaclab_arena.utils.pose import Pose
 
 
@@ -148,28 +148,17 @@ class AgibotRightArmActionsCfg:
     )
 
 
-def reset_robot_to_default(env, env_ids, asset_cfg: SceneEntityCfg = SceneEntityCfg("robot")):
-    """Restore the robot's default joint state and joint targets on reset."""
-    asset = env.scene[asset_cfg.name]
-    default_joint_pos = asset.data.default_joint_pos.torch[env_ids].clone()
-    default_joint_vel = asset.data.default_joint_vel.torch[env_ids].clone()
-    asset.write_joint_position_to_sim_index(position=default_joint_pos, env_ids=env_ids)
-    asset.write_joint_velocity_to_sim_index(velocity=default_joint_vel, env_ids=env_ids)
-    asset.set_joint_position_target_index(target=default_joint_pos, env_ids=env_ids)
-    asset.set_joint_velocity_target_index(target=default_joint_vel, env_ids=env_ids)
-
-
 @configclass
 class AgibotEventCfg:
     """Reset events for the Agibot robot."""
 
     reset_robot_to_default_pose = EventTermCfg(
-        func=reset_robot_to_default,
+        func=reset_joint_position_and_velocity_to_defaults,
         mode="reset",
     )
     """Restore ``init_state`` joint values and targets on every reset.
 
-    RMPFlow expects `joint_lift_body`` and ``joint_body_pitch`` at those default values
+    RMPFlow expects ``joint_lift_body`` and ``joint_body_pitch`` at those default values
     through ``cspace_to_urdf_rules`` in ``agibot_left_arm_gripper.yaml``."""
 
 
--- a/isaaclab_arena/embodiments/agibot/agibot.py
+++ b/isaaclab_arena/embodiments/agibot/agibot.py
@@ -10,8 +10,10 @@ import isaaclab.envs.mdp as mdp
 import isaaclab.utils.math as PoseUtils
 from isaaclab.controllers.config.rmp_flow import AGIBOT_LEFT_ARM_RMPFLOW_CFG, AGIBOT_RIGHT_ARM_RMPFLOW_CFG
 from isaaclab.envs.mdp.actions.rmpflow_actions_cfg import RMPFlowActionCfg
+from isaaclab.managers import EventTermCfg
 from isaaclab.managers import ObservationGroupCfg as ObsGroup
 from isaaclab.managers import ObservationTermCfg as ObsTerm
+from isaaclab.managers import SceneEntityCfg
 from isaaclab.markers.config import FRAME_MARKER_CFG
 from isaaclab.sensors.frame_transformer.frame_transformer_cfg import FrameTransformerCfg, OffsetCfg
 from isaaclab.utils.configclass import configclass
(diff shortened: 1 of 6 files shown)


isaaclab_arena/embodiments/agibot/agibot.py
```diff
@@ -112,7 +118,15 @@ class AgibotLeftArmActionsCfg:
         body_name="gripper_center",
         controller=AGIBOT_LEFT_ARM_RMPFLOW_CFG,
         scale=1.0,
-        body_offset=RMPFlowActionCfg.OffsetCfg(rot=[0.7071, 0.0, -0.7071, 0.0]),
+        # -90 deg about y, in OffsetCfg's (x, y, z, w) order. The left end-effector frame is
```
qianl-nv: ditto, no need for inline comment here. putting the info in commit message (which you already did) is sufficient

isaaclab_arena/embodiments/agibot/agibot.py
```diff
@@ -65,7 +68,10 @@ class AgibotLeftArmSceneCfg(AgibotSceneCfg):
                 prim_path="{ENV_REGEX_NS}/Robot/gripper_center",
                 name="left_end_effector",
                 offset=OffsetCfg(
-                    rot=(0.7071, 0.0, -0.7071, 0.0),
+                    # -90 deg about y in OffsetCfg's (x, y, z, w) order. Previously written
```
qianl-nv: remove comment. 
the xyzw switch was introduced by Lab 3.0 beta1 in commit 7be783644. this seems to an overlook from that upgrade. not need for extra inline comment since xyzw is already the default order.

isaaclab_arena/embodiments/agibot/agibot.py
```diff
@@ -145,6 +159,44 @@ class AgibotRightArmActionsCfg:
...
+    """Restore the robot's default joint state and joint targets on reset.
+
+    Deliberately robot-only, matching the Franka embodiment's reset event.
+    ``mdp.reset_scene_to_default`` would also restore every other scene asset, which tramples
+    task-level contracts -- an object whose reset pose a task disabled (``disable_reset_pose``)
+    would get teleported back to its spawn state by the embodiment.
+    """
+    asset = env.scene[asset_cfg.name]
+    default_joint_pos = asset.data.default_joint_pos.torch[env_ids].clone()
+    default_joint_vel = asset.data.default_joint_vel.torch[env_ids].clone()
```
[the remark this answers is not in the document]
qianl-nv: this comment is not valid for normal execution
Isaac Lab creates env_ids on env.device before reset.
.torch is the official cached, zero-copy equivalent of wp.to_torch.
Isaac Lab’s own reset_scene_to_default uses the same pattern:

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -391,8 +391,12 @@ def compose_manager_cfg(self) -> tuple[IsaacLabArenaManagerBasedRLEnvCfg, dict[s
 
         # Apply the environment configuration callback if it is set
         # This can be used to modify the simulation configuration, etc.
+        # The callback may either mutate the config in place (returning None) or return a
+        # replacement config.
         if self.arena_env.env_cfg_callback is not None:
-            env_cfg = self.arena_env.env_cfg_callback(env_cfg)
+            callback_result = self.arena_env.env_cfg_callback(env_cfg)
```
qianl-nv: I would vote a enforcing the env_cfg_callback to always return the modified env_cfg (whether its modified in place or not). This makes the contract clear and avoids accident missing return statements. Could we do that?
Then we can revert the change here, and update the concept doc.

isaaclab_arena/environments/isaaclab_arena_environment.py
```diff
@@ -42,6 +42,8 @@ def __init__(
             task: The task to use in the environment.
             teleop_device: The teleop device to use in the environment.
             env_cfg_callback: A callback function that modifies the environment configuration.
+                It may mutate the config in place and return None, or return a replacement
```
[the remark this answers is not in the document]
qianl-nv: if we enforce return, then the type hint should be

env_cfg_callback: Callable[
    [IsaacLabArenaManagerBasedRLEnvCfg],
    IsaacLabArenaManagerBasedRLEnvCfg,
] | None = None,

isaaclab_arena/tests/test_place_upright_task.py
```diff
@@ -90,45 +90,47 @@ def assert_upright(env: ManagerBasedEnv, terminated: torch.Tensor):
 
 
 def _test_place_upright_mug_multi(simulation_app) -> bool:
-
-    from isaaclab_arena.tests.utils.simulation import step_zeros_and_call
-
+    """Checks that place_upright's env_ids / upright_percentage plumbing hits the right envs.
+
+    Asserts immediately after each ``place_upright`` call, without stepping physics in between.
```
qianl-nv: can shorten the comment to keep only the 1st sentence

qianl-nv (commented):
Thanks for the fixes, they are overall great.

I have a few nits, and a suggestion to enforce the env_cfg_callback to always return (and update the concept doc to match).
No current repository callback returns None; all return env_cfg. It's a clearer contract.

isaaclab_arena/embodiments/agibot/agibot.py
```diff
@@ -145,6 +148,44 @@ class AgibotRightArmActionsCfg:
...
+class AgibotEventCfg:
+    """Reset events for the Agibot robot."""
+
+    reset_robot_to_default_pose = EventTermCfg(
+        func=reset_robot_to_default,
+        mode="reset",
+    )
+    """Restore ``AGIBOT_A2D_CFG.init_state`` joint values and targets on every reset.
+
+    Without this the robot resets with every joint at 0, including ``joint_lift_body`` and
```
qianl-nv, line 178: Please rewrite and shorten this comment to 1~2 lines. In general, Arena prefers concise, short comments over lengthy agentic generated ones.

isaaclab_arena/embodiments/agibot/agibot.py
```diff
@@ -145,6 +148,44 @@ class AgibotRightArmActionsCfg:
     )
 
 
+def reset_robot_to_default(env, env_ids, asset_cfg: SceneEntityCfg = SceneEntityCfg("robot")):
+    """Restore the robot's default joint state and joint targets on reset.
+
+    Deliberately robot-only, matching the Franka embodiment's reset event.
```
qianl-nv, line 154: remove comment. Keep the summary line above should be sufficient

qianl-nv (approved):

isaaclab_arena/embodiments/agibot/agibot.py
```diff
@@ -145,6 +148,44 @@ class AgibotRightArmActionsCfg:
...
+    task-level contracts -- an object whose reset pose a task disabled (``disable_reset_pose``)
+    would get teleported back to its spawn state by the embodiment.
+    """
+    asset = env.scene[asset_cfg.name]
+    default_joint_pos = asset.data.default_joint_pos.torch[env_ids].clone()
+    default_joint_vel = asset.data.default_joint_vel.torch[env_ids].clone()
+    asset.write_joint_position_to_sim_index(position=default_joint_pos, env_ids=env_ids)
+    asset.write_joint_velocity_to_sim_index(velocity=default_joint_vel, env_ids=env_ids)
+    asset.set_joint_position_target_index(target=default_joint_pos, env_ids=env_ids)
+    asset.set_joint_velocity_target_index(target=default_joint_vel, env_ids=env_ids)
```
alexmillane: This is a general function, i.e. not Agibot specific. Could we move this to somewhere more appropriate?

Also this function is not robot specific either no? Would a move appropriate name be `reset_joint_position_and_velocity_to_defaults`?

alexmillane (commented):
Thank you for the fix.

One small comment. Ping me when you're ready for another review.

isaaclab_arena/embodiments/agibot/agibot.py
xingyoujun: Done in 81b96ad54 — moved to `isaaclab_arena/terms/events.py` as
`reset_joint_position_and_velocity_to_defaults`. Agreed it's neither Agibot- nor
robot-specific.

Why it doesn't reuse the existing terms: `reset_all_articulation_joints` resets *every*
articulation (plus root pose) and doesn't touch the joint drive targets, and
`mdp.reset_scene_to_default` resets every scene asset. This term is scoped to a single
asset and also resets the drive targets, which is the part the Agibot fix depends on.

PR #1104, opened 2026-08-18
Expand agent-generated PnP envs using LW kitchens

xyao-nv:
## Summary
Add 4 more agent-generated envs yaml in LW kitchens (diff layouts & styles than existing). Validated using Pi with Curobo reachability checker.

### DIFF
--- a/isaaclab_arena_environments/kitchen_bench/droid_place_bananas_in_wooden_bowl_u_shaped_with_island_farmhouse2.yaml
+++ b/isaaclab_arena_environments/kitchen_bench/droid_place_bananas_in_wooden_bowl_u_shaped_with_island_farmhouse2.yaml
@@ -14,7 +14,9 @@ embodiment:
 background:
   id: kitchen
   registry_name: lightwheel_kitchen_u_shaped_with_island_farmhouse2
-  params: {}
+  params:
+    collision_mode: mesh
+    repair_collision_mesh_non_watertight: false
 objects:
 - id: banana_1
   registry_name: banana_ycb_robolab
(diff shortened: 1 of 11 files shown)


isaaclab_arena_environments/kitchen_bench/droid_place_bananas_in_wooden_bowl_u_shaped_with_island_farmhouse2.yaml
```diff
@@ -0,0 +1,109 @@
...
+  id: droid
+  registry_name: droid_abs_joint_pos
+  params:
+    stand_height_m: 0.8
+    placement_bbox_stand_only: true
+    collision_mode: mesh
+background:
+  id: kitchen
+  registry_name: lightwheel_kitchen_u_shaped_with_island_farmhouse2
+  params: {}
```
qianl-nv: for the other lightwheel example we have

background:
  id: kitchen
  registry_name: lightwheel_kitchen_one_wall_coastal
  params:
    collision_mode: mesh
    repair_collision_mesh_non_watertight: false

does the new kitchen scenes has no existing collision relevant objects on the table so we can use bbox mode? if not let's set them to mesh mode?

qianl-nv (approved):
These tasks look great, thanks! 
one nit regarding background bbox vs mesh mode; there's also agent comment about broken doc link

isaaclab_arena_environments/kitchen_bench/droid_place_bananas_in_wooden_bowl_u_shaped_with_island_farmhouse2.yaml
xyao-nv: Isn't it already hardcoded when creating `FixedCollisionObject`?

xyao-nv: but added

PR #1110, opened 2026-08-19
Fix Agibot left-arm teleop defects and env_cfg_callback contract

qianl-nv:
## Summary
Fix three defects that make Agibot left-arm teleoperation unusable and crash documented `env_cfg_callback` usage.

## Detailed description
- Add a robot-only reset event to `AgibotEmbodiment`: `event_config` was never set, so every joint reset to 0 and RMPFlow's lula model no longer described the actual robot. Robot-only rather than `mdp.reset_scene_to_default`, which would also reset every scene asset and trample task-level contracts such as `disable_reset_pose`.
- Fix the left arm's frame/body offset quaternions, written as (w,x,y,z) in (x,y,z,w) fields: read as a 180° flip, the arm ran away ~270 mm under a zero command after every reset (corrected: 9.6 mm) and could not be teleoperated.
- `ArenaEnvBuilder` accept an `env_cfg_callback` that must return a cfg. Update the doc and comments to state so.
- Harden `test_place_upright_mug_multi` to assert the env_ids/percentage plumbing at the moment of the write: it previously depended on marginal rotate-in-place physics that any change to the robot's reset behaviour flips.

Resubmitted for https://github.com/isaac-sim/IsaacLab-Arena/pull/1097 since the CI doesn't work from fork.

### DIFF
--- a/isaaclab_arena/embodiments/agibot/agibot.py
+++ b/isaaclab_arena/embodiments/agibot/agibot.py
@@ -13,7 +13,6 @@ from isaaclab.envs.mdp.actions.rmpflow_actions_cfg import RMPFlowActionCfg
 from isaaclab.managers import EventTermCfg
 from isaaclab.managers import ObservationGroupCfg as ObsGroup
 from isaaclab.managers import ObservationTermCfg as ObsTerm
-from isaaclab.managers import SceneEntityCfg
 from isaaclab.markers.config import FRAME_MARKER_CFG
 from isaaclab.sensors.frame_transformer.frame_transformer_cfg import FrameTransformerCfg, OffsetCfg
 from isaaclab.utils.configclass import configclass
@@ -25,6 +24,7 @@ from isaaclab_arena.assets.register import register_asset
 from isaaclab_arena.embodiments.common.arm_mode import ArmMode
 from isaaclab_arena.embodiments.embodiment_base import EmbodimentBase
 from isaaclab_arena.embodiments.franka.franka import FrankaMimicEnv
+from isaaclab_arena.terms.events import reset_joint_position_and_velocity_to_defaults
 from isaaclab_arena.utils.pose import Pose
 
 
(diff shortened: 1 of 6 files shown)


isaaclab_arena/embodiments/agibot/agibot.py
```diff
@@ -145,6 +148,31 @@ class AgibotRightArmActionsCfg:
     )
 
 
+def reset_robot_to_default(env, env_ids, asset_cfg: SceneEntityCfg = SceneEntityCfg("robot")):
```
xyao-nv: `reset_joints_to_default`? Seems no root poses related. Shall it be shared among embodiments.py? Nothing specific to agibot.

isaaclab_arena/environments/isaaclab_arena_environment.py
```diff
@@ -28,7 +28,13 @@ def __init__(
         embodiment: EmbodimentBase | None = None,
         task: TaskBase | None = None,
         teleop_device: TeleopDeviceBase | None = None,
-        env_cfg_callback: Callable[IsaacLabArenaManagerBasedRLEnvCfg] | None = None,
+        env_cfg_callback: (
```
xyao-nv, line 31: Why returning? Is it a must? The callback can modify env_cfg in place and return nothing. Whats the reason behind?

xyao-nv (approved):

isaaclab_arena/environments/isaaclab_arena_environment.py
qianl-nv: Bug 3 — the env_cfg_callback doc example discards the config
Symptom: a callback written exactly as documented crashes with AttributeError: 'NoneType' object has no attribute 'seed' before the environment starts.
docs/pages/concepts/environment/environment_definition.rst:180-184 mutates env_cfg and returns nothing, but isaaclab_arena/environments/arena_env_builder.py:395 does env_cfg = self.arena_env.env_cfg_callback(env_cfg) and dereferences the result at line 398.
Fix: add return env_cfg to the doc example, or (preferably, so existing callbacks keep working) ignore a None return in the

isaaclab_arena/embodiments/agibot/agibot.py
qianl-nv: latest fix moves it into event. Franka/Droid currently reset with randomization so not using this

qianl-nv:
Closing this and move to https://github.com/isaac-sim/IsaacLab-Arena/pull/1097
There are some more discussion on that original MR, and CI also seems to run correctly there now

PR #1111, opened 2026-08-19
Document next_to side parameters & rename object in agentic env workflow

xyao-nv:
## Summary
Document next_to side parameters (fix to https://nvbugspro.nvidia.com/bug/6628702) and rename the Pepsi can.


qianl-nv (approved):
Thanks for the fix!

PR #1118, opened 2026-08-20
Pin GR00T doc links in interop workflows

xyao-nv:
## Summary
In IL interop workflow doc, pin GR00T documentation links to specific commit.Fix to https://nvbugspro.nvidia.com/bug/6640966.


qianl-nv (approved):
Thanks!

PR #1127, opened 2026-08-21
Clarify Experiment override paths

cvolkcvolk:
## Summary
Clarify local and OSMO override paths

## Detailed description
- Explain the local shared.* and runs.* override namespaces.
- Show that OSMO Experiment overrides require the experiment_cfg.* prefix.
- Add shared override examples and warn against the misleading +shared.* suggestion.


xyao-nv (approved):
Mainly doc and a quick glance LGTM.

PR #1129, opened 2026-08-21
Remove warning when no variation recorder is used

peterd-NV:
## Summary

The variation recorder is currently throwing a warning whenever the recorder is set to None to tell users that no variations are being recorded. A value of None for the recorder is a valid configuration and used for cases such as Isaac Lab recorder script interop. 

The current setup throws this warning message every env reset when no recorder is provided even when it's intended. This PR removes the warning to prevent confusion.


alexmillane (approved):
LGTM.

PR #1132, opened 2026-08-24
Fix geometric pick-and-place success checks

cvolkcvolk:
Provide a runtime lookup to provide a positional check `GeometricObjectOnDestinationTerm` that holds the cached aabb's of the objects and provide runtime position lookup.

Success now requires all three conditions:

- New `object_bounds_center_over_destination`: The center of the spawned object bounds is over the destination footprint and above its bottom.
- New `contact_force_is_upward_support`: The filtered contact force on the object points upward and exceeds the configured threshold.
- The object linear speed is below the configured threshold.

The old max_separation workaround is removed because the destination footprint now provides the geometric placement check


isaaclab_arena/tasks/predicates/object_on_destination_term.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+
+    def __call__(
+        self,
+        env: ManagerBasedEnv,
+        object_cfg: SceneEntityCfg,
+        destination_pose_cfg: SceneEntityCfg,
+        destination_prim_path: str,
+        contact_sensor_cfg: SceneEntityCfg,
+        force_threshold: float,
```
cvolkcvolk, line 80: Could those values be self. class member variables?

isaaclab_arena/tasks/predicates/geometric_object_on_destination_term.py
```diff
@@ -0,0 +1,120 @@
...
+        assert destination_pose_cfg.name == self._destination_pose_entity_name, (
+            f"This term cached geometry using destination pose entity '{self._destination_pose_entity_name}', "
+            f"but was called with '{destination_pose_cfg.name}'."
+        )
+        assert destination_prim_path == self._destination_prim_path, (
+            f"This term cached destination geometry at '{self._destination_prim_path}', "
+            f"but was called with '{destination_prim_path}'."
+        )
+
+        object_pose_w = self._object_geometry.get_pose_w()
```
cvolkcvolk: _object_geometry reads as if it is a fixed value instance and not that it is actually a live lookup to the env. Could we either redesign or at least rename such that is becomes more clear?

isaaclab_arena/tasks/predicates/object_on_destination_term.py
cvolkcvolk, line 80: Could or should most of those values be self. class member variables?

isaaclab_arena/tasks/predicates/live_scene_geometry.py
```diff
@@ -0,0 +1,306 @@
...
+
+    prim_path = prim.GetPath()
+    assert found_geometry, f"Prim '{prim_path}' has no default-purpose geometry."
+    return AxisAlignedBoundingBox(
+        min_point=tuple(float(value) for value in lower),
+        max_point=tuple(float(value) for value in upper),
+    )
+
+
+def build_spawned_entity_local_aabbs(
```
cvolkcvolk: Could we not name it entity or make it more expressive?

isaaclab_arena/tasks/predicates/live_scene_geometry.py
```diff
@@ -0,0 +1,306 @@
...
+            coverage_count[environment_id] += 1
+
+    assert all(count == 1 for count in coverage_count), (
+        f"Scene entity '{entity_name}' geometry must cover every environment exactly once; got coverage"
+        f" {coverage_count}."
+    )
+    return AxisAlignedBoundingBox(min_point=lower_by_environment, max_point=upper_by_environment)
+
+
+def _get_reference_pose_row_order(
```
cvolkcvolk: Are we sure we need this? Its not clear to me why we would

isaaclab_arena/tasks/predicates/live_scene_geometry.py
```diff
@@ -0,0 +1,306 @@
...
+            pose_row_by_environment[environment_id] == -1
+        ), f"Read-only scene reference '{entity_name}' has more than one pose for environment {environment_id}."
+        pose_row_by_environment[environment_id] = pose_row
+    assert all(
+        pose_row >= 0 for pose_row in pose_row_by_environment
+    ), f"Read-only scene reference '{entity_name}' does not provide exactly one pose per environment."
+    return torch.tensor(pose_row_by_environment, dtype=torch.long, device=env.device)
+
+
+class LiveSceneEntityGeometry:
```
cvolkcvolk: This class does two things, caching aabb and looking up a pose. Which both feel disconnected as the pose lookup does not even use the aabb. I feel like we should either split this  and make the intended boundary more clear overall.

isaaclab_arena/tasks/predicates/predicate_utils.py
cvolkcvolk, line 1: Do we need the changes in this file? If so why is this suddenly needed?

isaaclab_arena/tasks/composite_task_base.py
cvolkcvolk, line 1: This feels like a hack to me? Is this change really needed or could we postpone this?

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -23,22 +25,44 @@
...
 
 
 @agent_ready
 @register_task
 class PickAndPlaceTask(TaskBase):
-    """Pick-and-place task. Success fires when the pick-up object contacts the destination
-    with low velocity and, when ``max_separation`` is set, is within axis-aligned proximity
-    of the destination. Failure (object_dropped) fires when the object falls below the
-    background's ``object_min_z``.
+    """Pick an object up and place it on or in a destination.
```
cvolkcvolk, line 38: Please reduce the verbosity in this class docstring

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -23,22 +25,44 @@
...
+
+    Success requires the center of the object's spawned bounds to be over the destination's
+    footprint, the destination's filtered normal force on the object to point near world +Z,
+    and the object's current linear speed to be below ``velocity_threshold``. This is an
+    instantaneous check; the conditions do not need to persist for a minimum duration. Failure
+    occurs when the object falls below the background's ``object_min_z``.
+
+    An ordinary ``destination_location`` must be included in the environment's ``Scene``. A rigid
```
cvolkcvolk: This is not clear to me and reads like a lot of complexity. Did a requirement change now with our implementation or is this just documentation verbosity?

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -106,40 +131,21 @@ def get_scene_cfg(self):
...
-                        "object_cfg": SceneEntityCfg(self.pick_up_object.name),
-                        "target_object_cfg": SceneEntityCfg(self.destination_location.name),
-                        "max_x_separation": max_x_separation,
-                        "max_y_separation": max_y_separation,
-                        "max_z_separation": max_z_separation,
-                    },
-                )
-            )
+        # The class-backed term caches spawned object and destination bounds once for the runtime environment.
+        destination_pose_cfg = self._make_destination_pose_cfg()
```
cvolkcvolk: Do we really need this?

cvolkcvolk (commented):
self review

isaaclab_arena/tasks/predicates/object_on_destination_term.py
cvolkcvolk: Addressed in 3ab5d6653. I kept these as call parameters because Isaac Lab validates and passes all TerminationTermCfg params to class terms on every call. Instead, I removed their duplicate binding and validation from the constructor; only data that is actually cached remains instance state. The velocity threshold is now validated when PickAndPlaceTask is created.

cvolkcvolk: Addressed in 3ab5d6653. I kept these as call parameters because Isaac Lab validates and passes all TerminationTermCfg params to class terms on every call. Instead, I removed their duplicate binding and validation from the constructor; only data that is actually cached remains instance state. The velocity threshold is now validated when PickAndPlaceTask is created.

isaaclab_arena/tasks/predicates/live_scene_geometry.py
cvolkcvolk: Addressed in 5a4fe7e4d. Spawned AABB construction is now a separate function, while SceneEntityPoseReader has the single responsibility of reading current poses. GeometricObjectOnDestinationTerm owns the cached bounds and combines them with the current poses.

isaaclab_arena/tasks/predicates/geometric_object_on_destination_term.py
cvolkcvolk: Addressed in 5a4fe7e4d. Cached spawned AABBs and current pose access are now separate. The term stores the object and destination AABBs directly and uses SceneEntityPoseReader only for current poses, so the misleading geometry member no longer exists.

isaaclab_arena/tasks/predicates/live_scene_geometry.py
cvolkcvolk: Addressed in 2fa4961d8. The function is now named compute_spawned_geometry_aabbs_relative_to_pose. This makes clear that it computes bounds from the spawned geometry and expresses them relative to the pose selected by the caller.

isaaclab_arena/tasks/predicates/live_scene_geometry.py
cvolkcvolk: Addressed in f58a202c7. I removed the clone-plan row reconstruction and its mocked Newton test. The current PhysX path now validates that its post-clone FrameView exposes exactly one prim path per environment in environment order, then reads the pose rows directly. Newton pick-and-place is not verified in this release, and its FrameView does not expose row identity; that contract should be added in Isaac Lab before Arena supports this reference path on Newton.

isaaclab_arena/tasks/predicates/predicate_utils.py
cvolkcvolk: Addressed in 17be7e91e. These buffer-access changes were unrelated to the geometric success fix, so I restored predicate_utils.py exactly to main. The ProxyArray accessor modernization can be handled separately.

isaaclab_arena/tasks/composite_task_base.py
cvolkcvolk: Narrowed in 7350c0883. I restored the reset API and event parameters to main; that refactor was unrelated. The remaining change is required because atomic managers construct ManagerTermBase predicates, while composite tasks previously hid child success configs inside TaskBase objects and then called the unconstructed class directly. The composite now passes child TerminationTermCfg values through Isaac Lab manager resolution before evaluating them. This is covered by the composite and sequential manager regression test.

isaaclab_arena/tasks/pick_and_place_task.py
cvolkcvolk: The requirement came from the geometric check: it needs the spawned destination pose and geometry at runtime. I simplified the contract so every destination follows one rule: it is a named scene entity. The microwave turntable is now registered as a read-only reference, while the articulation remains its physics and reset owner.

isaaclab_arena/tasks/pick_and_place_task.py
cvolkcvolk: Reduced this to a short success/failure summary and a concise destination contract. I also removed the detailed Mimic example from the class docstring.

isaaclab_arena/tasks/pick_and_place_task.py
cvolkcvolk: No. I removed the helper and the separate destination pose config and prim path. The term now receives one destination SceneEntityCfg and derives both current pose and spawned geometry from that registered entity.

isaaclab_arena/tasks/predicates/spatial.py
```diff
@@ -103,27 +195,32 @@ def object_on_destination(
...
     # and ``M`` is the number of filtered bodies.
     # We assume B = 1 and M = 1
-    assert sensor.data.force_matrix_w.shape[2] == 1
-    assert sensor.data.force_matrix_w.shape[1] == 1
+    assert contact_sensor.data.force_matrix_w.shape[2] == 1
+    assert contact_sensor.data.force_matrix_w.shape[1] == 1
     # NOTE(alexmillane, 2025-08-04): We expect the binary flags to have shape (N, )
     # where N is the number of envs.
-    force_matrix_norm = torch.norm(wp.to_torch(sensor.data.force_matrix_w), dim=-1).reshape(-1)
+    force_matrix_norm = torch.norm(contact_sensor.data.force_matrix_w.torch, dim=-1).reshape(-1)
```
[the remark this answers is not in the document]
cvolkcvolk: Adressed

isaaclab_arena/tasks/predicates/geometric_object_on_destination_term.py
```diff
@@ -0,0 +1,125 @@
...
+
+        assert (
+            object_cfg.name in env.scene.rigid_objects
+        ), f"GeometricObjectOnDestinationTerm requires rigid object '{object_cfg.name}'."
+
+        self._object_name = object_cfg.name
+        self._destination_name = destination_cfg.name
+        self._object_rigid_object: RigidObject = env.scene[self._object_name]
+        self._object_aabbs = compute_spawned_geometry_aabbs_relative_to_pose(env, object_cfg)
+        self._destination_aabbs = compute_spawned_geometry_aabbs_relative_to_pose(env, destination_cfg)
```
alexmillane: A few questions about these AABBs:

- They computed in the object frame correct? Suggestion to suffix them with the frame name (suggestion `_O`).
- They're computed under the initial rotation right? The AABBs are later _translated_ by the runtime pose, however, could we not have issues when the object rotates? Take the extreme case of a long, narrow object, if it rotates after the sim is up and running, the AABB could be off by quite a bit.

isaaclab_arena/tasks/predicates/geometric_object_on_destination_term.py
```diff
@@ -0,0 +1,125 @@
...
+            f"This term cached geometry for destination '{self._destination_name}', "
+            f"but was called with '{destination_cfg.name}'."
+        )
+
+        object_pose_w = self._object_rigid_object.data.root_pose_w.torch
+        destination_pose_w = self._get_destination_pose_w()
+        object_center_over_destination = object_bounds_center_over_destination(
+            object_pose_w=object_pose_w,
+            object_bounds=self._object_aabbs,
+            destination_pose_w=destination_pose_w,
```
alexmillane: Would be good to align on frame conventions. I've used capitals in my code. I've also used the "active" convention where frames are labelled as the frames they take point from and to.

`T_W_O` would be the pose/transform taking points from the object frame `O` to the world frame `W`.

I feel that labelling like this makes transform composition less error prone.

isaaclab_arena/tasks/predicates/geometric_object_on_destination_term.py
```diff
@@ -0,0 +1,125 @@
...
+)
+from isaaclab_arena.tasks.predicates.spatial import (
+    contact_force_is_upward_support,
+    object_bounds_center_over_destination,
+)
+
+
+# TODO(cvolk, 2026-08-24): [arena-world-migration] Replace term-owned geometry caching and direct live-state reads
+# with env.arena_world queries, then make this a plain predicate function.
+class GeometricObjectOnDestinationTerm(ManagerTermBase):
```
alexmillane: nit: I'm wondering about the naming "geometric". This term uses force (which is totally fine). Our old term is also based on force. Am I correct that this term is no more geometric than our old term?

Is there any thing to speak against just making this always the `ObjectOnDestinationTerm`? Maybe we don't always want the geometric part of the check?

isaaclab_arena/tasks/predicates/object_on_destination_term.py
```diff
@@ -0,0 +1,125 @@
...
+
+        contact_sensor: ContactSensor = env.scene[contact_sensor_cfg.name]
+        force_matrix_w = contact_sensor.data.force_matrix_w
+        assert force_matrix_w is not None, f"Contact sensor '{contact_sensor_cfg.name}' has no filtered force matrix."
+        force_matrix_w = force_matrix_w.torch
+        assert force_matrix_w.shape == (env.num_envs, 1, 1, 3), (
+            f"Contact sensor '{contact_sensor_cfg.name}' must provide one sensed body and one filtered body; "
+            f"got force shape {tuple(force_matrix_w.shape)}."
+        )
+        support_force_on_object_w = force_matrix_w[:, 0, 0, :]
```
alexmillane, line 111: Suggestion ot add a comment on the meaning of these 0s

isaaclab_arena/tasks/predicates/geometric_object_on_destination_term.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+        support_force_on_object_w = force_matrix_w[:, 0, 0, :]
+        destination_provides_upward_support = contact_force_is_upward_support(
+            contact_force_w=support_force_on_object_w,
+            force_threshold=force_threshold,
+            support_cone_half_angle_deg=support_cone_half_angle_deg,
+        )
+
+        object_linear_velocity_w = self._object_rigid_object.data.root_lin_vel_w.torch
+        object_is_moving_slowly = torch.linalg.vector_norm(object_linear_velocity_w, dim=-1) < velocity_threshold
```
alexmillane: nit: Could move this out to a (simple) function `object_moving_slowly` for symmetry with the functions above?

isaaclab_arena/tasks/predicates/live_scene_geometry.py
```diff
@@ -0,0 +1,197 @@
...
+        representative_prim.GetPath(),
+        predicate=lambda prim: prim.HasAPI(UsdPhysics.RigidBodyAPI),
+        stage=representative_prim.GetStage(),
+        traverse_instance_prims=False,
+    )
+    assert len(rigid_body_prims) == 1, (
+        f"Rigid scene entity '{entity_name}' source '{representative_prim.GetPath()}' contains "
+        f"{len(rigid_body_prims)} rigid bodies; expected exactly one."
+    )
+    return rigid_body_prims[0]
```
alexmillane: Hard to tell what this function is doing?

Is is searching for a rigid body prim below `representative_prim`?

Suggestion to 
- `_find_rigid_body_prim` -> `_find_rigid_body_prim_below`
- `representative_prim` -> `parent`

isaaclab_arena/tasks/predicates/live_scene_geometry.py
```diff
@@ -0,0 +1,197 @@
...
+        lower = np.minimum(lower, np.asarray(geometry_range.GetMin(), dtype=np.float64))
+        upper = np.maximum(upper, np.asarray(geometry_range.GetMax(), dtype=np.float64))
+        found_geometry = True
+
+    prim_path = prim.GetPath()
+    assert found_geometry, f"Prim '{prim_path}' has no default-purpose geometry."
+    return AxisAlignedBoundingBox(
+        min_point=tuple(float(value) for value in lower),
+        max_point=tuple(float(value) for value in upper),
+    )
```
alexmillane: Is there any code sharing possible between this and `compute_local_bounding_box_from_prim`?

isaaclab_arena/tasks/predicates/live_scene_geometry.py
```diff
@@ -0,0 +1,197 @@
...
+    assert len(rigid_body_prims) == 1, (
+        f"Rigid scene entity '{entity_name}' source '{representative_prim.GetPath()}' contains "
+        f"{len(rigid_body_prims)} rigid bodies; expected exactly one."
+    )
+    return rigid_body_prims[0]
+
+
+def _compute_aabb_relative_to_prim(prim: Usd.Prim) -> AxisAlignedBoundingBox:
+    """Compute descendant geometry bounds relative to a prim's origin and axes."""
```
alexmillane: Is this AABB in the axis of the prim or the world axis?

isaaclab_arena/tasks/predicates/live_scene_geometry.py
```diff
@@ -0,0 +1,197 @@
...
+    with current poses rather than rebuilding it on each simulation step.
+
+    Args:
+        env: The runtime manager-based environment.
+        entity_cfg: Scene entity whose pose defines the returned coordinates.
+
+    Returns:
+        One AABB per environment, expressed relative to the entity pose used by
+        the predicate.
+    """
```
alexmillane: See comment above. When we say relative, I guess we mean that the AABB is axis aligned to the world, but is expressed with respect to the object center.

(this is relavent to my question about rotations in the term above).

isaaclab_arena/tasks/predicates/live_scene_geometry.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def get_pose_w(self) -> torch.Tensor:
+        """Return current poses as ``(x, y, z, qx, qy, qz, qw)``."""
+        position_w_buffer, orientation_w_buffer = self._frame_view.get_world_poses()
+        pose_w = torch.cat((position_w_buffer.torch, orientation_w_buffer.torch), dim=-1)
+        assert pose_w.shape == (self._num_envs, 7), (
+            f"AssetBaseCfg scene entry '{self._entity_name}' returned pose shape {tuple(pose_w.shape)}; "
+            f"expected ({self._num_envs}, 7)."
+        )
+        return pose_w
```
alexmillane: nit:

- name: `get_pose_in_world_frame`
- suggestion to expand the docstring to say "returns `T_W_O` the transform taking the points in the object frame `O` to the world frame `W`"

isaaclab_arena/tasks/predicates/spatial.py
```diff
@@ -3,18 +3,105 @@
...
+
+    The check requires the object center to be inside the destination's X/Y
+    footprint and above its lower Z bound. The upper Z bound is intentionally
+    ignored so the same check works for open containers and supporting surfaces.
+    This is a center-point test, not full-object containment.
+
+    Args:
+        object_pose_w: Object poses in world coordinates with shape
+            ``(num_envs, 7)`` and quaternion order ``(x, y, z, w)``.
+        object_bounds: Object bounds relative to ``object_pose_w``.
```
alexmillane: I'm thinking that this is likely relative to the object center, rather than relative to it's pose (which indicates it's rotated).

isaaclab_arena/tasks/predicates/spatial.py
```diff
@@ -3,18 +3,105 @@
...
+        destination_pose_w[:, :3],
+        destination_pose_w[:, 3:],
+        object_bounds_center_w,
+    )
+    center_inside_horizontal_bounds = (
+        (object_bounds_center_in_destination[:, :2] >= destination_bounds.min_point[:, :2])
+        & (object_bounds_center_in_destination[:, :2] <= destination_bounds.max_point[:, :2])
+    ).all(dim=-1)
+    center_above_destination_bottom = object_bounds_center_in_destination[:, 2] >= destination_bounds.min_point[:, 2]
+    return center_inside_horizontal_bounds & center_above_destination_bottom
```
alexmillane, line 70: cool!

isaaclab_arena/tasks/predicates/spatial.py
```diff
@@ -3,18 +3,105 @@
...
+    ), f"support_cone_half_angle_deg must be in [0, 90), got {support_cone_half_angle_deg}."
+
+    force_magnitude = torch.linalg.vector_norm(contact_force_w, dim=-1)
+    upward_force = contact_force_w[:, 2]
+    minimum_upward_fraction = math.cos(math.radians(support_cone_half_angle_deg))
+    return (
+        (force_magnitude >= force_threshold)
+        & (upward_force > 0.0)
+        & (upward_force >= force_magnitude * minimum_upward_fraction)
+    )
```
alexmillane, line 105: cool!

isaaclab_arena/tasks/composite_task_base.py
```diff
@@ -202,30 +201,33 @@ def _evaluate_subtask_successes(
     @staticmethod
     def composite_task_success_func(
         env,
-        subtasks: list[TaskBase],
+        resolved_subtask_success_cfgs: list[TerminationTermCfg],
```
alexmillane: Why the change here. Seems like it would be nicer to pass the subtasks?

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -107,39 +113,16 @@ def get_termination_cfg(self):
...
-                    params={
-                        "object_cfg": SceneEntityCfg(self.pick_up_object.name),
-                        "target_object_cfg": SceneEntityCfg(self.destination_location.name),
-                        "max_x_separation": max_x_separation,
-                        "max_y_separation": max_y_separation,
-                        "max_z_separation": max_z_separation,
-                    },
-                )
-            )
+        # The class-backed term caches spawned object and destination bounds once for the runtime environment.
```
alexmillane: remove superfluous comment.

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -107,39 +113,16 @@ def get_termination_cfg(self):
...
         success = TerminationTermCfg(
-            func=check_success,
+            func=GeometricObjectOnDestinationTerm,
```
alexmillane: Do we need to call this `GeometricObjectOnDestinationTerm`?

Do we need the old ObjectOnDestination? Could this just replace it?

isaaclab_arena/tasks/pick_and_place_task.py
alexmillane, line 180: This should match the success condition no? Otherwise this thing could fire, when the task success will not.

isaaclab_arena/tasks/sequential_task_base.py
```diff
@@ -25,43 +27,46 @@ class SequentialTaskBase(CompositeTaskBase):
     @staticmethod
     def composite_task_success_func(
         env,
-        subtasks: list[TaskBase],
+        resolved_subtask_success_cfgs: list[TerminationTermCfg],
```
alexmillane: struggling to see the reason here. I feel that the old way was cleaner.

isaaclab_arena_environments/robolab/tasks/clear_organic_objects.yaml
alexmillane, line 1: nice~

alexmillane (commented):
Thanks for solving this!

A few comments. The major one about the handling of rotation of destination objects.

isaaclab_arena/tasks/predicates/geometric_object_on_destination_term.py
cvolkcvolk: Yeah, we use IsaacLab's suffix "convention" convention like root_pose_w, object_pos_w etc. and agents like continuing with this. 
Shall we add a guideline to the agents.md to follow T_W_O convention going forward?

isaaclab_arena/tasks/predicates/geometric_object_on_destination_term.py
cvolkcvolk: Renamed to `ObjectOnDestinationTerm`

isaaclab_arena/tasks/predicates/geometric_object_on_destination_term.py
cvolkcvolk: Renamed to  `object_bounds_O` and `destination_bounds_D`

cvolkcvolk: Not quite. the AABB is computed in the object frame, so the object’s initial world translation and rotation are removed. At runtime its center is transformed using the full current pose. updated the code to make the transforms more explicit (rotations of both the object and destination are applied)

```
object_bounds_center_W = t_W_O + quat_apply(q_W_O, object_bounds_center_O)
object_bounds_center_D = quat_apply_inverse(q_W_D, object_bounds_center_W - t_W_D)
```

isaaclab_arena/tasks/predicates/geometric_object_on_destination_term.py
cvolkcvolk: Agreed. Updated

isaaclab_arena/tasks/predicates/live_scene_geometry.py
cvolkcvolk: Using ` _find_single_rigid_body_prim_in_subtree(root_prim, ...)` now

isaaclab_arena/tasks/predicates/live_scene_geometry.py
cvolkcvolk: Not really that it would be worth moving to split functions imo. This aligns the result with world axes and subtracts only translation and only

isaaclab_arena/tasks/predicates/live_scene_geometry.py
cvolkcvolk: To the prims axes and expressed relative to its origin. Added suffixes and renamed this function to make the frame more explicit

isaaclab_arena/tasks/predicates/live_scene_geometry.py
cvolkcvolk: renamed the functions to `..._in_prim_frame` and `..._in_entity_frame` to make that explicit

isaaclab_arena/tasks/predicates/spatial.py
cvolkcvolk: yes the point  tested is the center of the object’s (spawned) bounds

isaaclab_arena/tasks/sequential_task_base.py
cvolkcvolk: Mh for this fix, exposing the nested success configs to Isaac Lab seems to be smallest way to construct the class-based term correctly (Pick-and-place success is now a class-based manager term)

isaaclab_arena/tasks/pick_and_place_task.py
cvolkcvolk: - Renamed to `ObjectOnDestinationTerm` 

- This is a bit suboptimal as the old object_on_destination() cannot be removed yet as its used in the progress tracking, which evaluates plain callables. Updating this would blow the scope of this PR

isaaclab_arena/tasks/pick_and_place_task.py
cvolkcvolk: So, our new manager-owned success would go through composite and sequential progress tracking... and the implementation is broadening the scope of this PR considerably. I would suggest doing this in a follow up.

alexmillane (commented):
Thanks for fixing that!

alexmillane (approved):

PR #1133, opened 2026-08-24
Add empirical sensitivity reports

cvolkcvolk:
## Summary
Add empirical sensitivity reports

## Detailed description
- Let exact discrete observations use matching episodes directly without fitting NPE or MNPE.
- Compare posterior bin mass with the sampled baseline and show bootstrap uncertainty.
- Route empirical and fitted analysis through SensitivityAnalyzer.analyze(method=...), while retaining fit() and sample_posterior() for fitted reuse.
- Preserve fitted inference as the report default and add deterministic regression coverage.
- Recording declared factor bounds and display metadata remains a follow-up.


docs/adr/0001-select-sensitivity-posterior-method.md
cvolkcvolk: This should not live in this PR? Maybe move to another draft PR to park this idea

isaaclab_arena/analysis/sensitivity/generate_report.py
```diff
@@ -107,11 +130,34 @@ def main():
...
+            "'fitted' trains NPE or MNPE. Default: fitted."
+        ),
+    )
+    parser.add_argument(
+        "--bins",
+        type=int,
+        default=6,
+        help="Number of fixed bins per continuous factor for --method empirical. Default: 6.",
+    )
+    parser.add_argument(
```
cvolkcvolk: Do we need this?

isaaclab_arena/analysis/sensitivity/analyzer.py
```diff
@@ -6,17 +6,45 @@
...
 
 from sbi.inference import MNPE, NPE
 from sbi.utils import BoxUniform
 
 from isaaclab_arena.analysis.sensitivity.dataset import SensitivityDataset
+from isaaclab_arena.analysis.sensitivity.empirical import EmpiricalSensitivityResult, compute_empirical_marginals
+
+
+@dataclass(frozen=True)
+class FittedSensitivityResult:
```
cvolkcvolk: Why do we add this class for this PR?

isaaclab_arena/analysis/sensitivity/dataset.py
```diff
@@ -119,7 +119,31 @@ def default_observation(self) -> torch.Tensor:
         """
         is_binary = set(self._x.flatten().tolist()).issubset({0.0, 1.0})
         assert is_binary, "default_observation assumes binary (0/1) outcomes; pass an explicit observation otherwise."
-        return torch.ones(self._x.shape[1], dtype=torch.float32)
+        return torch.ones(self._x.shape[1], dtype=self._x.dtype, device=self._x.device)
+
+    def resolve_observation(
```
cvolkcvolk: Could we have a more explicit name than "resolve"?

PR #1166, opened 2026-08-31
Upgrade to Isaac Lab 3.0 GA

peterd-NV:
## Summary
Upgrade to Isaac Lab 3.0 GA release/3.0.0 branch.

Main changes:

- Bring in Lab 3.0 GA and Newton 1.5
- Update Isaac Lab import paths that were reorged
- Update changed Lab APIs (renderer settings, object pose access, etc.)
- Move torch/numpy imports to be after kit startup (new Lab requirement)
- Add warp array patch to enable warp to torch conversion for empty CPU buffers
- Remove deprecated torchvision.io video writer
- Update tests for new APIs, import ordering
- Update uv versions and lock
- Switch CI to use lab 3.0 ga image (to be removed in future PR)
- Disable uv wheel tests since Lab 3.0 GA wheel is not available yet
- Migrate RL training to Lab's new unified entry point
- Removes` --headless` cli arg (lack of `--viz` now defaults to headless) 

KNOWN ISSUES: 

1. Lab's RTX per-env scene partitioning prevents kit viewport from being assigned a visible partition. Result is that the kit viewport can appear blank even if simulation is running in the background. Lab source has fix but also requires Sim 6.1 with latest kit build. Fix: Add `ISAAC_LAB_ENABLE_ISAAC_RTX_PER_ENV_SCENE_PARTITION=0 python your_script.py ...` before running.


isaaclab_arena/utils/isaaclab_utils/resolve_clone_plan_source_patch.py
```diff
@@ -3,115 +3,23 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""Runtime patch for ``isaaclab.cloner.cloner_utils.resolve_clone_plan_source``.
+"""Backward-compatible no-op for Arena's retired nested-clone workaround.
```
qianl-nv: Are there particular reason we need to keep this backward-compatible API? Feel like a this stage Arena is generally ok with breaking changes?

isaaclab_arena/scripts/train.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) 2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
```
qianl-nv, line 1: Looks like this wrapper is needed due to lab breaking the external callback for env registration in their unified trainer. Do we plan to fix this in Lab before GA and remove this script then, or we expect this to stick around for 0.3.1 release at least?

peterd-NV: I suggest we prepare to keep this fix for 0.3.1 but simultaneously make a fix on the lab side and hopefully get it in before Lab GA goes live

isaaclab_arena/utils/isaaclab_utils/resolve_clone_plan_source_patch.py
peterd-NV: removed

qianl-nv (approved):
Thanks for the update! LGTM.

.github/workflows/ci.yml
```diff
@@ -184,7 +185,8 @@ jobs:
     needs: [pre_commit]
 
     container:
-      image: nvcr.io/nvstaging/isaac-amr/isaaclab_arena:latest
+      # TODO(peterd, 2026-08-31): Restore :latest after the official Lab 3.0 GA CI image is published.
+      image: nvcr.io/nvstaging/isaac-amr/isaaclab_arena:peterd-lab3-ga-ae8a9aa3db
```
alexmillane, line 189: I guess we want to push this to our latest tag `nvcr.io/nvstaging/isaac-amr/isaaclab_arena:latest` before merge?

alexmillane: Is this image built against Isaac Sim 6.1 based on an internal image?

isaaclab_arena/assets/object_base.py
```diff
@@ -51,6 +52,20 @@ def __init__(
             self.add_variation(ObjectMassVariation(self.name))
         self.initial_velocity: Velocity | None = None
         self.object_cfg = None
+        self._base_frame_view: FrameView | None = None
+        self._base_frame_view_stage = None
+
+    def _close_base_frame_view(self) -> None:
+        """Release the cached frame view. Safe to call more than once."""
+        frame_view = getattr(self, "_base_frame_view", None)
```
alexmillane: superfluous `getattr`

isaaclab_arena/assets/object_base.py
```diff
@@ -162,7 +177,31 @@ def get_object_pose(self, env: ManagerBasedEnv, is_relative: bool = True) -> tor
...
+                            f"AssetBaseCfg scene entry '{self.name}' pose row {environment_id} belongs to"
+                            f" '{prim_path}', not environment '{environment_prim_path}'."
+                        )
+                except Exception:
+                    frame_view.close()
+                    raise
+                self._base_frame_view = frame_view
+                self._base_frame_view_stage = stage
+            position_w, orientation_w = self._base_frame_view.get_world_poses()
+            object_pose = torch.cat((position_w.torch, orientation_w.torch), dim=-1)
```
alexmillane, line 200: This is looking really complex compared to what things used to look like.

A couple of questions:
- Is this what Isaac Lab 3.0 expects to look up the pose of an AssetBase?
- Is this caching of `_base_frame_view` and `_base_frame_view_stage` required?
- What's going on in the middle section of the code where we're iterating over the environments?

Something looks... suspect about this code. I would like to check with the lab team that this is really what's required.

@peterd-NV do you know why all of this is required?

isaaclab_arena/assets/object_base.py
```diff
@@ -51,6 +52,20 @@ def __init__(
...
+        """Release the cached frame view. Safe to call more than once."""
+        frame_view = getattr(self, "_base_frame_view", None)
+        if frame_view is not None:
+            frame_view.close()
+            self._base_frame_view = None
+            self._base_frame_view_stage = None
+
+    def __del__(self):
+        """Release the cached frame view when this object is dropped."""
+        self._close_base_frame_view()
```
alexmillane: See comment below. Is this caching required? It's making the class a lot more complex.

isaaclab_arena/cli/isaaclab_arena_cli.py
```diff
@@ -70,6 +70,21 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
...
     )
 
+    # Isaac Lab no longer exposes this programmatic AppLauncher option through argparse,
+    # but Arena still accepts it for its existing scripts, docs, and external integrations.
+    arena_group.add_argument(
+        "--headless",
+        action="store_true",
+        default=False,
+        help="Run without a local GUI. This is also the default when no visualizer is selected.",
+    )
```
alexmillane: Do we still need this? I thought by default our stuff runs headless. We activate the visualizer through `--viz kit`?

isaaclab_arena/cli/isaaclab_arena_cli.py
```diff
@@ -70,6 +70,21 @@ def add_isaaclab_arena_cli_args(parser: argparse.ArgumentParser) -> None:
...
+    arena_group.add_argument(
+        "--enable_cameras",
+        action="store_true",
+        default=False,
+        help="Add the environment's camera sensors and observations.",
+    )
+
```
alexmillane, line 81: Do we understand why this was removed in Lab/Sim? If they don't need it, that seems like an indication that we may not need it?

isaaclab_arena/assets/object_base.py
```diff
@@ -162,7 +177,31 @@ def get_object_pose(self, env: ManagerBasedEnv, is_relative: bool = True) -> tor
...
+                    assert len(prim_paths) == env.unwrapped.num_envs, (
+                        f"AssetBaseCfg scene entry '{self.name}' resolved to {len(prim_paths)} prims; "
+                        f"expected {env.unwrapped.num_envs}."
+                    )
+                    for environment_id, prim_path in enumerate(prim_paths):
+                        environment_prim_path = scene.env_prim_paths[environment_id]
+                        assert str(prim_path).startswith(f"{environment_prim_path}/"), (
+                            f"AssetBaseCfg scene entry '{self.name}' pose row {environment_id} belongs to"
+                            f" '{prim_path}', not environment '{environment_prim_path}'."
+                        )
```
alexmillane, line 193: Something seems a bit wrong here. On the one hand, we're caching the frame view, presumably for speed; on the other hand, we're iterating through lists in Python in the pose look up and doing string comparisons.

Why do we have to do this iteration? Maybe a comment is worth it?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -462,6 +463,8 @@ def build_registered(
         Returns:
             A ``(name, cfg, env_kwargs)`` tuple.
         """
+        install_empty_cpu_warp_to_torch_patch()
```
alexmillane, line 466: Is there a bug/issue link we could attach here? Let's also maybe add a `TODO` (with a tag) for us to remove this when the issue/bug is solved.

isaaclab_arena/evaluation/policy_runner.py
```diff
@@ -69,6 +68,8 @@ def rollout_policy(
     num_steps: int | None,
     num_episodes: int | None,
 ) -> MetricsDataCollection | None:
+    import torch
```
alexmillane, line 71: 😬 Do we have to move our torch imports local now?

isaaclab_arena/tests/test_kuka_allegro_embodiment.py
```diff
@@ -24,9 +22,7 @@ def test_kuka_allegro_registered_in_asset_registry() -> None:
 
 @pytest.mark.with_newton
 def test_scene_cfg_has_robot_and_fingertip_contact_sensors() -> None:
-    pytest.importorskip(
-        "isaaclab_tasks.manager_based.manipulation.dexsuite.config.kuka_allegro.dexsuite_kuka_allegro_env_cfg"
-    )
+    pytest.importorskip("isaaclab_tasks.core.lift.config.kuka_allegro.kuka_allegro_env_cfg")
```
alexmillane: Why do we have to do this?

Suggestion to add a comment a comment why we have to do this.

isaaclab_arena/tests/test_object_on_destination_term.py
```diff
@@ -232,6 +233,40 @@ def compute_geometry_bounds(_geometry_env, entity_cfg):
...
+    assert term_reader.close_calls == 1
+
+    # Dropping the term releases the reader, so no env-level bookkeeping is needed. This
+    # mirrors RecorderManager driving RecorderTerm.close from its own destructor.
+    dropped_reader = DummyClosable()
+    dropped_term = object.__new__(object_on_destination_term_module.ObjectOnDestinationTerm)
+    dropped_term._destination_asset_base_pose_reader = dropped_reader
+    del dropped_term
+    gc.collect()
+    assert dropped_reader.close_calls == 1
```
alexmillane: I'm fine to merge all the stuff related to the FrameViews and their garbage collection, but I feel we need to ask the Lab team about this. This can't be the expected operation of these things.

isaaclab_arena/utils/isaaclab_utils/warp_patch.py
```diff
@@ -0,0 +1,55 @@
...
+            if requires_grad and array.grad is not None:
+                tensor.grad = torch.from_dlpack(wp.to_dlpack(array.grad))
+                tensor.grad._warp_grad_array = array.grad
+
+            return tensor
+
+        return original_to_torch(array, requires_grad=requires_grad)
+
+    wp.to_torch = to_torch_with_empty_cpu_fallback
+    _EMPTY_CPU_TO_TORCH_WORKAROUND_INSTALLED = True
```
alexmillane, line 59: Is this a know issue to the Lab/warp teams?

I'm wondering if we could:
a) Add a comment, preferably to a link which tracks this issue if one exists.
b) a test that test if this patch is still required. That way when this is no long needed, we'll know about it?

alexmillane (commented):
Thank you for doing this. This is hard work, I know.

I have a few comments. I know that we need to merge this soon, however, I think we should have answers for some of the less nice stuff here, at least that we know where we're headed.

.github/workflows/ci.yml
peterd-NV: The image is built against the public 6.0.1 image. I checked with Ruchik and the public sim 6.1 image is coming Sept 8 so we can upgrade after that. 

My plan was to: push this PR with the images pointing to the one-off tag so other CI works in the meantime aren't broken. Open a new PR to change this back to latest once it's merged and the main branch image build job runs.

isaaclab_arena/environments/arena_env_builder.py
peterd-NV: Currently only a slack thread to try and figure out how to resolve the issue. https://nvidia.slack.com/archives/C06HLQ6CB41/p1788293669009869 

I'll add a TODO and test to catch when this can be removed.

isaaclab_arena/evaluation/policy_runner.py
peterd-NV: Yes, torch and numpy imports need to happen after kit startup now. Doing otherwise will cause a conflict and make kit hang indefinitely.

isaaclab_arena/utils/isaaclab_utils/warp_patch.py
peterd-NV: Same as comment above. Issue is being discussed in slack atm. Will add a todo and test to tell us when this can be removed.

isaaclab_arena/assets/object_base.py
peterd-NV: This was added for cases when __del__ calls it before the attribute was initialized (e.g. constructor failure). It makes sense to remove once the frameviews are removed from objectbase and into Clemen's "Arena World" but should still be here for the purpose of this PR with the object_base attached FrameViews.

isaaclab_arena/assets/object_base.py
peterd-NV: What happened in Lab 3.0 is roughly:

- Lab got rid of their FrameViews for AssetBase
- We are now creating the FrameViews manually for AssetBase
- The caching here is to avoid having to rebuild a FrameView every time get_object_pose is called
- The middle part where it iterates over envs is a guard against a condition that codex flagged. The FrameView is not guaranteed to be created in env order (e.g. env0, env1, env2). Arena expects the poses to be in that order so the check verifies that. It currently runs once during FrameView creation and not during every call to pose getter.

The comments from @cvolkcvolk on moving this out of object base and into an "arena world" object will mean we can remove all of this.

isaaclab_arena/assets/object_base.py
peterd-NV: See comment above: this can all be removed if we refactor the FrameView into an arena world object. For now, the caching is there to prevent needing to create a new Frameview on every call.

isaaclab_arena/cli/isaaclab_arena_cli.py
peterd-NV: Lab removed it and made headless default if no `--viz` argument is supplied. We can remove it too to be the same and just update any existing documented commands.

isaaclab_arena/cli/isaaclab_arena_cli.py
peterd-NV: We have a few more uses for this flag than lab had. In Arena we are using it to tell whether to add embodiment cameras and camera obs. Emboidments use this flag to know whether or not it needs to construct cameras. Arena also uses it to tell applauncher to enable camera rendering. 

Lab was able to get rid of it because they build the env config first, check if there are camera cfgs, then launches sim with the correct rendering settings. 

Arena builds the env after sim launch so we can't automatically parse the correct rendering settings before sim starts.

isaaclab_arena/cli/isaaclab_arena_cli.py
peterd-NV: Removed

isaaclab_arena/assets/object_base.py
peterd-NV: Removed now that are also destructors removed.

isaaclab_arena/assets/object_base.py
peterd-NV: Follow up: FrameView will be created and cached in object base for this upgrade PR. Will be moved to Arena world with Clemens' follow up changes.

isaaclab_arena/assets/object_base.py
peterd-NV: From slack thread: will keep for now (the check only runs on creation and not every time), will be removed with Arena world

isaaclab_arena/environments/arena_env_builder.py
peterd-NV: Added todo and test. Issue will be fixed in warp 1.18

isaaclab_arena/tests/test_object_on_destination_term.py
peterd-NV: Following the discussion, this has been removed and is unnecessary.

isaaclab_arena/utils/isaaclab_utils/warp_patch.py
peterd-NV: Added todo to remove and test to catch when it's fixed. Warp team is aware of issue and will be fix in warp 1.18

isaaclab_arena/tests/test_kuka_allegro_embodiment.py
peterd-NV: These came from when the original kuka/dexsuite was added. Removed them as I think we no longer need them.

PR #1187, opened 2026-09-06
Centralize object pose queries

cvolkcvolk:
## Summary

Route ObjectBase pose reads through ArenaWorld and removes ObjectBase's FrameView workaround

## Detailed description

- Adds live articulation root-link pose support to ArenaWorld.
- Keeps FrameView access in the scene-extra pose reader, where it is shared by all callers.

### DIFF
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -7,13 +7,12 @@ from __future__ import annotations
 
 import torch
 from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
 
-import warp as wp
 from isaaclab.assets import ArticulationCfg, AssetBaseCfg, RigidObjectCfg
 from isaaclab.envs import ManagerBasedEnv
 from isaaclab.managers import EventTermCfg, SceneEntityCfg
 from isaaclab.sensors.contact_sensor.contact_sensor_cfg import ContactSensorCfg
-from isaaclab.sim.views import FrameView
 from isaaclab_tasks.contrib.stack.mdp.franka_stack_events import randomize_object_pose
 
 # Re-export ObjectType from the lightweight module so existing
@@ -26,6 +25,9 @@ from isaaclab_arena.terms.events import set_object_pose, set_object_pose_per_env
 from isaaclab_arena.utils.pose import Pose, PosePerEnv, PoseRange
 from isaaclab_arena.utils.velocity import Velocity
 from isaaclab_arena.variations.object_mass_variation import ObjectMassVariation
+
+if TYPE_CHECKING:
+    from isaaclab_arena.environments.isaaclab_arena_manager_based_env import IsaacLabArenaManagerBasedRLEnv
 
 __all__ = [
     "ObjectBase",
@@ -52,16 +54,6 @@ class ObjectBase(PlaceableAsset, ABC):
             self.add_variation(ObjectMassVariation(self.name))
         self.initial_velocity: Velocity | None = None
         self.object_cfg = None
-        self._base_frame_view: FrameView | None = None
-        self._base_frame_view_stage = None
-
-    def _close_base_frame_view(self) -> None:
-        """Release the cached frame view. Safe to call more than once."""
-        frame_view = self._base_frame_view
-        if frame_view is not None:
-            frame_view.close()
-            self._base_frame_view = None
-            self._base_frame_view_stage = None
 
     def _set_initial_pose(self, pose: Pose | PoseRange | PosePerEnv) -> None:
         """Store the pose and write its construction values into the object config."""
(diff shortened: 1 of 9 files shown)


isaaclab_arena/assets/object_base.py
```diff
@@ -170,36 +162,9 @@ def get_object_pose(self, env: ManagerBasedEnv, is_relative: bool = True) -> tor
...
-                except Exception:
-                    frame_view.close()
-                    raise
-                self._base_frame_view = frame_view
-                self._base_frame_view_stage = stage
-            position_w, orientation_w = self._base_frame_view.get_world_poses()
-            object_pose = torch.cat((position_w.torch, orientation_w.torch), dim=-1)
-        else:
+        if self.object_type not in (ObjectType.RIGID, ObjectType.ARTICULATION, ObjectType.BASE):
             raise ValueError(f"Function not implemented for object type: {self.object_type}")
```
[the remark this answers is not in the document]
cvolkcvolk: Agree. Dropped.

isaaclab_arena/tests/test_arena_world_scene_access.py
```diff
@@ -124,6 +125,35 @@ def compute_geometry_bounds(_scene, scene_key: str):
     assert geometry_build_calls == ["object", "destination"]
 
 
+def _check_articulation_pose_reads(arena_world_module) -> None:
+    """Check that articulation root poses are read live."""
+    import torch
+
+    class RuntimeBufferDouble:
+        def __init__(self, tensor: torch.Tensor):
+            self.torch = tensor
```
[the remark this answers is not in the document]
cvolkcvolk: Done

isaaclab_arena/assets/object_base.py
```diff
@@ -52,16 +50,6 @@ def __init__(
             self.add_variation(ObjectMassVariation(self.name))
         self.initial_velocity: Velocity | None = None
         self.object_cfg = None
-        self._base_frame_view: FrameView | None = None
```
qianl-nv, line 55: nice!

isaaclab_arena/tests/test_events.py
```diff
@@ -88,7 +88,7 @@ def _test_set_object_pose_per_env_event(simulation_app):
 
         # Check the pose right after reset, before the box is stepped: the reset event is the unit under
         # test, so we verify each env received its assigned pose rather than where physics carries it.
-        cracker_box_poses = cracker_box.get_object_pose(env)
+        cracker_box_poses = env.unwrapped.arena_world.get_pose_e(cracker_box.name)
```
qianl-nv, line 91: can we add this wrapper back into ObjectBase so caller like this cracker_box.get_pose_w(env) still works? env.unwrapped.arena_world.get_pose_e(cracker_box.name) is a bit tedious to put everywhere. 

Although at this point only unit tests access are affected, and it's also arguable that we don't want to add an extra API interface to ObjectBase that's only used in unit test, unless we plan to use this new API in production.

WDYT?

```
def get_pose_w(self, env: ManagerBasedEnv) -> torch.Tensor:
    """Return T_W_F for this scene entry, shape (num_envs, 7)."""
    return env.unwrapped.arena_world.get_pose_w(self.name)

def get_pose_e(self, env: ManagerBasedEnv) -> torch.Tensor:
    """Return T_E_F for this scene entry, shape (num_envs, 7)."""
    return env.unwrapped.arena_world.get_pose_e(self.name)
```

qianl-nv (approved):
Thanks for cleaning this up! One optional comment

isaaclab_arena/tests/test_events.py
cvolkcvolk: Good question and you are right that the direct runtime expression is verbose and `cracker_box.get_pose_e(env)` looks nice...

Here is the catch: The bigger picture is that we use ArenaWorld as the only entry point for live environment state lookup while our Arena Objects stay build time configs only.
Eg. ObjectBase describes what gets added to the scene, while ArenaWorld provides access to the live simulation state (After Arena compilation, the live object is actually a Isaac Lab scene entry and not an ObjectBase anymore).
Given that assumption of Arena Objects being build time configs, cracker_box.get_pose_e(env) would introduce a second runtime query API on the configuration object... which is suboptimal imo

isaaclab_arena/assets/object_base.py
```diff
@@ -157,53 +145,6 @@ def _init_object_cfg(self) -> RigidObjectCfg | ArticulationCfg | AssetBaseCfg:
...
-                self._base_frame_view = frame_view
-                self._base_frame_view_stage = stage
-            position_w, orientation_w = self._base_frame_view.get_world_poses()
-            object_pose = torch.cat((position_w.torch, orientation_w.torch), dim=-1)
-        else:
-            raise ValueError(f"Function not implemented for object type: {self.object_type}")
-        if is_relative:
-            object_pose[:, :3] -= env.unwrapped.scene.env_origins
-        return object_pose
-
```
alexmillane, line 206: <3

alexmillane (commented):
LGTM.

I notice some deletions of manual pose reads. I'm thinking that there are more pose reads we could route through the Arena world? Maybe I'm wrong though.

PR #1168, opened 2026-09-01
Force CPU to fix rendering after rebuild bug

alexmillane:
## Summary
Force CPU to fix rendering after rebuild bug

## Detailed description
- After stage reset, some object pases used by the renderer could differ from the poses used by the physics, leading to parts of the robot rendering in the wrong locations. 
- This MR contains a workaround that forces the system to run on the CPU with fabrics disabled in cases where a stage rebuild is required.
- Addresses: [6657464](https://nvbugs.nvidia.com/6657464) and [6628837](https://nvbugs.nvidia.com/6628837)

## Images

Before:

https://github.com/user-attachments/assets/dabca106-da1c-4c6d-8f15-aa35ef72cce8

After:

https://github.com/user-attachments/assets/5c33234d-4ec9-4a66-b833-e35a032a3f9f

PR #1192, opened 2026-09-07
Reactivate fabics after they were fixed in Isaac Sim 6.1

alexmillane:
## Summary
Reactivate fabrics in workflows requiring stage rebuilds.

## Detailed description
- We disabled fabrics in #1168 due to bugs in Isaac Sim 6.0.
- With the upgrade to Isaac Sim 6.1 we can re-enable fabrics and re-enable the corresponding tests.


qianl-nv (approved):
Thanks for the clean up!

cvolkcvolk:
Thanks!

PR #1193, opened 2026-09-08
Align agentic environment warnings

sangeetas-nv:
## Summary

Align agentic environment warnings

## Detailed description

The concept and workflow pages currently show different warnings for agentic environment generation. This change standardizes the warning so both pages display the same guidance.

Documentation-only change with no runtime impact.


qianl-nv (approved):
Thanks for the clean up!

PR #1199, opened 2026-09-08
Fix procedural cube support in graph-spec pick-and-place

qianl-nv:
## Summary
Fix procedural cube in graph-spec pick-and-place

## Detailed description
- Graph-spec pick-and-place tasks failed when a scene used `procedural_cube` because the asset had no USD path for lighting, placement, and contact-sensor setup.
- Procedural cube/table now use `spawner_cfg`; `Object` enables contact reporters on spawner-only rigid bodies and skips empty USD probes.
- `rubiks_cube_bowl.yaml` now references `procedural_cube` instead of the Hot3D rubiks cube asset.

### DIFF
--- a/isaaclab_arena/assets/object.py
+++ b/isaaclab_arena/assets/object.py
@@ -71,10 +71,7 @@ class Object(ObjectBase):
         return self.bounding_box
 
     def get_corners(self, pos: torch.Tensor) -> torch.Tensor:
-        assert self.usd_path is not None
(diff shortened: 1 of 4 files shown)


qianl-nv:
[Exclude procedural objects from agent asset catalogue](https://github.com/isaac-sim/IsaacLab-Arena/pull/1200) merged this alternative solution.

PR #1204, opened 2026-09-08
Add cable asset class

peterd-NV:
## Summary
Isaac Lab introduced a dedicated cable asset in 3.0 EA, this PR adds a native cable asset to Arena using Lab's new cable API. This change supports the Isaac CAP cable routing env. 

Cables in Lab are procedurally generated and not rigid bodies or articulations. Thus the Cable arena asset is not an object base which has API assumptions for rigid/articulations. Cables are also not supported by the Arena placer in it's current form. Changes to the placer required for working with segment based assets will be left for the future. 

In this PR:

1. Adds Cable asset class
2. Updates Arena's usd export function to work with Lab cable assets (cables are not from a USD file)
3. Adds tests for cpu/gpu cable spawning and export


isaaclab_arena/tests/test_cable_asset.py
```diff
@@ -0,0 +1,31 @@
...
+        initial_pose=initial_pose,
+    )
+
+    scene_key, cable_cfg = cable.get_object_cfg()
+    assert scene_key == "test_cable"
+    assert isinstance(cable_cfg, CableObjectCfg)
+    assert cable_cfg.prim_path == "{ENV_REGEX_NS}/Cable"
+    assert cable_cfg.init_state.pos == initial_pose.position_xyz
+    assert cable_cfg.init_state.rot == initial_pose.rotation_xyzw
+    assert cable.get_event_cfg() == ("test_cable", None)
```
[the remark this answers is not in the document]
peterd-NV: runtime test added in later commit

isaaclab_arena/assets/cable.py
```diff
@@ -0,0 +1,47 @@
...
+        self.object_cfg = CableObjectCfg(prim_path=prim_path, spawn=spawn)
+        if initial_pose is not None:
+            self.object_cfg.init_state.pos = initial_pose.position_xyz
+            self.object_cfg.init_state.rot = initial_pose.rotation_xyzw
+
+    def get_object_cfg(self) -> tuple[str, CableObjectCfg]:
+        """Return the scene key and Isaac Lab cable configuration."""
+        return self.name, self.object_cfg
+
+    def get_event_cfg(self) -> tuple[str, None]:
```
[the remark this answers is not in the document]
peterd-NV: resolved

isaaclab_arena/tests/test_cable_asset.py
```diff
@@ -0,0 +1,31 @@
...
+        ),
+        initial_pose=initial_pose,
+    )
+
+    scene_key, cable_cfg = cable.get_object_cfg()
+    assert scene_key == "test_cable"
+    assert isinstance(cable_cfg, CableObjectCfg)
+    assert cable_cfg.prim_path == "{ENV_REGEX_NS}/Cable"
+    assert cable_cfg.init_state.pos == initial_pose.position_xyz
+    assert cable_cfg.init_state.rot == initial_pose.rotation_xyzw
```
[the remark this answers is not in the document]
peterd-NV: added

isaaclab_arena/assets/cable.py
```diff
@@ -0,0 +1,47 @@
...
+from __future__ import annotations
+
+from isaaclab.assets import CableObjectCfg
+from isaaclab.sim.spawners.spawner_cfg import SpawnerCfg
+
+from isaaclab_arena.assets.asset import Asset
+from isaaclab_arena.utils.pose import Pose
+
+
+class Cable(Asset):
```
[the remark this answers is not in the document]
peterd-NV: Cables should not be objectbase as that assumes rigid/articulation root state APIs. Cables are also not automatically placeable just by deriving from PlaceableAsset due to missing ability to write to root pose. Proper relation support requires extending the placement/reset in Arena to support segment based assets which is out of scope for this PR.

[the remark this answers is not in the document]
peterd-NV: fixed

isaaclab_arena/assets/cable.py
```diff
@@ -0,0 +1,47 @@
...
+    """A procedurally spawned Isaac Lab cable managed as an Arena scene asset."""
```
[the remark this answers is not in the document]
peterd-NV: added note

isaaclab_arena/assets/cable.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __init__(
+        self,
+        name: str,
+        prim_path: str,
+        spawn: SpawnerCfg,
```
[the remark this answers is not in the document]
peterd-NV: fixed

isaaclab_arena/assets/cable.py
```diff
@@ -0,0 +1,58 @@
...
+from isaaclab_arena.assets.asset import Asset
+from isaaclab_arena.terms.events import reset_cable_to_default
+from isaaclab_arena.utils.pose import Pose
+
+
+class Cable(Asset):
+    """A procedurally spawned Isaac Lab cable managed as an Arena scene asset.
+
+    Cable simulation requires the Newton physics backend. Build environments containing this asset
+    with ``presets="newton"``.
```
xyao-nv: Can this be enforced at code level thru asseting somewhere?

isaaclab_arena/assets/cable.py
```diff
@@ -0,0 +1,58 @@
...
+from isaaclab.assets import CableObjectCfg
+from isaaclab.managers import EventTermCfg, SceneEntityCfg
+from isaaclab.sim.spawners.shapes import CableCfg
+
+from isaaclab_arena.assets.asset import Asset
+from isaaclab_arena.terms.events import reset_cable_to_default
+from isaaclab_arena.utils.pose import Pose
+
+
+class Cable(Asset):
```
xyao-nv: hmm, so cable is not an ObjectBase and does not enjoy the benefits from ObjectBase class anymore. E.g. agentic env gen & object placement all need to do special machineries to support this cable.

Can we minimize the changes to other modules by Cable? Technically it's still a foreground/task-relevant "Object". How hard is it to add a new object type "cable" apart from, rigid, base, articulation in ObjectBase type?

peterd-NV: Refactored to be ObjectBase with unsupported methods raising explicit errors.

isaaclab_arena/assets/cable.py
peterd-NV: Added a `validate_simulation_cfg` feature to assets which is used to validate that the cable has the right config. This new validation is included for all other assets but defaults to a no-op unless overridden.

isaaclab_arena/assets/cable.py
```diff
@@ -0,0 +1,106 @@
...
+
+        physics_cfg = sim_cfg.physics
+        assert isinstance(physics_cfg, NewtonCfg), (
+            f"Cable asset '{self.name}' requires the Newton physics backend; "
+            f"got {type(physics_cfg).__name__ if physics_cfg is not None else 'the default PhysX backend'}."
+        )
+        pending_solver_cfgs = [physics_cfg.solver_cfg]
+        has_vbd_solver = False
+        while pending_solver_cfgs:
+            solver_cfg = pending_solver_cfgs.pop()
```
xyao-nv: Does it need to tie the solver to this asset? Could it be two sets of VBDsolver for cable & deformables in the env?

isaaclab_arena/assets/cable.py
```diff
@@ -0,0 +1,106 @@
...
+        from isaaclab_newton.physics import NewtonCfg, VBDSolverCfg
+
+        physics_cfg = sim_cfg.physics
+        assert isinstance(physics_cfg, NewtonCfg), (
+            f"Cable asset '{self.name}' requires the Newton physics backend; "
+            f"got {type(physics_cfg).__name__ if physics_cfg is not None else 'the default PhysX backend'}."
+        )
+        pending_solver_cfgs = [physics_cfg.solver_cfg]
+        has_vbd_solver = False
+        while pending_solver_cfgs:
```
qianl-nv: this can go into a common util function
physics_cfg_has_solver(phsics_cfg, VBDSolverCfg)
we might need this for other assets soon

qianl-nv (approved):
Looks much cleaner, and thanks for adding the validate_simulation_cfg to Asset!

isaaclab_arena/assets/cable.py
qianl-nv: According to my agent, nested couplers are explicitly rejected by Isaac Lab, so recursion only adds complexity and could temporarily accept an invalid nested configuration. So the recursion here is a kinda unnecessary. 

A direct check of the top-level solver plus its immediate entries is sufficient.

xyao-nv (approved):

isaaclab_arena/assets/cable.py
peterd-NV: resolved in slack: Keeping this as only physics backend check and removing vdb specific check as the machinery to do this should not live in the cable asset class

isaaclab_arena/assets/cable.py
peterd-NV: resolved in slack: Keeping this as only physics backend check and removing vdb specific check as the machinery to do this should not live in the cable asset class. A newton env may have mulitple VDB configs so checking like this is not fully correct.

PR #1215, opened 2026-09-09
(release/0.3.0) Release viewport capture before stage teardown when Replicator is included

xyao-nv:
## Summary

## Detailed description
- Issue reported by VDR: When omni.Replicator is included, it hangs during sim shutdown after video is written, html does not output.
Fixes: 
- Stop video recording cleanly before shutting down the simulator.
- Prevent completed evaluations from hanging before the HTML report is written

### Before
MP4 is written, process remains around “Closing simulation app,” and timeout eventually returns 124; index.html is absent.

### After
Process exits 0, and both the MP4 and index.html are present.

```
timeout --signal=TERM --kill-after=30s 180s \
python isaaclab_arena/evaluation/policy_runner.py \
--viz kit --policy_type zero_action \
--record_viewport_video \
--num_episodes 20 \ 
lift_object
```

### DIFF
--- a/isaaclab_arena/environments/isaaclab_arena_manager_based_env.py
+++ b/isaaclab_arena/environments/isaaclab_arena_manager_based_env.py
@@ -15,6 +15,7 @@ from isaaclab_arena.metrics.metrics_manager import MetricsManager
 from isaaclab_arena.recording.episode_recorder_manager import EpisodeRecorderManager
 from isaaclab_arena.tasks.predicates.object_settling import ObjectInitialRestPoseRecorder
 from isaaclab_arena.variations.variation_recorder import VariationRecorder
+from isaaclab_arena.video.video_recording import close_viewport_video_recorder
 
 
 class IsaacLabArenaManagerBasedRLEnv(ManagerBasedRLEnv):
@@ -95,3 +96,10 @@ class IsaacLabArenaManagerBasedRLEnv(ManagerBasedRLEnv):
             A MetricsDataCollection instance.
         """
         return self.metrics_manager.compute()
+
+    def close(self) -> None:
+        """Release viewport capture before closing the simulation environment."""
+        try:
+            close_viewport_video_recorder(getattr(self, "video_recorder", None))
+        finally:
+            super().close()
--- a/isaaclab_arena/evaluation/resource_cleanup.py
+++ b/isaaclab_arena/evaluation/resource_cleanup.py
@@ -13,6 +13,7 @@ from isaaclab_arena.utils.isaaclab_utils.simulation_app import (
     collect_garbage_and_clear_cuda_cache,
     teardown_simulation_app,
 )
+from isaaclab_arena.video.video_recording import close_viewport_video_recorder
 
 if TYPE_CHECKING:
     import gymnasium as gym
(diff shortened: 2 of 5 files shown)


isaaclab_arena/environments/isaaclab_arena_manager_based_env.py
cvolkcvolk, line 1: The ideal fix looks like would belong to IsaacLab where `ManagerBasedEnv` owns the `VideoRecorder`. So `VideoRecorder.close()` should release its capture, and `ManagerBasedEnv.close()` should call it... Shall we add a TODO() to eventually clean this up properly in Lab?

isaaclab_arena/environments/isaaclab_arena_manager_based_env.py
```diff
@@ -95,3 +96,10 @@ def compute_metrics(self) -> MetricsDataCollection:
             A MetricsDataCollection instance.
         """
         return self.metrics_manager.compute()
+
+    def close(self) -> None:
+        """Release viewport capture before closing the simulation environment."""
+        try:
+            close_viewport_video_recorder(getattr(self, "video_recorder", None))
```
cvolkcvolk, line 103: We should have access to self.recorder here directly:
`close_viewport_video_recorder(self.video_recorder)`

If Isaac Lab removes or renames the attribute, the code fails visibly instead of silently skipping the cleanup

isaaclab_arena/video/video_recording.py
```diff
@@ -40,6 +41,48 @@ def render_mode(self) -> str | None:
         return "rgb_array" if self.record_viewport_video else None
 
 
+def close_viewport_video_recorder(video_recorder) -> None:
```
cvolkcvolk, line 44: Second on the use of `getattr()`... we could use the known fields directly:

```
 def close_viewport_video_recorder(
      video_recorder: VideoRecorder | None,
  ) -> None:
      """Release the Kit viewport capture used by the pinned Isaac Lab version."""
      if video_recorder is None or video_recorder._backend != "kit":
          return

      capture = video_recorder._capture
      if capture is None:
          return

      annotator = capture._rgb_annotator
      render_product = capture._render_product
```

isaaclab_arena/evaluation/resource_cleanup.py
```diff
@@ -30,9 +31,11 @@ def close_policy(policy: PolicyBase | None) -> None:
 
 
 def close_environment(env: gym.Env | None) -> None:
```
cvolkcvolk, line 33: Is this required explicetly to call close_viewport_video_recorder() here seperately? 

Suggestion to keep cleanup only in `IsaacLabArenaManagerBasedRLEnv.close() `, then the below call of `env.close()` in
```
    finally:
        try:
            env.close()
        finally:
            collect_garbage_and_clear_cuda_cache()
```
would triggered to close it

cvolkcvolk (approved):
Thanks for addressing this! 
I wasn't able to reproduce the shutdown hang locally, nor find Replicator to be the root cause. Left a few comments. Ok from my side to unblock for now but I think the root cause should be fixed in IsaacLab

PR #1218, opened 2026-09-09
Separate rooted object behavior from ObjectBase

qianl-nv:
## Summary
Split generic and rooted object behavior

## Detailed description
- Keep shared scene-object state and configuration access on `ObjectBase`.
- Move rigid, articulated, and static-root behavior into `RootedObjectBase`.
- Update object classes and callers to import `ObjectType` from its lightweight module.

### DIFF
--- a/isaaclab_arena/assets/object_base.py
+++ b/isaaclab_arena/assets/object_base.py
@@ -59,8 +59,19 @@ class ObjectBase(PlaceableAsset, ABC):
 class RootedObjectBase(ObjectBase):
     """Parent class for rigid, articulated, and static rooted objects."""
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(
+        self,
+        name: str,
+        prim_path: str | None = None,
+        object_type: ObjectType = ObjectType.BASE,
+        **kwargs,
+    ):
+        super().__init__(name=name, prim_path=prim_path, object_type=object_type, **kwargs)
+        assert self.object_type in {
+            ObjectType.BASE,
+            ObjectType.RIGID,
+            ObjectType.ARTICULATION,
+        }, f"RootedObjectBase does not support object type '{self.object_type}'."
         if self.object_type == ObjectType.RIGID:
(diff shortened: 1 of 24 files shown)


isaaclab_arena/assets/object_base.py
```diff
@@ -46,10 +37,30 @@ def __init__(
...
+
+    def get_object_cfg(self) -> tuple[str, AssetBaseCfg]:
+        """Return the scene key and concrete asset config."""
+        return self.name, self.object_cfg
+
+    def get_event_cfg(self) -> tuple[str, EventTermCfg | None]:
+        return self.name, self._pose_event_cfg
+
+
+class RootedObjectBase(ObjectBase):
```
xyao-nv, line 59: Minor.
`RootedObjectBase` sounds ambiguous because `ObjectReference` may refer to a nested prim, and deformables also have USD roots. How about `RootStateObjectBase`?

isaaclab_arena/assets/object_base.py
```diff
@@ -46,10 +37,30 @@ def __init__(
...
+    """Parent class for rigid, articulated, and static rooted objects."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
         if self.object_type == ObjectType.RIGID:
```
xyao-nv, line 75: Can we add
```
  assert self.object_type in {
      ObjectType.BASE,
      ObjectType.RIGID,
      ObjectType.ARTICULATION,
  }
```

xyao-nv (approved):

peterd-NV (approved):
Looks great, thanks for adding

isaaclab_arena/assets/object_base.py
qianl-nv: I think we are use 'root' for different things here.
By "Rooted" I meant that the transform has a well defined root point which is fixed to the object. For deformable this is not true. In this context rooted vs deformable makes sense.
I think you are using 'root' in the USD prim structure's sense.

PR #1198, opened 2026-09-08
Add timing to rollout.

alexmillane:
## Summary
Add timing to rollouts.

## Detailed description
- **Why**
  - We'd like to time evaluation rollouts as the code evolves over time.
- **What**
  - Use our built in `Timer` class, and
  - Wrap the largest blocks in the rollout section.
  - Output per-env timing results to block.

PR #1220, opened 2026-09-09
Plot where a rollout step's time goes

alexmillane:
## Summary
Plots the rollout timings that #1198 records.

## Detailed description
- We can read the timings JSON but had nothing that turns it into a picture, so comparing Runs meant reading numbers by hand.
- Adds `isaaclab_arena/visualization/plot_timings.py`, which draws one stacked bar per Run splitting the mean step into the timers nested below it, plus a median-to-p90 spread panel when percentiles survived.
- Walks the timer name hierarchy instead of a fixed list of names, so a timer added to the rollout shows up as a slice without touching the script; each parent's unattributed time becomes its own "(other)" slice so the slices add up to the mean step.
- Reads both shapes Arena writes: the Experiment Runner's own file and the Experiment file the OSMO collect task merges.


isaaclab_arena/tests/test_plot_timings.py
```diff
@@ -0,0 +1,174 @@
...
+def test_reads_the_collected_shape_and_groups_by_run_name(tmp_path):
+    timings_path = _write_timings(tmp_path / "timings.json", _nested_records(), run_names=["first", "second"])
+
+    records_by_run_name = read_run_timer_records(timings_path)
+
+    assert sorted(records_by_run_name) == ["first", "second"]
+    assert len(records_by_run_name["first"]) == 5
+
+
+def test_splits_the_root_timer_into_children_then_its_own_leftover():
```
alexmillane: Delete all the test below here as superfluous.

isaaclab_arena/visualization/plot_timings.py
```diff
@@ -0,0 +1,378 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Plot where a rollout step's time goes, from an Arena Experiment timings file.
+
+Timer names are hierarchical: a timer entered inside another records under ``<enclosing>/<name>``.
+This script follows that tree, so each bar is one Run's mean step split into the timers nested
+below it, and the time a timer does not attribute to its children becomes its own "(other)" slice.
+Adding a timer to the rollout therefore adds a slice here without changing this script.
+"""
```
alexmillane: Reduce docstring to the first level.

isaaclab_arena/visualization/plot_timings.py
```diff
@@ -0,0 +1,378 @@
...
+
+matplotlib.use("Agg")
+import matplotlib.pyplot as plt  # noqa: E402
+
+TIMER_NAME_SEPARATOR = "/"
+DEFAULT_ROOT_TIMER_NAME = "step"
+
+# Time a parent timer does not attribute to its children is allowed to go this far below zero
+# before the file is treated as inconsistent, which absorbs float error in the recorded totals.
+_UNATTRIBUTED_MS_TOLERANCE = 1e-6
```
alexmillane: I think we should remove this check. Our timers are not guarunteed that child timers add up to their parents.

isaaclab_arena/visualization/plot_timings.py
```diff
@@ -0,0 +1,378 @@
...
+        raise ValueError(f"Timings file '{timings_path}' must hold a list or an object")
+
+    records_by_run_name: dict[str, list[dict]] = defaultdict(list)
+    for record in records:
+        if "name" not in record or "total_ms" not in record:
+            raise ValueError(f"Timings file '{timings_path}' has a record without a name and total")
+        records_by_run_name[record.get("run_name", timings_path.parent.name)].append(record)
+    if not records_by_run_name:
+        raise ValueError(f"Timings file '{timings_path}' holds no timer records")
+    return dict(records_by_run_name)
```
alexmillane, line 101: Am I correct that this difference is due to the OSMO experiment and the local experiment runner producing files of different types? Could we instead align them to produce the same thing?

isaaclab_arena/visualization/plot_timings.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+def _child_timer_names(parent_name: str, timer_names: Sequence[str]) -> list[str]:
+    """Return the names nested exactly one level below the parent, in recorded order."""
+    prefix = parent_name + TIMER_NAME_SEPARATOR
+    return [
+        timer_name
+        for timer_name in timer_names
+        if timer_name.startswith(prefix) and TIMER_NAME_SEPARATOR not in timer_name[len(prefix) :]
+    ]
```
alexmillane: Break this up over multiple lines with intermediate variables to make it more readable.

alexmillane (commented):
Self review 1

isaaclab_arena/visualization/plot_timings.py
alexmillane: Done — module docstring is now just "Plot where a rollout step's time goes, from an Arena Experiment timings file."

isaaclab_arena/visualization/plot_timings.py
alexmillane: Done, and good catch — nested timers genuinely aren't guaranteed to sum to their parent (a child can be entered from more than one path). Removed `_UNATTRIBUTED_MS_TOLERANCE` and the ValueError; the leftover is just clamped at zero now.

isaaclab_arena/visualization/plot_timings.py
alexmillane: Not quite — the two routes already produce the same thing, as of #1198. Both the local Experiment Runner and the OSMO collect task write the identical layout:

```
<experiment-output>/arena_experiment_timings.json          {"totals": [...], "runs": [...]}
<experiment-output>/<run-name>/arena_experiment_timings.json   bare list, one per Run
```

So the two shapes this handles aren't OSMO-vs-local, they're **per-Run file vs combined Experiment file** — and both routes emit both. The combined file is the one you'd normally plot; per-Run support just lets you point at a single Run's file directly. Happy to drop per-Run support and accept only the combined file if you'd rather have the simpler reader — it's about 8 lines.

isaaclab_arena/visualization/plot_timings.py
alexmillane: Done — now an explicit loop with `prefix` and `name_below_parent` rather than the one-line comprehension.

isaaclab_arena/tests/test_plot_timings.py
alexmillane: Done — deleted everything from `test_splits_the_root_timer_into_children_then_its_own_leftover` down, leaving the two file-reading tests. Worth flagging that this also removed the only end-to-end test of `main()` and the one covering plotting without percentiles (the merged-totals case, where `merge_timer_stats_json` drops them); say the word if you want either kept.

PR #1196, opened 2026-09-08
Fix camera-extrinsics variation under Newton

alexmillane:
## Summary
Fix camera-extrinsics variation under Newton

## Detailed description
- **why**
  - camera extrinsics variations are broken in Newton.
  - writes to the camera pose do not take effect when that backend is selected.
  - The source of the issue is that `camera.set_world_poses()` does not take effect with newton on.
    - only the physics representation of the pose is writen, not the representation used for rendering.
- **What**
  - Adds `CameraLocalOffsetWriter` that correctly updates the pose while we wait for a fix in Lab.
  - Adds tests that capture the effect (they fail without this MR and pass with it).

PR #1236, opened 2026-09-09
Fix camera extrinsics variation for Newton

peterd-NV:
## Summary
Fix camera extrinsics variation so RTX-rendered cameras use the sampled poses with Newton.

- Synchronize each rendered USD camera prim with the Isaac Lab camera view.
- Refresh affected cameras after applying extrinsics variation.
- Add test coverage for PhysX and single- and multi-environment Newton.

Before, CameraExtrinsicsVariation updated only Isaac Lab’s internal camera view. With Newton,
that pose change did not propagate to the USD camera prim. As a result, camera pose queries
showed the sampled extrinsics offset, but RGB observations were still rendered from the original
pose. This made camera randomization ineffective.


alexmillane:
Already done in #1196

PR #1237, opened 2026-09-10
Upgrade Isaac Sim dependencies to 6.1

qianl-nv:
## Summary
Upgrade Arena to Isaac Sim 6.1

## Detailed description
- Use the public Isaac Sim 6.1 NGC base image and the refreshed `latest` CI image.
- Override the pinned IsaacLab source dependency with the 6.1 pip wheel stack without modifying the submodule.
- Refresh the uv lock for 6.1 dependencies, including `mujoco-usd-converter==0.5.0`.


alexmillane (approved):
Back into the fold! Thank you.

PR #1238, opened 2026-09-10
Instrument the rollout with nested timers (fresh branch of #1198)

alexmillane:
## Summary
Same commits as #1198, on a fresh branch to rule out branch-scoped CI state.

## Detailed description
- `Run tests with subprocess` has failed repeatedly on `alex/feature/timing_2`, always with an Isaac Sim startup crash (`appState = 'startup'`) that then stalls the crash reporter for the full 900s subprocess timeout.
- The same failure hits `main` and nine other branches (23 of 24 readable subprocess-job failures in the last week carry that signature), and 40 local runs of the most-affected test plus repeated local Phase 3 runs have not reproduced any crash.
- This branch points at the identical commit as #1198 (`0517d499c`, which already merges current `main`), so any difference in CI outcome is attributable to per-branch runner state — for example the `actions/cache` scope used by the *Setup Isaac Sim kit cache* step — rather than to the code.
- Draft, and not for merge: #1198 remains the PR under review. Close this once the question is settled.

### DIFF
--- a/isaaclab_arena/evaluation/arena_experiment_result.py
+++ b/isaaclab_arena/evaluation/arena_experiment_result.py
@@ -23,6 +23,7 @@ if TYPE_CHECKING:
     from isaaclab_arena.evaluation.arena_run import ArenaRunCfg
 
 ARENA_EXPERIMENT_RESULT_FILENAME = "arena_experiment_result.json"
+ARENA_EXPERIMENT_TIMINGS_FILENAME = "arena_experiment_timings.json"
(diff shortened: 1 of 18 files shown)


alexmillane:
Closing: this was a control experiment to test whether branch-scoped CI state explained the subprocess failures. It didn't — the cause was a module-scope `import torch` in timer.py loading before AppLauncher, fixed in #1198 (merged as dd91b2006).

PR #1242, opened 2026-09-10
Add clutter placement

zhx06:
## Summary
Add offline physics-settled clutter placement

## Detailed description
- Add a standalone example to settle ordinary Arena assets, preserving base rotations and generating independent layouts.
- Reject invalid layouts and save one exact-pose scene YAML per generated environment.
- Restore cached object and robot poses on reset without changing the relation solver.
- Validate output filesystem support before settling and publish caches without overwriting existing files.
- Validation: focused PhysX, graph, and CLI tests; `pre-commit --all-files`. Newton and CAP/Berkeley reproduction remain unverified.


isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
```diff
@@ -165,27 +182,58 @@ def _instantiate_object_reference(
     return _AFFORDANCE_REFERENCE_CLASSES[joint_param_names[0]](**common_kwargs)
 
 
+def _apply_initial_pose(asset: Asset, value: Any) -> None:
+    """Apply a fixed YAML pose at construction and on every reset."""
+    if value is None:
+        return
+    assert isinstance(value, dict), "initial_pose must be a mapping"
```
alexmillane, line 189: Can we correct the type annotation above. At the input we say `Any` but then check that it's `None | dict`

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
```diff
@@ -165,27 +182,58 @@ def _instantiate_object_reference(
...
+        rotation = value.get("rotation_xyzw", default.rotation_xyzw)
+    else:
+        position, rotation = value["position_xyz"], value["rotation_xyzw"]
+    for name, values, size in (("position_xyz", position, 3), ("rotation_xyzw", rotation, 4)):
+        assert isinstance(values, (list, tuple)) and len(values) == size, f"{name} needs {size} numbers"
+        assert all(
+            isinstance(v, Real) and not isinstance(v, bool) and math.isfinite(v) for v in values
+        ), f"{name} must contain finite numbers"
+    assert math.isclose(sum(v * v for v in rotation), 1.0, abs_tol=1e-4), "rotation_xyzw must be a unit quaternion"
+    asset.set_initial_pose(Pose(tuple(float(v) for v in position), tuple(float(v) for v in rotation)))
```
alexmillane, line 205: 99% of this function is extracting a Pose from a dict.

Then we set it in half the final line.

Suggestion to break this function up. The first function should be

```
def _get_pose_from_dict(pose_dict: dict) -> Pose | None:
```

This

```
def _apply_initial_pose(pose: Pose | None):
```

isaaclab_arena_environments/isaac_cap/clutter/cache.py
```diff
@@ -0,0 +1,81 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from isaaclab_arena.assets.asset import Asset
+    from isaaclab_arena.environment_spec.arena_env_graph_spec import ArenaEnvGraphSpec
+    from isaaclab_arena.utils.pose import Pose
+
+
+def scene_with_cached_poses(
+    spec: ArenaEnvGraphSpec, poses: Mapping[str, Pose], assets_by_node_id: Mapping[str, Asset]
```
alexmillane, line 24: I'm interested why the agent wants to use `Mapping`. Do we expect something other than a `dict` here?

isaaclab_arena_environments/isaac_cap/clutter/clutter_scene.yaml
alexmillane, line 1: This is a test/example file correct?

I would suggest that this goes into `isaaclab_arena_environments` or `isaaclab_arena_examples`.

Generally we try to keep examples out of the core framework.

isaaclab_arena_environments/isaac_cap/clutter/cache.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def write_scene_cache(spec: ArenaEnvGraphSpec, path: Path) -> None:
+    """Publish a complete YAML file atomically without overwriting; requires hard-link support."""
+    path.parent.mkdir(parents=True, exist_ok=True)
+    with tempfile.TemporaryDirectory(dir=path.parent, prefix=".clutter-") as directory:
+        temporary = Path(directory) / path.name
+        spec.write_yaml(temporary)
+        # A same-filesystem hard link publishes the complete file without overwriting a raced writer.
+        os.link(temporary, path)
```
alexmillane, line 81: I'm wondering why we have to go through all this effort to write a file? What's wrong with just writing a file normally?

alexmillane (commented):
This is a good start.

However I have a few opinions on the design:
- **Output:** Right now the system outputs a single `environment.yaml` per placement. Generating lots of graphs in the case that we want many placements. I would suggest instead that we introduce a new type of yaml file that is an optional companion to our usual environment files. _If_ such a file is passed to the run-time (or specified in the environment.yaml) poses are taken from this companion file. This file would be a mapping between object names and a list of poses-per-object. With this design we can get 1000s of layouts with just two files. It also separates specification of the environment from the specification of the placement. Wdyt?
- **Settling specification**. Right now we specify which objects are to be settled via command line parameters. Which means that they're not captured in the `environment.yaml`. What do you think about instead adding a new relation-type called `ClutterOn` which is specified in the `environment.yaml` and captures that this object is the target of the clutter placement. This way the placement of the object travels with the environment spec, rather than being captured on a users CLI (and then immediately lost).
- **Before drop initialization:** Right now we use a parallel system to our `ObjectPlacer` to arrange the object prior to dropping them. Could we not implement this inside the object placer. `ClutterOn` could translate to an object being above a support surface, as low as possible, without colliding with other (`ClutterOn`) objects. This would leave to the objects forming a vertical collumn prior to the settling. It also would mean that `ClutterOn` would produce valid results even without running through `generate_clutter_scene.py`. They would just drop at the start of simulation.

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
zhx06: Changed the input annotation from `Any` to `dict | None`

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
zhx06: Code updated as suggested.

isaaclab_arena_environments/isaac_cap/clutter/cache.py
zhx06: Only `dict` is expected. This function is removed but the new one will use `dict`

isaaclab_arena_examples/relations/clutter/clutter_scene.yaml
zhx06: Addressed. This has been moved to` isaaclab_arena_environments/isaac_cap/clutter/`

isaaclab_arena_environments/isaac_cap/clutter/cache.py
zhx06: Addressed. This is removed.

docs/pages/concepts/object_placement/relations.rst
```diff
@@ -73,6 +73,25 @@ Most environments can be described with a small set of relations:
    starting pose; final solving and validation use each relation's actual
    parent.
 
+``ClutterOn(parent)``
+   Defines a clutter release above a fixed ``IsAnchor`` support. ``ObjectPlacer``
```
alexmillane: Why do we need the restriction that the cluttered-on object is an anchor? Perhaps this is coming from the fact that we need to know the placement of the object in order to do the dropping.

This restriction is fine for this MR, but I think we need to relax it. In the syringe task for example, Berkley would like to clutter on objects that are not archors. For example, the syringes on the tray.

I'm thinking for this, we'd need to do two passes. Once to place everything except the clutter objects, then another solve for the clutter objects, followed by the drop.

Maybe we have to do this in a follow up. Sounds like a bigger change.

docs/pages/concepts/object_placement/relations.rst
```diff
@@ -185,6 +204,49 @@ a Boolean value.
...
+   - position_xyz: [-0.1, 0.2, 0.8]
+     rotation_xyzw: [0.0, 0.0, 0.0, 1.0]
+
+Positions are in the local environment frame, in metres. One index selects a
+complete layout across all objects. Environment ``i`` starts at index
+``i % num_layouts`` and cycles on its own resets. Reusing a cache bypasses
+relation solving; it does not run physics settling. All non-anchor objects
+with spatial relations must be included, and object sets are unsupported.
+Assets must expose writable physics roots. Disable pose-changing variations
+and callbacks when exact replay is required.
```
alexmillane, line 247: Could you describe how "cycling" tasks place. I presume that these layout form a queue and as envs finish their episodes they draw from the queue. What happens when the queue is exhausted?

docs/pages/concepts/object_placement/relations.rst
```diff
@@ -185,6 +204,49 @@ a Boolean value.
...
+
+The CAP offline generator in
+``isaaclab_arena_environments/isaac_cap/clutter/generate_clutter_scene.py``
+creates such files from ``ClutterOn`` relations. Its adjacent ``README.rst``
+describes generation controls and validation limits.
```
alexmillane: I think we should move this script out of the CAP repo. This script is needed for any clutter-involving environment right? Not just cap environments. A good question is where to put this. I think it belongs in the core repo. Maybe in a (new) scripts folder.

alexmillane: Another option is to run clutter placement + settling automatically on environments in which `ClutterOn` is used, and a pose companion file is not passed. Actually I think that that might be cleaner, because it means that no additional scripts are required. Our experiment runner and policy runner are effectively the settling scripts.

isaaclab_arena/assets/object_reference.py
```diff
@@ -101,6 +101,11 @@ def get_bounding_box(self) -> AxisAlignedBoundingBox:
                 self._bounding_box = raw_bbox.scaled(self._parent_scale)
         return self._bounding_box
 
+    def get_bounding_box_rotation_xyzw(self) -> tuple[float, float, float, float]:
+        """World orientation of the parent USD axes used by the reference bounds."""
+        parent_pose = self.parent_asset.initial_pose
+        return parent_pose.rotation_xyzw if parent_pose is not None else (0.0, 0.0, 0.0, 1.0)
```
alexmillane: This function seems badly named. It gets the rotation of the parent asset. It doesn't appear to me to linked to the bounding box. Anyway, I feel that getting the rotation of parent is very specific for a function in this core class. What about `get_parent_pose() -> Pose`

alexmillane: Ah looking below I think the real purpose of this function is to handle cases where the pose hasn't been set. Suggestion to implement the function suggested above and then a small variant `get_parent_pose_or_identity() -> Pose | None`

isaaclab_arena/cli/isaaclab_arena_cli.py
```diff
@@ -125,6 +125,12 @@ def add_env_graph_spec_cli_args(parser: argparse.ArgumentParser) -> None:
     env_graph_spec_group = parser.add_argument_group(
         "Environment Graph Spec Arguments", "Arguments specific to environment graph spec"
     )
+    env_graph_spec_group.add_argument(
+        "--placement_layouts",
+        type=str,
+        default=None,
+        help="Companion pose YAML overriding the environment's placement_layouts file (requires --env_spec)",
```
alexmillane: Does this mean that we don't support these in the python environments? Is there a good reason for that? Could we add support there?

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
```diff
@@ -36,14 +38,34 @@
...
 
     Args:
         graph_spec: A validated graph spec (asset refs exist, ids unique, etc.).
         enable_cameras: Forwarded to the embodiment so its cameras are added.
+        placement_layouts: Companion pose file overriding the graph reference, relative to the working directory.
+    """
+    return build_arena_env_with_assets_from_graph_spec(graph_spec, enable_cameras, placement_layouts)[0]
+
+
+def build_arena_env_with_assets_from_graph_spec(
```
alexmillane, line 63: Do we need 2 functions here? One that just calls the other and indexes by zero? Seems superfluous.

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
```diff
@@ -53,13 +75,28 @@ def build_arena_env_from_graph_spec(graph_spec: ArenaEnvGraphSpec, enable_camera
...
+    if cache_path is not None:
+        from isaaclab_arena.relations.placement_layouts import PlacementLayouts
+
+        assert not graph_spec.object_sets, "Cached layouts require concrete assets, not object sets"
+        layouts = PlacementLayouts.from_yaml(cache_path)
+        unknown = set(layouts.poses) - set(assets_by_node_id)
+        assert not unknown, f"Unknown cached graph objects: {unknown}"
+        runtime_poses = {assets_by_node_id[key].get_scene_key(): poses for key, poses in layouts.poses.items()}
+        assert len(runtime_poses) == len(layouts.poses), "Cached graph nodes must have distinct scene keys"
+        arena_env.placement_layouts = PlacementLayouts(runtime_poses)
```
alexmillane: Suggestion to move the loading of the placement layouts into its own function, above the creation of the `IsaacLabArenaEnvironment` and pass it into its constructor.

alexmillane: There are a bunch of other helpers above that are doing similar things.

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
```diff
@@ -165,27 +201,68 @@ def _instantiate_object_reference(
...
+    if pose_dict is None:
+        return None
+    assert isinstance(pose_dict, dict), "initial_pose must be a mapping"
+    if "position_xyz" not in pose_dict or "rotation_xyzw" not in pose_dict:
+        assert default_pose is None or isinstance(
+            default_pose, Pose
+        ), "Partial initial_pose requires a fixed default pose"
+        default = default_pose if default_pose is not None else Pose.identity()
+        pose_dict = default.to_dict() | pose_dict
+    return Pose.from_dict(pose_dict)
```
alexmillane, line 242: In my syringe MR I'm adding a function `Pose.from_dict` factory that handles this such that we don't have to keep implementing it everywhere.

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
```diff
@@ -165,27 +201,68 @@ def _instantiate_object_reference(
...
+
+
+def _apply_initial_pose(asset: PlaceableAsset, pose: Pose | None) -> None:
+    """Set an asset's fixed pose at construction and on reset."""
+    if pose is not None:
+        asset.set_initial_pose(pose)
```
alexmillane: This function does nothing. It just called `set_initial_pose`. I suppose that it makes the underlying function tolerant to a `None`?

Suggestion to either a) make the original function tolerant to a None or b) add another function to the class `maybe_apply_pose` that is the same as `apply_pose` but is tolerant to a `None` value.

otherwise every single agent reimplements this basically empty function in every file in the codebase.

alexmillane (commented):
Partial review. I got bogged down on nits. I'll come back when I have more time and review from a high level.

docs/pages/concepts/object_placement/relations.rst
zhx06: If we remove the IsAnchor requirement directly, our current clutter initialization may use a support pose that changes during solving, which could lead to invalid release positions or some instability after the simulation starts.

For syringes on the tray, we might need a chain of object placement: place objects without `ClutterOn` first, then hold their solved poses fixed and place the `ClutterOn` children. I’m not sure if we want to introduce and maintain this staged placement logic in Arena in this MR, or leave it as a follow-up.

docs/pages/concepts/object_placement/clutter_placement.rst
alexmillane, line 1: This looks quite long and angenty.

Prefer short docs + images.

docs/pages/concepts/object_placement/relations.rst
```diff
@@ -195,6 +216,56 @@ a Boolean value.
...
+   - position_xyz: [0.1, 0.2, 0.8]
+     rotation_xyzw: [0.0, 0.0, 0.0, 1.0]
+   - position_xyz: [-0.1, 0.2, 0.8]
+     rotation_xyzw: [0.0, 0.0, 0.0, 1.0]
+
+Positions are in the local environment frame, in metres. One index selects a
+complete layout across all objects. Environment ``i`` starts at index
+``i % num_layouts`` and advances independently on each reset, wrapping to zero
+after the last layout. With two layouts, environment 0 selects 0, 1, 0, ...;
+environment 1 selects 1, 0, 1, ... . Partial resets advance only the resetting
```
alexmillane, line 241: I would question this approach.

This means that environment 0's second draw is the same as the environment 1's first draw.

This guaruntees that you will have repeats (num_envs repeats), even if you had an infinitely sizes pool.

Was there a reason for this approach. It would make more sense to me that you have a queue of layouts, loaded from the file, then environments draw from this queue as required. This would mean that you don't get repeats (until you run out of layouts).

isaaclab_arena/assets/object_reference.py
```diff
@@ -55,13 +55,21 @@ def _build_reset_event(self):
...
-            T_W_O = self.initial_pose_relative_to_parent
-        else:
-            T_P_O = self.initial_pose_relative_to_parent
-            T_W_P = self.parent_asset.initial_pose
-            T_W_O = T_W_P.multiply(T_P_O)
-        return T_W_O
+        parent_pose = self.get_parent_pose()
+        if parent_pose is None:
+            return self.initial_pose_relative_to_parent
+        return parent_pose.multiply(self.initial_pose_relative_to_parent)
```
alexmillane, line 61: Suggestion to return the notation to what it was. This makes the transformation between frames correct.

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
```diff
@@ -38,14 +40,38 @@
...
+        placement_layouts: Companion pose file overriding the graph reference, relative to the working directory.
+    """
+    arena_env, _ = build_arena_env_with_assets_from_graph_spec(
+        graph_spec, enable_cameras=enable_cameras, placement_layouts=placement_layouts
+    )
+    return arena_env
+
+
+def build_arena_env_with_assets_from_graph_spec(
+    graph_spec: ArenaEnvGraphSpec, enable_cameras: bool = False, placement_layouts: str | Path | None = None
```
alexmillane, line 64: `placement_layouts` -> `placement_layouts_path`

isaaclab_arena/environment_spec/arena_env_graph_conversion_utils.py
```diff
@@ -175,28 +225,58 @@ def _instantiate_object_reference(
...
     for obj in graph_spec.objects:
         params = dict(obj.params)
+        params.pop("initial_pose", None)
         params.setdefault("instance_name", obj.id)
         assets_by_node_id[obj.id] = asset_registry.get_asset_by_name(obj.registry_name)(**params)
 
+    for node in nodes:
+        if "initial_pose" in node.params:
+            asset = assets_by_node_id[node.id]
+            asset.maybe_set_initial_pose(_get_pose_from_dict(node.params["initial_pose"], asset.get_initial_pose()))
```
alexmillane, line 278: This is a weird statement. Do we need the final `asset.get_initial_pose()`. Isn't the function resillient to getting a `None`.

isaaclab_arena/environment_spec/arena_env_graph_spec.py
```diff
@@ -50,6 +50,11 @@ class ArenaEnvGraphSpec(BaseModel):
     relations: list[SpatialRelationSpec] = Field(
         default_factory=list, description="Spatial layout relations across all assets."
     )
+    placement_layouts: str | None = Field(
```
alexmillane, line 53: `placement_layout_path` This is a path to the layouts, not the layouts themselves.

isaaclab_arena/tests/clutter/test_clutter_cli.py
```diff
@@ -0,0 +1,177 @@
...
+    from isaaclab_arena.embodiments.no_embodiment import NoEmbodiment
+
+    AssetRegistry().register(NoEmbodiment, key="clutter_test_no_embodiment")
+
+
+def test_help_lists_settling_controls():
+    result = _run(["--help"])
+    assert result.returncode == 0, result.stderr
+    assert "--passive_move_thresh_m" in result.stdout
+    assert "--required_quiet_windows" in result.stdout
```
alexmillane, line 48: This test just tests the CLI args dont change. We don't really want that.

isaaclab_arena/scripts/generate_clutter_scene.py
```diff
@@ -0,0 +1,166 @@
...
+        timeout_s=args.timeout_s,
+        poll_interval_s=args.poll_interval_s,
+        move_thresh_m=args.move_thresh_m,
+        turn_thresh_deg=args.turn_thresh_deg,
+        required_quiet_windows=args.required_quiet_windows,
+        fall_through_tolerance_m=args.fall_through_tolerance_m,
+        containment_margin_m=args.containment_margin_m,
+        passive_move_thresh_m=args.passive_move_thresh_m,
+        passive_turn_thresh_deg=args.passive_turn_thresh_deg,
+    )
```
alexmillane, line 49: Could we use hydra to simplify this? This seems like a prototypical use case. We have a dataclass, and we want to reflect all it's members to the CLI.

If we can just use hydra for stuff like this it would simplify the code a lot.

@cvolkcvolk

isaaclab_arena/relations/relations.py
```diff
@@ -220,6 +222,68 @@ def __init__(
...
+        gap_m: float = 0.03,
+        random_yaw: bool = True,
+        edge_margin_m: float = 0.0,
+        relation_loss_weight: float = 1.0,
+    ):
+        """Declare a clutter release.
+
+        Args:
+            parent: Fixed support carrying IsAnchor.
+            spread: Fraction of the support's width and depth available for release, in (0, 1].
```
alexmillane, line 249: Not quite sure what this does. Suggestion to expand the comment.

isaaclab_arena/tests/clutter/test_clutter_cli.py
```diff
@@ -0,0 +1,177 @@
...
+    output.write_text("existing scene")
+    result = _run(_arguments(output))
+    assert result.returncode != 0
+    assert "Output already exists" in result.stderr
+    assert output.read_text() == "existing scene"
+
+
+@pytest.mark.with_subprocess
+@pytest.mark.parametrize("preset", [None, "newton"])
+def test_cli_generates_scene_cache(tmp_path, preset):
```
alexmillane, line 102: Suggestion to remove the tests in the file before this one. They don't really test anything.

isaaclab_arena/relations/clutter/settle.py
```diff
@@ -0,0 +1,304 @@
...
+    members: dict[str, list[str]] = {}
+    for asset in assets:
+        relation = get_relation(asset, ClutterOn)
+        if relation is not None:
+            members.setdefault(relation.parent.get_scene_key(), []).append(asset.get_scene_key())
+    assert members, "Environment must declare at least one ClutterOn relation"
+    return [ClutterGroup(support, tuple(objects)) for support, objects in members.items()]
+
+
+def settle_clutter(
```
alexmillane, line 62: Please break this function up. Large functions are pretty much always wrong. Also many-fold nested loops are also a sign that something has gone wrong.

isaaclab_arena/relations/object_placer.py
```diff
@@ -247,6 +257,25 @@ def _place_ranked(
...
+                    object_uses_mesh_collision(obstacle, self.params.solver_params.collision_mode)
+                    and mesh_cache.get_collision_mesh(obstacle) is not None
+                ):
+                    continue
+                collision_bboxes.append(obstacle.get_world_bounding_box())
+
+        for i, positions in enumerate(initial_positions):
+            self._initialize_clutter_positions(
+                positions, self._get_bounding_boxes_for_candidate_index(candidate_bboxes, i), collision_bboxes
+            )
```
alexmillane, line 277: What's going on here? I'm surprised to see special treatment of the clutter objects in the placer.

Perhaps a comment?

isaaclab_arena/relations/object_placer.py
```diff
@@ -262,6 +291,14 @@ def _place_ranked(
...
         assert self._solver.last_loss_per_env is not None
         all_losses: list[float] = self._solver.last_loss_per_env.cpu().tolist()
+        for index, (positions, orientations, loss) in enumerate(
+            zip(all_positions, orientations_per_candidate, all_losses, strict=True)
+        ):
+            values = [loss, *orientations.values(), *(value for position in positions.values() for value in position)]
+            assert all(math.isfinite(value) for value in values), (
+                f"Non-finite solver output for environment {index // candidates_per_env}, "
+                f"candidate {index % candidates_per_env}"
+            )
```
alexmillane, line 301: I'm surprised to see additions here. What's going on? Is this addition to check that the results are finite? Added for a reason? 

Can we move this out to function?

isaaclab_arena/relations/object_placer.py
```diff
@@ -555,6 +596,50 @@ def _compute_on_guided_position(
...
         return (x, y, z)
 
+    def _initialize_clutter_positions(
+        self,
+        positions: dict[PlaceableAsset, tuple[float, float, float]],
+        bboxes: dict[PlaceableAsset, AxisAlignedBoundingBox],
+        collision_bboxes: list[AxisAlignedBoundingBox],
+    ) -> None:
+        """Lower clutter releases into the first free vertical interval, in asset order."""
+        unplaced_clutter = {obj for obj in positions if get_relation(obj, ClutterOn) is not None}
```
alexmillane, line 606: `clutter_objects`

isaaclab_arena/relations/object_placer.py
```diff
@@ -555,6 +596,50 @@ def _compute_on_guided_position(
...
+                    for axis in range(2)
+                )
+                if (
+                    overlaps_xy
+                    and bottom + height > float(obstacle.min_point[0, 2]) - gap
+                    and bottom < float(obstacle.max_point[0, 2]) + gap
+                ):
+                    bottom = float(obstacle.max_point[0, 2]) + gap
+            positions[obj] = (xy[0], xy[1], bottom - float(box.min_point[0, 2]))
+            unplaced_clutter.remove(obj)
```
alexmillane, line 641: Please break this function up. Either by comments or (better) by moving pieces out to helper functions. It's very hard to read as is.

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -127,6 +128,38 @@ def _solve_relations(self) -> None:
...
+                for scene_name, scene_pose in asset.layout_pose_to_scene_writes(pose):
+                    scene_poses.setdefault(scene_name, []).append(
+                        list(scene_pose.position_xyz + scene_pose.rotation_xyzw)
+                    )
+        assert scene_poses and all(
+            len(poses) == layouts.num_layouts for poses in scene_poses.values()
+        ), "Cached assets must write distinct scene entities in every layout"
+        self._placement_event_cfg = EventTermCfg(
+            func=ResetPlacementLayouts, mode="reset", params={"poses": scene_poses}
+        )
```
alexmillane, line 161: Let's move this out of this file. The equivalent for the online placement, `_solve_relations` calls out to another file to generate the event term.  Suggestion to follow that strategy.

alexmillane (commented):
Looking good. It's getting there. I've left a bunch of comments. 

Could you break this MR up. It's a bit difficult to review at its current length.

Could you reduce the number of tests? Right now this add ~120 tests. Most of which don't really test the behaviour that this MR introduces. I would suggest a handful of tests that check that settling in a test scene actually works.

PR #1255, opened 2026-09-14
Unify task success and progress tracking

cvolkcvolk:
The bigger idea is: progress and task success should come from the same definition and use the same state.

Previously, progress tracking recorded milestones, while a separate success check decided when to finish. Those two could disagree—for example, placement could trigger success even though the tracked lift stage never happened.

Now the responsibilities are clearer:

1. **The task declares what ends the episode.**
  A new `TaskTerminationCfg` puts success objectives, failure conditions, and the time limit together in `get_termination_cfg()`. 

2. **The progress tracker remembers what has happened.**
   `ProgressTracker` knows which stage is active and which stages have finished. Its completion state is now the source of task success, not just a record for progress reporting.

3. **The builder connects this to Isaac Lab.**
   `ArenaEnvBuilder` translates the task's criteria into Isaac Lab termination terms. For tasks with success objectives, it automatically installs one shared `ProgressBasedSuccessTerm` term, which owns the tracker. 

This removes local `TerminationsCfg` classes and its separate `get_progress_objectives()` hook. The task's completion criteria now live in one place.

During termination evaluation, the shared term advances the tracker.

This prepares the next change: stateful predicates. For settle → lift → rest for 10 seconds → place, the tracker could maintain the rest counter while that stage is active. That counter could use the same episode-reset path.

**Scope and merge dependency:** This MVP migrates the existing PickAndPlaceTask, OpenDoorTask, and NoTask. OpenDoor uses its existing sequence: move away from the reset position → open past the threshold. Other tasks, composite/sequential task support, and documentation updates are in #1256. Merging this PR alone breaks environment construction for unmigrated tasks, so both PRs should be ready before merging. Mimic is outside this work's scope.


isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -188,6 +187,44 @@ def _compose_metrics_cfg(self, metrics: list[MetricBase] | None) -> object | Non
...
+            assert (
+                getattr(component_termination_cfg, "success", None) is None
+            ), "Define success objectives in the task's TaskTerminationCfg; the builder owns the success term."
+        task_termination_fields = [
+            (name, TerminationTermCfg, failure) for name, failure in task_termination_cfg.failures.items()
+        ]
+        task_termination_fields.append(
+            ("time_out", TerminationTermCfg, TerminationTermCfg(func=time_out, time_out=True))
+        )
+        if task_termination_cfg.success:
```
cvolkcvolk: Could we make this more descriptive? Maybe a short comment here would help

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -188,6 +187,44 @@ def _compose_metrics_cfg(self, metrics: list[MetricBase] | None) -> object | Non
         fields = [(m.name, MetricTermCfg, m.get_metric_term_cfg()) for m in metrics]
         return make_configclass("MetricsCfg", fields)()
 
+    def _build_termination_manager_cfg(
+        self,
+        task_termination_cfg: TaskTerminationCfg,
+        scene_termination_cfg: object | None,
+        embodiment_termination_cfg: object | None,
+    ) -> object:
```
cvolkcvolk: Anything more explicit than -> object ?

isaaclab_arena/progress_tracking/task_success.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class TaskSuccessFromProgress(ManagerTermBase):
+    """Advance required task progress and report completion on the same environment step."""
+
+    def __init__(self, cfg: TerminationTermCfg, env):
+        super().__init__(cfg, env)
+        progress_objectives = cfg.params["progress_objectives"]
+        assert progress_objectives, "Task success requires at least one progress objective."
+        assert getattr(env, "_progress_tracker", None) is None, "Only one root term may own task progress."
```
cvolkcvolk: Could we try to avoid a getattr() call here?

isaaclab_arena/progress_tracking/task_success.py
```diff
@@ -0,0 +1,40 @@
...
+from isaaclab_arena.progress_tracking.progress_tracker import ProgressTracker
+from isaaclab_arena.tasks.predicates.object_settling import reset_rest_pose_recorder
+
+
+class TaskSuccessFromProgress(ManagerTermBase):
+    """Advance required task progress and report completion on the same environment step."""
+
+    def __init__(self, cfg: TerminationTermCfg, env):
+        super().__init__(cfg, env)
+        progress_objectives = cfg.params["progress_objectives"]
```
cvolkcvolk: How do we know that cfg.params holds a "progress_objectives"?

isaaclab_arena/progress_tracking/task_success.py
```diff
@@ -0,0 +1,40 @@
...
+        assert progress_objectives, "Task success requires at least one progress objective."
+        assert getattr(env, "_progress_tracker", None) is None, "Only one root term may own task progress."
+        self._tracker = ProgressTracker(progress_objectives, num_envs=env.num_envs, device=env.device)
```
cvolkcvolk: No abbreviations, call it directly _progress_tracker

cvolkcvolk (commented):
Self review

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -422,54 +407,15 @@ def record_post_step(self):
         return None, None
 
 
-def progress_tracking_reset_func(env, env_ids, progress_objectives: list[ProgressObjective]) -> None:
```
peterd-NV, line 425: I'm slightly confused as to how the removal of this works. The reset function is used to reset the internal state machine of the progress tracker. With it gone, how does the state machine run/stay consistent between env resets?

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -188,6 +188,47 @@ def _compose_metrics_cfg(self, metrics: list[MetricBase] | None) -> object | Non
         fields = [(m.name, MetricTermCfg, m.get_metric_term_cfg()) for m in metrics]
         return make_configclass("MetricsCfg", fields)()
 
+    def _build_termination_manager_cfg(
+        self,
+        task_termination_cfg: TaskTerminationCfg,
+        scene_termination_cfg: object | None,
+        embodiment_termination_cfg: object | None,
```
alexmillane: I wonder if we could simplify by getting rid of these other sources of termination.

I think we've had them in there for symmetry but we're not using them so far and...  YAGNI.

isaaclab_arena/progress_tracking/progress_objective.py
```diff
@@ -31,19 +33,16 @@ class ProgressObjectiveCompletionMode(str, Enum):
...
-    group, ANY requires at least one, and CHOOSE requires at least K.
+    Supply exactly one of sequence or predicate_groups. Predicates within each sequence
+    must hold in order. Named groups advance independently; logical determines how many
+    groups must complete.
 
     Args:
         name: Identifies the ProgressObjective within the TaskBase.
-        predicate_groups: The sequential predicate chains that define the ProgressObjective.
+        sequence: One ordered list of predicates, optionally paired with score weights.
+        predicate_groups: Named independent predicate sequences, each expressed as a list.
```
alexmillane: I guess we've added sequence as a convenience argument. Should we make a single argument which is either and then detect inside?

isaaclab_arena/progress_tracking/task_success.py
```diff
@@ -0,0 +1,46 @@
...
+    def __call__(self, env, progress_objectives: list[ProgressObjective]) -> torch.Tensor:
+        """Advance each active stage once and return task completion per environment."""
+        self._progress_tracker.step(env, step_index=env.episode_length_buf)
+        return self._progress_tracker.is_complete()
+
+    def reset(self, env_ids=None) -> None:
+        """Clear progress and initial resting positions for the restarting environments."""
+        selected_env_ids = self._environment_ids if env_ids is None else self._environment_ids[env_ids]
+        self._progress_tracker.reset(selected_env_ids)
+        reset_rest_pose_recorder(self._env, selected_env_ids)
```
alexmillane, line 63: I wonder if we should add a generic `on_reset(env, env_ids)` to our `ArenaManagerBasedEnv` such that we can group all Arena-specific reset behaviour in one place.

isaaclab_arena/tasks/open_door_task.py
```diff
@@ -50,61 +48,40 @@ def __init__(
...
-                ],
-                description=f"Move the {self.openable_object.name} door, then open it past the success threshold.",
-            ),
-        ]
+                        partial(self.openable_object.is_open, **is_open_params),
+                    ],
+                    description=f"Move the {self.openable_object.name} door, then open it past the success threshold.",
+                ),
+            ],
+        )
```
alexmillane, line 81: Amazing simplification!

alexmillane: One question, Do we want to maintain the ability to return a normal `TerminationCfg`?

isaaclab_arena/tasks/pick_and_place_task.py
```diff
@@ -157,32 +130,43 @@ def get_mimic_env_cfg(self, arm_mode: ArmMode):
...
+        return TaskTerminationCfg(
+            timeout_s=self.episode_length_s,
+            success=[
+                ProgressObjective(
+                    name="pick_and_place",
+                    sequence=[
+                        partial(
+                            objects_settled,
+                            object_names=[self.pick_up_object.name],
+                        ),
```
alexmillane, line 162: It's probably the subject of another MR. But we'd like to get the `object_settled` out og the ProgressObjectives, for the reasons we've spoken about before.

I wonder if we can reuse some of the `ProgressObjectives` machinery/underlying classes, to track some generic set of events (not attached to task success).

But that's probably the subject of another MR.

alexmillane (commented):
I love the design intent. Both simplifies and removes the potential for an error.

I left a couple of small comments.

isaaclab_arena/progress_tracking/progress_tracker.py
cvolkcvolk: We’re keeping `ProgressTracker.reset()`. What’s removed is `progress_tracking_reset_func()`, the separate reset-event callback. `TaskSuccessTerm` now owns `ProgressTracker`. 
When an environment resets, Isaac Lab’s TerminationManager calls `TaskSuccessTerm.reset(env_ids)`, which forwards the reset to ProgressTracker

Progress updates also move into `TaskSuccessTerm`, during termination evaluation

isaaclab_arena/environments/arena_env_builder.py
cvolkcvolk: Uhh thats a good point. That would leave `TaskTerminationCfg` as the only source for terminations

cvolkcvolk: We indeed don't use it currently and task could still declare a condition via `TaskTerminationCfg.failures`

cvolkcvolk: Love it

isaaclab_arena/progress_tracking/progress_objective.py
cvolkcvolk: Agree its just a convenience argument to make it more clear from the call-site what it means. Either its a list or a dict. I will merge it to one argument.


Eg its curretnly that a list defines one sequence:

```
  ProgressObjective(
      name="pick_and_place",
      predicate_groups=[settled, lifted, placed],
  )
```

A dictionary defines named sequences that advance independently:

```
  ProgressObjective(
      name="place_both_objects",
      predicate_groups={
          "left_object": [left_lifted, left_placed],
          "right_object": [right_lifted, right_placed],
      },
      logical="all",
  )
```



Sidenote, What do you think about renaming  `PredicateGroups` to `PredicateSequences` as its essentially just a type alias on a list of `PredicateSequence` @peterd-NV 

```
PredicateSequence = list[Callable] | list[tuple[Callable, float]]
PredicateGroups = dict[str, PredicateSequence]
```
to 
```
PredicateSequences = dict[str, PredicateSequence]
```

isaaclab_arena/progress_tracking/task_success.py
cvolkcvolk: I think that could make sense as a broader cleanup. But maybe worth a follow up as this PR will touch already quite a few places. Added TODO beside the reset_rest_pose_recorder() call

isaaclab_arena/tasks/open_door_task.py
cvolkcvolk: I don't see a reason for this other than backcompatibility. I’d prefer to keep `TaskTerminationCfg` as the only return type

isaaclab_arena/tasks/pick_and_place_task.py
cvolkcvolk: Agreed, let’s handle that separately. Added a TODO

isaaclab_arena/progress_tracking/progress_tracker.py
peterd-NV: Make's sense to me after reviewing design doc.

isaaclab_arena/tasks/open_door_task.py
```diff
@@ -50,61 +48,40 @@ def __init__(
...
-                name="open_door",
-                predicate_groups=[
-                    partial(
-                        is_away_from_rest_openness,
-                        asset_cfg=SceneEntityCfg(
-                            self.openable_object.name, joint_names=[self.openable_object.openable_joint_name]
+        return TaskTerminationCfg(
+            timeout_s=self.episode_length_s,
+            success=[
+                ProgressObjective(
```
cvolkcvolk, line 64: @alexmillane revisiting your earlier comment: 
I removed the convenience "predicate_sequence" argument and made it only "predicate_sequences". 

Now the plural fits the dictionary case, but reads awkwardly for the more common single-list case (Where it reads if each list entry is a sequence, wheras its just one sequence). And renaming it to singular would create the opposite problem...


So I think I prefer to make the two forms explicit:

```
# One ordered sequence
  ProgressObjective(
      name="pick_and_place",
      predicate_sequence=[settled, lifted, placed],
  )
```


```
# Several named sequences that progress independently
  ProgressObjective(
      name="clear_table",
      predicate_sequences={
          "can": [can_lifted, can_placed],
          "bottle": [bottle_lifted, bottle_placed],
      },
      logical="all",
  )
```





And background, here `predicate_sequences` is just a rename of `predicate_groups` 

It was
```
PredicateSequence = list[Predicate] | list[tuple[Predicate, float]]
PredicateGroups = dict[str, PredicateSequence]
```
and now

```
PredicateSequence = list[Predicate] | list[tuple[Predicate, float]]
PredicateSequences = dict[str, PredicateSequence]

```

However I don't fully like the we pass

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -258,10 +283,10 @@ def compose_manager_cfg(self) -> tuple[IsaacLabArenaManagerBasedRLEnvCfg, dict[s
...
             placement_event_cfg = PlacementEventCfg()
         variations_event_cfg = self._compose_variations_event_cfg()
-        progress_objectives = task.get_progress_objectives()
-        progress_tracking_events_cfg: Any = (
-            make_progress_tracking_events_cfg(progress_objectives) if progress_objectives else None
-        )
+        task_termination_cfg = task.get_termination_cfg()
+        assert isinstance(
+            task_termination_cfg, TaskTerminationCfg
+        ), "Tasks must return TaskTerminationCfg with success objectives, failures, and timeout_s."
```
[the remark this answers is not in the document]
cvolkcvolk: Will follow in #1256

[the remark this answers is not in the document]
cvolkcvolk: Will follow #1256

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -188,6 +188,31 @@ def _compose_metrics_cfg(self, metrics: list[MetricBase] | None) -> object | Non
...
+    ) -> dict[str, TerminationTermCfg]:
+        """Translate TaskTerminationCfg into Isaac Lab termination terms.
+
+        Args:
+            task_termination_cfg: Task-owned success objectives, failures, and timeout.
+
+        Returns:
+            Named failure and timeout terms, plus a success term when objectives are defined.
+        """
+        termination_terms = dict(task_termination_cfg.failures)
```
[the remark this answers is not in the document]
cvolkcvolk: The builder now returns a generated configclass using make_configclass(), so .success and .time_out remain accessible as attributes.

isaaclab_arena/environments/isaaclab_arena_manager_based_env.py
```diff
@@ -35,6 +39,7 @@ def __init__(
     ):
         apply_arena_global_settings()
         self._arena_world: ArenaWorld | None = None
+        self._progress_tracker: ProgressTracker | None = None
```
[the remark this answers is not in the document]
cvolkcvolk: Done

isaaclab_arena/environments/arena_env_builder.py
```diff
@@ -188,6 +188,31 @@ def _compose_metrics_cfg(self, metrics: list[MetricBase] | None) -> object | Non
...
+        termination_terms["time_out"] = TerminationTermCfg(func=time_out, time_out=True)
```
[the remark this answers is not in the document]
cvolkcvolk: Updated

isaaclab_arena/tasks/task_termination_cfg.py
```diff
@@ -0,0 +1,36 @@
...
+from isaaclab.managers import TerminationTermCfg
+
+from isaaclab_arena.progress_tracking.progress_objective import ProgressObjective
+
+
+@dataclass
+class TaskTerminationCfg:
+    """Declare when a task succeeds, fails, or runs out of time."""
+
+    timeout_s: float
```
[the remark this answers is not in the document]
cvolkcvolk: Adressed in #1256. Will merge into #1255

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -8,61 +8,58 @@
...
 from isaaclab_arena.tasks.predicates.consecutive import ConsecutivePredicate
-from isaaclab_arena.tasks.predicates.object_settling import reset_rest_pose_recorder
 
-_PROGRESS_TRACKER_ATTR = "_progress_tracker"
-
-
-def _contains_nested_termination_term_cfg(value) -> bool:
-    """Return whether a nested parameter contains a managed term config."""
 
+def _initialize_predicate_parameters(value, env) -> None:
```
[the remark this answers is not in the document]
cvolkcvolk: Proper reuse would need an Isaac Lab resolver that can be called independently of an installed manager.

alexmillane (approved):
Re-approve. Looks great!

PR #1259, opened 2026-09-14
Update packages flagged for vulnerabilities

peterd-NV:
## Summary
Update package versions that were flagged for vulnerabilities by Nspect.


xyao-nv (approved):
Thanks for fixing it!

PR #1268, opened 2026-09-16
Add run-time distractor-disappear variation

viiik-inside:
Adds ObjectDisappearVariation, a RunTimeVariationBase that teleports a distractor object to a far-away pose with a configurable probability.

## Detailed description
  - Distractor clutter was fixed for every episode, so a policy always saw the same set of non-task objects. This adds a per-object knob to thin it out at random.
  - Adds `ObjectDisappearVariation`, a run-time variation that teleports an object far out of the scene with a configurable probability, drawn per environment on every reset. Backed by a new reusable `BernoulliSampler`.
  - Attached automatically to every rigid object and disabled by default, mirroring `ObjectMassVariation`, so it is reachable as `<object>.disappear.enabled=true` from the CLI or an experiment config without editing the scene.
  - Teleporting happens in a reset event rather than a spawn pose: relation placement rewrites non-anchor poses on reset, and variation events are composed after it, so the teleport is what survives


isaaclab_arena/tests/test_object_disappear_variation.py
```diff
@@ -0,0 +1,204 @@
...
+from isaaclab_arena.tests.utils.persistent_simulation_app import run_function_with_persistent_simulation_app
+
+HEADLESS = True
+
+TEST_ASSET_NAME = "sphere"
+TEST_EVENT_NAME = f"{TEST_ASSET_NAME}_disappear"
+
+# Relation-placement regression case: a rigid object placed on an anchored table.
+TEST_TABLE_NAME = "table"
+TEST_BOX_NAME = "cracker_box"
```
alexmillane, line 15: Suggestion to save these variables at the top of the file for things we expect to change. e.g. HEADLESS. Otherwise just put constants in line.

isaaclab_arena/variations/object_disappear_variation.py
```diff
@@ -0,0 +1,111 @@
...
+    ground is an infinite half-space, so there is no free space under it. Past ~100 km, float32
+    coordinates are too coarse for contact resolution and the object jitters and sinks through
+    instead of resting.
+    """
+
+    sampler_cfg: BernoulliSamplerCfg = field(default_factory=BernoulliSamplerCfg)
+    """Probability that the object disappears, drawn per environment on every reset."""
+
+
+def hold_object_away(
```
alexmillane, line 57: This function just sets the pose of selected objects if the sampler says so.

What about `set_object_poses_with_probability`?

alexmillane (approved):
LGTM. A couple of nits.

isaaclab_arena/tests/test_object_disappear_variation.py
viiik-inside: Done in 8151b4f9f: only HEADLESS stays at the top; the asset names and the 100 m threshold are inlined where they are used.

isaaclab_arena/variations/object_disappear_variation.py
viiik-inside: Renamed to set_object_poses_with_probability in 8151b4f9f.

PR #1272, opened 2026-09-16
Fix collision bounds for RotateAroundSolution yaw

zhx06:
Fix collision bounds for objects rotated by `RotateAroundSolution`. Add regression tests for marker-only yaw rotations and overlapping placements.

## Why this change?

- `RotateAroundSolution` could rotate an object without rotating the bounding box used for collision checks.
- This let overlapping objects pass placement validation.

## What changes?

- Include the `RotateAroundSolution` rotation when computing collision bounds, so checks account for the object's rotated shape.
- Add regression tests for rotated bounds and for an overlapping placement that previously passed validation.

## Validation

- 170 focused placement tests passed.
- Of the nine new regression cases, seven fail without the fix; all nine pass with it.
- `pre-commit run --all-files` passed.

Initial placement comparison: incorrect bounds on the left; corrected bounds with mesh collision checking on the right

Initial placement: incorrect bounds allow overlap (left); corrected bounds with mesh collision checking avoid it (right). Mesh configuration is separate from this PR.


xyao-nv (approved):
good catch!

isaaclab_arena/relations/object_placer.py
```diff
@@ -465,7 +465,7 @@ def _rotate_candidate_bboxes(
             extra_yaws = [
                 orientations_per_candidate[c].get(obj, marker_yaw) - marker_yaw for c in range(num_candidates)
             ]
-            if not has_roll_pitch and all(yaw == 0.0 for yaw in extra_yaws):
+            if not has_roll_pitch and marker_yaw == 0.0 and all(yaw == 0.0 for yaw in extra_yaws):
```
xyao-nv, line 472: Can you add inline docstring explaining why if <><><> this needed

zhx06: Addressed.

PR #1276, opened 2026-09-16
Add jaw gripper release and withdrawal predicates for insertion task

xyao-nv:
## Summary
Add release and withdrawal predicates for insertion task

## Detailed description
- Insertion tasks could require gripper release and gripper withdraw as success criteria
- Add articulation state query into ArenaWorld
- Include configurable release thresholds and TCP withdrawn distance


isaaclab_arena/tasks/predicates/gripper.py
```diff
@@ -0,0 +1,88 @@
...
+    Returns:
+        Boolean tensor with one result per environment.
+    """
+    measured = env.arena_world.get_joint_position(robot_name, gripper_joint_name)
+    commanded = env.arena_world.get_processed_actions(gripper_action_name)[:, 0]
+    # Each finger moves inward by (open_joint_m - measured), reducing the fully open gap by twice that amount.
+    gap = span_m - 2.0 * (open_joint_m - measured)
+    # A finger stalled short of its closing target stays more open than commanded.
+    # Also require the jaw gap to match the object's width, indicating a grasp rather than another obstruction.
+    gripped = ((measured - commanded) > stall_threshold_m) & (torch.abs(gap - grasp_width_m) < grasp_width_tolerance_m)
```
peterd-NV: Does this work for when the gripper is opening? For example, if `measured` is 0.5 and `commanded` is 1.0, then mesaured - commanded < 0. `gripped` is then false and the predicate returns True immediately. 

I think adding abs(measured - commanded) here might help fix this directional issue.

peterd-NV (approved):
LGTM, one comment on the math for gripper released function that I think should be looked at.

alexmillane (commented):
Thanks for adding this. I also ran into this and would like to use it.

I wonder if we could add an example of how this will be used? I'm wondering how to get this new predicate into a task while keeping our tasks multi-embodiment.

What's your idea there?

PR #1262, opened 2026-09-15
Add YAML env_cfg_override and gear graph specs

qianl-nv:
Add validated graph-level environment configuration overrides and physics-backend defaults. Migrate the gear-insertion environments from registered Python factories to reusable graph YAML specifications.

## Why this change?
Environment graph YAML describes scenes and tasks but previously could not select a default physics backend or tune the generated Isaac Lab environment configuration. Gear insertion therefore relied on separate Python environment factories that duplicated graph configuration and solver setup.

## What changes?
Add `default_physics_backend` and `env_cfg_override` to `ArenaEnvGraphSpec`. Overrides are applied after backend resolution, support validated nested `_target_` configclasses, and preserve backend preset defaults when fields are patched without changing their concrete type.
Move the easy and medium gear environments to graph YAML files that include shared Newton configuration. The previous registered gear environment names and Python factory are removed; run the variants through the behavior demo or load their graph YAML directly.
Document backend resolution, Newton `replicate_physics` behavior, and graph override usage, including nested MJWarp/VBD coupling.

## Usage
```bash
/isaac-sim/python.sh \
  isaaclab_arena_environments/isaac_cap/gear_insertion/gear_env_behaviour_demo.py medium
```
A graph can select and tune Newton as follows:
```yaml
default_physics_backend: newton
env_cfg_override:
  sim:
    physics:
      solver_cfg:
        njmax: 8192
        nconmax: 4096
      collision_cfg:
        _target_: isaaclab_newton.physics.NewtonCollisionPipelineCfg
        reduce_contacts: true
```

## Validation

- `test_physics_presets.py -k env_cfg_override`: 4 passed
- Manually verified that the gear behavior demo loads the graph YAML and begins Newton environment creation.

PR #1281, opened 2026-09-17
Move syringe physics settings into shared YAML

alexmillane:
## Summary
Move syringe physics settings into YAML

## Detailed description
- Move physics params manual overrides to YAML (enabled by #1262).
- Share physics settings across single, both, and cluttered variants.
- Keep Python adaptations where YAML or Isaac Lab support is missing.
- Fix small bug with hydra overrides.


isaaclab_arena_environments/isaac_cap/syringe_sort/environments/environment.py
```diff
@@ -75,6 +42,10 @@ def build(self, cfg: SyringeSortEnvironmentCfg):
 
         spec = ArenaEnvGraphSpec.from_yaml(str(Path(__file__).with_name(self.yaml_file)))
         arena_env = spec.to_arena_env(enable_cameras=cfg.enable_cameras)
+        arena_env.placer_params.random_yaw_init = False
```
alexmillane, line 51: Add the following comment:

```
# NOTE(alexmillane, 2028.09.17): The placement of 4 syringes on the tray is tight. Below ensures that they are correctly placed.
# NOTE(alexmillane, 2028.09.17) [arena-parameters-overrides-missing-feature]: Currently solver params only adjustable from python. Move these overrides to yaml when this is possible.
# NOTE(alexmillane, 2028.09.17) [clutter-placement-missing-feature]: Move to clutter-based placement when that feature is enabled.
```

isaaclab_arena_environments/isaac_cap/syringe_sort/environments/syringe_env_config.yaml
```diff
@@ -0,0 +1,39 @@
...
+        integrator: implicitfast
+        nconmax: 5000
+        njmax: 5000
+        iterations: 100
+        ls_iterations: 50
+        impratio: 20.0
+        cone: elliptic
+        use_mujoco_contacts: true
+        # Preserve Isaac Lab defaults used by the previous task-owned solver,
+        # rather than Arena's Newton preset values (15000 and 2).
```
alexmillane: Remove this comment.

alexmillane (commented):
Self review.

qianl-nv (approved):
Thanks for the fix! Much cleaner now!

isaaclab_arena_environments/isaac_cap/syringe_sort/environments/syringe_env_config.yaml
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) 2026, The Isaac Lab Arena Project Developers (https://github.com/isaac-sim/IsaacLab-Arena/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Shared simulation configuration for all syringe variants.
+default_physics_backend: newton
+env_cfg_override:
+  scene:
+    replicate_physics: false
```
qianl-nv, line 10: BTW Lab explicitly state that Newton + replicate_physics: false is not supported
https://github.com/isaac-sim/IsaacLab/blob/develop/source/isaaclab/isaaclab/scene/interactive_scene_cfg.py#L117 
I think cap only every run with single env so it's not a problem, but this might not work for multi-env

alexmillane: good catch. I've removed this as it's ineffective. I guess at some point we may see some issues. Let's cross that bridge when it comes :sweat_smile:

PR #1292, opened 2026-09-18
Consolidate USD helper funcs into one utils.usd package

xyao-nv:
## Why
USD utilities were split between top-level `utils/usd_*.py` modules and the existing `utils/usd/` package.

## What

- Consolidates four USD utility modules under `isaaclab_arena.utils.usd `and updates all package, test, and
  example imports. No intended behavior change.

## Before

USD utilities used mixed locations such as:

  - utils.usd_articulation
  - utils.usd_helpers
  - utils.usd_pose_helpers
  - utils.usd_prim_tree

## After

USD utilities share one consistent namespace:

  - utils.usd.articulation
  - utils.usd.helpers
  - utils.usd.pose
  - utils.usd.prim_tree


qianl-nv (approved):
LGTM. Thanks for the clean up!

PR #1304, opened 2026-09-21
Add temporal predicates / "stateful predicates"

cvolkcvolk:
Add runner-owned consecutive-step requirements in `tasks/predicates/temporal.py` and use them in `PickAndPlaceTask`. Support environment-initialized instantaneous checks without requiring manager inheritance.

## Why this change?

Some tasks need a condition to remain true, not just become true briefly. For example, an object should stay correctly placed and stable for several steps before success.

[#1255](https://github.com/isaac-sim/IsaacLab-Arena/pull/1255) gave task progress and success one update and reset lifecycle. This change uses that lifecycle for consecutive-step counters, without adding another manager term or reset callback.

## What changes?

- `TrueForConsecutiveStepsCfg` declares the predicate and required number of control steps. It holds no runtime state.
- `ProgressObjectiveRunner` creates an internal `_TrueForConsecutiveSteps` instance for each occurrence. Counting starts when that requirement becomes active; a false result clears the streak. Episode resets clear the selected environments' counters through `TaskSuccessTerm` and `ProgressTracker`.
- A requirement can wrap a callable or an instantaneous `TerminationTermCfg`. Configured callable classes receive `(cfg, env)`; they do not need to inherit from `ManagerTermBase`. Scene references are resolved before construction.
- `ProgressTracker` prevents repeated updates for the same step from advancing progress twice. Skipped steps interrupt a streak. Direct calls to `ProgressTracker.step()` with temporal requirements must supply per-environment step indices; `TaskSuccessTerm` already supplies them.
- `PickAndPlaceTask` can require its complete placement condition to hold for multiple steps. `placement_consecutive_steps=1` remains the default.
- `objects_below_velocity_thresholds()` exposes a rest check without pose recording or counters. The existing `objects_settled()` behavior is unchanged.
- The existing `ProgressTracker.get_predicate()` accessor unwraps temporal requirements for passive diagnostic reads. The unused `termination_term_result()` adapter is removed.

Requirements work in the middle or at the end of a sequence. To require overlapping conditions, combine their instantaneous results and wrap that combined predicate once. Composite tasks' final-condition checks continue updating the final requirement without counting twice.

Follow up [#1305](https://github.com/isaac-sim/IsaacLab-Arena/pull/1305) migrates  `ConsecutivePredicate`, `CompositePredicate`

## Usage example

Here `settled`, `lifted`, and `placed_and_stable` are already-configured predicate callables. Their arguments are unchanged.

Before: one successful placement check completes the sequence.

```python
[...]

PR #1307, opened 2026-09-21
Simplify progress updates to one call per control step

cvolkcvolk:
Simplify `ProgressTracker` to advance once per control step through `TaskSuccessTerm`. Remove duplicate-step filtering and skipped-step recovery while keeping predicate-result reuse within each update and per-environment resets.

## Why this change?

Stacked on #1304. Isaac Lab already calls `TaskSuccessTerm` once per control step, and `ProgressTrackingRecorder` only reads the result. The extra step-index bookkeeping supports repeated or skipped tracker updates that the normal environment loop does not perform.

## What changes?

- Remove the stored last-step indices, update masks for repeated calls, and skipped-step counter resets.
- Keep `step_index` as optional event metadata. It no longer controls whether progress advances.
- Keep active-predicate masks and the shared per-update cache, so progress advancement and final-condition checks do not count the same requirement twice.
- Update tests and documentation to use one update per control step and passive reads afterward.

Task declarations do not change. Direct callers must call `ProgressTracker.step()` exactly once per control step; repeated calls are no longer ignored. Manually invoking `TaskSuccessTerm` is not a passive success check. The outdated cable-routing-v2 demo's direct success-term call remains outside this change and needs migration with its existing API usage.

## Usage example

Before, repeating the same index was ignored:

```python
tracker.step(env, step_index=env.episode_length_buf)
tracker.step(env, step_index=env.episode_length_buf)
success = tracker.is_complete()
```

After, update once and read the stored result as needed:

```python
tracker.step(env, step_index=env.episode_length_buf)
success = tracker.is_complete()
progress = tracker.get_state()
```

## Validation

- All pre-commit hooks passed with `pre-commit run --all-files`.
- 50 simulation-independent test bodies passed in Docker, covering temporal requirements, objective tracking, composition, and success-term updates/resets. These were called directly, without the simulator wrapper.
- Normal focused pytest was blocked before the test body by duplicate `--enable_cameras` registration, followed by missing `omni.timeline` during teardown. The same failure reproduces on unchanged #1304.
- Five broader task/builder checks could not import the locally missing `isaaclab_tasks.contrib` dependency. The unrelated CAP cable-task check was excluded. Full simulator validation remains outstanding.


isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -476,32 +467,18 @@ def _all_objectives_complete(runners: list[ProgressObjectiveRunner]) -> torch.Te
...
-            ), "Reset progress before restarting step indices."
-            updated_envs = step_index > self._last_processed_step
-            if not bool(updated_envs.any()):
-                return
-            # Unobserved control steps cannot contribute to a consecutive streak.
-            skipped_steps = (self._last_processed_step >= 0) & (step_index > self._last_processed_step + 1)
-            for runner in self.runners:
-                runner._reset_consecutive_step_requirements(skipped_steps)
-        active_envs = updated_envs.clone()
+        active_envs = torch.ones(self.num_envs, dtype=torch.bool, device=self.device)
```
xyao-nv: Is there any specific reason why the special handling of step = 1, 1, 2 is removed?
`skipped_steps = (self._last_processed_step >= 0) & (step_index > self._last_processed_step + 1)`

isaaclab_arena/progress_tracking/progress_tracker.py
```diff
@@ -476,32 +467,18 @@ def _all_objectives_complete(runners: list[ProgressObjectiveRunner]) -> torch.Te
...
-            step_index is not None or not self._requires_step_index
-        ), "TrueForConsecutiveStepsCfg requires a per-environment step_index."
-        updated_envs = torch.ones(self.num_envs, dtype=torch.bool, device=self.device)
         if step_index is not None:
             assert step_index.shape == (self.num_envs,), "step_index must contain one index per environment."
             assert step_index.dtype in (torch.int32, torch.int64), "step_index must contain integer indices."
-            step_index = step_index.to(device=self.device)
             assert bool((step_index >= 0).all()), "step_index must be non-negative."
-            assert bool(
-                (step_index >= self._last_processed_step).all()
```
xyao-nv: How about this index jumping case? step =1, 3, 4

See full discussion on slack.

PR #987, opened 2026-07-31
Add directional lighting to env-graph scenes

alexmillane:
## Summary
Directional light + shadows for env-graph scenes. Includes refactor for optional dimming of dome light when lighting direction variations are turned on.

## Detailed description
- **Why:** Env-graph/robolab scenes had no directional light, so `directional_light.direction` lighting variations had nothing to act on.
- **What: yaml env-graphs** 
  - We now inject a `directional_light` alongside the dome.
  - It is **off by default** (new `LightBase.on()`/`off()`)
- **What: lighting direction variation**
  - Directional light turned on when `directional_light.direction` variation is enabled
  - Optional dimming of a dome light when the variation is activated
    - Needed to be able to cast shadows
  - Default dome intensity raised 500 → 1500.

PR #1309, opened 2026-09-21
Copy the default spawner cfg per instance

### LINKED ISSUE #1044: [Bug Report] DomeLight instances share one DomeLightCfg via a mutable default argument
### Describe the bug

`DomeLight.__init__` uses the class-level `default_spawner_cfg` object as its `spawner_cfg` default argument:

```python
# isaaclab_arena/assets/object_library.py:458,468
default_spawner_cfg = sim_utils.DomeLightCfg(color=(0.75, 0.75, 0.75), intensity=500.0)

def __init__(
    self,
    ...
    spawner_cfg: sim_utils.DomeLightCfg = default_spawner_cfg,
```

`Object.__init__` then stores it by reference without copying (`isaaclab_arena/assets/object.py:55`, `self.spawner_cfg = spawner_cfg`). Since `set_intensity`, `set_color`, `set_color_temperature` and `add_hdr` mutate `self.spawner_cfg` in place (`object_library.py:415`, `422`, `433-434`, and in `add_hdr`), every `DomeLight` constructed without an explicit `spawner_cfg` shares one config object.

Consequences within a single process:

- Adjusting the light in one environment silently changes it in every other environment that used the default.
- The mutation persists into the class attribute, so lights constructed *later* start from the mutated value rather than the documented default of 500.
- `LightIntensityVariation` / `LightColorVariation` / `HDRImageVariation` operate through these same setters, so variation sweeps leak across environments too.

This is silent — nothing raises, the light is simply the wrong brightness or colour. It matters most for multi-environment processes such as `test_all_environments.py` and `experiment_runner`, and for any workflow that tunes lighting per environment.

The same pattern is worth checking on the sibling light classes (`DirectionalLight` and others deriving from `LightBase`), which follow the same `default_spawner_cfg` idiom.

### Steps to reproduce

```python
from isaacsim import SimulationApp
app = SimulationApp({"headless": True})

from isaaclab_arena.assets.object_library import DomeLight

a = DomeLight()
b = DomeLight()

print("same cfg object:", a.spawner_cfg is b.spawner_cfg)
print("same as class default:", a.spawner_cfg is DomeLight.default_spawner_cfg)

a.set_intensity(385.0)

print("a.intensity =", a.spawner_cfg.intensity)
print("b.intensity =", b.spawner_cfg.intensity)
print("DomeLight.default_spawner_cfg.intensity =", DomeLight.default_spawner_cfg.intensity)
print("freshly constructed c.intensity =", DomeLight().spawner_cfg.intensity)

app.close()
```

Output:

```
same cfg object: True
same as class default: True
a.intensity = 385.0
b.intensity = 385.0
DomeLight.default_spawner_cfg.intensity = 385.0
freshly constructed c.intensity = 385.0
```

Expected: `b` and the class default stay at 500.0, and only `a` becomes 385.0.

### System Info

- Arena Release: `main` at 094a69bfb (v0.2.x)
- Isaac Sim 6.0 / Isaac Lab 3.0 Beta, container image `isaaclab_arena:latest`

### Additional context

The straightforward fix is to default `spawner_cfg` to `None` and copy when falling back:

```python
spawner_cfg: sim_utils.DomeLightCfg | None = None,
...
super().__init__(..., spawner_cfg=spawner_cfg if spawner_cfg is not None else self.default_spawner_cfg.copy())
```

Copying a caller-supplied cfg as well would also stop two assets constructed from one shared cfg object from aliasing, though that is a broader change to `Object`.

### Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)
- [x] I have checked that the issue is not in running Isaac Sim itself and is related to the repo

### Acceptance Criteria

- [ ] Two `DomeLight` instances constructed without an explicit `spawner_cfg` do not share a config object, and mutating one does not affect the other or the class attribute.
- [ ] The same holds for the other `LightBase` subclasses using the `default_spawner_cfg` idiom.
- [ ] A regression test covers the isolation.

catplotlib:
Fix `GroundPlane`, `Sphere`, `DomeLight` and `DirectionalLight` sharing a single spawner cfg across instances. Default `spawner_cfg` to `None` and resolve it to a copy of the class default.

Fixes #1044

## Why this change?

Each of these four classes uses its class-level `default_spawner_cfg` as the default argument value, and `Object.__init__` stores what it receives by reference (`isaaclab_arena/assets/object.py:56`). Every instance built without an explicit cfg therefore points at the same object.

The light setters mutate that object in place (`set_intensity`, `set_color`, `set_color_temperature`, `add_hdr`), so three things follow:

- Adjusting one light changes every other light that used the default.
- The mutation lands on the class attribute, so instances built *later* start from the mutated value instead of the documented default.
- `LightIntensityVariation`, `LightColorVariation` and `HDRImageVariation` go through those setters, so variation sweeps leak across environments.

Nothing raises; the light is simply the wrong brightness or colour. It matters most for multi-environment processes such as `test_all_environments.py` and `experiment_runner`.

The issue reports this for `DomeLight`. The same pattern is in three more classes, so this PR covers all four:

| class | class attribute | default argument |
| --- | --- | --- |
| `GroundPlane` | L317 | L324 |
| `Sphere` | L340 | L361 |
| `DomeLight` | L443 | L453 |
| `DirectionalLight` | L497 | L507 |

## What changes?

`spawner_cfg` now defaults to `None`, and each `__init__` resolves it as `copy.deepcopy(self.default_spawner_cfg) if spawner_cfg is None else spawner_cfg`. That mirrors the way `initial_pose` already resolves against its class default in `DirectionalLight.__init__`, and `copy` was already imported.

Passing an explicit cfg is unchanged: it is still stored by reference, so callers that hold on to a cfg and mutate it keep working. Only the default path copies. `self.default_spawner_cfg` is used rather than the class name so subclass overrides continue to apply.

## Validation

Passing:

- `black --line-length 120 --unstable` and `flake8 7.0.0` on the changed file: clean, and unchanged from the same checks on `main`.
- The sharing behaviour and the fix, reproduced with a standalone script mirroring the same construction pattern (class-level default, stored by reference, in-place setter):

  ```
  BEFORE: shared=True   other=2000.0  later-built=2000.0  class-default=2000.0
  AFTER : shared=False  other=500.0   later-built=500.0   class-default=500.0
  explicit cfg still passed through by reference: True
  ```

Could not run:

- The repository test suite. `object_library.py` imports `isaaclab.sim` at module scope and the suite needs Isaac Sim, which I do not have access to, so I could not exercise the real `DomeLightCfg` or the variation classes.

Remaining unverified: behaviour of the real Isaac Lab cfg objects under `copy.deepcopy`. They are dataclass-style configs, so a deep copy should be safe, but I have not run it. I also did not add a test for the same reason; happy to add one if you can point me at a spot in the suite that runs without the simulator.


catplotlib:
Thanks, the `LightBase` point is correct and I had missed it. `LightBase.__init__` (L387) deep-copies whatever cfg it receives, and has since #987, so `DomeLight` and `DirectionalLight` never shared a cfg on `main` and my two light hunks were copying a copy. I checked the chain before changing anything: `GroundPlane` and `Sphere` go `LibraryObject` -> `Object`, and `LibraryObject.__init__` forwards `**kwargs` without copying, so those two did share. That half of #1044 is real; the light half is already fixed.

Reworked in 2543eef:

- The resolution now happens once in `LibraryObject.__init__`, next to the existing `name` and `scale` resolution, rather than being repeated per class. It covers `GroundPlane`, `Sphere`, and any future library object that sets `default_spawner_cfg`.
- No double copy for lights: `DomeLight` now passes `spawner_cfg=None`, `LightBase` deep-copies `None` (a no-op), and `LibraryObject` does the single copy.
- An explicitly passed cfg is still stored by reference, so callers that keep a handle and mutate it are unaffected.
- Added `test_default_spawner_cfg_not_shared` to `tests/test_object_configuration.py`, using the existing `run_function_with_persistent_simulation_app` harness as suggested. It asserts two default-built spheres have distinct cfgs, that mutating one leaves the sibling, the class default and a later instance untouched, that the light setter path stays isolated, and that an explicit cfg is passed through.

On validation: `black --line-length 120 --unstable` and `flake8` are clean and unchanged from `main`. I still cannot run the suite itself as I do not have Isaac Sim, so the new test is unexecuted. Every assertion in it holds against a standalone model of the fixed construction chain, but it needs a run on your side before you trust it.

catplotlib:
Good point, broadened in 6ac3c54. The test now loops over `GroundPlane`, `Sphere`, `DomeLight` and `DirectionalLight`, asserting for each that two default-built instances hold distinct cfgs, that neither aliases the class default, and that the copy keeps the class's own cfg type. The setter path is now exercised for both light classes rather than just `DomeLight`. `GroundPlane` is checked by identity and type only, since its default is a bare `GroundPlaneCfg()` with no field I can assert a value on.

Still unrun on my side, for the same reason as above.

PR #1311, opened 2026-09-22
Clarify temporal predicates and validate placement step counts

cvolkcvolk:
Clarify sequential, independent, and overlapping temporal requirements with simple examples. Reject invalid placement step counts when `PickAndPlaceTask` is constructed.

## Why this change?

Complete the remaining review updates from #1304. Remembering that two conditions each completed is different from requiring them to hold during the same steps; the documentation now shows both explicitly.

## What changes?

- Show the three timing patterns and the supported predicate declaration formats.
- Validate `placement_consecutive_steps` in `PickAndPlaceTask.__init__()` and extend the existing task test. Invalid values now fail at construction instead of later in `get_termination_cfg()`; valid configurations are unchanged.
- Explain why `ProgressTracker` copies step indices and record a `TODO(cvolk)` for the separate `PredicateCfg` proposal. No new configuration format is introduced.

## Validation

- Passed all repository pre-commit hooks and `git diff --check`.
- Runtime tests were not rerun for this follow-up. Earlier local attempts were blocked by duplicate `--enable_cameras` registration and the missing `isaaclab_tasks.contrib` module.


xyao-nv (approved):
THx!

PR #1312, opened 2026-09-22
Skip non-RGB camera observations in video recorder

xyao-nv:
Record only RGB observations with `CameraObsVideoRecorder`. Ignore depth, segmentation, and other camera observations that cannot be encoded as RGB mp4. Used by CAP [here](https://github.com/isaac-sim/IsaacLab-Arena/commit/8accb0debc2c824a6d4f3f5d972c74e0408aa09f)

## Why this change?

The `camera_obs` observation group may contain non-RGB policy inputs. For example, depth observations have shape `[N, H, W, 1]`.

The video recorder currently treats every observation as an RGB stream. This can cause invalid encoding attempts or fail the
recorder's three-channel assertion.

## What changes?

Filter camera observations before creating video writers. Record only batched observations with shape `[N, H, W, 3]`.

Add a regression test containing both RGB and depth observations. The test verifies that only the RGB stream creates a video writer.

## Validation
- `test_camera_observation_video_recorder.py`` passed 
- `pre-commit run --all-files`: passed


isaaclab_arena/video/camera_observation_video_recorder.py
```diff
@@ -103,12 +110,17 @@ class EpisodeVideoWriter:
...
 
     Cameras are batched as ``[N_envs, H, W, C]``.  Each env is recorded
     independently; its encoder is finalised when that env resets (terminated
     or truncated), producing one file per completed episode:
     ``<name_prefix>-env<N>-<camera_name>-episode-<E>.mp4``.
+
+    RGB observations are supported by default. ``modality_adapters`` maps an
+    observation modality suffix to a callable that produces ``[H, W, 3]`` RGB
+    uint8 frames. This extension point can visualize depth, normals, or
+    segmentation while keeping the recording backend focused on mp4 video.
```
xyao-nv: Remove " while keeping the recording backend focused on mp4 video."

isaaclab_arena/video/camera_observation_video_recorder.py
```diff
@@ -103,12 +110,17 @@ class EpisodeVideoWriter:
 
 
 class CameraObsVideoRecorder(gym.Wrapper):
-    """Record one mp4 per (env, camera, episode) in ``obs['camera_obs']``.
+    """Record supported camera-observation modalities as one mp4 per episode.
```
xyao-nv: Record supported modalities as one mp4 per (env, camera, episode)  in ``obs['camera_obs']``.

isaaclab_arena/video/camera_observation_video_recorder.py
```diff
@@ -117,12 +129,19 @@ def __init__(
...
         os.makedirs(video_folder, exist_ok=True)
         self.video_folder = video_folder
         self.name_prefix = name_prefix
         self.fps = fps if fps is not None else int(env.metadata.get("render_fps", 30))
+        self.modality_adapters: dict[str, CameraObservationVideoAdapter] = {"rgb": _to_uint8}
+        if modality_adapters is not None:
+            self.modality_adapters.update(modality_adapters)
+        assert all(self.modality_adapters), "Video adapter modality names must not be empty."
+        self._video_adapter_modalities = sorted(self.modality_adapters, key=len, reverse=True)
+        self._warned_unadapted_observations: set[str] = set()
```
xyao-nv: do you need it here? how about just assert when init

isaaclab_arena/video/camera_observation_video_recorder.py
```diff
@@ -159,12 +182,37 @@ def step(self, action):
 
         return result
 
+    def _find_video_adapter(self, observation_name: str) -> CameraObservationVideoAdapter | None:
```
xyao-nv: Too much compelxities. How about just keeping the base class, + RGB supports

isaaclab_arena/video/camera_observation_video_recorder.py
```diff
@@ -146,6 +148,11 @@ def step(self, action):
 
             with Timer("record_camera_frames"):
                 for camera_name, frames in cam_obs.items():
+                    if not camera_name.endswith("_rgb"):
```
qianl-nv: would it be better to use  observation term’s params["data_type"] == "rgb" instead of the naming subfix here?

qianl-nv (approved):
LGTM! Approved with a nit

PR #1313, opened 2026-09-22
Update public env gen model to DeepSeek Flash

xyao-nv:
Update the public agentic environment-generation endpoint to use `deepseek-ai/deepseek-v4.1-flash`. Refresh the model-selection guidance with the new benchmark snapshot.

## Why this change?

The previous public default, `deepseek-ai/deepseek-v4-pro-0813`, is deprecated. DeepSeek V4.1 Flash passed the structural checks for all five documented example workflows. It's faster compared with Kimi K3 during inference.

## What changes?

- Set `PUBLIC_ENDPOINT.model` to `deepseek-ai/deepseek-v4.1-flash`.
- Replace the deprecated model and its statistics in the model-selection documentation.

## Validation
- Passed tests
- 15/15 passed on benchmark on documented prompts


qianl-nv (approved):
Thanks for taking care of this!

